PR #5, opened 2015-10-22
Purge generated files

pavoljuhas:
unit tests would modify log and temporary files that are tracked in git.  Remove and ignore them.
setup.py overwrites setup.cfg, which has no active settings.  Remove and ignore it too.


pkienzle:
I remember wanting to do this earlier, but being told there is some reason the files were needed in the repo.  I just looked through the tests, and the files are only used for writing, so it can't be something like them being needed for other tests (and even if it were, they should be generated anew each time).

I'm applying the pull request, and if we are wrong we can always revert.

PR #15, opened 2016-10-11
Adding basic SESANS functionality to Sasview

jhbakker:
Data1D datatype generalized to allow arbitrary (x,y) formatted data.
Added hankel transform for SAS models to Sasview.
Warning: only works for models for which Hankel transform converges over x= (0,Inf)
This is combined with changes to sassmodels and may yield unexpected results if not tested together with 
sasmodels 'Jurrian1D' branch.


ajj:
Pull request done wrong way round. Closing. Should be reopened correctly.

PR #17, opened 2016-11-09
ticket-688: horizontal splitter to enable control of Data/Theory boxes

rozyczko:
Addressing ticket #688
This commit replaces sizers with a splitter widget for Data and Theory CustomTreeCtrl controls, allowing mutual resizing of those controls.
One side effect - "Data" and "Theory" labels had to be moved to inside the CustomTreeCtrl as top-level items.


wpotrzebowski:
Tested on OSX 10.11. Works fine. 

Minor code issues:
Lines 528 and 529 
file_panel = wx.Panel(splitter, -1)
theory_panel = wx.Panel(splitter, -1) 
(Re)initialized but not used. Do we need them?

Lines 579, 582 
Assignment to "_=". Is it necessary?

wpotrzebowski:
Can one of the admins verify this patch?

butlerpd:
Hummm.... testing on my windows 10 development machine is not very satisfactory.  The main real problem is that when I try to drag the center bar a long blinking black line shows up in the data box.  Probably due to some redraw being done too often?

Thinking about the design some more would it not be better to turn the whole data management panel into a SplitterWindow? Perhaps using nested SplitterWindows or the MultiSplitterWindow maybe?

rozyczko:
How do you envision the new look and behaviour?
Using MultiSplitterWindow for the entirety of the Data Explorer - how would you see splitting the controls? Data being one, Theory being another and the buttons and comboboxes being yet another window?
Creating a nice design is not something we should do hastily and the question is if we should be doing it in 4.1? Fixing the blinking black line is one thing but a semi-major GUI rework is probably a bigger enhancement request.

pkienzle:
Mac was displaying the root label, which is why I put the label in the root.  Windows appears to not do so.   I'll see if I can convince windows to display the root.

Mac wasn't having any artifacts with the horizontal bar, nor is windows.

PR #21, opened 2016-12-07
Ticket 824

krzywon:
For some reason, the box slicers, box sum, annular averaging and sector averaging routines need data2d objects with one and only one detector given within the data set. I'm curious to know the reason behind this. This fix allows data sets with no detector information to use the slicers, sum and averaging.

### DIFF
--- a/src/sas/sascalc/dataloader/manipulations.py
+++ b/src/sas/sascalc/dataloader/manipulations.py
@@ -142,7 +142,7 @@ class _Slab(object):
         :param maj_min: min value on the major axis
         :return: Data1D object
         """
-        if len(data2D.detector) != 1:
+        if len(data2D.detector) > 1:
             msg = "_Slab._avg: invalid number of "
             msg += " detectors: %g" % len(data2D.detector)
             raise RuntimeError, msg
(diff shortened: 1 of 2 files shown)


smk78:
Replying to @krzywon:

Interesting...

The structure of the file that did not work (14250.h5) is:
sasentry01
  sasdata
  sasinstrument
    sassource
  sasprocess
  title

The structure of the file in the 2dtest folder that does work (33837..._v3.h5) is:
sasentry01
  sasdata
  sasinstrument
    sasdetectorrear_detector
    sassource
  sasprocess
  sastransmission_spectrum_sample
  title

Are you saying that the routines failed because the 14250 file did not have the sasdetectorrear_detector block?

Both files were written by Mantid, albeit slightly different versions, 3.8.0 and a 3.7.nightlybuild, respectively.

wpotrzebowski:
Can one of the admins verify this patch?

smk78:
I’m happy to have a look but it won’t be today.

Steve

wpotrzebowski:
Interesting - I didn't write this comment: "Can one of the admins verify this patch?"
I agree though that one of the admins can verify it.

krzywon:
That is exactly what I am saying. The 14250 Larmor data set that does not have access to the slicers and other averaging tools because it has no detector information given in the header. There is a check for exactly one detector within the code in master to (1) shows these options in the drop down and (2) even run the routines if they are reached through scripting.

My main question, mostly for Mathieu who wrote some of this, is why is this check here? Do the routines only work if there is a single continuous detector? Are you also barring zero detectors in case there is a multi-detector array that wasn't specified in the data file?

mdoucet:
I think this may be a case where the detector info was used in the initial implementation but dropped at a later point. If the info is not used, I would remove the check altogether.

PR #22, opened 2016-12-09
Ticket 795

krzywon:
The save file converter is complete and ready for testing by others. This is dependent on sasmodels branch ticket-795.


wpotrzebowski:
ready for testing

src/sas/sascalc/data_util/calcthread.py
```diff
@@ -269,7 +270,9 @@ def _run(self):
             except KeyboardInterrupt:
                 pass
             except:
-                self.exception()
+                errors += 1
+        if errors > 0:
+            self.exception()
```
krzywon: This is no longer really an issue. I am going to revert my changes back to the original. This method only runs a single time and many of the errors I am seeing (cannot plot negative numbers on log axis) are coming from the plotting perspective.

src/sas/sasgui/perspectives/fitting/pagestate.py
```diff
@@ -348,13 +350,146 @@ def clone(self):
...
+            elif self.categorycombobox == 'Parallelepiped':
+                self.formfactorcombobox = 'core_shell_parallelepiped'
+            elif self.categorycombobox == 'Shape Independent':
+                self.formfactorcombobox = 'be_polyelectrolyte'
+            elif self.categorycombobox == 'Sphere':
+                self.formfactorcombobox = 'adsorbed_layer'
+            elif self.categorycombobox == 'Structure Factor':
+                self.formfactorcombobox = 'hardsphere'
+
+    @staticmethod
```
pkienzle: Extra carriage return in the file, and one below on the next static method.

krzywon: I'm not sure how to get rid of this character. It doesn't show up in any of my editors and completely removing and adding the line again puts it back.

pkienzle:
Ignore it for now.  As far as I can tell there is no difference
between this line and the next, and I think it is a github display
issue.  So long as the line endings are consistent there is no
problem.

To check I did the following:

* set autocrlf=false in ~/.gitconfig
* clone repo and checkout ticket-795
* examine the content of the file with python
```
>>> data = open('pagestate.py','rb').read()
>>> idx = data.find('staticmethod')
>>> data[idx-20:idx+20]
"hardsphere'\r\n\r\n    @staticmethod\r\n    de"
```

Roughly half the files have DOS line endings, 155 out of 292; if we
are going to change this, it should be as a separate 4.2 ticket.  In
sasmodels there are some as well (15 out of 136).

PR #29, opened 2016-12-20
Opencl dialog

wpotrzebowski:
Simple dialog to set SAS_OPENCL variable from SasView GUI. It looks for available opencl drivers and provides menu to choose from. A corresponding ticket: #823


wpotrzebowski:
@pkienzle: I would appreciate if you can take a particular look on the function that retrieves drivers information (_get_clinfo) and the function that interacts with sasmodels (on_OK). GUI functionality has been reviewed by @rozyczko

butlerpd:
Who is assigned on to accept this pull request?  Please assign that person as soon as agreed.

src/sas/sasgui/perspectives/fitting/gpu_options.py
```diff
@@ -0,0 +1,173 @@
...
+                for device in devices:
+                    if len(devices) > 1 and len(platforms) > 1:
+                        combined_index = ":".join([str(p_index),str(d_index)])
+                    elif len(platforms) > 1:
+                        combined_index = str(p_index)
+                    else:
+                        combined_index = str(d_index)
+                    #combined_index = ":".join([str(p_index),str(d_index)]) \
+                    #    if len(platforms) > 1 else str(d_index)
+                    clinfo.append((combined_index, ":".join([platform.name,device.name])))
```
pkienzle: Please indicate "single precision" if double precision is not available (if "cl_khr_fp64" not in device.extensions).

sasmodels uses the first available double precision device is single = False in the model definition.  If I select the HD 4000 single precision device on my laptop and run a double precision model it falls back to the Intel CPU device, even though I have not selected it in the dialog.

This behaviour should be okay (if the device supports double precision it  can probably run sasview); otherwise we have to make the user select both the single and the double device.  I guess you could list pairs of devices, such as

  X  HD 4000 (single)   Intel CPU (double)
  _   Intel CPU (single)  Intel CPU (double)
  _   No OpenCL

Have you tested this on a card which presents itself as multiple identical devices?

src/sas/sasgui/perspectives/fitting/gpu_options.py
```diff
@@ -0,0 +1,173 @@
...
+        else:
+            self.sas_opencl = None
+
+    def on_OK(self, event):
+        """
+        Close window on accpetance
+        """
+        import sasmodels
+        #If statement added to handle Reset
+        if self.sas_opencl:
```
pkienzle: Should save the selection somewhere in ~/.sasview/config, and reload it on sasview startup.  Otherwise you have to set it at the beginning of every session.

src/sas/sasgui/perspectives/fitting/gpu_options.py
```diff
@@ -0,0 +1,173 @@
...
+            os.environ["SAS_OPENCL"] = self.sas_opencl
+        else:
+            if "SAS_OPENCL" in os.environ:
+                del(os.environ["SAS_OPENCL"])
+        sasmodels.kernelcl.ENV = None
+        #Need to reload sasmodels.core module to account SAS_OPENCL = "None"
+        reload(sasmodels.core)
```
pkienzle: Maybe better to add an interface into sasmodels to select a different OpenCL device in rather than writing it a note in the SAS_OPENCL environment variable then knocking it on the head...

wpotrzebowski:
Some issues resolved but still a few needs to be addressed. Not ready for testing yet!

wpotrzebowski:
SAS_OPENCL is now being remembered from session to session but it fails to initiate opencl if SAS_OPENCL=None. This happens because sasmodels.kernelcl cannot be found. I will most likely need to change interface to samodels to overcome this problem

rozyczko:
All issues resolved. Ready for merge

PR #31, opened 2017-01-19
Sesans41

jhbakker:


jhbakker:
Ok, I have adressed all the issues pointed out by Jeff. 
I do not understand the error that Travis CI throws at me. It seems pip fails to install a required package?
Can anyone make sense of it?

jhbakker:
Shouldn't build_tools/requirements.txt specify a version of pylint? 
Now it just says install pylint, does that mean it always takes the latest version?

krzywon:
For the original testing I did for this, I used the newest file in the sesans example folder, se008731_01_40pcorr.ses, but I just went back and tried some of the other files as a test. They are not loading properly because the file format and the reader have changed since the reader was first written. I think we should remove or possibly replace the files in sasview/test/sesans_data/ and sasmodels/example/ that are out of date.

jhbakker:
Paul (x2), Andrew,

Friendly reminder, just so you don't forget!
Could you please review the code, or remove yourselves from the Reviewers list before code freeze?

src/sas/sascalc/dataloader/data_info.py
```diff
@@ -92,14 +39,17 @@ class plottable_1D(object):
     dxl = None
     ## Slit smearing width
     dxw = None
+    ## SESANS specific params (wavelengths for spin echo length calculation)
+    lam = None
+    dlam = None
```
pkienzle: .lam and .dlam don't appear to be used anywhere, so why pack them around?

src/sas/sascalc/data_util/qsmearing.py
```diff
@@ -35,18 +39,42 @@ def smear_selection(data, model = None):
...
+            raise ValueError('one or more of your dx values are negative, please check the data file!')
+
+    if _found_sesans == True:
+        #Pre-compute the Hankel matrix (H)
+        qmax, qunits = data.sample.zacceptance
+        hankel = sesans.SesansTransform()
+        sesans.SesansTransform.set_transform(hankel,
+        SE = Converter(data._xunit)(data.x, "A"),
+        zaccept = Converter(qunits)(qmax, "1/A"),
+        Rmax = 10000000)
```
jhbakker: Renamed set_transform to __init__.
Modified hankel call in qsmearing to your suggestion.

src/sas/sascalc/data_util/qsmearing.py
```diff
@@ -35,18 +39,42 @@ def smear_selection(data, model = None):
...
+        # Then return the actual transform, as if it were a smearing function
+        return PySmear(SESANS1D(data, hankel._H0, hankel._H, hankel.q), model)
```
jhbakker: I tried that before making these 2 classes.
SESANS1D inherit from Sesanstransform and adds some stuff to make it a Resolution object. 
This can then be used without decreasing the efficiency of the Sesanstransform calculation.
Merging the classes reduces efficiency (I'm sure there's a way to merge them without this loss, but I don't know how...)

src/sas/sascalc/data_util/qsmearing.py
```diff
@@ -91,7 +119,17 @@ class PySmear(object):
...
+        if hasattr(self.resolution, 'data'):
+            if self.resolution.data.meta_data['loader'] == 'SESANS':  # Always True if file extension is '.ses'!
+                self.offset = 0
+            # This is default behaviour, for future resolution/transform functions this needs to be revisited.
+            else:
+                self.offset = numpy.searchsorted(self.resolution.q_calc, self.resolution.q[0])
+        else:
+            self.offset = numpy.searchsorted(self.resolution.q_calc, self.resolution.q[0])
+
+        # self.offset = numpy.searchsorted(self.resolution.q_calc, self.resolution.q[0])
```
jhbakker: That line is not my code, only the metadata check above it is mine.

q_calc must envelop q, so q_calc[0] < q[0] and q_calc[-1] > q[-1].
In words: The Hankel transform integration q-space (q_calc) must envelop the sasmodel q-space (q)
So, setting them equal will break the transform.

src/sas/sascalc/dataloader/data_info.py
jhbakker: Not used anywhere...yet. 
It will be needed for acceptance angle limits implementation. 
If it's a real problem I will remove it, otherwise I request to leave it in.

butlerpd: If it is a placeholder for future development I would argue It should stay but should be well documented in comments why it is there so that if plans change somebody else down the road knows it can safely be deleted.

pkienzle: We need to rethink our handling of data; the current code is far to complex with copies being made in several places with slightly different attributes and lots of special code asking if it is 1D data, 2D data and now SESANS data.  We will also need something similar for oriented USANS data so that the model can include orientation parameters even though it is a 1D dataset, with the resolution calculator returning a 1D data set from the 2D calculation.

My preference would be for you to add the lam/dlam attributes when you need them so that refactoring is less work.  Also, I suspect you won't need them everywhere that you've added them, but we won't know that until we have the code that uses them.

jhbakker:
Is there anything I need to do now?
Or can this pull request be left like this until code freeze?

src/sas/sascalc/dataloader/readers/cansas_reader.py
```diff
@@ -260,12 +260,20 @@ def _parse_entry(self, dom, recurse=False):
...
+                    if len(unit_list) > 1:
+                        self.current_dataset.yaxis(unit_list[0], unit_list[1])
+                    else:
+                        self.current_dataset.yaxis("Intensity", unit)
                     self.current_dataset.y = np.append(self.current_dataset.y, data_point)
                 elif tagname == 'Idev' and isinstance(self.current_dataset, plottable_1D):
                     self.current_dataset.dy = np.append(self.current_dataset.dy, data_point)
                 elif tagname == 'Q':
-                    self.current_dataset.xaxis("Q", unit)
+                    unit_list = unit.split(" | ")
```
pkienzle: Splitting on " | " may be a bit brittle.  You could split on "|", then use unit_list[0].strip(), unit_list[1].strip() to correct for spaces.

PR #33, opened 2017-02-14
no ticket: remove unnecessary sleep

pkienzle:
This pull request needs to be tested variety of platforms for a variety of models.  In particular, should be tested on mac and windows with and without OpenCL.

Requires ticket-851 pull request is applied to sasmodels.

Try to stress the system by hitting the calc button while fitting, the fit button while calculation, saving and loading complicated models, etc.

Should wait for the beginning of the release cycle: having the sleep() in the code shouldn't break anything, but removing it may exposes errors if the problem was not due to multiple threads in the calculator at the same time.


krzywon:
I tested this on Win10 and there doesn't seem to be any issue. I ran a few individual data set fits, ran a constrained fit, saved a constrained fit project, and then reloaded it without seeing any errors.

PR #34, opened 2017-02-21
ticket-756: Change category "Customized Models" to "Plugin Models"

butlerpd:
The original intent of this ticket was to **not** do what was actually done but go beyond and have category manager fully aware of, and properly manage, categories for all models, whether custom built plugins or the distributed built-in models.  Currently it only knows about the build in models.  However, it turns out to not be quite so straightforward and I believe there should be an agreement on the desired behavior and then a design produced to implement that behavior can be proposed.

The basic problem is that there are a variety of methods to keep the list of plugins up to date on the fly - plugin models can be updated, added, or deleted from within SasView so that every time the combo box is populated it needs to have the latest. This becomes trickier if the those models are dispersed through all the categories. The structure factor category is another problematic one since its members should be defined internally (depending on the model contents) and not by random user choices.

One option would be to have a Plugins and structure factor category that cannot be edited but allow models in both those categories to be added to other categories as well. That obviates the need for all lists to always be updated since a model cannot be added to a list (outside of plugin) without being first in the plugins list. However, deletion or updating of a plugin might still require some cross talk.

Bottom line is we need a consensus on the desired behavior before proceeding further.  Meanwhile this should fix the renaming from custom to plugin in a way that does not confuse category manager (part of how category manager is bypassed is to compare the category name to a string (like "plugin") and only if it does not match call category manager.


butlerpd:
Do we really have to pollute 2 modules (and any future ones that will might use plugin models) with the old name?  It would be far cleaner if the page state loader took care of this conversion.  That way it is not only in one place, but when the save/load state gets refactored next it is all self contained.

butlerpd:
I believe the comments have been addressed now.  Waiting for verification

krzywon:
I made the change Paul B. suggested to move the category name change into the conversion section of pagestate.

PR #36, opened 2017-02-22
Ticket 844: Smearing defaults to None. When data is sent to fitting, dQ is selected if pinhole or slit smeared. 

krzywon:


pkienzle:
Loading latex_smeared.xml data (usans with slit geometry), the pinhole geometry displays the pinhole parameters but the slit geometry displays height None and width 0.0. Is the information about the pinhole geometry available in the file?

krzywon:
Good catch. Yes, that should have the slit information in it. I will patch master today.

PR #37, opened 2017-02-27
Ticket #844 and #862: Smearing set properly when sent to fitting and fit plotted when model selected

krzywon:


src/sas/sascalc/data_util/qsmearing.py
```diff
@@ -107,9 +107,9 @@ def apply(self, iq_in, first_bin=0, last_bin=None):
         start, end = first_bin + self.offset, last_bin + self.offset
         q_calc = self.resolution.q_calc
         iq_calc = numpy.empty_like(q_calc)
-        if start > 0:
+        if start > 0 and self.model is not None:
             iq_calc[:start] = self.model.evalDistribution(q_calc[:start])
```
pkienzle: What is iq_in if model is None?  Is this code even reachable if model is None?  Maybe it is better to not apply resolution when there is no model to calculate?

Here's where it seems to be called from:
```
PySmear.apply from sas.sasgui.perspectives.fitting.model_thread.Calc1D.compute
PySmear.apply from sas.sascalc.fit.AbstractFitEngine.FitData1D.residuals
```

Is there somewhere else?

krzywon:
I applied the fix you suggested for _on_select_data. I meant to have that elif as an if, but as is usual, your suggestion is even cleaner.
I am looking into why we are in apply before a model is set. This only happens when a saved state is loaded, so it is likely somewhere in basepage or pagestate. I am closing this pull request until I can hunt this down.

PR #40, opened 2017-03-03
Add OSX Build Instructions

robertdfrench:
This is all that was necessary to build SasView from scratch on El Capitan. If it is suitable to merge this, I would like to update the docs to advise folks to:

1. Install homebrew
1. `source mac_setup.sh`

### DIFF
--- a/mac_setup.sh
+++ b/mac_setup.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+sudo pip install Pillow
+brew tap homebrew/science
+brew install hdf5
+sudo pip install h5py
+xcode-select --install
+sudo python setup.py install


wpotrzebowski:
Hi Robert, Can create a sasview branch and then reopen a pull request? I want to test it through jenkins server and I cannot access your repo - I suppose.

robertdfrench:
Ah, okay that makes sense. I think a homebrew package would be good for users, but necessarily amenable to setting up a development environment as per @pkienzle's concerns. What about a `mac_setup_development_environment.sh` or some such that relies on conda, and then I could (under a separate ticket) submit a SasView package to Homebrew?

robertdfrench:
@wpotrzebowski just to check, you want me to open a branch on this repo instead of submitting PR from a fork? Sounds good. I will do that when I hear back from @pkienzle or @ajj.

robertdfrench:
Ahh, no merges to master until 4.1. Got it. Fair enough!

PR #41, opened 2017-03-07
Add OSX Build Instructions

robertdfrench:
* Rename to 'mac_setup_homebrew.sh'

### DIFF
--- a/mac_setup_homebrew.sh
+++ b/mac_setup_homebrew.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+sudo pip install Pillow
+brew tap homebrew/science
+brew install hdf5
+sudo pip install h5py
+xcode-select --install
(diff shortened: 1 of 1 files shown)


butlerpd:
I believe that it was agreed we could merge this once renamed even before release of 4.1.  Am asking Andrew and Paul Kienzle to concur that my memory is not wrong.

pkienzle:
To clarify: the only request that I have is that some documentation go into the install script.  You are then welcome to merge it. 

I suspect the script is broken in that it would fail if you tried to run it on a fresh mac with only brew installed, but it is a good enough starting point for those who are already using a brew environment.  

Since we are shipping a dmg for mac users, and since we can steer new users without python experience toward anaconda over brew, getting this working right is not a priority.

PR #42, opened 2017-03-08
Ticket 850: Custom Pinhole Smearing Uses %Q for dQ

krzywon:
There is now a single input for setting custom pinhole smearing. This input is a percent of Q applied across the entire Q space. Saving and loading with the latest repository code works. Loading save states from previous versions (4.0.1 and 3.1.2 tested) works as well.


src/sas/sasgui/perspectives/fitting/basepage.py
```diff
@@ -1246,12 +1244,11 @@ def reset_page_helper(self, state):
...
-            self.dx_min = state.dx_min
-            self.dx_max = state.dx_max
-            if self.dx_min is not None:
-                self.smear_pinhole_min.SetValue(str(self.dx_min))
-            if self.dx_max is not None:
-                self.smear_pinhole_max.SetValue(str(self.dx_max))
+            self.dx_percent = state.dx_percent
+            if self.dx_percent is not None:
+                if state.dx_old:
+                    self.dx_percent = self.data.x[0] / self.dx_percent
```
pkienzle: Should probably clear dx_old since dx_percent no longer contains the dx_min from the saved state, and now contains a ratio.

Also, shouldn't this be 100*ratio rather than ratio since you divide by 100 before applying it?

krzywon: When a fit state is loaded (from a project or analysis file) the data set is loaded separately from the fit parameters and are only brought together again in basepage.py. I plan on cleaning this up for 4.2, but will need to make another large change to make this happen.

dx_old is a new flag to capture when dx_min and dx_max are found in older save states to let basepage know to convert the data from raw values to percent. Once I fully separate the computational and loading code from the gui, This will no longer be needed and get removed.

PR #43, opened 2017-03-08
Ticket 830: Modify the cansas HDF5 reader to comply NXcanSAS-1.0

krzywon:
I modified the NXcanSAS reader to comply with the v1.0 format published in December 2016. This can wait for after 4.1 is released unless there is a strong need to have it now.

Changes (in no particular order):
- dQl and dQw slit smearing data are now read in, whereas they aren't in SasView v4.0.1 (this may be the driving force for pulling now)
- roll/pitch/yaw instead of polar_angle/azimuthal_angle for multiple tilts
- Transmission Spectrum data all loaded in simultaneously, like I and Q are, to improve load times
- Lint


src/sas/sascalc/dataloader/readers/cansas_reader_HDF5.py
```diff
@@ -200,134 +223,180 @@ def read_children(self, data, parent_list):
...
                         self.detector.pixel_size.y = float(data_point)
                         self.detector.pixel_size_unit = unit
-                    elif key == u'SSD' and self.parent_class == u'SAScollimation':
+                    elif key == u'distance' and self.parent_class == \
+                            u'SAScollimation':
                         self.collimation.length = data_point
                         self.collimation.length_unit = unit
-                    elif key == u'name' and self.parent_class == u'SAScollimation':
+                    elif key == u'name' and self.parent_class == \
+                            u'SAScollimation':
```
pkienzle: PEP 8 suggests using parentheses around the condition when it doesn't fit on the line rather than backslash.  This is partly because whitespace after the backslash is a syntax error.

src/sas/sascalc/dataloader/readers/cansas_reader_HDF5.py
```diff
@@ -396,15 +468,16 @@ def final_data_cleanup(self):
...
             except:
                 self.errors.add(sys.exc_value)
             dataset.mask = zeros
-            ## Calculate the actual Q matrix
+            # Calculate the actual Q matrix
             try:
                 if dataset.q_data.size <= 1:
-                    dataset.q_data = np.sqrt(dataset.qx_data * dataset.qx_data + dataset.qy_data * dataset.qy_data)
+                    dataset.q_data = np.sqrt(dataset.qx_data * dataset.qx_data +
+                                             dataset.qy_data * dataset.qy_data)
```
pkienzle: PEP 8 suggests breaking the line before the operator rather than after.

src/sas/sascalc/dataloader/readers/cansas_reader_HDF5.py
```diff
@@ -192,17 +192,17 @@ def read_children(self, data, parent_list):
                     self.current_dataset.mask = data_set.flatten()
                     continue
                 # Transmission Spectrum
-                elif key == u'T' and self.parent_class == \
-                        u'SAStransmission_spectrum':
+                elif (key == u'T') \
+                        and (self.parent_class == u'SAStransmission_spectrum'):
```
pkienzle: You can wrap the entire condition in parentheses, then break before the operator:
```
              elif (key == u'T'
                      and self.parent_class == u'SAStransmission_spectrum'):
```

PR #44, opened 2017-03-13
Ticket 876: CanSAS Nexus as a default data reader

krzywon:
This is a pull request that can wait until after 4.1 release.

If the data file extension is not in the list of known extensions (defaults.json), this change sends the data to the cansas nexus (HDF5) reader as the 3rd and final default option before giving up. I tested loading times for 2 exact data files, but with the file extension changed. The loading times, each around 0.02 seconds, were exactly the same to the 7th decimal place.

### DIFF
--- a/src/sas/sascalc/dataloader/loader.py
+++ b/src/sas/sascalc/dataloader/loader.py
@@ -61,31 +61,32 @@ class Registry(ExtensionRegistry):
         :param format: explicit extension, to force the use
             of a particular reader
 
-        Defaults to the ascii (multi-column) reader
-        if no reader was registered for the file's
-        extension.
+        Defaults to the ascii (multi-column), cansas XML, and cansas NeXuS
+        readers if no reader was registered for the file's extension.
         """
         try:
             return super(Registry, self).load(path, format=format)
-        except:
-            try:
-                # No reader was found. Default to the ascii reader.
-                ascii_loader = ascii_reader.Reader()
-                return ascii_loader.read(path)
-            except:
-                try:
-                    cansas_loader = cansas_reader.Reader()
-                    return cansas_loader.read(path)
-                except:
-                    try:
-                        cansas_nexus_loader = cansas_reader_HDF5.Reader()
-                        return cansas_nexus_loader.read(path)
-                    except:
-                        msg = "\n\tUnknown data format: %s.\n" % path
-                        msg += "\tThe file is not a known format for SasView. "
-                        msg += "The most common formats are multi-column "
-                        msg += "ASCII, CanSAS XML, andCanSAS NeXuS."
-                        raise Exception(msg)
+        except Exception:
+            pass # try the ASCII reader
+        try:
+            ascii_loader = ascii_reader.Reader()
+            return ascii_loader.read(path)
+        except Exception:
+            pass # try the cansas XML reader
+        try:
+            cansas_loader = cansas_reader.Reader()
+            return cansas_loader.read(path)
+        except Exception:
+            pass # try the cansas NeXuS reader
+        try:
+            cansas_nexus_loader = cansas_reader_HDF5.Reader()
+            return cansas_nexus_loader.read(path)
+        except Exception:
+            # No known reader available. Give up and throw an error
+            msg = "\n\tUnknown data format: %s.\n\tThe file is not a " % path
+            msg += "known format for SasView. The most common formats are "
+            msg += "multi-column ASCII, CanSAS XML, and CanSAS NeXuS."
+            raise Exception(msg)
 
     def find_plugins(self, dir):
         """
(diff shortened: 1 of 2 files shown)


src/sas/sascalc/dataloader/loader.py
```diff
@@ -72,8 +73,19 @@ def load(self, path, format=None):
                 ascii_loader = ascii_reader.Reader()
                 return ascii_loader.read(path)
             except:
-                cansas_loader = cansas_reader.Reader()
-                return cansas_loader.read(path)
+                try:
+                    cansas_loader = cansas_reader.Reader()
+                    return cansas_loader.read(path)
+                except:
```
pkienzle: Shouldn't use bare "except:" since it will trap SystemExit and KeyboardInterrupt.  At a minimum use "except Exception:", but more specific is better if you know the specific failure possibilities for the called code.

Also, this looks like a case where "pass" is an acceptable exit from the exception.  This will keep down the indent level.  Example:
```
    try:
        load format 1
    except Exception:
        pass  # fall through to next format
    try:
        load format 2
    except Exception:
        pass  # fall through to next format
    ...
```

src/sas/sascalc/dataloader/loader.py
```diff
@@ -61,31 +61,32 @@ def load(self, path, format=None):
...
+            pass # try the cansas NeXuS reader
+        try:
+            cansas_nexus_loader = cansas_reader_HDF5.Reader()
+            return cansas_nexus_loader.read(path)
+        except Exception:
+            # No known reader available. Give up and throw an error
+            msg = "\n\tUnknown data format: %s.\n\tThe file is not a " % path
+            msg += "known format for SasView. The most common formats are "
+            msg += "multi-column ASCII, CanSAS XML, and CanSAS NeXuS."
+            raise Exception(msg)
```
pkienzle: Prefer RuntimeError(msg) over Exception(msg) for a non-specific error.  Looking at the standard errors available [https://docs.python.org/2.7/library/exceptions.html], nothing else seems to apply.  Don't use IOError, since that usually is defined using IOError(errno=code, strerror=msg), and refers to a specific OS defined error code.  You could instead make a specific error for sasview loaders which is a subclass of Exception.

It might be useful to display a reader-specific error message instead of just "Unknown data format".  You could use magic strings at the start of the file to identify cansas files, and otherwise try the ascii reader:

    HDF magic string: http://www.digitalpreservation.gov/formats/fdd/fdd000229.shtml#sign
    XML: "<?xml"  or "<SASroot"

Maybe save the traceback from the default reader, and include that in your RuntimeException if the alternate reader does not work.

krzywon: DataInfo has an 'errors' data member that is supposed to be used to store errors caught while using the default reader for a file being loaded. Unfortunately, the only 3 readers that do this are the cansas XML, cansas HDF5, and saxs ess readers. In loader.py the extra loading attempts are only supposed to happen if the extension registry cannot find a loader for the selected file.

I think making a handful of data loader specific exceptions is the way to go, but I will also have to apply them to all of the readers as well as the loader and extension registry. From there, I should be able to better infer what to do in loader.load() 

Exceptions I think would be useful to add:
NoKnownLoaderException(Exception) - If the extension registry cannot find a reader, throw this to try the 3 default types.
FileContentsException(Exception) - For data files where the registered loader is unable to open or recognize the format, or for failed attempts using the 3 default readers.
DataReaderException(Exception) - For data files that at least partially load, but had issues along the way. These should remain contained within the loader and return them with the data object in DataInfo.errors.

loader.load() would become something like:

    try:
        data = super(Registry, self).load(path, format=format)
        # Exceptions from super.load would skip this return, but data would be set
        return data
    except NoKnownLoaderException as e:
        # Nested try-except to keep e intact
        # The 
        try:
            ascii_loader = ascii_reader.Reader()
            return ascii_loader.read(path)
        except FileContentsException:
            pass # try the cansas XML reader
        # Repeat the above 5 lines for cansas XML and cansas HDF5 reader

    ...

        # If we get here, that means unreadable by SasView and then we can throw this exception
        msg = "\n\tUnknown data format: %s.\n\tThe file is not a " % path
        msg += "known format for SasView. The most common formats are "
        msg += "multi-column ASCII, CanSAS XML, and CanSAS NeXuS."
        raise Exception(msg)
    except FileContentsException as e:
        raise RuntimeException(e)
    except Exception as e:
        # For unknown or unexpected errors
        raise RunTimeException("Some useful message.")
    
    def handle_reader_errors(self, data, e=''):
        """
        Handle any errors thrown by the data loader and try to use as 
        much of the data as possible
        """
        if not isinstance(data, list):
            data = []
        first_data = data[0]
        if not isinstance(first_data, Data1D) and not isinstance(first_data, Data2D):
            x = numpy.zeroes(0)
            y = numpy.zeroes(0)
            first_data = Data1D(x,y)
        first_data.errors.append(str(e))
        data[0] = first_data
        return data

I am closing this pull request while I work on this. What I thought might be a 3 line change is turning into a much larger can of worms.

PR #46, opened 2017-03-17
MAINT: check_packages.py should work in py3

andyfaff:
Modified check_packages.py to work with python 3 as well. It checks for the major python version and warns that py2 is required for the GUI (not sure if py2 is required for any other mode of operation).
Several PEP8 fixes have been applied.

### DIFF
--- a/check_packages.py
+++ b/check_packages.py
@@ -1,7 +1,8 @@
 """
 Checking and reinstalling the external packages
 """
-import os
+from __future__ import print_function
+
 import sys
 
 # Fix for error: hash-collision-3-both-1-and-1/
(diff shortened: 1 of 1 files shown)


andyfaff:
Ok. Just thought it might be a good idea that the script that checks for
dependencies should be agnostic of the major Python version. Plus, it's
probably good for future conversion to Py 3.

>

pkienzle:
Agreed, which is why we should merge it, but not until after the 4.1 release.

It should be pretty easy to get the non-GUI bits working in python 3. The six package does a good job of providing a uniform interface to python 2 and 3. One caveat is that file read/write uses byte streams not unicode strings, so it will be a bit more involved than just fixing the syntax errors.

PR #49, opened 2017-03-24
MAINT: SlitlengthCalculator - removed float indexing

andyfaff:
This code was using float indexing which prevents recent numpy versions from working.

### DIFF
--- a/src/sas/sascalc/calculator/slit_length_calculator.py
+++ b/src/sas/sascalc/calculator/slit_length_calculator.py
@@ -15,7 +15,7 @@ class SlitlengthCalculator(object):
         self.x = None
         # y data
         self.y = None
-        #default slit length
+        # default slit length
         self.slit_length = 0.0
 
         # The unit is unknown from SAXSess profile:
@@ -41,7 +41,7 @@ class SlitlengthCalculator(object):
         :return: the slit length calculated value.
         """
         # None data do nothing
-        if self.y == None or self.x == None:
+        if self.y is None or self.x is None:
             return
         # set local variable
         y = self.y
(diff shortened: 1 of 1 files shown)


src/sas/sascalc/calculator/slit_length_calculator.py
```diff
@@ -80,19 +80,19 @@ def calculate_slit_length(self):
...
+        y_half_u = y[ind - 1]
 
         # get corresponding x values
         x_half_d = x[ind]
-        x_half_u = x[ind-1]
+        x_half_u = x[ind - 1]
 
         # calculate x at y = y_half using linear interpolation
         if y_half_u == y_half_d:
             x_half = (x_half_d + x_half_u)/2.0
```
krzywon: The only possible improvement I see is to add whitespace around the division sign here. Otherwise, everything else looks okay to me.

src/sas/sascalc/calculator/slit_length_calculator.py
```diff
@@ -80,19 +80,19 @@ def calculate_slit_length(self):
...
         else:
-            x_half = (x_half_u * (y_half - y_half_d)  \
-                       + x_half_d * (y_half_u - y_half)) \
-                        / (y_half_u - y_half_d)
+            x_half = ((x_half_u * (y_half - y_half_d) +
+                       x_half_d * (y_half_u - y_half)) /
+                      (y_half_u - y_half_d))
```
pkienzle: PEP8 now recommends putting the binary operator joining the expressions at the start of the line rather than the end of the previous line.

andyfaff: I agree. that it looks nicer on the start of the line. In my defence PyCharm told me to do it.

PR #52, opened 2017-03-25
Fixing acknowledgements box

ajj:
Changed:
- control used to display text so that users can copy/paste and so it autosizes
- order of elements to be a bit more logical

Needs testing on windows before merge.

Required for 4.1 (in previous versions, the citation and acknowledgment was truncated)

Expected layout (from mac) is:

screen shot 2017-03-25 at 23 02 02

### DIFF
--- a/sasview/local_config.py
+++ b/sasview/local_config.py
@@ -86,7 +86,7 @@ _danse_url = "http://www.cacr.caltech.edu/projects/danse/release/index.html"
 _inst_url = "http://www.utk.edu"
 _corner_image = os.path.join(icon_path, "angles_flat.png")
 _welcome_image = os.path.join(icon_path, "SVwelcome.png")
-_copyright = "(c) 2009 - 2016, UTK, UMD, NIST, ORNL, ISIS, ESS, ILL, ANSTO and TU Delft"
+_copyright = "(c) 2009 - 2017, UTK, UMD, NIST, ORNL, ISIS, ESS, ILL, ANSTO and TU Delft"
 marketplace_url = "http://marketplace.sasview.org/"
 
 #edit the list of file state your plugin can read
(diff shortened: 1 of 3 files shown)


pkienzle:
On windows, cuts off at 

    (3) .... see documentation for

butlerpd:
Actually not so on my windows - is perfect.  Moreover the outlined box is far longer than 3 on the one I see. Are you sure you are looking at the correct code Paul?

butlerpd:
Fair enough ... but are you saying that it fails for you because of that and nobody else? why would that be? and what are the two configs for? don't understand the difference between those files do you?

pkienzle:
wx version is 3.0.0.0

I'm using the wxpython from conda, very recent pull (I'm testing the travel laptop before going).

Definitely using the ajj-4.1-acknowl-2 branch.

Maybe a question of the fonts available on this machine?

Anyway, I made the box slightly larger and removed the tabs so that it works on my windows box.  The box is too large on the mac, but that is less bad (and anyway, we aren't perfectionists).

Regarding config, I think if it gets confused about config files it might forget that local config exists and try the config in guiframe.

I tried to make sense of this before last code camp, rearranging the source tree.  See ticket 887

    http://trac.sasview.org/ticket/887

butlerpd:
Humm.... interesting.  About to test your changes on my pc.  will have to
look more deeply at your github repo to see what you did (and hope that
will help me understand what is going on:-)

Paul

butlerpd:
Actually, Andrew's fix made it "too large" on my PC (#3 that you say cut off,  I had lots of room left at the end) --- even more so now that you expanded.... but seems ok.  Interestingly the box in bullet 1 just keeps expanding so that it fills whatever size the about box is (which hides the fact that otherwise the box is way to big).... but shouldn't the expanding box have taken care of different fonts on different machines?  Guess not.

Andrew should now double check on his mac I guess.  Anyway I am going to correct the dates in the about box and add it to this branch so he can verify that as well

ajj:
@pkienzle - thanks for the clean up. Will replicate in src/sas/sasgui/guiframe/config.py for now and look to clean up at code camp.

butlerpd:
Still don't understand why config.py is even needed but guess we are ready to merge into master? ... and release?

PR #55, opened 2017-03-27
MAINT: import numpy as np

andyfaff:
The standard way of importing numpy is `import numpy as np`. This PR changes all the numpy imports to reflect this.


andyfaff:
@ajj was there a reason you backed out the changes to manipulations.py?

ajj:
Yes - I screwed up the merge with changes from the Igor reader PR. Will reinstate changes tomorrow

PR #59, opened 2017-04-04
Simplify SESANS reading code

rprospero:
The new SESANS reader code should be more legible, more robust, better documented, and properly handle units.


src/sas/sascalc/dataloader/readers/sesans_reader.py
```diff
@@ -17,158 +17,163 @@
...
-                zaccept_unit_split = paramnames[7].split("[")
-                zaccept_unit = zaccept_unit_split[1].replace("]","")
-                if zaccept_unit.strip() == r'\AA^-1' or zaccept_unit.strip() == r'\A^-1':
-                    zaccept_unit = "1/A"
-                output.sample.zacceptance=(float(paramvals[7]),zaccept_unit)
-                output.vars = varheader
+                    line = input_f.readline()
+                    params = {}
+                    while line.strip() != "":
+                        if line.strip() == "":
```
pkienzle: Unreachable code

src/sas/sascalc/dataloader/readers/sesans_reader.py
```diff
@@ -17,158 +17,163 @@
...
+        Performs unit conversion on a measurement.
+
+        :param value: The magnitude of the measurement
+        :type value: Number
+        :param value_unit: a string containing the final desired unit
+        :type value_unit: String
+        :param default_unit: a string containing the units of the original measurement
+        :type default_unit: String
+        :return: The magnitude of the measurement in the new units
+        """
```
pkienzle: We ought to be using consistent markup across the project for function docs (though something is definitely better than nothing!)  Maybe add a ticket if there is not already one there?

For types I suggest using pep 484 style type hints embedded in the comments:

    # type: (float, str, str) -> float

I've been using "*parameter name* description" for a description of the inputs to the function.  I haven't tried rendering the above format with sphinx and automod.

src/sas/sascalc/dataloader/readers/sesans_reader.py
```diff
@@ -17,158 +17,161 @@
...
         :return: SESANSData1D object, or None
-        
+
         :raise RuntimeError: when the file can't be opened
         :raise ValueError: when the length of the data vectors are inconsistent
         """
         if os.path.isfile(path):
             basename = os.path.basename(path)
             _, extension = os.path.splitext(basename)
             if self.allow_all or extension.lower() in self.ext:
```
andyfaff: There are three indentation levels here, we can reduce that:

    if os.path.isfile(path):
        basename = os.path.basename(path)
         _, extension = os.path.splitext(basename)
         if not(self.allow_all or extension.lower() in self.ext):
             raise RuntimeError()
    with open(path, 'r') as input_f:
        ...

src/sas/sascalc/dataloader/readers/sesans_reader.py
```diff
@@ -17,158 +17,161 @@
...
+
+                    output.sample.zacceptance = (
+                        float(self._header_fetch(params, "Q_zmax")),
+                        self._fetch_unit(params, "Q_zmax"))
+
+                    output.sample.yacceptance = (
+                        float(self._header_fetch(params, "Q_ymax")),
+                        self._fetch_unit(params, "Q_ymax"))
 
                 if len(output.x) < 1:
```
andyfaff: `if not len(output.x):`?

src/sas/sascalc/dataloader/readers/sesans_reader.py
rprospero: Yep.  I switched from a generator based approach to a while loop and left in that short circuit condition.  I've taken it out now.

src/sas/sascalc/dataloader/readers/sesans_reader.py
rprospero: I've attempted to move to the embedded type hints, if I've understood properly

src/sas/sascalc/dataloader/readers/sesans_reader.py
rprospero: That empty file check really needed to be done right after the data was read.  I've moved it there in commit 56e2e3e

andyfaff:
As a minimum the tests should:

- check that this reader can read SESANS data, and that it reads everything correctly
- this reader raises the correct error is a SESANS dataset isn't provided.
- if the general reader is supplied with a SESANS file, that it's this reader that is used to read it.

(see `igor_reader` in `utest_abs_reader`)

src/sas/sascalc/dataloader/readers/sesans_reader.py
rprospero: That's certainly easier to follow.  Some of the lines might be turned back into one liners, since they're no longer butting up against the 80 char limit, but I haven't made those changes yet.

rprospero:
After this morning's discussions, there will be some changes to the .SES file format, so there is no reason to merge the new reader code if it won't read the updated format.

PR #63, opened 2017-04-06
Add yacceptance

rprospero:
To eventually accomodate non-isotropic samples, SESANS needs a yacceptance variable to go with the zacceptance one.

### DIFF
--- a/src/sas/sascalc/dataloader/readers/cansas_reader.py
+++ b/src/sas/sascalc/dataloader/readers/cansas_reader.py
@@ -291,6 +291,8 @@ class Reader(XMLreader):
                     pass
                 elif tagname == 'Sesans':
                     self.current_datainfo.isSesans = bool(data_point)
+                elif tagname == 'yacceptance':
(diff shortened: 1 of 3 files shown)


src/sas/sascalc/dataloader/readers/cansas_reader.py
```diff
@@ -1056,6 +1058,8 @@ def _write_data(self, datainfo, entry_node):
             sesans = self.create_element("Sesans")
             sesans.text = str(datainfo.isSesans)
             node.append(sesans)
+            self.write_node(node, "yacceptance", datainfo.sample.yacceptance[0],
```
krzywon: There is currently no way to load in yacceptance into sasview. When saving a project or sesans data set as XML, the default value of None will throw an error here. Setting the default of yacceptance (zacceptance should also probably be changed) to a tuple of (0,"") might be the safer way to go.

rprospero: Good catch.  I've set the default values as you suggested.

PR #64, opened 2017-04-07
MAINT: remove has_key occurrences

andyfaff:
Remove `has_key` occurrences, they're deprecated


src/sas/sascalc/realspace/VolumeCanvas.py
```diff
@@ -409,38 +409,37 @@ def getParam(self, name):
...
-                raise ValueError, \
-                    "VolumeCanvas.getParam: Could not find %s" % name
+                raise ValueError("VolumeCanvas.getParam: Could not find "
+                                 "%s" % name)
 
             shapeinstance = self.shapes[toks[0]]
 
             try:
-                shapeinstance.params.has_key(toks[1])
+                a = shapeinstance.params[toks[1]]
```
andyfaff: Note: not sure why the `has_key` method would ever raise a `KeyError`

src/sas/sascalc/realspace/VolumeCanvas.py
```diff
@@ -409,38 +409,37 @@ def getParam(self, name):
         toks = name.split('.')
         if len(toks) == 1:
             try:
-                self.params.has_key(toks[0])
```
andyfaff: I don't think`has_key` would ever raise a `KeyError`. If the key isn't in the dict, then it returns False, otherwise it returns True.

```
>>> a = {'a':0, 'b':1, 'c':2}
>>> a.has_key('a')
True
>>> a.has_key([])
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
>>> a.has_key('s')
False
>>> 
```

PR #67, opened 2017-04-07
MAINT: use 'is None' instead of '== None'

andyfaff:
Note: done by search/replace all.

Searched for "== None" and replaced by "is None"


andyfaff:
So I rebased on sasview/master and there were rather a lot of merge issues which were too much for me to resolve. I therefore repeated the work and force pushed to the is_none branch.

andyfaff:
Ok, so this should be mergeable now, assuming it's re-approved.

PR #69, opened 2017-04-07
MAINT: 2to3 fix raise instances.

andyfaff:
Raising errors as `raise ValueError, msg` is now deprecated in favour of `raise ValueError(msg)`. This PR has used the `2to3 -w -f raise *.py` on each of these files to make these changes. I'm not sure if I've missed any. The changes were automated. I did not inspect each change.

### DIFF
--- a/src/sas/sascalc/dataloader/data_info.py
+++ b/src/sas/sascalc/dataloader/data_info.py
@@ -714,7 +714,7 @@ class Data1D(plottable_1D, DataInfo):
                 self.x_unit = '1/A'
                 self.y_unit = '1/cm'
         except: # the data is not recognized/supported, and the user is notified
-            raise(TypeError, 'data not recognized, check documentation for supported 1D data formats')
+            raise TypeError
 
     def __str__(self):
         """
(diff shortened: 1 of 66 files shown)


andyfaff:
Weirdly there are whole files that appear to have changed. I'm guessing it's the line endings that have been normalised in git or something like that.

pkienzle:
Yes, likely.  The patch that makes consistent line endings should be applied first so that the other changesets are reviewable.  I don't know how to indicate order on pull requests.

andyfaff:
>Yes, likely. The patch that makes consistent line endings should be applied first so that the other changesets are reviewable. I don't know how to indicate order on pull requests.

How about if the following is added to .gitattributes?

```
*.py text=auto eol=lf
```

ajj:
I thought I had set the repos to coerce line-endings. Seems not. Git attributes file now updated to do so.

andyfaff:
So there were going to be a whole lot of merge problems. So I've redone the changes off the current tip of master.
Changes applied via:
`2to3 -f raise -w ./`
Hopefully this is good to go.

src/sas/sascalc/dataloader/data_info.py
```diff
@@ -714,7 +714,7 @@ def __init__(self, x=None, y=None, dx=None, dy=None, lam=None, dlam=None, isSesa
                 self.x_unit = '1/A'
                 self.y_unit = '1/cm'
         except: # the data is not recognized/supported, and the user is notified
-            raise(TypeError, 'data not recognized, check documentation for supported 1D data formats')
+            raise TypeError
```
pkienzle: Apparently python 2 allowed the raising of a tuple, by 2to3 doesn't know about it:

http://stackoverflow.com/questions/9836756/why-raising-a-tuple-works-if-first-element-is-an-exception#9837750

src/sas/sasgui/perspectives/fitting/models.py
```diff
@@ -141,7 +141,7 @@ def __nonzero__(self):
         type, value, tb = sys.exc_info()
         if type is not None and issubclass(type, py_compile.PyCompileError):
             print "Problem with", repr(value)
-            raise type, value, tb
+            raise type(value).with_traceback(tb)
         return 1
```
pkienzle: Is this 2 and 3 compatible?

Also, I have code for annotating exceptions in sasmodels.exceptions that maybe useful here.

pkienzle:
py3 branch includes all necessary changes

PR #71, opened 2017-04-07
Tickets #869 & #14

timsnow:
Three things implemented:

- Decoupling of fitting GUI from computation and plotting to prevent GUI hanging

- Rapid submission (gap of <0.1 sec) of calculations ignored, to help with project loading

- If parameters are changed many times during a calculation, middle parameter changes are dropped and only the final change is subsequently re-calculated to prevent the inadvertent evolution of long, unstoppable, calculation pipelines


src/sas/sasgui/perspectives/fitting/basepage.py
```diff
@@ -1690,7 +1703,38 @@ def _draw_model(self, update_chisqr=True, source='model'):
...
+            pass
+
+        self.last_time_fit_submitted = current_time
+
+    def _threaded_draw_worker(self, threaded_draw_queue):
+        while True:
+            # Check to see is a manager is running and a calc is running
+            if ((self._manager.calc_1D is not None) and (self._manager.calc_1D.\
+                isrunning() == True)) or ((self._manager.calc_2D is not None)\
+                and (self._manager.calc_2D.isrunning() == True)):
```
andyfaff: No need for line continuation characters if the whole if statement is enclosed by brackets.

src/sas/sasgui/perspectives/fitting/basepage.py
```diff
@@ -1690,7 +1703,38 @@ def _draw_model(self, update_chisqr=True, source='model'):
...
+        else:
+            pass
+
+        self.last_time_fit_submitted = current_time
+
+    def _threaded_draw_worker(self, threaded_draw_queue):
+        while True:
+            # Check to see is a manager is running and a calc is running
+            if ((self._manager.calc_1D is not None) and (self._manager.calc_1D.\
+                isrunning() == True)) or ((self._manager.calc_2D is not None)\
```
andyfaff: no need for `(self._manager.calc_1D.isrunning == True)` can just be `(self._manager.calc_1D.isrunning())`

pkienzle:
Ticket 866, which shows the compute error twice, should also be addressed by this ticket, or a new ticket opened if it this fix does not correct the problem there.

See comments on the top of the following for details:

    https://github.com/SasView/sasview/pull/80

Here is ticket 859, which gives a parameter set showing the error:

    http://trac.sasview.org/ticket/859

timsnow:
I've been through Paul's suggestions which were a more robust implementation of my suggestion. I also added a copy operation when adding the update_chisqr and source objects into the queue as they were being lost when the queue was shortened.

pkienzle:
Reloading the project test/save_states/constrained_fit_project.svs fails with

    'plottable_1D' object has no attribute 'dqw'

I think this might have been fixed in a separate branch, but I don't remember where.

pkienzle:
Pull request #99 (Fix broken unit tests) will perhaps fix the problem.

butlerpd:
Now we need to wait for branch 915 to make fixes so we can test whether intended functionality has in fact been added.

pkienzle:
Merged into py3 PR #109.

PR #73, opened 2017-04-08
Quick fix for numpy that caused problems in batch fitting

wpotrzebowski:


src/sas/sasgui/guiframe/data_processor.py
```diff
@@ -18,7 +18,7 @@
 
 """
 import wx
-import numpy
+import numpy as np
 import math
 import re
```
andyfaff: importing standard packages should be done before user packages. So the order here should be:

import re
import math
import wx
import numpy as np

PR #74, opened 2017-04-08
fixes #865

### LINKED ISSUE #865: Also allow the Python Shell/Editor to be invoked from the Fitting > Edit Custom Models sub-menu (Trac #749)
Also allow the Python !Shell/Editor to be invoked from the Fitting > Edit Custom Models sub-menu because it is now used more with the enhancements to custom/plugin models.

Migrated from http://trac.sasview.org/ticket/749
```json
{
    "status": "closed",
    "changetime": "2016-10-10T14:09:48",
    "_ts": "2016-10-10 14:09:48.955472+00:00",
    "description": "Also allow the Python !Shell/Editor to be invoked from the Fitting > Edit Custom Models sub-menu because it is now used more with the enhancements to custom/plugin models.",
    "reporter": "richardh",
    "cc": "",
    "resolution": "fixed",
    "workpackage": "SasView GUI Enhancements",
    "time": "2016-10-09T17:00:51",
    "component": "SasView",
    "summary": "Also allow the Python Shell/Editor to be invoked from the Fitting > Edit Custom Models sub-menu",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView 4.1.0",
    "owner": "smk78",
    "type": "enhancement"
}
```

rprospero:
I've tracked the namespace bug down to a single function call and I've managed to write a work around, but I'm at a bit of a loss as to what exactly is causing the bug.

If you look at the original code, you can call self.model.calculate_Iq() before line 1241 and get your Iq value back.  If you call it after line 1241, then you run into the namespace issue.  Debugging further than that is problematic because

a) self.model is not in scope within the function, so it' impossible to tell where it's being corrupted
b) the model loaded BY update_model_list is not corrupted.  In fact, this is a necessary component of the work around.

What this code does is store the ID of the (currently functional) model before calling self.parent.update_model_list().  It then calls the update and, somehow, corrupts the model.  When then find that model ID in the update model dictionary and use that valid model for the rest of the routine.

### DIFF
--- a/src/sas/sasgui/perspectives/fitting/fitpage.py
+++ b/src/sas/sasgui/perspectives/fitting/fitpage.py
@@ -1238,7 +1238,6 @@ class FitPage(BasicPage):
             custom_model = CUSTOM_MODEL
             mod_cat = self.categorybox.GetStringSelection()
             if mod_cat == custom_model:
-                ################This is where the model dies###############
                 temp_id = self.model.id
                 temp = self.parent.update_model_list()
                 for v in self.parent.model_dictionary.values():
--- a/src/sas/sasgui/perspectives/fitting/fitpage.py
+++ b/src/sas/sasgui/perspectives/fitting/fitpage.py
@@ -1238,7 +1238,12 @@ class FitPage(BasicPage):
             custom_model = CUSTOM_MODEL
             mod_cat = self.categorybox.GetStringSelection()
             if mod_cat == custom_model:
+                ################This is where the model dies###############
+                temp_id = self.model.id
                 temp = self.parent.update_model_list()
+                for v in self.parent.model_dictionary.values():
+                    if v.id == temp_id:
+                        self.model = v()
                 if temp:
                     self.model_list_box = temp
                     current_val = self.formfactorbox.GetLabel()
--- a/src/sas/sasgui/perspectives/fitting/models.py
+++ b/src/sas/sasgui/perspectives/fitting/models.py
@@ -318,8 +318,8 @@ class ModelManagerBase:
         new models were added else return empty dictionary
         """
         new_plugins = self.findModels()
-        if len(new_plugins) > 0:
-            for name, plug in  new_plugins.iteritems():
+        if new_plugins:
+            for name, plug in  new_plugins.items():
                 self.stored_plugins[name] = plug
                 self.plugins.append(plug)
                 self.model_dictionary[name] = plug
--- a/src/sas/sasgui/perspectives/fitting/models.py
+++ b/src/sas/sasgui/perspectives/fitting/models.py
@@ -320,10 +320,9 @@ class ModelManagerBase:
         new_plugins = self.findModels()
         if len(new_plugins) > 0:
             for name, plug in  new_plugins.iteritems():
-                if name not in self.stored_plugins.keys():
-                    self.stored_plugins[name] = plug
-                    self.plugins.append(plug)
-                    self.model_dictionary[name] = plug
+                self.stored_plugins[name] = plug
+                self.plugins.append(plug)
+                self.model_dictionary[name] = plug
             self.model_combobox.set_list("Plugin Models", self.plugins)
             return self.model_combobox.get_list()
         else:


wpotrzebowski:
Do you know if the same behavior is observed with and without opencl?

rprospero:
I only know that the behaviour appears without opencl.  I don't currently have a working opencl implementation on my machine to test with opencl.

src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -1238,7 +1238,12 @@ def _on_select_model(self, event=None):
             custom_model = CUSTOM_MODEL
             mod_cat = self.categorybox.GetStringSelection()
             if mod_cat == custom_model:
+                ################This is where the model dies###############
```
andyfaff: space after hash for comments.

src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -1238,7 +1238,12 @@ def _on_select_model(self, event=None):
...
+                temp_id = self.model.id
                 temp = self.parent.update_model_list()
+                for v in self.parent.model_dictionary.values():
```
andyfaff: How about:
```
for k in self.parent.model_dictionary:
    v = self.parent.model_dictionary[k]
    if v.id == temp_id:
        self.model = v()
```
the reason being that in Python 2.x asking for dictionary values returns a list, which can be inefficient. If this isn't a problem then perhaps the way you have it is better.

src/sas/sasgui/perspectives/fitting/models.py
```diff
@@ -320,10 +320,9 @@ def update(self):
         new_plugins = self.findModels()
         if len(new_plugins) > 0:
```
andyfaff: What is new_plugins? A list?
If it's a simple list or dict you can test for len > 0 by simply doing `if new_plugins:`. Like the following:
```
a = []
if a:
    print(a)
a = [1]
if a:
    print(a)
```

src/sas/sasgui/perspectives/fitting/models.py
```diff
@@ -320,10 +320,9 @@ def update(self):
...
             for name, plug in  new_plugins.iteritems():
```
andyfaff: `iteritems` is removed in Python 3. This should be changed to:
```for name, plug in new_plugins.items():```

File: src/sas/sasgui/perspectives/fitting/fitpage.py (hunk 1 above)
Comment (krzywon) [inline src/sas/sasgui/perspectives/fitting/fitpage.py]: I'm guessing this was a placeholder comment to show where the issue was. Can it be removed?

Comment (rprospero) [inline src/sas/sasgui/perspectives/fitting/fitpage.py]: Agreed.  We no longer need the comment.

File: src/sas/sasgui/perspectives/fitting/fitpage.py (hunk 2 above)
Comment (rprospero) [inline src/sas/sasgui/perspectives/fitting/fitpage.py]: The size of model_dictionary should be small enough and this function called rarely enough that the delay from generating the value list should be imperceptible.

It's only adding a single line of code, so I'm not opposed to it if anyone has a strong opinion on it, but I find the current version marginally more legible.

File: src/sas/sasgui/perspectives/fitting/models.py (hunk 3 above)
Comment (rprospero) [inline src/sas/sasgui/perspectives/fitting/models.py]: Python thinks of everything.  I've made the change.

File: src/sas/sasgui/perspectives/fitting/models.py (hunk 4 above)
Comment (rprospero) [inline src/sas/sasgui/perspectives/fitting/models.py]: Changed

File: src/sas/sasgui/perspectives/fitting/models.py [hunk 5]
```diff
@@ -318,12 +318,11 @@ def update(self):
...
         """
         new_plugins = self.findModels()
-        if len(new_plugins) > 0:
-            for name, plug in  new_plugins.iteritems():
-                if name not in self.stored_plugins.keys():
-                    self.stored_plugins[name] = plug
-                    self.plugins.append(plug)
-                    self.model_dictionary[name] = plug
+        if new_plugins:
+            for name, plug in  new_plugins.items():
```
andyfaff: two spaces after in

PR #75, opened 2017-04-09
TST: ensure specific readers work

andyfaff:
This PR ensures that the specific reader works for each type of file. Previously all the reading was being done by `Loader`, which would direct the file towards the specific reader. If the specific reader was failing (which was seen recently), then another reader would be tried. The read with the alternate reader would fail, indicating that it was the alternate readers fault. Here the specific reader is checked, it also checks that `Loader` directs the file towards the correct specific reader.


pkienzle:
I believe export2d.h5 is no longer part of the repo, so once you remove it from the pull request you will be able to merge it.

andyfaff:
Checks have passed, I've removed export2d.h5, so merging.

PR #77, opened 2017-04-09
Ticket #920: Log binning

ricleal:
This opens the possibility to bin the data in a logarithm of any base.
Only available for Q Slicer. The GUI needs a new implementation: E.g. logarithm base field in "Slice Parameters" to set the base of the logarithm.

Edit: Ticket #920 in Trac: http://trac.sasview.org/ticket/920


andyfaff:
Can `np.geomspace` help out with this PR?

ricleal:
Can someone merge this PR?

src/sas/sascalc/dataloader/manipulations.py
```diff
@@ -747,12 +800,14 @@ class _Sector(object):
...
     """
-    def __init__(self, r_min, r_max, phi_min=0, phi_max=2 * math.pi, nbins=20):
+
+    def __init__(self, r_min, r_max, phi_min=0, phi_max=2 * math.pi, nbins=20, base = None):
         self.r_min = r_min
         self.r_max = r_max
         self.phi_min = phi_min
         self.phi_max = phi_max
         self.nbins = nbins
+        self.base = base
```
ricleal: This was done only for scripting. Thus no input checking. I'm a great fan of _design by contract_ avoiding `if` in the core methods for performance issues. The validation is indeed better in the calls. I should have documented better though with acceptable values for the base...
It would be great if you added this in GUI with the batch slicer you are doing (the validation for the base goes here!).
I did not write a test because the current tests are based on files with linear binning. I can try to test generate some binning in scipy and cross validate. I will update the PR accordingly.

ricleal:
I added a test for the X axis. I will see tomorrow if I can do something about generating the same results with scipy for the Y axis.

ricleal:
I only could test the X axis. Since we have no files with log binning is quite difficult to test this. I tried `scipy.stats.binned_statistic` but it would need more time to have a radial integration...
Can you merge it like this?
Let me know

PR #80, opened 2017-04-10
Ticket 866

gonzalezma:
Generates an error message when the model does not return any finite value. This is the case e.g. with the Hayter-MSA model for some parameters (ticket 859). If there are only some non-finite values, then this is indicated in the console message. It seems to work as expected when the Compute button is used to calculate the model, but if a parameter is changed and then the user press 'Enter' the error window
appears twice. It seems that in this case the function _complete1D (or _complete2D) is called twice. Is there a known/good reason for this or should a new ticket be opened to investigate this behaviour?


src/sas/sasgui/perspectives/fitting/fitting.py
```diff
@@ -1732,6 +1732,12 @@ def _complete1D(self, x, y, page_id, elapsed, index, model,
             @param unsmeared_error: data error, rescaled to unsmeared model
         """
         try:
+            number_finite = np.count_nonzero(np.isfinite(y))
+            if number_finite == 0:
```
andyfaff: `if not number_finite:`
or
`if not np.any(np.isfinite(y)):`

src/sas/sasgui/perspectives/fitting/fitting.py
```diff
@@ -1732,6 +1732,12 @@ def _complete1D(self, x, y, page_id, elapsed, index, model,
...
+                logger.error("Using the present parameters the model does not return any finite value. ")
+                wx.PostEvent(self.parent, StatusEvent(status=\
```
andyfaff: please don't use line continuation. You are using parentheses, you shouldn't need to use line continuation.

src/sas/sasgui/perspectives/fitting/fitting.py
```diff
@@ -1788,8 +1794,13 @@ def _complete1D(self, x, y, page_id, elapsed, index, model,
                     self._plot_residuals(page_id=page_id, data=data, fid=fid,
                                          index=index, weight=weight)
 
-            msg = "Computation  completed!"
+            msg = "Computation completed!"
+            if number_finite != len(y):
```
andyfaff: if `y` is not 1 dimensional (perhaps not out of bounds of possibility) then this statement will be True.
Use `if number_finite != y.size`

src/sas/sasgui/perspectives/fitting/fitting.py
```diff
@@ -1816,6 +1827,12 @@ def _complete2D(self, image, data, model, page_id, elapsed, index, qmin,
         Complete get the result of modelthread and create model 2D
         that can be plot.
         """
+        number_finite = np.count_nonzero(np.isfinite(image))
+        if number_finite == 0:
```
andyfaff: `if not number_finite:`

src/sas/sasgui/perspectives/fitting/fitting.py
```diff
@@ -1875,6 +1892,9 @@ def _complete2D(self, image, data, model, page_id, elapsed, index, qmin,
                 self._plot_residuals(page_id=page_id, data=data, fid=fid,
                                       index=index, weight=weight)
         msg = "Computation  completed!"
+        if number_finite != len(image):
```
andyfaff: if `y` is not 1 dimensional (perhaps not out of bounds of possibility) then this statement will be True.
Use `if number_finite != y.size`

src/sas/sasgui/perspectives/fitting/fitting.py
```diff
@@ -1788,8 +1794,13 @@ def _complete1D(self, x, y, page_id, elapsed, index, model,
...
 
-            msg = "Computation  completed!"
+            msg = "Computation completed!"
+            if number_finite != y.size:
+                msg += ' PROBLEM: For some Q values the model returns non finite intensities!'
+                logger.error("For some Q values the model returns non finite intensities.")
+            
             wx.PostEvent(self.parent, StatusEvent(status=msg, type="stop"))
+            
         except:
```
pkienzle: No need for try/accept block in

    try:
        ... code ...
    except:
        raise

src/sas/sasgui/perspectives/fitting/models.py
```diff
@@ -153,7 +153,7 @@ def compile_file(dir):
     """
     try:
         import compileall
-        compileall.compile_dir(dir=dir, ddir=dir, force=1,
+        compileall.compile_dir(dir=dir, ddir=dir, force=0,
                                quiet=report_problem)
```
pkienzle: (1) why is a forced recompile needed?

(2) why is this patch performing the change?  It seems to go beyond the bounds of checking model return values,.

(3) looking at models.py, the call to compile_file() is commented out, and grep does not show other instances, so this appears to be dead code.  Still shouldn't touch it in this patch.  There are other tickets for improving handling of plugin models which would be more appropriate.

wpotrzebowski:
ready for testing on Linux

wpotrzebowski:
ready for testing on Linux

src/sas/sasgui/perspectives/fitting/models.py
gonzalezma: Agree. This was completely unintended (an old commit to models.py that I don't know how it ended here). Now went back to the original version of fitting.py in order to clear the plot (show zeros) when the model does not return any finite value. With this the only change is that the number of finite values returned by the model is stored and used at the end of the _complete1d and _complete2d functions to show the appropriate messages in the screen.

gonzalezma:
Back to a version of fitting.py closer to the initial one, where nans are converted using nan_to_num and thus sasview shows an empty plot if the model does not return any finite value. Then the only change is that the number of finite values returned by the model is stored and used at the end of the _complete1d and _complete2d functions to show the appropriate messages in the screen.

PR #82, opened 2017-04-10
Ticket 467: Apply a slicer to multiple data files through the slicer parameters panel.

krzywon:
Adds the ability to directly apply sector averaging, annular averaging, box sum X averaging and box sum Y averaging to multiple loaded data sets through the slicer parameters panel. The panel will also save the sliced data (in XML format), load the saved data and send it to individual fits or batch fitting.
I also updated the box sum slicer to use some of the code repeated between the two panels, but did not apply the batch slicing to it. The two panels are nearly the same and could have a common parent, but I did not go that far, because it is not (yet) needed and can wait for the new Qt GUI.

This works on trac ticket 467 - batch slicing.


krzywon:
Closing until I can resolve all of the master merge issues.

PR #83, opened 2017-04-10
Improve model reloading bug

rprospero:
There was an awkward bug where, if a plugin model was modified (so that the .py and the .pyc were out of sync), it would cause an error message until the user reloaded the models.

### DIFF
--- a/src/sas/sasgui/perspectives/fitting/fitpage.py
+++ b/src/sas/sasgui/perspectives/fitting/fitpage.py
@@ -1238,7 +1238,6 @@ class FitPage(BasicPage):
             custom_model = CUSTOM_MODEL
             mod_cat = self.categorybox.GetStringSelection()
             if mod_cat == custom_model:
-                ################This is where the model dies###############
                 temp_id = self.model.id
                 temp = self.parent.update_model_list()
                 for v in self.parent.model_dictionary.values():
--- a/src/sas/sasgui/perspectives/fitting/fitpage.py
+++ b/src/sas/sasgui/perspectives/fitting/fitpage.py
@@ -1235,9 +1235,8 @@ class FitPage(BasicPage):
             new_event = PageInfoEvent(page=self)
             wx.PostEvent(self.parent, new_event)
             # update list of plugins if new plugin is available
-            custom_model = CUSTOM_MODEL
             mod_cat = self.categorybox.GetStringSelection()
-            if mod_cat == custom_model:
+            if mod_cat == CUSTOM_MODEL:
                 temp_id = self.model.id
                 temp = self.parent.update_model_list()
                 for v in self.parent.model_dictionary.values():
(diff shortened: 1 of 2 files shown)


src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -1235,10 +1235,13 @@ def _on_select_model(self, event=None):
...
             # update list of plugins if new plugin is available
-            custom_model = CUSTOM_MODEL
             mod_cat = self.categorybox.GetStringSelection()
-            if mod_cat == custom_model:
+            if mod_cat == CUSTOM_MODEL:
+                temp_id = self.model.id
                 temp = self.parent.update_model_list()
+                for v in self.parent.model_dictionary.values():
+                    if v.id == temp_id:
+                        self.model = v()
```
krzywon: This seems like a single model is desired. A break here would exit out of the for loop as soon as it is found.

src/sas/sasgui/perspectives/fitting/models.py
```diff
@@ -259,7 +259,9 @@ def findModels(self):
         """
         temp = {}
         if self.is_changed():
-            return  _findModels(dir)
+            temp =  _findModels(dir)
+            self.last_time_dir_modified = time.time()
```
krzywon: Is this change necessary? Your changes to 311-313 should already cover the last time change.

src/sas/sasgui/perspectives/fitting/models.py
```diff
@@ -317,13 +319,13 @@ def update(self):
...
-            for name, plug in  new_plugins.iteritems():
-                if name not in self.stored_plugins.keys():
-                    self.stored_plugins[name] = plug
-                    self.plugins.append(plug)
-                    self.model_dictionary[name] = plug
+        if new_plugins:
+            for name, plug in  new_plugins.items():
+                self.stored_plugins[name] = plug
+                self.plugins.append(plug)
+                self.model_dictionary[name] = plug
```
krzywon: import CUSTOM_MODELS from fitpage.py and use it in 329.

src/sas/sasgui/perspectives/fitting/models.py
rprospero: Unfortunately, this change is critically necessary.  

When _findModels reads a ".py" file, it compiles it into a ".pyc" file.  This changes the last last modified time for the directory.  The actual variable, however, is not updated, so SasView was assuming that the directory had changed since its last read and was reloading again.

src/sas/sasgui/perspectives/fitting/models.py
rprospero: changed

src/sas/sasgui/perspectives/fitting/fitpage.py
rprospero: changed

PR #86, opened 2017-04-11
URGENT: Logging conf file working! This was breaking master.

ricleal:
This is finally the `logging.ini` working.
The file in production is packaged in `sas.sasview`.
In the future we have to put all the configuration files in `Resources` (OSx). I haven't figured it out for windows...


run.py
```diff
@@ -13,29 +14,16 @@
 the given module or script.
 """
 
-import os
-import sys
 import imp
```
wpotrzebowski: Are logging modules used at all?

ricleal: Nope.. They were forgotten there. Just run `run.py` without them and there were no issues... Shall I push the changes?

wpotrzebowski: There are also some other things pylint complains about, so you may want to fix some of this in one go.

PR #88, opened 2017-04-11
Remove all files dynamically generated during unit testing.

krzywon:
Any unit test that dynamically generates a file or files will delete the file(s) once the test is finished.


test/sasdataloader/test/utest_cansas.py
```diff
@@ -104,20 +101,18 @@ def _check_data(self, data):
...
     def test_cansas_xml(self):
-        filename = "isis_1_1_write_test.xml"
         xmlreader = XMLreader(self.isis_1_1, self.schema_1_1)
         valid = xmlreader.validate_xml()
         xmlreader.set_processing_instructions()
         self.assertTrue(valid)
         fo = open(self.isis_1_1)
-        str = fo.read()
+        fo.read()
+        fo.close()
```
pkienzle: Should be:

    with open(self.isis_1_1) as fo:
        text = fo.read()

Neither matter in this case because it is a test, but for long running programs it is better to use the with ... construct so that the file is closed immediately even in the event of an error in reading.  Also, replacing builtins like 'str' is usually a bad idea (though min() and max() are sitting on some nice name realestate).

pkienzle: Also, why it the code reading the file but not keeping what is read?  There are easier ways to check if a file exists.

PR #92, opened 2017-05-03
Ticket 887 reorg

pkienzle:
Reorganization of the tree, moving sasview into src/sas/sasview.

The new version seems to run but I haven't done complete testing.  In particular, I did not test that installers build properly.

There are other things that could be improved:
* The configuration info that is needed by sascalc such as the plugins directory needs to be moved from sas/sasgui/__init__.py to sas/__init__.py.  
* Duplicate files sasview/local_config.py and sasgui/guiframe/config.py need to be eliminated, and the associated images.


ricleal:
I think this is a good starting point. Probably we could do better if we followed some standard python project structure. Apparently the folder `src` is not part of that
> Don't: put your source in a directory called `src` or `lib`. This makes it hard to run without installing.
What about something like that:

```
# Main code
/sasview/sas/sascalc
/sasview/sas/sasgui
/sasview/sas/__init__.py
/sasview/__init__.py
#
/tests
#
/docs
/buid_tools
# This will go into Resources (in OSx for example) and not bundle with code ?
/static/images
# this could go into ~/.sasview
/configuration/config.ini
/configuration/logging.ini
# Files in the root folder
README.md
LICENSE.md
setup.py
requirements.txt
```
Just an idea...
Shall I approve the changes for now or do you want to revise tem again?
Let me know

pkienzle:
Your proposed structure adds an additional level: sasview.sas.sascalc...  so I wouldn't do that.

The biggest stake holders are those maintaining jenkins since they have to deal with the broken installers.  I don't have a working installer setup so I have not tested the changes.  Make sure they approve the changes before applying.

I would suggest doing more radical changes in stages, particularly with so many pull requests outstanding.  For example, you could raise sas up a level if you wanted.  Also, consider renaming sas.sascalc and sas.sasgui to sas.calc and sas.gui.  If you put the project layout proposal in a shared document (e.g., a file in the repo or as a trac wiki page) and send a message on the sasview developer's list then those who care can comment on it before you do a lot of work.

There is a bit of a question with how to best handle perspectives.  The code exists in /src/sas/sascalc/corfunc /src/sas/sasgui/perspectives/corfunc and /test/corfunc/test, with gui resources and docs in /src/sas/sasgui/perspectives/corfunc/media.  Moving the last into /static/corfunc makes it even more spread out. May want to instead arrange the tree so that perspectives are in one place: /src/sas/corfunc for the calculation, with subdirectories for test, wxgui, qtgui, media and help.  Some routines at the top level sas for interacting with ini files and finding resources in the installation would be useful.

If you go for a more radical deployment approach, you can leave the media files where they are relative to the source by setting up a mini virtual environment in the install directory and installing packages into it.  This will simplify the installer code since you just need to install the packages into a virtual environment on the build machine then package up that environment when you are done.  No distinction between frozen and non-frozen installs since nothing will be stuffed into a zip file.

ricleal:
I like your suggestions. But I agree this is a radical change and probably needs to get more people involved. Your comments should serve as input for the next change.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win32

wpotrzebowski:
ready for testing on Linux

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Linux

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win32

pkienzle:
ready for testing on Win32

pkienzle:
ready for testing on OSX

pkienzle:
Merged into py3 [PR #109]

PR #95, opened 2017-08-02
Refactor File Loaders

lewisodriscoll:
All file readers have been refactored to inherit from a base class [FileReader](https://github.com/lewisodriscoll/sasview/blob/ticket-876/src/sas/sascalc/dataloader/file_reader_base_class.py) that handles the functions common to all readers. Errors when loading files are now caught correctly and an appropriate error message is displayed. This closes tickets [889](http://trac.sasview.org/ticket/889), [950](http://trac.sasview.org/ticket/950), and [849](http://trac.sasview.org/ticket/849).

If a file with an unrecognised extension is loaded, we first try to load it with the ASCII reader, if that fails we try the CanSAS XML reader, then finally the NXcanSAS reader. If all of these fail then a useful error message is displayed. This closes ticket [876](http://trac.sasview.org/ticket/876).

Support for Igor 2D (.ASC) and Nexus (.nxs) files has been removed. This closes tickets [922](http://trac.sasview.org/ticket/922) and [983](http://trac.sasview.org/ticket/983). Support for HFIR .d1d files has also been removed, as they no longer use this file format.

The ability to convert ISIS 2D ASCII data to NXcanSAS has been added to the file converter. If a user tries to load an ISIS 2D ASCII file without first converting it, and error message is shown asking them to convert it first. This closes ticket [970](http://trac.sasview.org/ticket/970).


src/sas/sascalc/file_converter/ascii2d_loader.py
```diff
@@ -0,0 +1,143 @@
...
+            num_qs = int(all_lines[current_line].strip())
+            current_line += 1
+            current_line, qy = _load_points(all_lines, current_line, num_qs)
+        except ValueError as e:
+            err_msg = "File incorrectly formatted.\n"
+            if str(e).find('broadcast') != -1:
+                err_msg += "Incorrect number of q data points provided. "
+                err_msg += "Expected {}.".format(num_qs)
+            elif str(e).find('invalid literal') != -1:
+                err_msg += "Expected integer on line {}. Instead got '{}'".format(current_line + 1,
```
krzywon: This is one of a handful of lines that exceed the standard line length within the ascii2d_loader.

src/sas/sascalc/dataloader/readers/cansas_reader_HDF5.py
```diff
@@ -67,14 +63,24 @@ def read(self, filename):
...
-                self.raw_data = h5py.File(filename, 'r')
+                try:
+                    self.raw_data = h5py.File(filename, 'r')
+                except Exception as e:
+                    if extension not in self.ext:
+                        msg = "CanSAS2.0 HDF5 Reader could not load file {}".format(basename + extension)
+                        raise DefaultReaderException(msg)
+                    raise FileContentsException(e.message)
                 # Read in all child elements of top level SASroot
                 self.read_children(self.raw_data, [])
```
krzywon: Returning self.output below here is no longer needed as the parent class now handles that.

src/sas/sascalc/file_converter/ascii2d_loader.py
```diff
@@ -0,0 +1,143 @@
...
+
+            # Read qy data
+            num_qs = int(all_lines[current_line].strip())
+            current_line += 1
+            current_line, qy = _load_points(all_lines, current_line, num_qs)
+        except ValueError as e:
+            err_msg = "File incorrectly formatted.\n"
+            if str(e).find('broadcast') != -1:
+                err_msg += "Incorrect number of q data points provided. "
+                err_msg += "Expected {}.".format(num_qs)
```
lewisodriscoll: If a ValueError is raised on line 58, the error message is set on line 75, so num_qs is never used

test/sasdataloader/test/utest_generic_file_reader_class.py
```diff
@@ -0,0 +1,46 @@
+"""
```
lewisodriscoll: These tests have been added to test/sasdataloader/test/utest_extension_registry.py

PR #100, opened 2017-09-01
Ticket 767

lewisodriscoll:
The Sum/Multi models were using an old style sasview model. I've updated them to use the MixtureModel from sasmodels, which performs the correct arithmetic (specified in ticket [#767](http://trac.sasview.org/ticket/767))

See sasmodels pull request [#49](https://github.com/SasView/sasmodels/pull/49) (now merged)


lewisodriscoll:
Note: The required sasmodels code has now been merged into master

lewisodriscoll:
The documentation for custom models will probably need updating too. Now if a user creates a sum/multi model, a file that looks like the one below will be created in the plugin_model directory:
```python
from sasmodels.sasview_model import make_model_from_info
from sasmodels.core import load_model_info

model_info = load_model_info('sphere+cylinder')
model_info.name = 'MyModel'
Model = make_model_from_info(model_info)
```
By changing the string that's the argument to the `load_model_info` function, the user can change the model. Any combination of model names, `+` and `*` will work. The use of brackets isn't supported, and operations will be executed in the correct order (i.e. multiplication first, then addition). Custom model names can also be used in the model string by prepending the model name with 'custom.', e.g. `sphere+custom.MyModel`

Structure factor models can also be created using the `@` symbol, e.g. `cylinder@hardsphere`

PR #104, opened 2017-09-14
Ticket 861

lewisodriscoll:
Allows users to define structure factor models by adding `structure_factor = True` to an 'advanced' custom model.py file. Similarly, form factor models can be defined by adding `form_factor = True`

Fixes ticket [#861](http://trac.sasview.org/ticket/861)


smk78:
Just pulled & compiled and added an old style plugin model back into .sasview\plugin_models, then re-started SasView. Commit 632fda90d2ac7346a8a5e823a4e5ad1fff44b05f has fixed the problem of the fitting perspective not starting if an old plugin was present.

PR #106, opened 2017-09-17
969 about box logo

butlerpd:
Piotr fixed this a while back and closed the associated ticket #969 but failed to make a pull request it seems.  The important change is to add the dls.png logo file to the sasview/images directory (which already contains the other logo images.

The only other change was to remove all the logo files from the src/sas/sasgui/guiframe/images.  They were in both directories (except for dls).   Part of the confusion may come from the fact that when running from source in the development environment either location seems to work.  However when bundled the directories move around some and only logos appearing in the sasview/images directory are found.


butlerpd:
given the nature of this fix I am going to merge it now that it has passed Travis checks and delete the branch.

PR #108, opened 2017-09-18
Corfunc3d

pkienzle:
Calculate the 3D correlation function and IDF

Closes ticket #784


pkienzle:
Note: tests should be updated to check the new functionality.

pkienzle:
I replaced the iterated call to trapz with a call to cumtrapz.  Since the corfunc3d features are not being tested, I don't know if this breaks anything.

smk78:
Just tested this. It is throwing an error when it tries to plot the 3d function:

Plotting Error: Plottable.View: Given y and x are not of the same length

Attached file has lewis' trapz version on the left, and pkienzle cumtrapz on the right.
corfunc3d-lews-vs-paulk

butlerpd:
I don't understand this -- am I reading right that using the trapz function does not throw an error while using cumtrapz fails because of a data error (q and i arrays are of different length)?

smk78:
Yes, butler of, that is what is currently happening.

smk78:
Sorry, replace 'butler of' by 'butlerpd' in previous comment! Autocorrect!

smk78:
ready for testing on osx

smk78:
On advice from pkienzle am merging this branch now. With the caveat that unit tests for the corfunc perspective are currently disabled because of precision issues. This is noted for future action in #998

PR #110, opened 2017-09-29
Ticket 947 sasmodels docs

pkienzle:
This completes the integration of sasmodels docs into sasview.  In particular, it pulls in the sasmodels api docs and also moves the documentation for resolution, polydispersity and magnetism into sasmodels.  It cleans up (almost) all sphinx warnings.

It simplifies build_sphinx.py.  Now doc build will fail hard if sasmodels docs are not built properly rather than reporting the missing paths and continuing.  Since we do not regularly review the build output to scan for warnings, a hard failure is preferable.

* Revisits closed ticket [#555](http://trac.sasview.org/ticket/555)), 
* Closes ticket [#947](http://trac.sasview.org/ticket/947)).
* Maybe closes ticket [#554](http://trac.sasview.org/ticket/947)).

Note: This PR depends on PR [#109](https://github.com/SasView/sasmodels/pull/109), so I set the comparison base to py3 branch to simplify the diff.  We may need to modify this PR when #109 is merged and the py3 branch is deleted.


pkienzle:
ready for testing on Win32

wpotrzebowski:
ready for testing on Win64

wpotrzebowski:
ready for testing on Win64

smk78:
Just tried https://jenkins.esss.dk/sasview/job/SasView_Win7_32bit_PullRequest/32/

fitting.html and working.html should contain a link to Adding your own Models (which is currently only accessible from fitting_help.html) as well as the link to Writing a Plugin Model (already present), because these are aimed a different audiences.

Writing a Plugin Model should also cross-link to Adding your own Models. The latter already cross-links to the former.

Otherwise docs look ok!

PR #111, opened 2017-10-25
Ticket 1013 - Reset FileReader class state on every data load

krzywon:
I added a reset_state class variable to the FileReader class that empties class variable output every time a new read is initiated. I also added unit tests to be sure the proper number of data sets are being returned from each reader. This should resolve issue [1013](http://trac.sasview.org/ticket/1013).

### DIFF
--- a/src/sas/sascalc/dataloader/file_reader_base_class.py
+++ b/src/sas/sascalc/dataloader/file_reader_base_class.py
@@ -53,7 +53,6 @@ class FileReader(object):
 
         :param filepath: The full or relative path to a file to be loaded
         """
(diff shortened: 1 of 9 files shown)


pkienzle:
Can you instead have a reader only operating on a single file (e.g., by setting the path in the init call)?  The current scheme is not thread safe.

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on OSX

butlerpd:
ready for testing on OSX

butlerpd:
closes #1013

PR #114, opened 2017-10-28
ticket-873 Lower entry barrier for new developers

wpotrzebowski:
Lowering new developers entry barrier by enabling installations with anaconda. 
The change to code is non-invasive (3 yml files added). 
Instructions can be found at http://trac.sasview.org/wiki/CondaDevSetup. If okayed, should be linked to main page


pkienzle:
1) Fails on mac os x with 
```
$ conda env create -f sasview-env-build_osx.yml
Using Anaconda Cloud api site https://api.anaconda.org
Error: Can't process without a name
sasview-env-build_osx.yml is not a valid yaml file.
Environment with requierements.txt file needs a name
```

2) the line ```prefix: /Users/sasview/miniconda2/envs/sasview``` is suspicious because the user was not instructed to create a sasview account (nor would they want to), and it contains the miniconda2 path when the user was instructed to install anaconda, and keep the default suggested path (which I've seen as ~/anaconda and ~/anacoda2 (and maybe ~/anaconda3?).

wpotrzebowski:
I would assume this is due to local conda/enviorment setup. We've been deploying it on fresh vms and works perfectly fine, so it does on mine laptop. Works also for @ajj.  
@pkienzle: What osx and conda version do you have?

butlerpd:
pretty good but still needs tweaking on Windows as well.  I did slightly update the wiki instructions for step 2 regarding getting git if you don't already have it.  Two main issues:

1. Similar I think to comment from @pkienzle regarding accepting defaults and paths.  The Windows default is to install for a single user as opposed to everyone (which requires admin access to the computer).  The problem is that if you do so conda is **not put into the path** and therefore the next instruction fails with "conda is not a recognized command."
2.  python run.py fails with a ton f gibberish which will totally confuse many people (such as myself and Richard) but which basically come from the fact that we don't have a gcc compiler in the path (or in our case any at all since I was using the NIST loaner which is a clean machine)

Re the compiler was wondering about using TinyCC since it is in the repo and is bundled with the installer version...   Am thinking that for Windows maybe what we need is a batch file on top that checks for things like git and gcc compilers and conda being in the path and takes care of those as necessary before running the yml file through conda?  Will do my own OSX check n a cloud VM as well.

In the short term we may want to simply adjust the instructions.  This I think would still be much better than what we have.  In the long term though it would be nice to simplify more

ajj:
@pkienzle I've removed the prefix line - the version that worked on my machine did not have that line.

wpotrzebowski:
I see. I've missed this one when cleaning up yml files.

pkienzle:
Re: tinycc, apparently you can build python modules with tinycc
(https://stackoverflow.com/questions/743044/trying-to-embed-python-into-tinycc-says-python-symbols-are-undefined).

Note that calculator/c_extensions/sld2i.cpp will need to be converted from C++ to C in order to use tinycc.  It is a C89 compiler with some C99 support, not a C++ compiler.  Shouldn't be too hard: sld2i already uses function calls for complex arithmetic, so it is just a matter of passing a structure to the methods and dereferencing it for each attribute.

pkienzle:
The conda env create command fails if I try running it from within a conda virtual environment, but not when I run it from outside.  ```conda``` refers to the same executable path in both cases.  May want to throw that note in the docs.

pkienzle:
Tested and working on mac.  

Last concern: we now have three more places where version numbers on packages have to be changed, and we will have to update them whenever anaconda decides to drop support for older versions of the package.  Maybe use version comparisons, such as numpy>=1.1?  The conda infrastructure should support this.  But then users may be getting failure because of version skew.

Going back to an earlier proposal, can we use the distributed application as the python library for sasview developers?  It already contains all the third party packages we need, and in the correct versions.  Just need to provide a run-like script in the installed path that can use the code from the repo for the app rather than the code in the installed path.  If we ship the app as an installed python environment then we get this for free, and we simplify our build in the process since we no longer will need to look in different places for the supporting data files depending on where we are running from.

wpotrzebowski:
(1) I've made a few changes to http://trac.sasview.org/wiki/CondaDevSetup, so it explains how to install windows compiler. Ideally everything is compiled with tinycc but as pointed out sld2i needs to be rewritten either to c or python

(2) Regarding keeping minimum requiments in a few different places it is our ambition at DMSC to minimize number of files as well as list of packages (e.g. in yml file). 

(3) We've been also discussing using conda constructor https://anaconda.org/anaconda/constructor for producing a single installation script. We will need however convert pip packages to conda. There is solution in place but requires some time

(4) Lastly, we should treat this a first approach towards lowering entry barrier for new devlopers.

pkienzle:
Re (1): sldi is now in C.  The conversion was quick, but fixing the bug from my Py2 to Py3 conversion took quite a bit longer.

wpotrzebowski:
Great! I need to test entire procedure now but should work with tinycc only. In any event will wait for pull request #127 to be merged before merging this one.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win64

pkienzle:
```sasview-env-build_win.yml``` does not list ```tinycc==1.1``` as a dependency.   Should it be added to ```-pip:``` section?

wpotrzebowski:
We build tinycc from repository (at least accroding to instrcutions at http://trac.sasview.org/wiki/CondaDevSetup) and therefore doesn't need to be specified in yml file.

pkienzle:
If you add tinycc to the yml file for windows then you can drop a few lines from the install instructions.   Furthermore, if we can test for ```os.name='nt'``` and add the ```--compiler=tinycc``` option in ```run.py``` we can drop the whole section.  

BTW, the dev instructions say ```--compile=tinycc``` instead of ```---compiler=tinycc```.

I added a change to run.py so that it always runs build.  Otherwise, the instructions should have noted that one should run ```python setup.py build update``` after changing the C code to update the compiled extensions.

pkienzle:
Okay, I added the ```--compiler=tinycc``` to ```run.py``` for windows.  Unfortunately I can't test run.py without pulling #127 since this branch is still using C++ code for some of the extensions.  Need to make sure all the C extensions are producing the correct results under tinycc before we pull it.

pkienzle:
I moved the tinycc build fiddling to #127 to remove future merge difficulties.  

You should still update the .yml file to pip install tinycc==1.1 and update the dev docs to remove the explicit windows build instructions.

wpotrzebowski:
I've linked related wiki page (http://trac.sasview.org/wiki/CondaDevSetup) to developers notes: http://trac.sasview.org/wiki/DevNotes

It should be ok to merge now

pkienzle:
Somebody needs to test the instructions on a empty machine to see that they work.  Paul has offered to do so.

butlerpd:
spun up a blank windows machine on paperspace.com and followed instructions and it mostly works as advertised.

The only caveat is that if you follow directions and accept Anaconda install defaults it installs for single use and not "everyone" which requires admin privileges.  Doing so means you ''cannot'' run the conda command from an arbitrary command window.  You need to follow the instructions (that clip by wayyy too fast) during anaconda install and go to the start menu/anaconda to start the conda command window.

The wiki instructions should be tweaked to make this clear otherwise windows passes

Alternatively of course you can use the full path to the anaconda command (not appropriate to add to instructions).  Also you could just make sure to install the sasview directory under the appropriate conda folder (also not to suggest).  Of course if you change the option to install for all users of the computer everything works as instructed.

Will do same by spinning up a mac

wpotrzebowski:
Clarified instructions on http://trac.sasview.org/wiki/CondaDevSetup for Windiws.
@butlerpd: does mac work you?

butlerpd:
Turns out to be harder to spin up a clean mac for me.  Will try that later but seemed to work ok on a "dirty" mac and as mentioned is a simple change to add yml files. Since the instructions are already public lets go ahead and merge.

RichardHeenan:
(A) On Windows7, on my laptop the wiki instructions below work  from git bash using
Source activate sasview
Which then says
running build_ext
Compiling with tinycc (64bit=True)

Then it runs sasview seemingly OK, but If I do

source activate sasviewqt

I get errors, starting with

$ source activate sasviewqt
(sasviewqt)
MINGW64 /c/stfc/sasview4/sasview (master)
$ python run.py
12:12:24 - DEBUG : __main__: 185: Starting SASVIEW in debug mode.
running build
running build_py
package init file 'src\sas\sascalc\dataloader\readers\schema\__init__.py' not found (or not a regular file)
running build_ext
error: don't know how to compile C/C++ code on platform 'nt' with 'tinycc' compiler

Is this supposed to work for the new qt gui , or do I need to set up something else first? I don’t think that I need to pull a different branch any more or have a different yml file????

(B) So I tried the new yml on my Windows 7 desktop pc (where I have also only ever previously built 32 bit sasview) , but this fails to even create a sasview environment

c:\sasview42\sasview\build_tools\conda\ymls>conda env create -f sasview-env-build_win.yml
Fetching package metadata ...................

NoPackagesFoundError: Package missing in current win-32 channels:
  - packaging 16.8 py27hae1a450_1

I tried from cmd as well as from git bash.

Anyone got any ideas re (A) or (B) ?

Thank you
Richard

(PS note that the wiki page below does not yet mention that you have to do
python setup.py docs
To build the manual. )

PR #115, opened 2017-10-28
Ess dmax

rprospero:
The main files that you'll care about are

src/sas/qtgui/Perspectives/PrInversion/UI/dmax.ui
src/sas/qtgui/Perspectives/PrInversion/dmax.py

the other files are just line ending fixes or minor changes to get the plotting system to behave itself.

The only real import that you should need is

from sas.qtgui.Perspective.PrInversion.dmax import DmaxWindow

The DmaxWindow object expects three arguments:  the pr_state object from sascalc, the number of functions, and, optionally, the window's parent.  After that point, it should be completely self contained.


krzywon:
This should play well with the P(r) perspective I've been building. I already have the Invertor() and nfunc as class variables. If there is anything I find needs changing, I'll either take care of it or contact you with questions.

PR #116, opened 2017-10-29
consistent $-math expression for sasmodels and sasview, now allowing 1-$\sigma$

pkienzle:
The new orientation docs use ```$x$-$y$-$z$```, etc., instead of ```$x$\ -\ $y$\ -\ $z$```, so the sasmodels docs that are included in sasview will not render correctly without this patch.

### DIFF
--- a/docs/sphinx-docs/source/_extensions/dollarmath.py
+++ b/docs/sphinx-docs/source/_extensions/dollarmath.py
@@ -10,7 +10,7 @@ semicolon).
 """
 
 import re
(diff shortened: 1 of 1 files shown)


wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Linux

wpotrzebowski:
ready for testing on Win64

wpotrzebowski:
It looks good to me. 

There are some latex warnings in output when pdf is created, e.g.:
"""
LaTeX Warning: Hyper reference `user/sasgui/perspectives/fitting/magnetism/magn
etism:magnetism' on page 14 undefined on input line 1273.
"""
but thery seem to have no consequences in produced output. 

OSX build is failing but this due to test_core_shell_parallelepiped

PR #117, opened 2017-10-29
Ticket1025

gonzalezma:
Just going back to the previous code for m_list in order to restore sum/multiply models. It seems that in the commit 'clean up plugin-model handling code' model was changed to model.name in the line:
m_list = [model + ".py" for model in self.model_list], but as self.model_list is a list of strings, this failed.
I assume this was a just mistake, so going back to original code.


src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -1732,10 +1732,7 @@ def _set_fun_box_list(self, fun_box):
         # build function (combo)box
         ind = 0
         while(ind < len(list)):
```
krzywon: Taking this one step further could eliminate the need for indexing:
`for item in list:
    fun_box.Append(item)
`

gonzalezma: True, although I don't know why fitpage.py and basepage.py appear here. Only the single change in model_editor.py is needed to restore the functionality for sum/multiply. The changes in fitpage and basepage are part of another branch that I created (ticket1007) which tries to solve the problem with the spherical_sld model (this is not yet finished, so no pull request yet).

krzywon: Looking at the commits, you may have accidentally branched off of your 1007 branch instead of master. The changes to fitpage and basepage should either be removed or the work for 1007 can be added here when ready.

PR #120, opened 2017-11-01
Fix location of image in sasview-api docs

llimeht:
Updating the location of the BatchGridClass.png image permits sphinx to generate the PDF once more. (A slightly prettier relative path might also be possible...)

### DIFF
--- a/src/sas/sasgui/guiframe/data_processor.py
+++ b/src/sas/sasgui/guiframe/data_processor.py
@@ -13,7 +13,7 @@ The organization of the classes goes as:
 
 .. note:: Path to image is: /sasview/src/sas/sasgui/guiframe/media/BatchGridClassLayout.png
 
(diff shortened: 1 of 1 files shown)


pkienzle:
I'm not seeing the problem with the latex builds, and the path seems correct for the location of the files within the source-temp tree created by the build_sphinx.py.  Please explain why you need this patch a little better, or close this pull request without merging.

Note that for building the docs, ```make``` does not work.  You need to use ```build_sphinx.py```.  

Note also that the docs incorporate some of the docs for sasmodels and bumps, so they need to be in sister directories when the build starts, with docs already built for those packages. I can see this causing difficulty for your Debian build if you are treating sasmodels, bumps and periodictable as separate debian packages (sorry!) but we would otherwise not be able to build a coherent set of documentation for standalone sasview.

llimeht:
This looks like it was a side-effect of trying to deal with the difficulties of stitching together the 3 different documents. Sorry for the noise.

PR #121, opened 2017-11-02
Fix typo in P(r) peaks description

llimeht:
One of Debian's QA tools (`lintian`) runs a checker for common typos and spotted this one: 

> distrubution


llimeht:
(also lots of additional ones now added)

PR #122, opened 2017-11-02
Fix instance vs class variables in FileReader

llimeht:
Class variables get inherited and shared between different instantiations
of the class and this means that the current usage of `TestFileReader` in
`utest_generic_file_reader_class.py` has side-effects that would cause
subsequent tests to fail.

(See an upcoming PR for why being able to run lots of tests at once is useful)

### DIFF
--- a/src/sas/sascalc/dataloader/file_reader_base_class.py
+++ b/src/sas/sascalc/dataloader/file_reader_base_class.py
@@ -26,12 +26,6 @@ else:
         return s.decode() if isinstance(s, bytes) else s
 
 class FileReader(object):
-    # List of Data1D and Data2D objects to be sent back to data_loader
-    output = []
-    # Current plottable_(1D/2D) object being loaded in
-    current_dataset = None
-    # Current DataInfo object being loaded in
-    current_datainfo = None
     # String to describe the type of data this reader can load
     type_name = "ASCII"
     # Wildcards to display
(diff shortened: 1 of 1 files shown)


llimeht:
It seems unlikely that `current_dataset`, `current_datainfo` or `f_open` should be shared between readers either. They sound like (and are used like) instance variables not class variables. Shadowing a class variable with an instance variable by redefining `output` definitely doesn't feel like the right thing to do either. If the goal is make life easier for users of the API and providing hints to IDEs, then mixing class and instance variables isn't the right thing to do. A decent editor can find `self.foo` in the constructor and will also learn that it is supposed to be an instance variable at the same time (to be accessed via `self.foo` and not `MyClass.foo`).

Type annotations are a nice next addition to the code although I don't think I can infer the types myself from the code that is there to write them. (All I did was hit tab, after all!) I'll look at that later.

pkienzle:
I'm not suggesting that class attributes be used for instance information, only for non-mutable default values and as a place-holder for documentation and type information.  The code doesn't care if ```foo``` is the class default or an instance specific value when you write ```self.foo```.

You can get  the same effect by moving the declaration to be within __init__ (though the default value will be missing from the docs), so the decision is largely stylistic.  Since the style guide [http://trac.sasview.org/wiki/DevNotes/DevGuide/CodingRules] doesn't say anything, I guess that leaves it up to the author.  I switched to class level declarations when learning mypy and find it convenient since it puts all public attributes at the top of the class definition rather than spreading them throughout the initialization code, but YMMV.

Regarding type annotation, this is aspirational.  At one point sasmodels was mostly annotated, but I never went so far as enforcing it with a mypy check as part of the build, so I'm sure it has degraded a bit.  We have not even begun to do so for sasview.

PR #125, opened 2017-11-03
translate equation images back into latex for sas_calculator_help

pkienzle:


pkienzle:
Note: requires dollarmath pr #116 in order to handle ```$x$-$y$-$z$``` axes specification.

smk78:
This looks fine _except_ that line 110 in sas_calculator_help which now says

+    M_{\perp y'} &= M_{0q_y}\cos\theta_\text{up} + M_{0q_x}\sin\theta_\text{up} \\

I think, looking at the original html, should say

+    M_{\perp y'} &= M_{0q_y}\cos\theta_\text{up} - M_{0q_x}\sin\theta_\text{up} \\

PR #129, opened 2017-11-29
remove infinite loop from sesans reader

pkienzle:


rprospero:
Does self.nextline() return None at the end of the file?  I ask because the current version of line 48 of sesans_reader.py

    while line and not line.startswith("BEGIN_DATA"):

will kill the loop if there's a blank line in the file, which is perfectly legal.  I'm thinking that the desired behaviour is

    while line is not None and not line.startswith("BEGIN_DATA"):

pkienzle:
nextline() returns "" at the end of file and "\n" for a blank line.

PR #131, opened 2017-11-30
Fixing ticket1007 and hopefully other similar sources of error

gonzalezma:
Tested for spherical_sld. I corrected also the code in the other places where fun_list is used, but could not test. If Paul K thinks is OK, then my 'to fix' comments in the code should be deleted.


src/sas/sasgui/perspectives/fitting/basepage.py
```diff
@@ -1642,7 +1642,13 @@ def _reset_parameters_state(self, listtorestore, statelist):
...
                         if item_page[2].__class__.__name__ == "ComboBox":
                             if item_page_info[2] in self.model.fun_list:
-                                fun_val = self.model.fun_list[item_page_info[2]]
+                                # to fix: fun_list is not a dictionary, but a tuple
+                                # so the following line (commented) will fail
+                                # fun_val = self.model.fun_list[item_page_info[2]]
+                                # I guess the following should work, but 
+                                # could not test as I don't know when this
+                                # part is used by SasView.
+                                fun_val = self.model.fun_list.index(item_page_info[2])
```
pkienzle: Try saving a project with spherical_sld (several layers with different functions in each layer) and see if it can be reloaded.  

Try fitting and see if the function boxes remain unchanged after the fit (both the value and the list of available functions).  

Try creating a plugin model with a choice list and see that you can edit the plugin without the GUI doing silly things.  E.g.:
```
parameters = [
   ["option", "", 0, [["option 0", "option 1", "option 2"]], "", "function selection"],
]

def Iq(q, option):
    return q**option
```

src/sas/sasgui/perspectives/fitting/basepage.py
```diff
@@ -1686,7 +1692,12 @@ def _reset_strparam_state(self, listtorestore, statelist):
                 value = item_page_info[2]
                 selection = value
                 if value in self.model.fun_list:
-                    selection = self.model.fun_list[value]
+                    # to fix: fun_list is not a dictionary, so 
+                    # I commented the following original line 
+                    #selection = self.model.fun_list[value]
+                    # and replaced by this.
+                    # I think this should work, but could not test it.
+                    selection = self.model.fun_list.index(value)
```
pkienzle: _reset_strparam_state is never referenced.  Maybe put an exception or a logging statement at the beginning while you are testing to see if it is being triggered in some other way: 
```
    logger.error("calling _reset_strparam_state")
    raise Exception("unused function")
```

src/sas/sasgui/perspectives/fitting/basepage.py
```diff
@@ -3364,7 +3375,13 @@ def _get_paste_helper(self, param, orient_param, content):
...
                     if item[2].__class__.__name__ == "ComboBox":
                         if content[name][1] in self.model.fun_list:
-                            fun_val = self.model.fun_list[content[name][1]]
+                            # to fix: fun_list is not a dictionary, but a tuple
+                            # so the following line (commented) will fail
+                            # fun_val = self.model.fun_list[content[name][1]]
+                            # I guess the following should work, but 
+                            # could not test as I don't know when this
+                            # part is used by SasView.
+                            fun_val = self.model.fun_list.index(content[name][1])
```
pkienzle: Try copying and pasting parameters for spherical_sld

src/sas/sasgui/perspectives/fitting/basepage.py
gonzalezma: Tests OK. I could fit, change repeatedly the number of shells, save and load and create a plugin model with a combo box, and did not observe any anomalous behaviour.

src/sas/sasgui/perspectives/fitting/basepage.py
gonzalezma: Copying and pasting standard parameters works OK in spherical_sld model, while for the combo box the copy/paste is not really useful --> Should this  if item[2].__class__.__name__ == "ComboBox": block be removed?

gonzalezma:
Manual tests playing with the combo boxes (fitting/saving/loading/adding or removing shells) worked fine.

PR #132, opened 2017-12-12
ticket-995 OpenCL required on Linux even if turned off in GUI

rprospero:
On Linux, if there is no OpenCL driver, then PyOpenCL will load properly, but then throw a logic error when attempting to enumerate the platforms.  This patch catches that exception and allows sasview to continue on without pyopencl.

### DIFF
--- a/src/sas/sasgui/perspectives/fitting/gpu_options.py
+++ b/src/sas/sasgui/perspectives/fitting/gpu_options.py
@@ -7,12 +7,14 @@ Created on Nov 29, 2016
 @author: wpotrzebowski
 '''
 
+import logging
 import os
 import sys
-import warnings
 import wx
 import sasmodels
 from sas.sasgui.guiframe.documentation_window import DocumentationWindow
+
+logger = logging.getLogger(__name__)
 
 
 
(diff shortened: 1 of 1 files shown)


src/sas/sasgui/perspectives/fitting/gpu_options.py
```diff
@@ -167,7 +167,13 @@ def _get_clinfo(self):
...
             platforms = cl.get_platforms()
         except ImportError:
-            warnings.warn("pyopencl import failed. Using only CPU computations")
+            warnings.warn("Unable to import the pyopencl package.  It may not \
+            have been installed.  If you wish to use OpenCL, try running pip \
+            install --user pyopencl")
+        except cl.LogicError as e:
+            warnings.warn("Unable to fetch the OpenCL platforms.  This likely \
+            means that the opencl drivers for your system are not installed.")
+            warnings.warn(e)
```
pkienzle: The spaces on the beginning of the line following the backslash will be included in the string.  You can instead do
```
    warn("part 1 "
              "part 2 "
              "part 3")
```
and the python interpreter will join them into a single string.  You can include ```+``` characters between the string parts; python does constant folding in both cases so they are compiled into a single value.

Check that the warnings are showing up in sasview logs and in the GUI.  May want to instead define the logger at the top of the file:
```
import logging
```
and use:
```
     logger.warn(e)
```

rprospero:
I've cleaned up the logging code.

PR #133, opened 2017-12-15
ticket 993

rozyczko:
Added explicit 64 bit installer directives to the INNO script generator

### DIFF
--- a/installers/installer_generator.py
+++ b/installers/installer_generator.py
@@ -9,6 +9,8 @@ import os
 import sys
 import string
 
+is_64bit = sys.maxsize > 2**32
+
 root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.insert(0, os.path.join(root, 'sasview-install', 'Lib', 'site-packages'))
 from sas.sasview import local_config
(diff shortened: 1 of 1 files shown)


rozyczko:
Ready for testing on Win64

rozyczko:
Ready for testing on Win32

krzywon:
Both the Win32 and Win64 installers now default to the x64 Program Files directory. We are going away from the 32 bit build, so long term this isn't an issue, but could lead to confusion for the remainder of the 4.x releases.

rozyczko:
Ready for testing on Win32

rozyczko:
Ready for testing on Win64

butlerpd:
LOL ... I thought we'd agreed we were not going to worry about a conditional for now? BTW does anybody even have a 32 bit machine to test that it does the right thing in that case?  Also moot I suppose if we are not going to deliver a 32 bit version of 4.2.

rozyczko:
We did agree but the change was trivial and considering we don't know whether we're releasing only 64bit on Windows or keep the 32bit installer for one more release, I decided to make it behave correctly.

butlerpd:
Now checked also on an old windows XP (32 bit) machine and it seems to work. At least it installs, loads properly and the documentation also loads. Did no real functionality testing. If we DO release a 32 bit we should test a bit more.

PR #134, opened 2017-12-19
Ticket 489: Properly handle negative dx values in the ABS reader.

krzywon:
This addresses ticket 489 and allows the NIST ABS reader to read in negative dxl values that are saved in USANS data files. The .cor extension is now associated with the ABS reader but will fall back to the XML reader if necessary.


src/sas/sascalc/dataloader/readers/abs_reader.py
```diff
@@ -196,9 +203,12 @@ def get_file_contents(self):
                     # skip it.
                     pass
 
+            # SANS Data:
             # The 6 columns are | Q (1/A) | I(Q) (1/cm) | std. dev.
             # I(Q) (1/cm) | sigmaQ | meanQ | ShadowFactor|
-            if line.count("The 6 columns") > 0:
+            # USANS Data:
+            # EMP LEVEL: <value> ; BKG LEVEL: <value>
+            if line.count("The 6 columns") > 0 or line.count("EMP LEVEL"):
```
pkienzle: Trivial point: ```line.index(target) >= 0``` will stop after the first target whereas ```line.count(target) > 0``` will continue to scan the entire line even after finding the target.  The efficiency difference is insignificant in this case since the line is short.  Even better is ```line.startswith(target)``` if target is guaranteed to be at the start of the line.

pkienzle:
Tests look good.  Should be able to merge.

butlerpd:
Ready for testing on Win32

butlerpd:
Ready for testing on Win64

butlerpd:
ready for testing on Win32

wpotrzebowski:
ready for testing on Linux

wpotrzebowski:
ready for testing on Win32

wpotrzebowski:
ready for testing on Win64

butlerpd:
functionality seems to work well. COR and ABS both handle neg dq.  *.txt does not.  ABS also handles standard SANS files correctly.  A couple of minor GUI issues noted but not relevant to this ticket so will merge this pull request since code has already been approved.

PR #135, opened 2017-12-22
Ticket 1037: Remove nan values from 1D and 2D data objects

krzywon:
The generic file reader class has been modified to coerce any nan values associated with the fittable data to 0. This resolves the issue associated with [#1037](http://trac.sasview.org/ticket/1037).


src/sas/sascalc/dataloader/file_reader_base_class.py
```diff
@@ -157,33 +158,41 @@ def sort_one_d_data(self):
...
                     data.ymax = np.max(data.y)
 
+    def sort_1d_array(self, array=[]):
+        if array.any():
+            for i in self.ind:
+                if math.isnan(array[i]):
+                    array[i] = 0
+            array = np.asarray([array[i] for i in self.ind]).astype(np.float64)
+        return array
+
```
pkienzle: 1. Don't need a default value array.  What would it mean to sort an array without passing in an array?  If you do want an default value make it a tuple rather than a list so that lint doesn't complain about using a mutable object as the default.

2. Use numpy indexing to reorder the elements rather than list comprehension or for loop.  Similarly for replacing NaN with zero.

3. Can combine the conversion from list to array with the coercion from int to float and maybe save one pass through the list, depending on how numpy implements it.

4. Is sort_1d_array part of the public api for this class?  If not, add a leading underscore.

5. The sort_1d_array method doesn't actually sort.  "_reorder_1d_array" would be a better name.

6. Do you need to keep ind around?  If not, pass it as a parameter.  Lint will then complain that this doesn't need to be a method (it doesn't reference self), so you could make the method a @staticmethod or move it out of the class as a global function.

7. If data.dx contains NaN they will be pushed to the end of dx and replaced with 0., and dx will no longer be sorted.  Is this behaviour desired?

The following code addresses 1-6.
```
@staticmethod
def _reorder_1d_array(array, ind):
    array = np.asarray(array, dtype=np.float64)
    array[np.isnan(array)] = 0
    return array[ind]
```

krzywon:
I changed the behavior to remove any data points with nans and extended the removal to 2D data. The mask option is not viable because the fitting engine does not seem to apply a mask to 1D data, at least not when I tried to do it.

pkienzle:
I added a ticket to say that masking is ignored for 1D data during fits.

butlerpd:
I am confused: If we are now removing the lines with any nans as implied by the method name (_remove_nans_in_data) why does the doc string in line 212 say

    :return: data with mask=0 for any value of nan in data .x, .y, .dx, .dy" 
and why does the comment in line 224 say

    # Set mask[i] to 0 when data.<param> is nan

What am I missing?

krzywon:
You're not missing anything. I didn't update the documentation with my last change.

butlerpd:
Ready for testing on Win64

RichardHeenan:
Have tested the 64 bit Windows build 23, of this branch, from Jenkins.
Now loads the original problem data set from Soleil, and without any knock on effects on other data sets as far I can tell from comparing fits to sets loaded before and after the problem one.
So looks fixed to me.
Richard

PR #136, opened 2018-01-09
Remove executable mode from data files

llimeht:
Some of the files in git have odd executable modes which are unnecessary and probably accidental. Having executable bits set on non-executables can confuse build systems, packages, users etc, so it seems worth tidying. (`dh_fixperms` and `dh_python2` fix most of these but aren't sure what to do with others so we have to manually fix them in building Debian packages.)


mdoucet:
I checked that everything installed on RHEL at SNS as per our teleconference conversation.

PR #139, opened 2018-01-12
Remove all instances of \text from the repo (Ticket #977)

rprospero:
Ignore the branch name.  This fixes 977 and not 997.

### DIFF
--- a/src/sas/sasgui/perspectives/calculator/media/resolution_calculator_help.rst
+++ b/src/sas/sasgui/perspectives/calculator/media/resolution_calculator_help.rst
@@ -57,7 +57,7 @@ Using the tool
    Note that the effect from the beam block/stop is ignored, so in the small $Q$
    region near the beam block/stop
 
-   [i.e., $Q < (2 \pi \cdot \w) / (d_s \cdot \lambda_\mathrm{min})$, where $w$ is the beam block diameter, $d_s$ is the sample-to-detector distance, and $\lambda_\mathrm{min}$ is the minimum wavelength.]
+   [i.e., $Q < (2 \pi \cdot w) / (d_s \cdot \lambda_\mathrm{min})$, where $w$ is the beam block diameter, $d_s$ is the sample-to-detector distance, and $\lambda_\mathrm{min}$ is the minimum wavelength.]
 
    the variance is slightly under estimated.
 
--- a/src/sas/sasgui/perspectives/calculator/media/resolution_calculator_help.rst
+++ b/src/sas/sasgui/perspectives/calculator/media/resolution_calculator_help.rst
@@ -57,7 +57,7 @@ Using the tool
    Note that the effect from the beam block/stop is ignored, so in the small $Q$
    region near the beam block/stop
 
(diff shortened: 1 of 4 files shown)


butlerpd:
I think we are confused Paul.  The ticket 977 was submitted by you (admitedly back in June) and specifically says to replace \text with \mathrm because latex and mathjax handle it differently and it causes unexpected behavior. Is that no longer true?

pkienzle:
IIRC I was fighting problems with underscores in ```\text{}``` blocks.  LaTeX wanted them rendered as ```\_``` but mathjax/katex left the rendering to html.  Use whatever looks best.

The [answer from TeX stack exchange](https://tex.stackexchange.com/a/19503) is to prefer ```\text``` for text and ```\mathrm``` for math symbols.   I guess ```\mathrm{scale}```, ```\mathrm{background}``` and ```\rho_\mathrm{core}``` would be appropriate since "scale", "background" and "core" count as symbols in this context, but "for" in ```x \text{ for } x < 3``` would count as a word.  

If you have multiple words, such as "beam block diameter", then you will need to use ```\text{beam block diameter}``` since ```\mathrm{beam block diameter}``` eats spaces.  In cases like this it might be better to introduce a symbol for the phrase, such as
```
[i.e., $Q < (2 \pi w) / (d_s \cdot \lambda_\mathrm{min})$ where $w$ is the beam block diameter and $d_s$ is the sample-to-detector distance.]
```

rprospero:
I've performed a more constrained change, keep the \text for elements that were textual, but moving to mathrm for things that appeared to be symbols.

src/sas/sasgui/perspectives/calculator/media/resolution_calculator_help.rst
```diff
@@ -57,7 +57,7 @@ Using the tool
    Note that the effect from the beam block/stop is ignored, so in the small $Q$
    region near the beam block/stop
 
-   [i.e., $Q < (2 \pi \cdot \text{beam block diameter}) / (\text{sample-to-detector distance} \cdot \lambda_\text{min})$]
+   [i.e., $Q < (2 \pi \cdot \w) / (d_s \cdot \lambda_\mathrm{min})$, where $w$ is the beam block diameter, $d_s$ is the sample-to-detector distance, and $\lambda_\mathrm{min}$ is the minimum wavelength.]
```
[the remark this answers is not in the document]
rprospero: I've added this to my latest commit.

src/sas/sasgui/perspectives/corfunc/media/corfunc_help.rst
```diff
@@ -187,8 +187,8 @@ The structural parameters extracted are:
 *   Long Period :math:`= L_p`
 *   Average Hard Block Thickness :math:`= L_c`
 *   Average Core Thickness :math:`= D_0`
-*   Average Interface Thickness :math:`\text{} = D_{tr}`
-*   Polydispersity :math:`= \Gamma_{\text{min}}/\Gamma_{\text{max}}`
+*   Average Interface Thickness :math:`\mathrm{} = D_{tr}`
```
pkienzle: Why is this one not like the others?  I suggest removing the ```\mathrm{} ``` here.

rprospero: Good point

PR #141, opened 2018-01-30
Fix loading of projects with the Uniform distribution

krzywon:
This PR fixes the issue with loading projects saved using the Uniform distribution introduced in the boltzmann sasmodels branch. The project loader was using the index of the distribution combobox to check for a distribution array. The Uniform distribution was put into the position the Array distribution used to hold causing the issue seen in [ticket 1018](http://trac.sasview.org/ticket/1018#comment:5).

This PR is tied to [sasmodels PR59](https://github.com/SasView/sasmodels/pull/59).

### DIFF
--- a/src/sas/sasgui/perspectives/fitting/basepage.py
+++ b/src/sas/sasgui/perspectives/fitting/basepage.py
@@ -2678,7 +2678,7 @@ class BasicPage(ScrolledPanel, PanelBase):
                        disp_func.type)
             except ValueError:
                 pass  # Fall through to default class
-        return POLYDISPERSITY_MODELS.keys().index('gaussian'), 'gaussian'
+        return (list(POLYDISPERSITY_MODELS).index('gaussian'), 'gaussian')
 
     def on_reset_clicked(self, event):
         """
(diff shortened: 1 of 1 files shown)


src/sas/sasgui/perspectives/fitting/basepage.py
```diff
@@ -2667,17 +2667,18 @@ def _lay_out(self):
...
 
         :param disp_function: dispersion distr. function
         """
         # Find the selection
         if disp_func is not None:
             try:
-                return POLYDISPERSITY_MODELS.values().index(disp_func.__class__)
+                index = POLYDISPERSITY_MODELS.values().index(disp_func.__class__)
+                return index, POLYDISPERSITY_MODELS.keys()[index]
             except ValueError:
```
pkienzle: Can you use ```return index, disp_func.type``` instead?

Note:```dict.values().index()``` does not work in python 3.  You could instead use the following, which will work for both:
```
index = list(POLYDISPERSITY_MODELS).index(disp_func.type)
```
This works because ```list(OrderedDict)``` is just the list of the keys, which in this case are created from ```distribution.type``` for each of the distributions in the ordered dict.

PR #142, opened 2018-02-06
Ticket 1068: Throw an explicit error message when a pixel space data set is sent to the loader.

krzywon:
An explicit error message is thrown when the data loader sees a data set read previously by the IgorReader (.asc) or NexusReader (.nxs). The error is immediately thrown and no data loading operations are attempted.

### DIFF
--- a/src/sas/sascalc/dataloader/loader.py
+++ b/src/sas/sascalc/dataloader/loader.py
@@ -54,7 +54,7 @@ class Registry(ExtensionRegistry):
         self.wildcards = ['All (*.*)|*.*']
 
         # Deprecated extensions
-        self.deprecated_extensions = ['.asc', '.ASC', '.nxs', '.NXS']
+        self.deprecated_extensions = ['.asc', '.nxs']
 
         # Creation time, for testing
         self._created = time.time()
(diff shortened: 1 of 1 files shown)


src/sas/sascalc/dataloader/loader.py
```diff
@@ -72,6 +75,16 @@ def load(self, path, format=None):
         """
         # Gets set to a string if the file has an associated reader that fails
         msg_from_reader = None
+        extlist = [ext for ext in self.deprecated_extensions if
+                   path.lower().endswith(ext)]
+        if extlist:
```
pkienzle: The code looks fine, but you could use
```
if any(path.lower().endswith(ext) for ext in self.deprecated_extensions):
   ...
```
or
```
self.deprecated_extensions = {'.asc', '.nxs'}
...
if os.path.splitext(path).lower() in self.deprecated_extensions:
    ...
```

butlerpd:
If a "bog standard" text file uses one of the special extensions or the standard xml or hdf files uses something other than xml or hdf as the extension,  the special reader is supposed to try first but if it fails then the file is supposed to be dropped through to the default readers.  So what happens if this is caught as a deprecated extension? does it still fall through? or is it stopped at the error?

krzywon:
@butlerpd If the file has an extension in the list of deprecated extensions, an error will be thrown and the file loading will not happen, regardless of the format.

PR #143, opened 2018-02-14
Ticket 1068: Loading data sets with file extensions no longer supported

krzywon:
This is a rework to how data sets are handled on loading when the file extension was once associated with a file reader that has since been removed from SasView. The data loader attempts to load the data set, and if a 1D or 2D data object is created, an error is added to that file to warn the user the loaded data might not be what they expect. All error handling and logic is performed within the sascalc. A unit test was added that checks for the error message when loading a file with the .asc extension.

I created a competing branch where all of the logic is done within sasgui, but I don't think that is the best way to handle this. To see the gui changes I made: https://github.com/SasView/sasview/compare/ticket-1068-1


butlerpd:
so what is the behavior if it fails to load? Does it still explain that this is a deprecated extension?

krzywon:
The check is now performed and the error is displayed regardless if the data is loaded.

butlerpd:
This may not be easily fixable but I note that a standard ascii file (2 to 6 columns) loads fine but the big red x starting with Error occured:; Error  - is confusing ... "warning" might be better?

If it really cannot be read it comes with Error occurred:; Loading data failed; followed by the warning which ends with "an attempt was made to load the file, but SasView cannot guarantee the accuracy of the data."  This is also confusing... why does the failed notice come so early?  At the very least maybe replace the text with "an attempt was made to load the file.  However, even if a data set is created in SasView cannot guarantee ...."

butlerpd:
Looks like we now generate a warning followed by a load followed by an error with the exact same message.  However the user only sees the error unless they look at the console.  Don' think this duplication is useful since mostly invisible. 

Also a cansas1d file fails to read if the extension is set to ASC for example (one of the deprecated extensions)  what are the default readers being tried?  This also suggests that the last sentence of the error message should add "should it be successful" as in 
"An attempt to load the file was made, but[, should it be successful,] SasView cannot guarantee the accuracy of the data.

Note that the general text seems to pick up even if the extension was changed to ASC.

krzywon:
I have fixed the issue loading XML (and NXcanSAS, indirectly) files with non-standard extensions. I updated the warning message to add the wording suggested by @butlerpd. Unit tests checking the same file contents with different file extensions are now included as well.

krzywon:
ready for testing on OSX

krzywon:
ready for testing on Win64

krzywon:
ready for testing on Linux

krzywon:
ready for testing on OSX

krzywon:
The OSX build seems to be having some issues finding the pyparsing module. It sometimes works and sometimes doesn't. I think this is ready to merge, despite the build issue.

wpotrzebowski:
Yes, pyparsing is problematic now. I am looking into this.

PR #147, opened 2018-03-10
Fix finding of test resource

llimeht:
A newly added test missed the call to `find()`, breaking the use of pytest.

(Would you be open to using pytest to run the tests on travis? Is there any functionality that `utest_sasview.py` provides that a simple call to pytest doesn't? Example of this failure being caught on travis at https://travis-ci.org/llimeht/sasview/jobs/351702734 )


krzywon:
I don't think anyone would be opposed to a test suite change, but there has never been a driving force to move away from unittest. I think the biggest hurdle to changing to pytest might be the upcoming switch to QT in the ESS_GUI branch. All of the tests for the new GUI are done using unittest.

PR #150, opened 2018-04-08
Drop incorrect version req on pyparsing

llimeht:
matplotlib has a minimum version of pyparsing ("Matplotlib requires pyparsing>=2.0.1;") that is considerably greater than and also incompatible with the 1.5.5 that pyparsing is currently pinned to. Travis currently installs 1.5.5 and 1.5.7 and 2.2.0 within the system via apt, conda and pip. 

This patch fixes up the sasview requirement so that matplotlib and sasview are co-installable, possibly addressing the pyparsing issues described in #143.

(As a general rule, avoiding `==` dependencies is really important for maintaining the ability to install the required packages. Use of `>=` to ensure a particular feature is available or a particular bug is fixed is best. Most maintainers are of dependencies will work to ensure that API compatibility is maintained, at least across minor and micro releases so `==` is overkill.)

### DIFF
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -2,7 +2,7 @@
 # Commented items are generally easier to install separately.
 pylint
 unittest-xml-reporting==1.10.0
-pyparsing==1.5.5
+pyparsing>=2.0
 html5lib==0.95
 reportlab==2.5
 lxml==2.3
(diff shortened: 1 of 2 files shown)


krzywon:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
I've updated a few packages on the build VM and it seems to be working now

PR #153, opened 2018-04-22
Fix race condition in test

llimeht:
`transform_isrunning()` returns `False` both before the calculation has started and also after the calculation has finished. The test of the transformation was only considering the latter case and so was regularly failing on the reproducible-builds.org test infrastructure. (This patch has been in place for a couple of months without failure.)

There is, perhaps, a broader discussion about whether the `transform_isrunning()` should return `False` when the calculation is yet to start (`complete` or `completed` would be a better test), but I don't know the wider implications of that suggestion throughout the code.

### DIFF
--- a/test/corfunc/test/utest_corfunc.py
+++ b/test/corfunc/test/utest_corfunc.py
@@ -73,14 +73,16 @@ class TestCalculator(unittest.TestCase):
         # Transform is performed asynchronously; give it time to run
         while True:
             time.sleep(0.001)
-            if not self.calculator.transform_isrunning():
+            if (not self.calculator.transform_isrunning() and
+                self.transformation is not None):
                 break
 
-    def transform_callback(self, transforms):
-        transform1, transform3, idf = transforms
+        transform1, transform3, idf = self.transformation
         self.assertIsNotNone(transform1)
         self.assertAlmostEqual(transform1.y[0], 1)
         self.assertAlmostEqual(transform1.y[-1], 0, 5)
+
+    def transform_callback(self, transforms):
         self.transformation = transforms
 
     def extract_params(self):


pkienzle:
Looks like self.transforms will not be set if the asserts in transform_callback are triggered.  This means that the loop will never complete.  Maybe move the assignment before the tests?

A separate question: which thread are the asserts running in?  Should they be captured in the compute thread forwarded to the test thread before raising an assert?

llimeht:
To clarify, we were seeing test errors (not test failures) at test/corfunc/test/utest_corfunc.py:87
```
params = self.calculator.extract_parameters(self.transformation[0])

or
TypeError: 'NoneType' object has no attribute '__getitem__'
```
which were because `self.transformation` was `None` still (I don't have access to the old logs any more to be precise as to which of these two TypeError messages we got). 

The reason `self.transformation is None` is that the loop can exit because `transform_isrunning()` is `False` before the calculation even starts but the completion function `completefn=self.transform_callback` has never been called and so `self.transformation` remains `None`. This is rooted in a semantic difference between `isrunning` and `iscomplete`. The latter is the desired exit condition for the loop while the former is the coded exit condition. 
```
    def transform_isrunning(self):
        if self._transform_thread is None: return False
        return self._transform_thread.isrunning()
```
This patch makes the loop in the main (test) thread continue until the completefn callback (`self.transform_callback`) has actually been called. This is ensured by testing for a side effect of that completefn calculation, namely that `self.transformation` is neither `None` nor zero-length.

The compute thread calls the callback function while everything else within the test code is in the main thread. As you say, assertion failures within `transform_callback()` would not be communicated back to unittest because there doesn't seem to be a call to `join()` where the main thread would wait until the calculation thread ends. While this is not the source of the failure we have seen (there's much stderr noise in this eventuality), this patch would mean that the assertion failure within the callback would cause an infinite loop.

Making the callback too small to fail and moving the tests from within the callback and into the calling code could work, perhaps like this:
```
    def transform(self):
        self.calculator.compute_transform(self.extrapolation, 'fourier',
            completefn=self.transform_callback)
        # Transform is performed asynchronously; give it time to run
        while True:
            time.sleep(0.001)
            if not self.calculator.transform_isrunning() and self.transformation is not None:
                break

        transform1, transform3, idf = self.transformation
        self.assertIsNotNone(transform1)
        self.assertAlmostEqual(transform1.y[0], 1)
        self.assertAlmostEqual(transform1.y[-1], 0, 5)

    def transform_callback(self, transforms):
        self.transformation = transforms
```
Also putting a reasonably generous maximum execution time on the loop might be worthwhile, calling `self.fail()` in the event that the timeout is hit.

pkienzle:
Your proposed change looks fine.  Please update the PR.

llimeht:
Updated but without putting an explicit timeout on the computation. Are you happy that the callback is now too small to fail and so failures in the compute thread can be neglected? The next step would be:

```
         self.calculator.compute_transform(self.extrapolation, 'fourier',
             completefn=self.transform_callback)
         # Transform is performed asynchronously; give it time to run
+        start_time = time.time()
+        calculation_timeout = 60  # seconds
         while True:
             time.sleep(0.001)
             if (not self.calculator.transform_isrunning() and
                 self.transformation is not None):
                 break
+            if time.time() > start_time + calculation_timeout:
+                self.fail("Calculation timed out. Did calc thread crash?")
 
         transform1, transform3, idf = self.transformation
         self.assertIsNotNone(transform1)

```

PR #156, opened 2018-06-18
Ticket 1069: Fix for orphaned error boxes

krzywon:
This fixes the issue with the orphaned error boxes that show up on the top-left of the fit page when first assigning data (not in the ticket) as well as when the polydispersity distribution is changed. Also, I made the error value read-only. They appear similar to other values, but cannot be edited now.


butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on OSX

butlerpd:
ready for testing on Win64

PR #159, opened 2018-07-25
Batch slicer documentation - 1139

krzywon:
Documentation for batch slicing added in 4.2b. A help button was added to the slicer parameters window that should link to the 2D data averaging section, but anchors are currently broken as outlined in [Trac ticket 1104](http://trac.sasview.org/ticket/1104).


pkienzle:
Just noticed that the Slicer Parameters box is modal.  This means that you can't interact with the help window while the Slicer Parameters box is displayed.  

Making the box non-modal has some good UI benefits.  E.g., you could interact with the slicer parameters from the graph as well as typing in the numbers in the box.  You could put the slicer in the Tools menu where it is easy to find rather than having to select a slicer on a graph before the user even knows the slicer box exists.  You could also make slicing available to data sets that are not currently plotted, which may be useful for batch slicing.

I've added this as [ticket 1150](http://trac.sasview.org/ticket/1150) since it is too much work for the 4.2 release, or any future wx release.

smk78:
Just to add to what @pkienzle has said, the modality of the slicer box is irritating.

PR #161, opened 2018-07-30
Ticket 1142 plugin reload

pkienzle:
Hide the module reload problem that was breaking the plugin editor tests for pure python models with global references.

We could do better by refactoring the sasview model interface.  See comments in code for details.

### DIFF
--- a/src/sas/sasgui/perspectives/calculator/pyconsole.py
+++ b/src/sas/sasgui/perspectives/calculator/pyconsole.py
@@ -27,6 +27,21 @@ def check_model(path):
     """
     Check that the model on the path can run.
     """
+    # TODO: fix model caching
+    # model_test.run_one() is directly forcing a reload of the module, but
+    # sasview_model is caching models that have already been loaded.
+    # If the sasview load happens before the test, then the module is
+    # reloaded out from under it, which causes the global variables in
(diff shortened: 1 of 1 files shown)


butlerpd:
ready for testing on Win64

butlerpd:
Failed to build on Jenkins.   ... needs investigation.

pkienzle:
ready for testing on Win64

butlerpd:
Interestingly, this seems to fix the problem with models written using the 4.2 editor but not for those written in 4.1.2.  Aside from backward compatibility this seems very strange. the error when loading the 4.1.2 version is the same as before suggesting that math is being overwritten with a null object?

`File "sas\sascalc\data_util\calcthread.pyc", line 274, in _run`
`TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'`

I will document the ticket with uploaded files

butlerpd:
NOTE; there is little difference between the new version and the old except for more comments and a couple of extra import statements ... and no documentation template - so don't understand this.

pkienzle:
Both versions failed for me, but the problem was elsewhere.  Should be fixed now.  I'll trigger a new windows build.

pkienzle:
ready for testing on Win64

PR #162, opened 2018-08-01
Beta Load Project Issue Fix

krzywon:
Saved projects hold the SasView version number in them for backward compatibility checks. The version is  put into a tuple of integers as (major, minor, point) e.g. (4, 1, 2). Adding '-beta' caused the int conversion to fail. This fix removes all non numeric characters from the version number when loading the project.

If desired, I can also remove the non-numeric characters when saving projects, but that isn't included in this PR as of yet.

### DIFF
--- a/src/sas/sascalc/fit/pagestate.py
+++ b/src/sas/sascalc/fit/pagestate.py
@@ -12,6 +12,7 @@ Class that holds a fit page state
 # copyright 2009, University of Tennessee
 ################################################################################
 import time
+import re
 import os
 import sys
 import copy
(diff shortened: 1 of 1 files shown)


butlerpd:
How do previous versions of SasView handle saved projects with a future version number in the file? Assuming it still works with maybe a warning that it may not behave as expected, removing the string from saving would be helpful.  Otherwise it seems like the stripping of the non numeric values could be done at either point - and leaving the non numeric in the save could be helpful for debugging purposes?

krzywon:
Prior to v4.0, the version number saved in projects was always 1.0 and all projects prior to then are treated equally. The converter checks the version number against a dictionary of known changes. If the saved project version number is lower than the version found in the conversion table, the model parameters are converted, otherwise loading moves on. The version number could be 99.99.99, but so long as the model hasn't changed, the project will load.

Thinking about this more, keeping the version number intact on saving would help, as you suggest, for debugging purposes.

PR #163, opened 2018-08-01
SASVIEW-971: Constraining an S(Q) param to its current value crashes the app

tcbennun:
Fixes for:

- [SASVIEW-960](https://jira.esss.lu.se/browse/SASVIEW-960) &ndash; Cannot fit any S(Q) params if using P(Q)S(Q) model
- [SASVIEW-971](https://jira.esss.lu.se/browse/SASVIEW-971) &ndash; Constraining an S(Q) param to its current value crashes the app

Beforehand, only the `model_info` structures were being updated when the user selected a S(Q); without updating the `kernel_module` object (which is really a `SasviewModel` object), the application could not access the S(Q) parameters for fitting or constraints.

----

Additionally, any parameters in the table which are not found in the kernel module's parameter list (i.e. they cannot be currently fitted &ndash; this is exclusively `radius_effective`, at the moment) are now greyed-out in the table, and cannot be checked, selected or edited. This is a temporary measure; at some point `radius_effective` will be toggle-able as either a fitting or internally calculated parameter.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
@@ -150,9 +150,10 @@ def addSimpleParametersToModel(parameters, is2D):
     return item
 
 def markParameterDisabled(model, row):
-    """Given the QModel row number, format to show it is not available for
-    fitting"""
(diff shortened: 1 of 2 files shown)


tcbennun:
Agreed - thanks!

PR #164, opened 2018-08-08
Load data for slit size calculator - 1132

krzywon:
As of 4.2-beta, all data loaded into SasView is returned as a list to allow for any number of data sets to be loaded simultaneously. The calculator perspective expected a Data1D object instead of a list, causing the issue outlined in [ticket 1132](http://trac.sasview.org/ticket/1132).


butlerpd:
ready for testing on Win64

PR #165, opened 2018-08-15
Added dependency required for pyopencl on win64 - ticket 1131

rozyczko:
Added dependency on pkg_resources to setup_exe.py in order for pyopencl to find the correct path to the include directory.

### DIFF
--- a/installers/setup_exe.py
+++ b/installers/setup_exe.py
@@ -286,6 +286,7 @@ if msvcrtdll:
 packages = [
     'matplotlib', 'scipy', 'encodings', 'comtypes', 'h5py',
     'win32com', 'xhtml2pdf', 'bumps', 'sasmodels', 'sas',
+    'pkg_resources'
     ]
 packages.extend([
     'reportlab',
--- a/src/sas/sasgui/perspectives/fitting/gpu_options.py
+++ b/src/sas/sasgui/perspectives/fitting/gpu_options.py
@@ -138,6 +138,9 @@ class GpuOptions(wx.Dialog):
         self.Bind(wx.EVT_BUTTON, self.on_help, help_btn)
 
         test_text = wx.StaticText(self, -1, "WARNING: Running tests can take a few minutes!")
+        test_text2 = wx.StaticText(self, -1, "NOTE: No test will run if No OpenCL is checked")
+        test_text.SetForegroundColour(wx.RED)
+        self.vbox.Add(test_text2, 0, wx.LEFT|wx.EXPAND|wx.ADJUST_MINSIZE, 10)
         self.vbox.Add(test_text, 0, wx.LEFT|wx.EXPAND|wx.ADJUST_MINSIZE, 10)
 
         btn_sizer = wx.BoxSizer(wx.HORIZONTAL)


rozyczko:
Ready for testing on Win64

rozyczko:
ready for testing on Win64

butlerpd:
I notice that this change also causes a test to fail (off by 5% I think).  Would this b because the tests are now using the GPU on the build machine?

For the record the test failures reported is:

`Error Message`
`5.365857206387022 != 5.62 within 2 places`

`Stacktrace`
`Traceback (most recent call last):`
`  File "utest_sld.py", line 57, in test_neutron_sld`
`    self.assertAlmostEquals(incoh, 5.62, 2)`
`AssertionError: 5.365857206387022 != 5.62 within 2 places`

Also this seems to work on GPU machine, but clicking test when "no OpenCl" is checked produces nothing.  It did produce a test in version 4.1.1.  Not sure if this is an intentional change, in which case thee should be a pop up that says "nothing to test" so people don't wait, or is it an error?

krzywon:
I believe the failing test is due to the incorrect version of periodictable. Version 1.5.0 must be used, but I just tested on my local system and this particular test failure occurs when version 1.4.1 is used. The console log says 1.5.0 is bundled with the build, but if multiple versions are installed on the build system, the lower version might be used for the tests.

wpotrzebowski:
ready for testing on Win64

wpotrzebowski:
periodictable issue has been fixed, so test doesn't fail now.

butlerpd:
After some more testing, I think the behavior may be by design.  Here are the behaviors depending on the type of windows machine:

* On machines with GPUs and OpenCl, no errors are thrown in the log file and three options are presented: OpenCl on GPU, OpenCL on CPU, and No OpenCl 
* On machines without GPU but with OpenCl drivers, no errors are thrown in the log file and two options are presented: OpenCl on CPU and No OpenCl
* On Machines without GPU and without OpenCl , the error noted previously in the ticket #1131 that states that pyopencl cannot be imported and suggesting it may not have been installed -- a bit misleading but fair enough, and only one "option" is presented which is the No OpenCl option.

In all cases however, clicking on `test` when `No OpenCl` option is clicked does nothing. This may be confusing since it did pop up some test results in the past.  However I suspect that what has changed is that the old `No OpenCl` was actually running the CPU tests?  and that now there is a new option of OpenCl on CPUs so that there is nothing different to test with `No OpenCl`?  In this case it would only be DLLs I guess which would be being tested against themselves?

If this is true we should document this new behavior in some way.  If that is the case (and this is not a new bug) then we can merge this PR and create a new ticket for the documentation.  If on the other hand this behavior is actually part of the "open Cl dialog not working" bug I think we should leave this open till it is fixed?

rozyczko:
ready for testing on Win64

rozyczko:
ready for testing on Win64

tcbennun:
This doesn't appear to fix the problem on Windows 10 64-bit; dialog still does not open. (I have a GPU and OpenCL drivers.)

butlerpd:
Indeed, the latest fiddle to the build server seems to have done the trick.  On a Windows 10 no GPU OpenCL all 134 models pass.  On my Windows 7 Enterprise with an Intell HD4000 graphics card, all 134 models also successfully pass both the GPU and CPU OpenCL testing.  Note these same machines were having some 58 models failing all tests in previous versions of this pull request build.

`No OpenCL` still does nothing.  Not clear if this is intentional but seems reasonable.  Suggest we add a line to the pop up where it gives the WARNING .. this will take a few minutes.  Something like:

`NOTE: No tests will be attempted if NO OpenCL is checked`

We can then call this fixed and merge to master

wpotrzebowski:
ready for testing on  OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
tested on OSX from pull request build and works as expected

PR #167, opened 2018-09-04
SASVIEW-966: Organise parameter table by model component

tcbennun:
These changes update the main param table:

- Subheadings for P(Q) and S(Q) are added to make the origin of each parameter clear
- radius_effective is no longer shown (if using sasmodels master) because it is not manipulable yet
- Duplicate parameter names from different models (i.e. if P(Q) and S(Q) each have a parameter with the same name) are no longer displayed with the conjured suffix '_S' (the subheadings make it clear which is which)


tcbennun:
Need to ensure shell parameters are placed correctly.

PR #168, opened 2018-09-04
Ticket 1164 slit calculator docs

pkienzle:
Fix the problem with sphinx superscript and substitution (Ticket [1164](http://trac.sasview.org/ticket/1164)) in an emphasized environment by removing the emphasis.

Not the best solution.  May want to try the most recent version of sphinx under python 3 to see if this problem has already been fixed upstream.  

Other option is to turn off emphasis for the superscript and the substitution.


smk78:
closes #1164

PR #169, opened 2018-09-06
ModelThread: passing calc results by dict instead of tuple

ibressler:
- prevents error when unpacking the tuple in completeXY methods later
- and makes it easier to get just a single value from the result
- related to 2df558ebcc19b99ebc1900efaa872a981b31a67d
- tested locally on my box with and without twisted enabled


ibressler:
ok, pushed it directly to ESS_GUI to speed up the process

tcbennun:
Already pushed directly, so closing

PR #170, opened 2018-09-06
DataExplorerWindow.displayFile() using updatePlot() avoids duplicate code

ibressler:
just a tiny mod which might be useful while figuring out polydispersity plot


ibressler:
ok, pushed it directly to ESS_GUI to speed up the process

tcbennun:
Already pushed directly, so closing

PR #172, opened 2018-09-06
Ticket 1166: Fix for report generation issue

krzywon:
This is a fix for the issue outlined in [1166](http://trac.sasview.org/ticket/1166). The issue stems from a memory clash when new reports are generated. Images of plots are stored in virtual memory when generating reports, but the same memory space(s) and file name(s) are used without ensuring the space is free prior to creating new reports and images. Release 4.1.2 and earlier also experience issues, but not in such obvious ways as to the beta release. In 4.1.2, when opening multiple reports without closing any of them, the image(s) from the most recently opened report are saved in the PDFs, regardless of which report is saved, because the old images are wiped from memory and replaced with the latest ones.

To fix this, I moved the file handler from the fitting perspective to its own class in guiframe.  ensure any new image stored in memory has a unique name, and attempt to free up memory when reports are closed. The file handler is now used by the fitting and invariant perspectives.

I have tested this locally, with multiple reports open simultaneously and all save appropriately, for the fitting and invariant perspective. No unit tests were added.

This is a semi-large change for a beta, but report generation is quite broken.

refs #1166

### DIFF
--- a/src/sas/sasgui/perspectives/fitting/report_dialog.py
+++ b/src/sas/sasgui/perspectives/fitting/report_dialog.py
@@ -73,14 +73,17 @@ class ReportDialog(BaseReportDialog):
 
         # save figures
         pictures = []
-        for num in range(self.nimages):
-            pic_name = basename + '_img%s.png' % num
+        i = 0
+        for url in self.fig_urls:
+            append_me = url.split(':')[1]
+            pic_name = basename + '_{}'.format(append_me)
             # save the image for use with pdf writer
-            self.report_list[2][num].savefig(pic_name)
+            self.report_list[2][i].savefig(pic_name)
             pictures.append(pic_name)
+            i += 1
 
         # translate png references int html from in-memory name to on-disk name
-        html = self.report_html.replace("memory:img_fit", basename+'_img')
+        html = self.report_html.replace("memory:", basename+'_')
 
         #set file extensions
         img_ext = []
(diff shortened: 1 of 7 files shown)


src/sas/sasgui/perspectives/fitting/report_dialog.py
```diff
@@ -73,14 +73,17 @@ def onSave(self, event=None):
 
         # save figures
         pictures = []
```
pkienzle: Can use enumerate(sequence) and [list comprehension] to remove a bit of boilerplate:
```
    # save the images for use with pdf writer
    pictures = ['_'.join((basename, url.split(':')[1]) for url in self.fig_urls]
    for i, pic in in enumerate(pictures):
       self.report_list[2][i].savefig(pic)
```
Maybe the following is a bit easier to read:
```
    # save the images for use with pdf writer
    def pic_name(basename, url):
        target = url.split(':')[1]
        return '_'.join((basename, target))
    pictures = [pic_name(basename, url) for url in self.fig_urls]
    for i, pic in in enumerate(pictures):
       self.report_list[2][i].savefig(pic)
```

src/sas/sasgui/perspectives/invariant/invariant_panel.py
```diff
@@ -782,7 +783,9 @@ def on_preview(self, event=None):
         report_text_str = self.state.__str__()
         report_img = self.state.image
         report_list = [report_html_str, report_text_str, report_img]
```
pkienzle: May want to make sure that ReportImageHandler has been initialized, either using:
```
ReportImageHandler.check_for_empty_instance()
imgRAM = ReportImageHandler.instance.img_holder
```
or with module as singleton as above:
```
imgRAM = report_image_handler.get_handler().img_holder
```

You probably don't need to since the code is only called after a report is generated(?), but if it is easy then code defensively.

PR #173, opened 2018-09-07
SASVIEW-1033: Support for sasmodels' new 'lazy' intermediate results

tcbennun:
This sasmodels feature is implemented in the branch beta_approx and (I assume) is relatively stable; SasViewModel._intermediate_results is a callable which, when called, returns a dict of intermediate results, keyed by name.

Support is retained for sasmodels master, whereby P(Q) and S(Q) data are specifically returned directly.

Note the conflict with PR #169 (easy to resolve but notable)

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/ModelThread.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ModelThread.py
@@ -214,24 +214,37 @@ class Calc1D(CalcThread):
         else:
             output[index] = self.model.evalDistribution(self.data.x[index])
 
-        sq_values = None
-        pq_values = None
-        s_model = None
-        p_model = None
-        if isinstance(self.model, MultiplicationModel):
-            s_model = self.model.s_model
-            p_model = self.model.p_model
-        elif hasattr(self.model, "calc_composition_models"):
-            results = self.model.calc_composition_models(self.data.x[index])
-            if results is not None:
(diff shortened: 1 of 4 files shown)


src/sas/qtgui/Perspectives/Fitting/ModelThread.py
```diff
@@ -214,24 +214,37 @@ def compute(self):
...
-        p_model = None
-        if isinstance(self.model, MultiplicationModel):
-            s_model = self.model.s_model
-            p_model = self.model.p_model
-        elif hasattr(self.model, "calc_composition_models"):
-            results = self.model.calc_composition_models(self.data.x[index])
-            if results is not None:
-                pq_values, sq_values = results
+        results_eval = {}
+        intermediate_results = getattr(self.model, "_intermediate_results", None)
```
pkienzle: I don't think sasmodels isn't yet saving the intermediate results in the model, even in the lazy branch.  

When it does, it should not store it with a variable name that starts with "_" since by convention that is a private variable subject to change, not a public part of the api to be used by code outside the class.

This interface still has threading issues.  If a separate thread evaluates the model it will override the immediate results saved within the model.  The interface within sasmodels could be changed so that calculate_Iq returns I(q) as well as the lazy results.  Sasview can be updated to call calculate_Iq directly, and get a handle directly to the lazy intermediate evaluator.

PR #175, opened 2018-09-09
FittingWidget: showing also new plots on recalc conditionally

ibressler:
A proposal on updating and creating new plots directly after (threaded) recalculation - prevents showPlot() to interfere with updatePlot() and thus plotting outdated data which will get overwritten by calc results anyway. Perhaps more streamlined this way, saves resources on the long run (when the UI grows).

New plots show up conditionally only, based on a flag. Data1D.hidden is used here, because it seems to not being used so far. This would also allow individual hide/show settings by the user on a per plot basis. It's tricky to update the data plot which is created only once, outside of any recalc action, that's what  updateDataVisibility() is for.

- cherry-picked from ESS_GUI_poly_plot branch to keep discussion separate, but included there as well and therefore, it might be best to agree on this proposal first before merging the poly_plot changes


ibressler:
This PR is obsolete, the recently introduced plot roles fix this to some degree.

PR #176, opened 2018-09-09
Update 4.2.0 release notes

butlerpd:
Addresses #1146 making release notes for 4.2.0


pkienzle:
Installation instructions need some work.  

* it is not clear what the "requirements below" are for the phrase (though I'm guessing it is the contents of the yml files):

     make sure the requirements below are already installed

* the git clone instructions are duplicated; remove the first one?

* warn that specific release will replace previous releases in python environment.

* instructions for starting sasview for the specific release are wrong.  Not sure what the correct instructions are, but given that there is an entry_point in setup.py, I'm guessing that it will drop a file in the bin directory for the environment into the path; maybe you just need to type "sasview"?

pkienzle:
I made the two files have the same line endings.  I went with LF rather than CRLF for consistency with the rest of the project, but if we want windows users to be able to read it with notepad, maybe we should change them to CRLF?

pkienzle:
Don't need the "python setup.py install" step if you are telling them to use run.py (and probably don't want it since it may cause confusion later when some script picks up an unexpected version of sasview from the python path.

"To build documentation" doesn't have a leading "-", unlike the rest of the docs.

butlerpd:
will check but waiting to see if anything needs adding based on remaining tickets - will hold as PR for now

butlerpd:
Looks like we do need to add a few more things to the "what's changed" but will wait till all tickets for 4.2 are closed.

PR #177, opened 2018-09-09
Fix blank page when clicking help on a plugin model

butlerpd:
Addresses #529.  Plungin models do not have model.html help pages so
check if model is plugin. If so send to general model help.  If regular
built-in model send to model.html help page. eventually need better
solution .. also for original problem with ticket which is the P*S.


butlerpd:
ready for testing on Win64

butlerpd:
My test of the download shows that it does not now show a blank page.

PR #179, opened 2018-09-14
Redirect Help >> Tutorial to Documentation tutorials page

smk78:
Have changed the action of Help >> Tutorial (which used to just bring up Tutorial.pdf - the old tutorial) to Help >> Tutorials, which now redirects to /doc/user/tutorial.html and gives access to all the tutorials. Addresses #1175. I have left the original code in place but commented out.


smk78:
Closes #1175

PR #181, opened 2018-09-17
plot polydispersity

ibressler:
Solves SASVIEW-1035 and trac ticket 17

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
@@ -591,7 +591,10 @@ def plotPolydispersities(model, disp_models):
         # create Data1D as in residualsData1D() and fill x/y members
         # similar to FittingLogic._create1DPlot() but different data/axes
         data1d = Data1D(x = arr[0], y = arr[1])
-        data1d.xaxis('\\rm{{{}}} '.format(name.replace('_', '\_')), 'A') # correct unit?
+        xunit = ""
+        if name in model.details and len(model.details[name]):
+            xunit = model.details[name][0]
+        data1d.xaxis('\\rm{{{}}}'.format(name.replace('_', '\_')), xunit)
         data1d.yaxis('\\rm{{{weight}}}', 'normalized')
         data1d.scale = 'linear'
         data1d.symbol = 'Line'
--- a/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
@@ -582,11 +582,9 @@ def plotPolydispersities(model, disp_models):
         # create Data1D as in residualsData1D() and fill x/y members
         # similar to FittingLogic._create1DPlot() but different data/axes
         data1d = Data1D(x=xarr, y=yarr)
-        xunit = ""
-        if name in model.details and len(model.details[name]):
-            xunit = model.details[name][0]
-        data1d.xaxis('\\rm{{{}}}'.format(name.replace('_', '\_')), xunit)
-        data1d.yaxis('\\rm{{{weight}}}', 'normalized')
+        xunit = model.details[name][0]
+        data1d.xaxis(r'\rm{{{}}}'.format(name.replace('_', '\_')), xunit)
+        data1d.yaxis(r'\rm{weight}', 'normalized')
         data1d.scale = 'linear'
         data1d.symbol = 'Line'
         data1d.name = "{} polydispersity".format(name)
--- a/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
@@ -570,6 +570,37 @@ def plotResiduals(reference_data, current_data):
 
     return residuals
 
+def plotPolydispersities(model, disp_models):
+    plots = []
+    if model is None:
+        return plots
+    # test for model being a sasmodels.sasview_model.SasviewModel?
+    for name, pars in model.dispersion.items():
+        if 0 == pars['width']:
+            continue # no actual polydispersity
+        disp_args = pars.copy()
+        disp_type = disp_args.pop('type', None)
+        if disp_type is None:
+            continue # show an error? "poly type ... not found"
+        disp_func = disp_models[disp_type](**disp_args)
+        center = model.params[name]
+        # same as in sasmodels.weights.Dispersion
+        lb = center - disp_args['nsigmas']*disp_args['width']*center
+        ub = center + disp_args['nsigmas']*disp_args['width']*center
+        arr = disp_func.get_weights(center, lb, ub, relative = True)
+        # create Data1D as in residualsData1D() and fill x/y members
+        # similar to FittingLogic._create1DPlot() but different data/axes
+        data1d = Data1D(x = arr[0], y = arr[1])
+        data1d.xaxis('\\rm{{{}}} '.format(name.replace('_', '\_')), 'A') # correct unit?
+        data1d.yaxis('\\rm{{{weight}}}', 'normalized')
+        data1d.scale = 'linear'
+        data1d.symbol = 'Line'
+        data1d.name = "{} polydispersity".format(name)
+        data1d.id = data1d.name # placeholder, has to be completed later
+        data1d.plot_role = Data1D.ROLE_DELETABLE
+        plots.append(data1d)
+    return plots
+
 def binary_encode(i, digits):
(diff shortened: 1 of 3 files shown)


src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
```diff
@@ -591,7 +591,10 @@ def plotPolydispersities(model, disp_models):
         # create Data1D as in residualsData1D() and fill x/y members
         # similar to FittingLogic._create1DPlot() but different data/axes
         data1d = Data1D(x = arr[0], y = arr[1])
-        data1d.xaxis('\\rm{{{}}} '.format(name.replace('_', '\_')), 'A') # correct unit?
+        xunit = ""
+        if name in model.details and len(model.details[name]):
+            xunit = model.details[name][0]
+        data1d.xaxis('\\rm{{{}}}'.format(name.replace('_', '\_')), xunit)
```
pkienzle: I don't believe you need to protect against parameter name not being included in model.details.  Do you know of cases where model.details are not defined, or name is not in them?  If so, how is the polydisperse parameter showing up in the GUI?

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -57,6 +57,21 @@
...
+if not hasattr(SasviewModel, 'get_weights'):
+    def get_weights(self, name):
+        """Returns model weights by parameter name.
+        type: (str) -> Tuple(float, np.ndarray, np.ndarray)
+        Supposed to be a member of SasviewModel."""
+        p = [p for p in self._model_info.parameters.call_parameters
+                        if name is p.name]
+        if not len(p):
+            return None, None, None
+        return self._get_weights(p[0])
```
pkienzle: The parameter table has a `__getitem__()` method, so the following should suffice:
```
def get_weights(self, name):
    """
    Returns the polydispersity distribution for parameter *name* as *value* and *weight* arrays.
    """
    # type: (str) -> Tuple(np.ndarray, np.ndarray)
    _, x, w = self._get_weights(self._model_info.parameters[name])
    return x, w
```
I also simplified the interface so it doesn't return the centre, only the x, w arrays.

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
```diff
@@ -576,21 +576,14 @@ def plotPolydispersities(model, disp_models):
...
-        arr = disp_func.get_weights(center, lb, ub, relative = True)
+        if 0 == pars['width']: # no actual polydispersity
+            continue
+        _, xarr, yarr = model.get_weights(name)
+        if xarr is None or yarr is None: # param name not found
+            continue
         # create Data1D as in residualsData1D() and fill x/y members
         # similar to FittingLogic._create1DPlot() but different data/axes
-        data1d = Data1D(x = arr[0], y = arr[1])
+        data1d = Data1D(x = xarr, y = yarr)
```
pkienzle: PEP 8: Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value.

https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
```diff
@@ -576,21 +576,14 @@ def plotPolydispersities(model, disp_models):
...
-        center = model.params[name]
-        # same as in sasmodels.weights.Dispersion
-        lb = center - disp_args['nsigmas']*disp_args['width']*center
-        ub = center + disp_args['nsigmas']*disp_args['width']*center
-        arr = disp_func.get_weights(center, lb, ub, relative = True)
+        if 0 == pars['width']: # no actual polydispersity
+            continue
+        _, xarr, yarr = model.get_weights(name)
+        if xarr is None or yarr is None: # param name not found
+            continue
```
pkienzle: Unfortunately, there is a case (the Maier-Saupe distribution) where a "width" of zero indicates that the particle is unoriented, and so the it ought to return a distribution.  While we do not treat this correctly yet (various code takes shortcuts when width is zero), we may eventually, so don't make this assumption here. 

Also, the get_weights function will never return None for the array, only sequences of length one if there is no distribution.  

And again, you are not referencing `pars` in the loop, so no need to loop over it.

So something like:
```
    for name model.dispersion:
        xarr, yarr = model.get_weights(name)
        if len(xarr) > 1:
            data1d = Data1D(x=xarr, y=yarr)
            ...
```

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
ibressler: No problem.
It might useful to add a static code checker configuration to the project. Such as
https://www.pylint.org or https://pypi.org/project/pycodestyle/
The configuration file will then define what is expected in the project such as the line length and which rules from PEP8. Furthermore, files could be checked locally before commit.

pkienzle: We have it turned on for sasmodels.  I guess there are too many problems in the sasview codebase for us to turn it on yet.

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
pkienzle: I find that problems are easier to debug if the program fails early. If the model interface really does change, then code which protects against unexpected inputs will remain quietly broken.

For the GUI, trap any errors that occur and display them to the user, complete with traceback so we can find out where they are.  This keeps the program robust.

The assumption that details exists and is initialized properly is no different from the unchecked assumption a few lines above in `if 0 == pars['width']` that width is a member of pars.  If we tried to check that every data structure has the correct form before performing any operation our code would become immensely unweildy.  Python has optional type checking via mypy which we can use to verify that the right types are being passed to functions.  I've added this as ticket 1180.

Minor note:  you can use the more natural `if x == 0:` in python rather than `if 0 == x:`.   Since `if x = 0:` is invalid, you don't get the accidental assignment bug that occasionally confuses C programmers.  Of course, if you have been doing defensive C coding for years then `if (0 == x)` may look more natural to you.

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
```diff
@@ -570,6 +570,31 @@ def plotResiduals(reference_data, current_data):
...
+        if len(xarr) <= 1: # param name not found or no polydisp.
+            continue
+        # create Data1D as in residualsData1D() and fill x/y members
+        # similar to FittingLogic._create1DPlot() but different data/axes
+        data1d = Data1D(x=xarr, y=yarr)
+        xunit = ""
+        if name in model.details and len(model.details[name]):
+            xunit = model.details[name][0]
+        data1d.xaxis('\\rm{{{}}}'.format(name.replace('_', '\_')), xunit)
+        data1d.yaxis('\\rm{{{weight}}}', 'normalized')
```
pkienzle: Too many braces. And could use an r-string.  `'\\rm{{{weight}}}'` => `r'\rm{weight}'`.  I think the latex interpreter just eats them so you won't see it in the output.

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
ibressler: Thanks for the feedback!
Regarding *model.get_weights()* there is a problem now: It can't detect if the polydispersity was turned off again as the previously calculated weights remain stored in the model. Even from the GUI point of view it is not possible to detect if a polydispersity has to be plotted - assuming that a width == 0 is a valid value.
For example, the ellipsoid model offers 4 parameters which can be polydisperse in theory. At the moment, all 4 are plotted. How to tell which one is actually 'active' polydisperse?
Can you give a hint, what should be changed (GUI or backend) in order to make this work?

pkienzle: I don't see how weights remain stored in the model. `model.get_weights()` is calling  `SasviewModel._get_weights()` which calls `weights.get_weights()` each time and does not update `self`.  `get_weights()` always returns a distribution, though it may be only a single point of weight 1.  This is the monodisperse case that doesn't need to be plotted.  So the test for `len(xarr) <= 1` that you have should be enough.  

Maybe you need to clear the existing plots from the GUI before entering your loop?
 
Note that there are some potential threading issues that can cause problems.  There is only one `SasviewModel` object and it is used by the GUI, the calculator, and the fit engine, each of which is running in a separate thread.  The fit engine in particular is stuffing values into the model independent of the GUI, so the state within the model may not match the state displayed on the GUI.   Don't think that is happening here though.

PR #183, opened 2018-09-18
Continue loading data files if a single file cannot be loaded.

krzywon:
This is a cherry pick from ticket-1111.

### DIFF
--- a/src/sas/sasgui/guiframe/local_perspectives/data_loader/data_loader.py
+++ b/src/sas/sasgui/guiframe/local_perspectives/data_loader/data_loader.py
@@ -204,7 +204,7 @@ class Plugin(PluginBase):
                                  info="info")
             except NoKnownLoaderException as e:
                 exception_occurred = True
-                error_message = "Loading data failed!\n" + e.message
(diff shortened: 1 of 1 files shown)


butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

src/sas/sasgui/guiframe/local_perspectives/data_loader/data_loader.py
```diff
@@ -204,7 +204,7 @@ def get_data(self, path, format=None):
                                  info="info")
             except NoKnownLoaderException as e:
                 exception_occurred = True
-                error_message = "Loading data failed!\n" + e.message
+                error_message = "Loading data failed!" + e.message
```
pkienzle: Don't you need a space after `failed!`? Or do you know that e.message starts with a space?

krzywon: A NoKnownLoaderException message always starts with a leading carriage return.

PR #184, opened 2018-09-24
SASVIEW-1052: Let user choose if radius_effective is computed from form factor, or free/fittable

tcbennun:
([See JIRA issue](https://jira.esss.lu.se/browse/SASVIEW-1052).)

The radius_effective_mode functionality was merged into sasmodels' beta_approx branch, so it is as stable as the rest of the beta approximation-related features which I have already merged into ESS_GUI. Therefore this should be merged in too.

In summary, the value of radius_effective_mode is checked, and the radius_effective parameter is made editable/non-editable appropriately. Additionally, when a model has been evaluated and radius_effective has been calculated in the model and passed back as an intermediate result, it is placed into the parameter table so the user can view it.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -2593,35 +2593,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             self.communicate.plotUpdateSignal.emit([plot])
 
         # Update radius_effective if relevant
-        def updateRadiusEffective():
-            ER_mode_row = self.getRowFromName("radius_effective_mode")
-            if ER_mode_row is None:
-                return
-            try:
-                ER_mode = int(self._model_model.item(ER_mode_row, 1).text())
-            except ValueError:
-                logging.error("radius_effective_mode was set to an invalid value.")
-                return
-            if ER_mode < 1:
-                # does not need updating if it is not being computed
-                return
-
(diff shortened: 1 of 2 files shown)


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -2546,6 +2613,38 @@ def complete2D(self, return_data):
...
+        scalar_results = self.logic.getScalarIntermediateResults(return_data)
+        ER_value = scalar_results.get("effective_radius") # note name of key
+        if ER_value is None:
+            return
+        # ensure the model does not recompute when updating the value
+        self._model_model.blockSignals(True)
+        self._model_model.item(ER_row, 1).setText(str(ER_value))
+        self._model_model.blockSignals(False)
+        # ensure the view is updated immediately
+        self._model_model.layoutChanged.emit()
```
rozyczko: Why do you feel you need layoutChanged to be emitted? If you are trying to make sure the view has enough time to respond to the model, maybe qApp.processEvent(), if anything.

tcbennun: The parameter value wouldn't change on-screen at all until the user hovered over it, otherwise.

tcbennun:
That's a good point, but I'm struggling to see a way to name the methods neatly. 'setParamUneditableByRow' doesn't seem great to me...

Thanks though - I'll merge for now as it's high-priority!

PR #185, opened 2018-09-26
Ticket 1183: Don't clear parameters on structure box content reset

pkienzle:
See [Ticket #1183](http://trac.sasview.org/ticket/1183).

When model reload triggered, all models with structure factor boxes had their parameters reset to the default value.

### DIFF
--- a/src/sas/sasgui/perspectives/fitting/fitting.py
+++ b/src/sas/sasgui/perspectives/fitting/fitting.py
@@ -376,7 +376,7 @@ class Plugin(PluginBase):
                         if index == -1:
                             index = 0
                         page.structurebox.SetSelection(index)
-                        page._on_select_model()
+                        page._on_select_model(keep_pars=True)
         except Exception:
             logger.error("update_custom_combo: %s", sys.exc_value)
 


pkienzle:
As noted on ticket, the current code recalculates models more than necessary.  This could be fixed as part of the PR, or could wait until 4.3.

butlerpd:
Noticed that I had several other SasView windows open and worried that there may have been some polution that tainted the results so redid with slightly different data sets.  Result was the same except that the P*S did not generate a new graph.  Also, the plugin tab was completely unusable and illegible and going off to different tabs did not help this time.  choosing the model in the dropdown did properly redraw the tab but of course rest the parameters to default.  So comments above remain valid I believe

pkienzle:
The new update seems to work better.  I wasn't able to reproduce the problem where data was being plotted with default values.  Other issues appear to be resolved.  For example, if the plugin is a structure factor that is used on a fit page, then that page is updated when the plugin is updated.  When a new structure factor model is added, it will show up in the structure factor combo box.  When a plugin model is converted to a form factor model, the structure factor box will appear.

pkienzle:
On mac, the previous version would show a different page in the window than was selected in the tab.  This commit identifies that the last drawn page may be different from the one selected on the tab, and then switch to the last drawn page and then switch back to the current page.  This may only be an issue on the mac... on windows/linux may want to see if it works without the patch.

To reproduce:

* put a model on a page (e.g., cylinder)
* create a new page and set it with a plugin model
* return to the first page (cylinder)
* modify the plugin model with the advanced editor
* on model check, the cylinder page should be displayed

On my version of wx for mac this shifts the scroll bar on the cylinder page to the bottom.  Not ideal, but better than the previous behaviour where the incorrect tab was highlighted, and it was different than the tab the user was working on.

May want to check what happens when two pages contain the edited plugin, both different from the currently displayed page.

butlerpd:
ready for testing on Win64

pkienzle:
Yet another try.  Turns out that putting the parameters onto the page was not putting the parameter values into the model.  I now call `_update_paramv_on_fit` which seems to be the easiest way to post the values back to the model after it is reloaded.

Turns out I'm redrawing every page with plugin models on reload.  Even the the module isn't being reloaded, the model is rebuilt from its info, and a new model object is created.  This is fixed in [sasmodels PR #43](https://github.com/SasView/sasmodels/pull/83), which now keeps a cache of modules associated with each file so it can see if the module has changed since the model was last reloaded.

The new sasmodels PR also detects changes to c code for triggering reload.  Without it, you will need to save the py file before the model will reload itself even if it is only the c code which is changed.

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

smk78:
Just pulled this branch and https://github.com/SasView/sasmodels/pull/83 and ran locally on Win64. Repeated my testing described in #1183. No issues observed. Also recommend merging after code has been reviewed.

timsnow:
I have pulled this on RHEL Linux and all seems to be working. I have also reviewed the changes and they all make sense within the scope of this ticket. Additionally lines 370-380 in fitting.py will help us to bridge the gap between the old and new methodologies.

I will merge this now.

PR #186, opened 2018-09-28
Ticket 1179: Add the SasView version to Report Files

krzywon:
The SasView version number is added to the title bar in Fit and Invariant reports. This handles ticket 1179 for v4.2, but I can't speak for v5.0 and beyond.

### DIFF
--- a/src/sas/sascalc/fit/pagestate.py
+++ b/src/sas/sascalc/fit/pagestate.py
@@ -637,11 +637,10 @@ class PageState(object):
                 chi2 = ("Chi2/Npts = " + value)
                 chi2_string = CENTRE % chi2
             elif name == "Title":
-                from sas.sasview.__init__ import __version__ as sasview_version
                 if len(value.strip()) == 0:
                     continue
                 title = (value + " [" + repo_time + "] [SasView v" +
-                         sasview_version + "]")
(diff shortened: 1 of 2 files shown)


src/sas/sascalc/fit/pagestate.py
```diff
@@ -637,9 +637,11 @@ def _get_report_string(self):
                 chi2 = ("Chi2/Npts = " + value)
                 chi2_string = CENTRE % chi2
             elif name == "Title":
+                from sas.sasview.__init__ import __version__ as sasview_version
```
pkienzle: Having sascalc depend on sasview is not ideal. 

sas.sasview depends on sas.sasgui which depends on sas.sascalc which depends on sas.sasview, so it is a bit circular.  

Not a technical problem in this case, since sas.sasview.__init__ doesn't import anything, but conceptually this in not clean.  

Should probably move version number to `sas.__init__` (but not for 4.2!). Add a ticket?

pkienzle: BTW, line 31 already imports __version__ as SASVIEW_VERSION.

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

PR #189, opened 2018-10-04
Revert "Ticket 1004: remove e-06 from default magnetism values."

rozyczko:
Reverts SasView/sasview#188
Mistakenly merged assuming the PR goes into ESS_GUI.
Should be approved by others as well for merging into master.


smk78:
So I see that this would change the _default_ magnetic SLD's.

The question I have is, is the remainder of the codebase that uses magnetic SLD's set up to use values with or without the 10^-6 ?

butlerpd:
The code change is quite simple and looks to me to be taking the input from the GUI box and multiplying by e-6 so that for the rest of the code base remains unchanged since it is receiving the same thing?

PR #190, opened 2018-10-11
simplify C build

pkienzle:
Move c extensions from core/module.so to _module.so.  This allows an inplace build of the C models so that run.py doesn't need to do tricks to load the module from the build path while the rest of the package is loaded from the src path.

Also included is some cleanup for improved python 3 support.

### DIFF
--- a/run.py
+++ b/run.py
@@ -66,7 +66,7 @@ def import_dll(modname, build_path):
     return imp.load_dynamic(modname, path)
 
 
-def prepare():
+def prepare(rebuild=True):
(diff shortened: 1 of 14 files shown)


pkienzle:
Also includes partial support for pytest (PR #193)

pkienzle:
Numpy version error.  Should be fixed now.

pkienzle:
ready for testing on OSX

pkienzle:
ready for testing on Win64

PR #193, opened 2018-10-11
support pytest

pkienzle:
Note: based off of simplify-c-build, which already contains mods to allow `python setup.py test`.

The .eggs directory is created during the test to pull in pytest helpers.  Without an internet connection, need to use `pip install pytest_runner`, which should perhaps be added to the requirements files and yaml files for setting up conda environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ Thumbs.db
 /docs/sphinx-docs/node-package.json
 
 # test outputs
+/.eggs
 /test/pr_inversion/test/test_output.txt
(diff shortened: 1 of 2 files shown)


pkienzle:
pytest and pytest-runner are conda packages, so don't need to use pip.

I'll leave the updates to the folks who are maintaining the build machine.  The yaml files have specific hard-coded version numbers which should correspond to the particular versions that are used to run the jenkins tests.

butlerpd:
seems trivial and innocuous enough to merge now? or should it wait till 4.2.1 is released?

PR #194, opened 2018-10-11
improved support for py37 in sasgui

pkienzle:
Finish transitioning code base to python 2-3 compatibility.

This allows pytest to scan the .py files in all directories; without this patch, src/sas/sasgui will need to be added to the pytest excluded directory lists.

Some minor tweaks to allow wx 4 to startup, but didn't test extensively.


butlerpd:
This should not be merged till 4.2.1 is released.  It should be trivial enough but touches **beaucoup** de files

src/sas/sasgui/perspectives/fitting/basepage.py
```diff
@@ -11,7 +11,10 @@
 import json
 import logging
 import traceback
-from Queue import Queue
+try:
+    from Queue import Queue
+except ImportError:
+    from queue import Queue
```
krzywon: Nit pick: A comment on python version compatibility would be helpful.

src/sas/sasgui/plottools/toolbar.py
```diff
@@ -64,11 +64,20 @@ def _init_toolbar(self):
...
                            'Forward', 'Forward navigation view')
         # todo: get new bitmap
-        self.AddCheckTool(self._NTB2_PAN, _load_bitmap('move.png'),
-                          shortHelp='Pan',
-                          longHelp='Pan with left, zoom with right')
-        self.AddCheckTool(self._NTB2_ZOOM, _load_bitmap('zoom_to_rect.png'),
-                          shortHelp='Zoom', longHelp='Zoom to rectangle')
+        # todo: update with code from matplotlib/backends/backend_wx.py
+        is_phoenix = 'phoenix' in wx.PlatformInfo
+        if is_phoenix: # wx phoenix >= 4.0.0b2
```
krzywon: Nit pick: Unless you plan on using is_phoenix again.
`if 'phoenix' in wx.PlatformInfo`

src/sas/sasgui/guiframe/CategoryInstaller.py
```diff
@@ -170,4 +170,4 @@ def check_install(homedir = None, model_list=None):
                 CategoryInstaller._regenerate_master_dict(by_model_dict,
                                                           model_enabled_dict)
 
-            json.dump(master_category_dict, open(serialized_file, 'wb'))
+            json.dump(master_category_dict, open(serialized_file, 'w'))
```
krzywon: Out of curiosity, why was the binary flag removed?

pkienzle: No, it should still be writing with binary.  I learned recently that we can use `io.open` to get the same file support in python 2 and 3.  This will allow us to open it as ascii with `'w'`, but control line endings with `newline='\n'`.

pkienzle:
Added new ticket [#1233](http://trac.sasview.org/ticket/1233) saying more testing is needed before we claim SasView 4.3 works with python 3.x/wx 4.x, particularly for file I/O.

Meanwhile, with this PR the sascalc tests will run via pytest on python 2 and 3.

krzywon:
Ready for testing on OSX

krzywon:
Ready for testing on Win64

PR #195, opened 2018-10-31
Ticket 1015.  Clean out existing kernels when changing compute engine

pkienzle:
See [ticket #1015](http://trac.sasview.org/ticket/1015).

Must have corresponding sasmodels PR applied.


wpotrzebowski:
This PR requires merging sasmodels PR #93. Should we do it before 4.2.1?

pkienzle:
It is a "quick fix" to the problem that occurs for new users.  I suggest applying it for 4.2.1 (after the sasmodels PR).

PR #198, opened 2018-11-13
Ticket 1094 headless

pkienzle:
Build bumps plugin model for sasview project/analysis files so that fits can be run headless.

See fit.py in the root directory for instructions.


pkienzle:
Goal is to load project/analysis without the GUI.  This should be done in a way that is compatible with SasView 5 project save/load.

pkienzle:
Ideally would like to be able to install sascalc without wx/qt dependency when deploying on HPC systems.  This PR does not (yet) update setup.py to allow this.

src/sas/sascalc/fit/fitstate.py
```diff
@@ -0,0 +1,407 @@
...
+            pass
+
+    def __str__(self):
+        return '<SasFit %s>'%self.fitfile
+
+    def show(self):
+        """
+        Summarize the fit pages in the state object.
+        """
+        for k, fit in enumerate(self.fits):
```
wpotrzebowski: The following loops contain some repetition that can be removed. It is not critical though

src/sas/sascalc/fit/fitstate.py
```diff
@@ -0,0 +1,407 @@
...
+        # No way to uniquely map the page id (M1, M2, etc.) to the different
+        # items in fits; neither fit_number nor fit_page_source is in the
+        # pagestate for the fit, and neither model_name nor data name are
+        # unique.  The usual case of one model per data file will get us most
+        # of the way there but there will be ambiguity when the data file
+        # is not unique, e.g., when different parts of the data set are
+        # fit with different models.  If the same model and same data are
+        # used (e.g., with different background, scale or resolution in
+        # different segments) then the model-fit association will be assigned
+        # arbitrarily based on whichever happens to come first.
```
wpotrzebowski: Just want to make sure.  Will the current setup work in simultaneous fitting to multiple data sets?

src/sas/sascalc/fit/fitstate.py
pkienzle: Stuff 81-87/91-97 into a helper function?

src/sas/sascalc/fit/fitstate.py
pkienzle: From my comment it seems there are ambiguities in the saved format, so some will fail to reload properly.  I'm guessing this is true for the 4.x GUI as well (since I presumably took the code from there in the first place).  Maybe Jeff knows the answer to this?

Are these ambiguities gone in the 5.0 save format?

src/sas/sascalc/fit/fitstate.py
```diff
@@ -0,0 +1,407 @@
...
+
+    # Set the data weighting (dI, sqrt(I), I, or uniform)
+    weight = get_data_weight(state)
+
+    # Make fit data object and set the data weights
+    # TODO: check 2D masked data
+    if state.enable2D:
+        fitdata = FitData2D(sas_data2d=data, data=data.data,
+                            err_data=data.err_data)
+        fitdata.err_data = weight
```
wpotrzebowski: Is fitdata.err_data = weight correct? There is also data.err_data and therefore I am asking

src/sas/sascalc/fit/fitstate.py
```diff
@@ -0,0 +1,407 @@
...
+    bumps interface.
+    """
+    #@staticmethod
+    #def data_view():
+    #    pass
+
+    #@staticmethod
+    #def model_view():
+    #    pass
+
```
wpotrzebowski: Why staticmethod is used here? I just cannot see it but there is probably a good reason.

pkienzle: `BumpsPlugin` is defining a namespace for a set of functions which is passed to `bumps.cli.install_plugin` (line 393 below).  `install_plugin` is expecting a module, and using `getattr` to retrieve particular named functions from the module.  A set of `staticmethod` in a class is duck-type equivalent.

src/sas/sascalc/fit/fitstate.py
pkienzle: This looks wrong.  Going back to the fitproblem.py code where I pulled this from, I think weight was being assigned to a copy of the original data object rather than the FitData2D object.  So, yes, it needs to be fixed and tested against a wider variety of cases (including masking, resolution, and weighting variants).

wpotrzebowski:
I am running into this error: /sas/sascalc/dataloader/readers/cansas_reader.py", line 495, in _parse_entry
    self.sort_one_d_data()
AttributeError: 'Reader' object has no attribute 'sort_one_d_data'

Is it something that should be fixed with PR #210

krzywon:
@wpotrzebowski Correct. That is the same issue being addressed in #210.

src/sas/sascalc/fit/fitstate.py
```diff
@@ -0,0 +1,407 @@
...
+        #print(problem.show())
+        return problem
+
+    #@staticmethod
+    #def new_model():
+    #    pass
+
+
+def setup_sasview():
+    from sas.sasview.sasview import setup_logging, setup_mpl, setup_sasmodels
```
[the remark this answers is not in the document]
pkienzle: Yes.  I'm not sure if we want them, so I'm leaving them commented out for now.

pkienzle:
Rebasing off ticket-1243, PR #210

src/sas/sascalc/fit/fitstate.py
pkienzle: I now raise an error if the save file is ambiguous.

pkienzle:
I pulled in the 1243 changes because otherwise the saved model wouldn't load.  However, 1243 branches from  master from before the lint and py3 changes, so the diff was too large.  Instead this PR is relative to master, and so includes both the headless changes as well as a couple of lines included in the 1243 PR.

src/sas/sascalc/fit/fitstate.py
```diff
@@ -0,0 +1,408 @@
...
+
+def parse_optional_float(value):
+    """
+    Convert optional floating point from string to value, returning None
+    if string is None, empty or contains the word "None" (case insensitive).
+    """
+    if value is not None and value != "" and value.lower() != "none":
+        return float(value)
+    else:
+        return None
```
rozyczko: or just try... except on float(value) catching ValueErrors. Otherwise you risk not just two return values (float or None) but also unchecked exception if 'value' is weird.

src/sas/sascalc/fit/fitstate.py
```diff
@@ -0,0 +1,408 @@
...
+        plugins_dir = find_plugins_dir()
+        path = os.path.abspath(os.path.join(plugins_dir, name + ".py"))
+        #print("loading custom", path)
+        model = load_custom_model(path)
+    elif name and name is not None and name.lower() != "none":
+        #print("loading standard", name)
+        model = _make_standard_model(name)
+    else:
+        model = None
+    _MODEL_CACHE[name] = model
```
rozyczko: Is creating a _model_cache dict entry for None really a good idea? What does it mean?
It would always be ```_MODEL_CACHE[None] = None``` anyway.

src/sas/sascalc/fit/pagestate.py
```diff
@@ -556,12 +686,9 @@ def __repr__(self):
         rep += "model  : %s\n\n" % str(self.model)
         temp_parameters = []
         temp_fittable_param = []
-        if self.data.__class__.__name__ == "Data2D":
-            self.is_2D = True
-        else:
-            self.is_2D = False
         if self.data is not None:
-            if not self.is_2D:
+            is_2D = (self.data.__class__.__name__ == "Data2D")
```
rozyczko: isinstance(self.data, Data2D)? You're importing Data2D anyway.

src/sas/sascalc/fit/fitstate.py
pkienzle: I wouldn't want to silently ignore badly formatted float values, which is what would happen with `try: return float(value) except: return None`.   We could do finer grained checking as we are loading in each field, but in this case I'm happy to reject the entire file if the any data in it is bad.

butlerpd: What do you mean by "silently ignore?"  I would think that would in fact be the right answer?  assuming it replaces with a NAN so it could be caught later if so desired?  I have found users get rather miffed if an otherwise perfectly good file barfs because of one bad value that could have been ignored and allowed them to continue working.  But as I said ... I probably am missing the point entirely here?

pkienzle: The code in question is used for parsing the value, min, and max fields for each parameter in a saved fit.  Since these are values that we've written to the save file, they ought to be in a format that is parseable.

One case that should not be parsed as a float is enumerated parameters, which store the value as a string such as `Rexp(-|nu|z)`.  If we are not handling this properly, I want it to raise an error so that we know rather than silently treating it as None which would then get interpreted as the default first choice.  If we treat it as NaN this would then raise an error when we try to convert it to an int, which is far away from where the actual problem occurred.    If we are going to fail, then fail early.  

If we are going to accept the input then we better be sure it is correct, otherwise it will require an observant user to notice that the value has changed.  Particularly in this case, where the saved file may be processed on a remote fitting server and the user never sees the value.

martintb:
This feature has been working great for me but I've recently run into what seems to be a bug. I'm getting the following exception when running a constrained fit with a _plugin model_. This error does not happen when I do constrained fits with a non-plugin model. I was able to reproduce this by copying the sasmodels "sphere.py" and "sphere.c" and making the plugin model "MySphere.py" and "MySphere.c". 

```
  File "/home/tbm/software/sasview/git/src/fit.py", line 45, in <module>
    bumps_cli()
  File "/home/tbm/software/sasview/git/src/src/sas/sascalc/fit/fitstate.py", line 405, in bumps_cli
    bumps.cli.main()
  File "/home/tbm/software/bumps/git/pylib/bumps/cli.py", line 485, in main
    problem = initial_model(opts)
  File "/home/tbm/software/bumps/git/pylib/bumps/cli.py", line 290, in initial_model
    problem = load_model(opts.args[0], opts.args[1:])
  File "/home/tbm/software/bumps/git/pylib/bumps/cli.py", line 82, in load_model
    problem = plugin.load_model(filename)
  File "/home/tbm/software/sasview/git/src/src/sas/sascalc/fit/fitstate.py", line 371, in load_model
    state = FitState(filename)
  File "/home/tbm/software/sasview/git/src/src/sas/sascalc/fit/fitstate.py", line 55, in __init__
    self._set_constraints()
  File "/home/tbm/software/sasview/git/src/src/sas/sascalc/fit/fitstate.py", line 159, in _set_constraints
    % model['fit_page_source'])
ValueError: could not find model M2 in file
```
[Here](https://github.com/SasView/sasview/files/3339847/Archive.zip) is a minimal svs file and MySphere plugin model that reproduces the error.

I'm using the ticket-1094-headless branch on macOS which I freshly pulled this morning and I'm attempting to run the fits on Ubuntu 16.04.6. Let me know if more information about my environment would be helpful and I'll provide it.

pkienzle:
@martintb, the latest checkin works around the mismatched names for the plug-in model in the individual fit pages and the simultaneous fit page.  The individual fit pages had models name prefixed with "[plug-in] ".

This probably means 4.2.2 won't reload the file either.  5.0 has it's own save format which isn't covered by this branch.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win64

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win64

martintb:
Is there any way to make bumps use the `<title>` tag or `data_name` attribute or similar from the .svs when doing simultaneous fits? I end up with a bunch of curves labeled 'Model 0' or 'M1' that are hard to distinguish. See example output from [fit.mon](https://github.com/SasView/sasview/files/3435944/fit.mon.txt).

Happy to file as a separate issue if that is more appropriate.

pkienzle:
Please file as a separate issue.

pkienzle:
@martintb, the new version now pulls the data name into the plot title as it does in sasview. I thought I saw a ticket for it, but I couldn't find it again.  If you opened one, you can now close it.

I also fixed the handling of 2D models.

PR #200, opened 2018-11-28
Added newlines to import statements in the plugin model template.

rozyczko:
Trac #1218
Copy of what is already in the ESS_GUI branch.
Added '\n' to import statements in the plugin model template text.


butlerpd:
ready for testing on Win64

PR #202, opened 2018-12-12
fix 2D log-scale plotting range when data contains zeros and negative values

pkienzle:
Figuring out proper handling of zmin_2D and zmax_2D will require more work.  

They do not appear to be settable by the user, so using the default should be fine, which is what this patch does.  However, it has not been checked against all uses of 2D plotting.

### DIFF
--- a/src/sas/sasgui/plottools/PlotPanel.py
+++ b/src/sas/sasgui/plottools/PlotPanel.py
@@ -1433,6 +1433,7 @@ class PlotPanel(wx.Panel):
         Render the current data
 
         """
+        # TODO: include mask info in plotter
         self.data = data
         self.qx_data = qx_data
         self.qy_data = qy_data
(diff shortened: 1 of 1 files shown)


pkienzle:
Note: for 5th percentile/2 cutoff rather than min, use:
```
cutoff = (np.quantile(output[index], 0.05) - np.log10(2) if index.any() else 0.)
output[output < cutoff] = cutoff
output[~index] = cutoff
```

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

rozyczko:
Minor issue with this is numpy versioning. quantile() shows up in >=1.15, whereas we had to downgrade numpy on ubuntu to 1.14.3 in order to fix other, more serious issues.
This only affects 5.0, though and by the time 5.0 is fully released, we need to address the numpy versioning properly.

PR #203, opened 2019-01-08
Ticket 1206 - fix problems in smearing sizer

butlerpd:
All changes are to fitpage.py.  Changed behavior so that both 1D and 2D smearing from data are shown as dq/q in % and changed several labels to more clearly indicate what is being displayed.  2D smearing had some serious logic problems (for example when using dq from data both boxes come up but showed "NONE" while for custom pinhole the same two boxes show up (non editable) even though only one input is allowed.  Finally issues with switching from 2D to 1D displayed options for smearing even if there was not model to smear. 

Also added a bunch of logging statements

Since some (not insignificant in this convoluted module) logic had to be changed the code should be checked carefully as well as tested. 

Addresses #1206
Addresses #1222

### DIFF
--- a/src/sas/sasgui/perspectives/fitting/fitpage.py
+++ b/src/sas/sasgui/perspectives/fitting/fitpage.py
@@ -17,6 +17,7 @@ import logging
 from sasmodels.weights import MODELS as POLYDISPERSITY_MODELS
 
 from sas.sascalc.fit.qsmearing import smear_selection
+from sas.sascalc.dataloader.data_info import Data1D, Data2D
 
 from sas.sasgui.guiframe.events import StatusEvent, NewPlotEvent, \
     PlotQrangeEvent
@@ -186,8 +187,7 @@ class FitPage(BasicPage):
 
         :return: True or False
         """
-        if self.data.__class__.__name__ == "Data2D" or \
-                        self.enable2D:
+        if isinstance(self.data, Data2D) or self.enable2D:
             return True
         return False
 
@@ -201,8 +201,7 @@ class FitPage(BasicPage):
         is_2d_data = False
 
         # Check if data is 2D
-        if self.data.__class__.__name__ == "Data2D" or \
-                        self.enable2D:
+        if isinstance(self.data, Data2D) or self.enable2D:
             is_2d_data = True
 
         title = "Fitting"
@@ -517,7 +516,7 @@ class FitPage(BasicPage):
         is_2d_data = False
 
         # check if it is 2D data
-        if self.data.__class__.__name__ == "Data2D" or self.enable2D:
+        if isinstance(self.data, Data2D) or self.enable2D:
             is_2d_data = True
 
         self.sizer5.Clear(True)
@@ -820,8 +819,7 @@ class FitPage(BasicPage):
                         wx.EVT_CHECKBOX(self, cb.GetId(), self.select_param)
                         self.sizer4_4.Add(cb, (iy, ix), (1, 1),
                                 wx.LEFT | wx.EXPAND | wx.ADJUST_MINSIZE, 5)
-                        if self.data.__class__.__name__ == "Data2D" or \
-                                    self.enable2D:
+                        if isinstance(self.data, Data2D) or self.enable2D:
                             cb.Show(True)
                         elif cb.IsShown():
                             cb.Hide()
@@ -833,8 +831,7 @@ class FitPage(BasicPage):
                         poly_tip = "Absolute Sigma for %s." % item
                         ctl1.SetToolTipString(poly_tip)
                         ctl1.SetValue(str(format_number(value, True)))
-                        if self.data.__class__.__name__ == "Data2D" or \
-                                    self.enable2D:
+                        if isinstance(self.data, Data2D) or self.enable2D:
                             if first_orient:
                                 values.SetLabel('PD[ratio], Sig[deg]')
                                 poly_text = "PD(polydispersity for lengths):\n"
@@ -868,8 +865,7 @@ class FitPage(BasicPage):
                                           wx.EXPAND | wx.ADJUST_MINSIZE, 0)
 
                         ctl2.Hide()
-                        if self.data.__class__.__name__ == "Data2D" or \
-                                self.enable2D:
+                        if isinstance(self.data, Data2D) or self.enable2D:
                             if self.is_mac:
                                 text2.Show(True)
                                 ctl2.Show(True)
@@ -894,8 +890,7 @@ class FitPage(BasicPage):
                                           wx.EXPAND | wx.ADJUST_MINSIZE, 0)
                         ctl4.Hide()
 
-                        if self.data.__class__.__name__ == "Data2D" or \
-                                self.enable2D:
+                        if isinstance(self.data, Data2D) or self.enable2D:
                             ctl3.Show(True)
                             ctl4.Show(True)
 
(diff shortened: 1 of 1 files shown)


butlerpd:
ready for testing on OSX

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on OSX

src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -212,9 +215,11 @@ def _fill_range_sizer(self):
               "Please enter only the value of interest to customize smearing..."
         smear_message_new_psmear = \
               "Please enter a fixed percentage to be applied to all Q values..."
-        smear_message_2d_x_title = "<dQp>[1/A]:"
-        smear_message_2d_y_title = "<dQs>[1/A]:"
-        smear_message_pinhole_percent_title = "dQ[%]:"
+        smear_message_2d_x_title = "<dQ/Q>p[%]:"
```
mdoucet: What do `p` and `s` mean? Maybe there should be a tool tip to help users figure out what is really meant?

src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -1600,91 +1615,129 @@ def _get_smear_info(self):
...
-                and data.dqx_data.any() != 0 \
-                and data.dqx_data.any() != 0:
+        #First check if data is 2D
+        #If so check that data set has smearing info and that none are zero.
+        #Otherwise no smearing can be applied using smear from data (a Gaussian
+        #width of zero will cause a divide by zero error)
+        if self.data.__class__.__name__ == "Data2D":
+            if data.dqx_data is not None and data.dqy_data is not None \
+              and data.dqx_data.all()and data.dqy_data.all():
                 self.smear_type = "Pinhole2d"
```
mdoucet: The smearing type appears for 1D data, but not for 2D data. Should it appear at least in the text when "Use dQ data" is selected, just to be clear?

src/sas/sasgui/perspectives/fitting/fitpage.py
butlerpd: Not sure why these were labelled this way - agree it makes no sense.. doubly since dqs is said to be "perpendicular."  Since 2D dq is in polar co-ordinates have replaced both the title and tool tip with dq_r and dq_phi.  Tooltip indicates this is polar coordinates. Does this resolve the issue?

src/sas/sasgui/perspectives/fitting/fitpage.py
butlerpd: Hadn't noticed and agree in principle.  However for 2D the original authors chose to replace the `type` with `accuracy`  on the grounds that the heavy computation here needed it I guess?  I suspect this was added much later at a user request and it was decided there really wasn't room for both? and/or that trying to keep both would be too ugly code wise?  I think this would be hard for space reasons.  Since 2D is not often used yet and ONLY allows pinhole I guess that is OK? perhaps at least OK enough to get us through to the new GUI of 5.0?  Or does anyone have a suggestions as to where/how to make the `pinhole2D` text appear?

src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -1600,91 +1615,129 @@ def _get_smear_info(self):
...
-            if data.dqx_data is None or data.dqy_data is None:
-                return
-            elif self.current_smearer is not None \
-                and data.dqx_data.any() != 0 \
-                and data.dqx_data.any() != 0:
+        #First check if data is 2D
+        #If so check that data set has smearing info and that none are zero.
+        #Otherwise no smearing can be applied using smear from data (a Gaussian
+        #width of zero will cause a divide by zero error)
+        if self.data.__class__.__name__ == "Data2D":
```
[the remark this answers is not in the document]
butlerpd: That's what I was looking for!! hated those ugly if statements :-)  Sadly need help here.  Was unable to make `isinstance` work, even after importing it and even importing `datainfo` for good measure.  Get the error that `Data2D is undefined global name`
Help?

butlerpd:
ready for testing on Win64

src/sas/sasgui/perspectives/fitting/fitpage.py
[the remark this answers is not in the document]
mdoucet: If you imported `data_info`, you would have to refer to the class as `data_info.Data2D`. 

Something like this might work (with the import at the top of the file...).

```
from sas.sascalc.dataloader.data_info import Data2D
if isinstance(self.data, Data2D): [...] 
```

src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -1600,91 +1615,129 @@ def _get_smear_info(self):
...
+                self.msg = "2D Data did not contain recognizable " \
+                           "resolution info."
+                logger.info(self.msg)
+        #If not check that data is 1D
+        #If so check for pinhole vs slit by veryfing whehter dx or dxl or dxw
+        #have data (currently sasview only supports either dx or dxl/dxw but
+        #not both simultaneously) and, as for 2D, are non zero .
+        #Otherwise no smearing can be applied using smear from data (a Gaussian
+        #width of zero will cause a divide by zero error)
+        elif self.data.__class__.__name__ == "Data1D":
```
mdoucet: Here's another place for an `isinstance` call.

src/sas/sasgui/perspectives/fitting/fitpage.py
```diff
@@ -3181,11 +3236,17 @@ def _set_smear_buttons(self):
         if self.data.__class__.__name__ == "Data2D" or \
                     self.enable2D:
             self.slit_smearer.Disable()
-            self.pinhole_smearer.Enable(True)
             self.default_mask = copy.deepcopy(self.data.mask)
+            if self.model is not None:
+                self.pinhole_smearer.Enable(True)
+
+        elif self.data.__class__.__name__ == "Data1D":
```
mdoucet: Here's another place for an `isinstance` call.

src/sas/sasgui/perspectives/fitting/fitpage.py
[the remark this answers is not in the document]
butlerpd: Ah  good point - that should fix it, but will need to import both Data2D and Data1D. Though why is it necessary to import from dataloader in the first place when the object clearly knows it is of class Data2(1)D seems silly :-)  Will give it a go. Thanks

src/sas/sasgui/perspectives/fitting/fitpage.py
butlerpd: yes ... many places that nasty thing is .... if I can get isinstance to work..

butlerpd:
ready for testing on Win64

smk78:
If one wanted to put a big fat cherry on top of this fix, it would be to make the dQ/Q boxes change colour if the value is 'excessive' in order to draw the users attention to it.

butlerpd:
Agreed that would be quite nice.  However, after some thought I think the "big fat cherry" begets much more topping :-)  basically we would want to be consistent with all the boxes (whether from data or user supplied and pinhole or slit), then one would of course want some kind of tool tip that indicates WHY it has gone red and what to do about it, and finally we would need to agree where to put the somewhat arbitrary line between red and not.  Since the point is to get this out sooner than later I will put in a new ticket for this against milestone 4.3 so we can close this PR

PR #204, opened 2019-01-22
Ticket 1223: Expand permitted range of transformed data in Corfunc implementation

rprospero:
This upgrades the default corfunc range from 200 to 1000.  For version 5, we should make this into a controlable parameter, but this will be enough for 4.2.1


smk78:
Closes #1223

PR #205, opened 2019-02-03
Prep4 release4 2 1

butlerpd:
Aside from the usual notes, the code was changed so that neither local_config.py nor the sphynxdocs config.py need updating during a release (have also updated the release instructions on the wiki).  What needs special checking:
1) Did I capture all the changes from 4.2.0 in the release notes (I had missed all of the sasmodels changes in the first round)
2) Are my "code" changes Ok - they seem to work locally on my machine

**NOTE:** need the new zenodo DOI inserted

addresses #946


butlerpd:
ready for testing on OSX

butlerpd:
ready for testing on Win64

docs/sphinx-docs/source/user/RELEASE.rst
```diff
@@ -9,6 +9,35 @@ Release Notes
 Features
 ========
 
+New in Version 4.2.1
+--------------------
+The major changes for this point release were to fix several problems with using
+the built in editor to create new models and to bring the NXcanSAS reader into
+compliance with the final published specification.  The orignal reader was
+based on a draft version of the specification.  As an eraly adopter,
```
krzywon: typo

src/sas/sasview/__init__.py
```diff
@@ -1,2 +1,6 @@
-__version__ = "4.2.0"
+__major_version__ = "4"
+__minor_version__ = "2"
+__version__ = __major_version__ + "." + __minor_version__ + ".1"
```
krzywon: If we are going to separate out versions, I think we should also do the point revision.
`__point_version__ = "1"`
`__version__ = __major_version__ + "." + __minor_version__ + "." + __point_version__ `

src/sas/sasview/local_config.py
```diff
@@ -46,12 +48,20 @@
...
+"Grant No 654000.")
 _acknowledgement_citation = \
-'''M. Doucet et al. SasView Version 4.2, Zenodo, 10.5281/zenodo.1412041'''
+'''M. Doucet et al. SasView Version %s, %s''' % (__version__, __DOI__)
 _acknowledgement =  \
-'''This work was originally developed as part of the DANSE project funded by the US NSF under Award DMR-0520547,\n but is currently maintained by a collaboration between UTK, UMD, NIST, ORNL, ISIS, ESS, ILL, ANSTO, TU Delft, DLS, BAM and the scattering community.\n\n SasView also contains code developed with funding from the EU Horizon 2020 programme under the SINE2020 project (Grant No 654000).\nA list of individual contributors can be found at: http://www.sasview.org/contact.html
-'''
+("This work was originally developed as part of the DANSE project funded by"
+"the US NSF under Award DMR-0520547,\n but is currently maintained by a"
+"collaboration between UTK, UMD, NIST, ORNL, ISIS, ESS, ILL, ANSTO, TU Delft,"
```
krzywon: Contributors appear multiple times. Separating them out could be useful, but might not be straight forward.

src/sas/sasview/__init__.py
butlerpd: indeed.  My real question is whether we need major and minor.  The reason is that the sphinx config uses 4.2 as the version number AND 4.2.1 as the release number -- which makes no sense to me and I have not found where that "version" number (4.2) is ever used.... but I didn't have the guts to remove it.   My feeling is we remove that? and get rid of major and minor here? or do as you say for full flexibility. Maybe something that can be discussed on a call?

src/sas/sasview/local_config.py
butlerpd: agreed .... on both points at this time. I would like to think a bit about what the right answer is on this one I think?

docs/sphinx-docs/source/user/RELEASE.rst
butlerpd: done

docs/sphinx-docs/source/user/RELEASE.rst
```diff
@@ -9,6 +9,35 @@ Release Notes
...
+* Fixes # 1142:	Plugin framework is broken
+* Fixes # 1183:	Test from creating new model reset all parameters to default in all open FitPages
+* Fixes # 1205:	4.2 set weighting choice seems to be ignored.
+* Fixes # 1206:	Incorrect (and confusing) presentation of dQ from data in instrumental smearing section
+* Fixes # 1212:	Bug in Iqxqy plotting non rectangular / square matrices?
+* Fixes # 1221:	ABS reader does not read in USANS data properly	GitHub
+* Fixes # 1222:	smearing options incorrect on show2D and show1D in fitpage14: Loading a saved project is really really slow
+* Fixes # 1223:	Expand permitted range of transformed data in Corfunc implementation
+
+
```
pkienzle: I removed the ':' from magnetic parameter names since the 0.98 sasmodels release.  The saved models should be adjusted automatically on load, but you may want to mention the change in the release notes.  Here's a diff that shows the renaming:
```
+def _rename_magnetic_pars(pars):
+    """
+    Change from M0:par to par_M0, etc.
+    """
+    keys = list(pars.items())
+    for k in keys:
+        if k.startswith('M0:'):
+            pars[k[3:]+'_M0'] = pars.pop(k)
+        elif k.startswith('mtheta:'):
+            pars[k[7:]+'_mtheta'] = pars.pop(k)
+        elif k.startswith('mphi:'):
+            pars[k[5:]+'_mphi'] = pars.pop(k)
+        elif k.startswith('up:'):
+            pars['up_'+k[3:]] = pars.pop(k)
+    return pars
```

I also twiddled the handling of multiple scattering, but this isn't available yet through the sasview interface.

Spinodal docs have been updated.

smk78:
I've explicitly mentioned the removal of the colons as a fix to #1188, as this wasn't clear in the release note for same on 4.2.0.

But the release note for 4.2.0 Fix #1167 is explicit enough about the Spinodal Model docs being updated. Or do you mean they have been updated again, PaulK? I looked but couldn't see a commit...

pkienzle:
Yes the magnetism name change (ticket 1188) is in 0.98, but the docs weren't updated until after. 

Ticket 1189 (some magnetic parameters need to be hidden for multiplicity models) was also fixed in 0.98.

The spinodal doc updates were posted after the release.

Since multiple scattering is not yet in sasview, we don't need to mention it.

Here is the commit log since the version number was changed to 0.98:
```
94bfa42 fix typo
d61465c Removing dummy test file
f9014c7 Dummy commit to test webhook
8803a38 Merge branch 'master' of github.com:sasview/sasmodels
2c4a190 simplify use of multiple scattering in bumps fits
899e050 Merge pull request #93 from SasView/ticket-1015-quick-fix
9d8a027 add docs for 'model_info=...' to plugin guide
a4f1a73 clean out existing kernels when changing compute engine
c6084f1 Merge pull request #75 from SasView/ticket-1142-plugin-reload
275b07d Merge pull request #87 from SasView/ticket-1074-gammainc
df87acf Replaced : in parameter names with _ in magnetism.rst
93fe8a1 Updated spinodal documentation
dcc51ee Merge pull request #89 from arm61/add_requirements
783e76f adding requiresments to setup.py
ce1eed5 Fixed conflict after merging master
74bcfe9 added a requirements.txt file for building from source
57c609b Note bad implementation of JN and Si in docs
fba9ca0 add gammaln and gammainc to the opencl math library
==== release 0.98  ===
353e899 Merge pull request #84 from SasView/ticket-1188-magnetic-colon
e39202b Merge pull request #85 from SasView/ticket-1188-magnetic_shells  <<< should be 1189
bd547d0 restrict magnetic parameters to visible shells. Refs #1188  <<< should be 1189
610ef23 rename magnetic parameters to not contain colon. Refs #1188
a1ec908 Sasmodels version number updated to 0.98
```

smk78:
Are the Release 0.98 commits above or below the ===?
And how do we want to incorporate these into the Release Notes? Have bullet like 'Updated to Sasmodels 0.98' and then underneath list these commits (without the commit tag)?

pkienzle:
Newest to oldest.  Release 0.98 is tagged at commit id 353e899.

This is the result of "git log --oneline", so easy enough for anyone to generate who cares enough about the details.  No need to put it in the release notes.

I also look at the diff between releases, e.g.,  "git diff 353e899", when I'm writing release notes, just to see if anything was not mentioned in the release notes.

PR #206, opened 2019-02-20
FIX: add a proper entry point

mrakitin:
Dear developers,

I attempted to install this package at NSLS-II, and had some problems with executing it once it's installed. This PR attempts to resolve the entry point, so that one does not have to run `python run.py`. Instead, simple calling of `sasview` will pop up the GUI window. The PR is based on the [v5.0-beta.2](https://github.com/SasView/sasview/releases/tag/v5.0-beta.2) tag.

I hope it will help to make the package easier to use. Our plan is to create a corresponding conda package in our channel https://anaconda.org/lightsource2-tag based on the updated installer. Please let me know if there is an interest in contribution.

Regards,
Maksim Rakitin
DAMA group, NSLS-II, BNL

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -196,6 +196,7 @@ class BuildSphinxCommand(Command):
         import build_sphinx
         build_sphinx.rebuild()
 
+_ = subprocess.call([sys.executable, "src/sas/qtgui/convertUI.py"])
 
 # sas module
 package_dir["sas"] = os.path.join("src", "sas")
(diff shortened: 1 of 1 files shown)


rozyczko:
The proposed improvement is to be able to run SasView after installing in python site-packages directory. Instead of running python ./run.py in the main sasview directory, the user needs to only type sasview, as long as the python path is properly defined.
This is a good idea and thanks for addressing the issue - we never attempted at running SasView this way, at least not 5.0.
However, minor additional work is needed, as some images (most notably, the initial splash screen) are not showing up. We'll have a look at it.
I will accept/merge the PR - thanks again.
-piotr

mrakitin:
Dear [Piotr](@rozyczko), thank you for accepting and merging the PR!

Please note, that during working on the conda package for sasview (https://github.com/NSLS-II/lightsource2-recipes/pull/567), I noticed that [this line](https://github.com/SasView/sasview/pull/206/files#diff-2eeaed663bd0d25b7e608891384b7298R199) is not necessarily will finish successfully as Python may not have the required tool from the PyQt5 dependency, as the dependencies are installed at a later time within `setup(...)`. Feel free to remove that line.

As for the splash screen, I realized that I didn't copy all images (e.g., `sas/qtgui/images/*` were not included). I also noticed `sas/qtgui/UI/res/SVwelcome_mini.png` wasn't copied either. What's confusing, there are a few inconsistent sets of images, but since it's a beta version, I hope it will be resolved before releasing the stable version.

By the way, did you consider using the [versioneer](https://github.com/warner/python-versioneer)? That package can automatically handle versioning based on the tags and commits in Git. Please let me know if you need an example of it.

PR #207, opened 2019-03-05
jenkins server for windows using older numpy

pkienzle:
Use the new default for lstsq(..., rcond) in a way that is compatible with older versions of numpy.


pkienzle:
ready for testing on Win64

PR #209, opened 2019-03-12
Use dQ/|Q| for 2-D resolution

pkienzle:
Custom pinhole smearing should use dQ/|Q| rather than dQ/Q [http://trac.sasview.org/ticket/1242](Ticket #1268).

### DIFF
--- a/src/sas/sasgui/perspectives/fitting/fitpage.py
+++ b/src/sas/sasgui/perspectives/fitting/fitpage.py
@@ -2363,8 +2363,8 @@ class FitPage(BasicPage):
             elif self.dx_percent is not None:
                 percent = self.dx_percent/100
                 if self._is_2D():
-                    data.dqx_data[data.dqx_data == 0] = percent * data.qx_data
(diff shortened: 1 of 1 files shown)


butlerpd:
ready for testing on Win64

butlerpd:
Given it is only for pinhole presumably that may be OK .... unless you start worrying about gravity?  So question remains if we want to allow asymmetry in dqx and dqy?

prjemian:
Bonse-Hart instruments have factor of 200x (or more) difference between dQx
and dQy.

prjemian:
... and... B H instruments can be used to collect 1D data.  Tedious process
that but there has been at least one series of such measurements.

pkienzle:
Any more complex resolution calculation is going to require more assumptions or more inputs.  I created [ticket 1271](http://trac.sasview.org/ticket/1271) which suggests defining resolution for the fit page from the resolution calculator panel.  

Meanwhile, is the current fix sufficient to merge and close the ticket?

ajj:
Looks fine to approve. Just make sure that #1271 has a reference to these commits to remind whoever works on it.

PR #210, opened 2019-03-12
Ticket 1243: Issue loading projects in v4.2.1

krzywon:
This is a fix for the issue found by Kilian Frank. Two methods were combined in file_reader_base_class.py but the old methods were still being called when loading projects. This will be an issue for all users attempting to load projects in v4.2.1.

This branch is based off the v4.2.1 release point.

Relevant ticket: [#1243](http://trac.sasview.org/ticket/1243)


pkienzle:
`FileReader` base class always calls sort_data after loading, so `get_file_contents()` shouldn't need to do so.

---

Further suggestions outside the scope of this ticket.

(1) It is impolite to change the function signature on a derived class.  Can you rename `get_file_contents` to `_get_file_contents` and add:
```
def get_file_contents(self):
    return _get_file_contents(self, xml_file=self.filepath, schema_path="", invalid=True)
```

(2) Similarly, you should be able to eliminate the `cansas_reader.Reader.read()` method, and instead recurse directly into `_get_file_contents` with:
```
126:     # Load data with less strict schema
127:     _, self.extension = os.path.splitext(os.path.basename(xml_file))
128:     self._get_file_contents(xml_file=xml_file, schema_path=invalid_schema, invalid=False)
```
Still somewhat confusing recursion, but at least it is obvious that the recursion is happening.

(3) Getting rid of `_is_call_local()` would also be nice, maybe with a `retry` flag sent  to `_get_file_contents` and then onto `_parse_entry` when retrying with the less strict schema.

(4) Is there a reason to distinguish between uppercase and lowercase extensions?  If not, then eliminate line 127 above since `self.extension` is already extracted and converted to lowercase in the `self.extension` attribute.

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

butlerpd:
I noticed that attempting to load files of unknown format still loads nothing but fails to give any indication that it failed or why.  There used to be a message that popped up in a window and was logged to the console indicating it was not loaded and why.  When was this changed? Guessing prior to 4.2.1?

I don't think it is really polite to fail without notification of any kind?  If this was a change for 4.2.1 (or 4.2.0)  I would suggest that
* If it is trivial to add back we do for 4.2.2.
* If not we fix in master

krzywon:
I'm looking into both of these issues. Thanks for bringing them to my attention.

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

krzywon:
ready for testing on Win64

krzywon:
ready for testing on Win64

krzywon:
ready for testing on Win64

krzywon:
ready for testing on Win64

krzywon:
The latest build has fixed the issues outlined by @butlerpd but the logging issue (#1297) still exists.

butlerpd:
It does now seem to work for 2D as well. I had some trouble with some but not all the svs saved during previous pull request tests so it may be that the writing was also not working entirely properly? or just that my file got corrupted.  As noted the logging stills seems broken, but I think this PR can be merged into 4.2.2.

butlerpd:
but note there have been quite a few code changes now (not just a "2 line change") Perhaps @pkienzle should take another look before merging?

krzywon:
Those changes are probably too big for the current PR. I'll open a new ticket for 4.3 to address them.

PR #211, opened 2019-03-20
py3/wx4 compatibility changes for gui

pkienzle:
#1260 (from trac [#1233](http://trac.sasview.org/ticket/1233))
#1273 (from trac [#1249](http://trac.sasview.org/ticket/1249))
#1286 (from trac [#1266](http://trac.sasview.org/ticket/1266))

* Python 3 is strict about backslash escape sequences.  In python 2, `\k` would be left alone since `k` is not a special character, but in python 3 backslash must be followed by a special character (see [table](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals)).

* wx 4 drops the `...Tuple()` methods; remove the `Tuple` from the name and is still works since the returned `Size` object supports `w, h = Size()` same as a tuple.

* simple lint cleanup, such as removing unnecessary backslash at the end of lines, etc.

* a **bug fix on mac**, which traps the user in an invisible modal dialog on > Image Viewer > Convert to Data, even in the released 4.2 series.  (File `src/sas/sasgui/perspectives/calculator/image_viewer.py`)

* remove a frequent **error in plot draw_idle**, wherein panel was being changed without checking if it was None first. (File `src/sas/sasgui/plottools/canvas.py`)

* FileDialog now uses wx.FD_SAVE rather than wx.SAVE (works in both wx 3 and 4).

* Model load/save works with wx4.

Tested somewhat on py2/wx3 and py3/wx4 for mac.

### DIFF
--- a/src/sas/sasview/wx4cruft.py
+++ b/src/sas/sasview/wx4cruft.py
@@ -19,12 +19,12 @@ def patch_py_editor():
 def read(self):
     """Return contents of file."""
     if self.filepath and os.path.exists(self.filepath):
-        with io.open(self.filepath, 'r', encoding='utf8') as f:
+        with io.open(self.filepath, mode='r', encoding='utf8') as f:
             return f.read()
     else:
         return ''
 
 def write(self, text):
     """Write text to file."""
-    with io.open(self.filepath, 'w', encoding='utf8') as f:
+    with io.open(self.filepath, mode='w', encoding='utf8') as f:
         f.write(text)
--- a/src/sas/sasview/wx4cruft.py
+++ b/src/sas/sasview/wx4cruft.py
@@ -1,0 +1,30 @@
+import os
+import sys
+import io
+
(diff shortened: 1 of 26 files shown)


src/sas/sasview/wx4cruft.py
```diff
@@ -0,0 +1,30 @@
...
+    to only support unicode text, whereas the old version of Document in
+    python 2.7 only supports bytes.
+    """
+    wx.py.document.Document.read = read
+    wx.py.document.Document.write = write
+
+def read(self):
+    """Return contents of file."""
+    if self.filepath and os.path.exists(self.filepath):
+        with io.open(self.filepath, 'r', encoding='utf8') as f:
```
rozyczko: All but the first arguments of io.open() are keyword not positional, so it would be consistent to use 
`io.open(self.filepath, mode='r', encoding='utf-8')`

PR #213, opened 2019-03-25
Sasview 1079

rprospero:
I've fixed the Corfunc perspective so that both the Q-Space extrapolation and the real space transform are both plotted.  Additionally, the plots are now included alongside the dialog on the right, instead of being squeezed in at the top, which make the plots easier to read.  Finally, I've added axes labels to the plots for additional clarity.

Ignore to odd colour scheme in the imags below.  That's a known issue with my computer alone.

Before

before

After

after

### DIFF
--- a/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
+++ b/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
@@ -117,6 +117,8 @@ class CorfuncWindow(QtWidgets.QDialog, Ui_CorfuncDialog):
 
         self._canvas = MyMplCanvas(self.model)
         self.mainVerticalLayout.insertWidget(0, self._canvas)
+        self._realplot = MyMplCanvas(self.model)
+        self.mainVerticalLayout.insertWidget(1, self._realplot)
 
         # Connect buttons to slots.
         # Needs to be done early so default values propagate properly.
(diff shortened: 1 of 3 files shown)


rprospero:
I've gone ahead and removed all of the limits from the corfunc transform, so that we don't need to worry about extending it further down the road.  Instead, there's a reasonable default plotting range and, more importantly, I've added the navigation toolbars so that the user can explore more of the plot range if they desire.

2019-03-26-094953_1918x1031_scrot

src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
```diff
@@ -6,6 +6,7 @@
 # global
 from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg \
     as FigureCanvas
+from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT
```
rozyczko: We don't use the Qt toolbar elsewhere, instead relying on our own Plotter class with a number of context menu items.

src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
```diff
@@ -77,14 +84,17 @@ def draw_real_space(self):
         self.axes = self.fig.add_subplot(111)
         self.axes.set_xscale("linear")
         self.axes.set_yscale("linear")
+        self.axes.set_xlabel("Z [$\AA$]")
+        self.axes.set_ylabel("Correlation")
+        self.fig.tight_layout()
```
rozyczko: Would also be great if the plots had titles as it makes it difficult to tell which graph is showing what, especially when no data is loaded.

src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
rozyczko: This might not be a serious issue, considering the relevant charts are added as indices to the dataset in the Data Explorer and can be viewed with "Create New" or "Quick Plot"

rprospero:
I've fixed the vertical spacing and added tiles to the plots, as well as a legend on the data plot.  I did also remove the matplotlib toolbars.  I took a quick look at replacing the two plots with a proper plotting widget and saw that it would be a significant rewrite, but, as you pointed out, the results are still in the data explorer.

2019-03-26-151006_1918x1031_scrot

PR #217, opened 2019-03-29
Ticket 1009: Added type of distribution used for PD parameters in report

gonzalezma:
This small change adresses the initial purpose of ticket 1009, i.e. adding the information about the type of polydispersity distribution used in the fit to the report. The following discussion about the report format and contents needs more discussion and work.

### DIFF
--- a/src/sas/sascalc/fit/pagestate.py
+++ b/src/sas/sascalc/fit/pagestate.py
@@ -490,7 +490,12 @@ class PageState(object):
         Helper method to print a state
         """
         for item in list:
-            rep += "parameter name: %s \n" % str(item[1])
+            if str(item[1][-5:]) == 'width':
+                par = str(item[1][:-6])
+                pd_type = str(self.model.dispersion[par]['type'])
+                rep += "parameter name: %s (%s) \n" % (str(item[1]), pd_type)
+            else:
+                rep += "parameter name: %s \n" % str(item[1])
             rep += "value: %s\n" % str(item[2])
             rep += "selected: %s\n" % str(item[0])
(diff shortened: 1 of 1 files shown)


src/sas/sascalc/fit/pagestate.py
```diff
@@ -490,7 +490,12 @@ def _repr_helper(self, list, rep):
         Helper method to print a state
         """
         for item in list:
-            rep += "parameter name: %s \n" % str(item[1])
+            if str(item[1][-5:]) == 'width':
+                par = str(item[1][:-6])
```
pkienzle: Please check that it works with a squarewell model, that has the parameter `wellwidth`.  You probably want to match against `'.width'` rather than `'width'` to protect against this.  Couldn't test since I don't know how to trigger the code.

gonzalezma: Thanks Paul. You were right, there was an error with the squarewell model. Changed the comparison to '.width'. To test, make a fit with at least 1 PD parameter and call Report results. Previous output was e.g.: radius.width = 0.1 +- 0.03. Now it will be radius.width (gaussian) = 0.1 +- 0.03.
Question for Richard: Do you think it is worthy to add also the number of points and sigmas? E.g.:
radius.width (gaussian, points=35, sigmas=3) = 0.1 +- 0.03.

pkienzle:
Not merging since the question of whether you want the show #sigma and #points is unresolved.

gonzalezma:
I added the information about npts and nsigmas, as suggested by Richard. Tested and worked as expected for me.

PR #218, opened 2019-03-29
Clean up obsolete code of magnetic scattering calculation in Sasview

dehoni:
I think that the code in sascalc/calculator/c_extensions/libfunc.c:cal_msld has no effect and is obsolete together with sascalc/calculator/c_extensions/sld2i.c:genicomXY.

### DIFF
--- a/src/sas/sascalc/calculator/c_extensions/libfunc.c
+++ b/src/sas/sascalc/calculator/c_extensions/libfunc.c
@@ -86,6 +86,8 @@ double sinc(double x)
   return sin(x)/x;
 }
 
+//The code calculating magnetic sld below seems to be not used.
+//Keeping it in order to check if it is not breaking anything.
 // calculate magnetic sld and return total sld
 // bn : contrast (not just sld of the layer)
 // m0: max mag of M; mtheta: angle from x-z plane;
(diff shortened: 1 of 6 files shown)


pkienzle:
It appears to be referenced from [sas_gen.py](https://github.com/SasView/sasview/blob/b36e7c726c1824fc66e2a5a76abb7351c49d66e6/src/sas/sascalc/calculator/sas_gen.py#L152).

rprospero:
Confirming that `cal_msld` is used by `genicomXY`, which is used by the `runXY` method of the `GenSAS` class in sas_gen.py, which is used by the `complete` method `SasGenPanel` class in gen_scatter_panel.py, with is used by the `SasGenWindow` class in the same file, which is used by the `Plugin` class in calculator.py.

As much as I'd like, I don't think that we can get rid of this code quite yet.

butlerpd:
So does this mean we should close this pull request?

butlerpd:
Just realized that the second commit in this pull request : *Make polarisation efficiency weights consistently the same as in SASmodels* is all done within the code that was commented out in the original commit.  

Given that the first comment by @pkienzle  against that original commit, and preceding the second commit, suggests that the removed code probably is used by the general scattering calculator, I am thinking that this was in preparation for restoring the code?  In fact I  have a vague recollection when some changes were made to the normalization of magnetic scattering code there was a comment that the same fix needed to be added to the magnetic code in the general sas calculator?  This also suggests that maybe these changes tested ok because it only affects the general scattering calculator?

If so, and there is no clever git command to do it, I am willing to take those changes (that only affect the one file: libfunc.c), kill the PR and then put the changes back into libfunc.c?

Is this correct?

pkienzle:
In the generic calculator the weight scale rho which is then squared to produce I(q).  In sasmodels the weights scale I(q) which contains rho^2.  Therefore there is no need to change gensas.

The bug in sasmodels came about when I rearranged the calculation for magnetic scattering, and blindly applied it to I(q) rather than rho(r).

I wrote a python+numpy version of gensas in [sasmodels/explore/realspace.py](https://github.com/SasView/sasmodels/blob/realspace_magnetic/explore/realspace.py#L394-L640) to test magnetic spheres and ellipsoids against analytic models (see SasView/sasmodels#363).  This could be used to replace the GenSas calculator in sascalc, or help with simplifying it.  Note that the absolute scaling doesn't match the analytical form; perhaps the units don't match, or some other scale factor is missing.

pkienzle:
Closing this PR since it is subsumed by #1403 and #1405.

PR #1298, opened 2019-04-02
Opencl drivier update when running Test 

wpotrzebowski:
OpenCl driver information wasn't available when running tests. These commit should fix it in 5.0. Progress bar will be implemented separately.


wpotrzebowski:
5.0 ready for testing on Win

Issue #1295, opened 2019-04-01
5.0 volfraction in S(Q)


RichardHeenan:
I am opening a new ticket for this, though it has in part been covered by a number of other tickets related to beta approx project and P(Q)S(Q) (most of which can now be closed).

The radius_effective and beta approx parts now work, though as usual a lot more unit tests are needed.   [Also some issues with vesicle mode.]

The final part of the "beta approx, plus improve S(Q)" project was not started as the UK student, Torin, ran out of time. It was intended that we sort out the the use of volfraction and scale when using S(Q).

What we have currently in v5.0 regarding scale & volfraction is the same as v4.2 , without S(Q) then:
I(Q) = scale.P(radius,Q)    where scale is volume fraction if working in absolute units
with S(Q) then:
I(Q) = scale.Svolfrac.P(radius,Q).S(Q,radius_effective,volfraction_effective) 

This is confusing for users as the value and meaning of scale changes on adding an S(Q), it would be much better is this did not happen.

We should in v5.0 add a third drop down in the fitting gui to choose whether to constrain volume fraction in S(Q) or let it freely adjust, i.e. volfraction_effective_mode=1 for constrained, 0 for freely adusting.  The drop down could appear alongside the one for radius_effective_mode. 

It has been suggested that any P(Q) which can be multiplied by an S(Q) should have an explicit volfraction parameter added, but I don't think that is actually necessary.

I would **propose** -     I(Q)=scale.P(radius,Q).S(Q,radius_effective,volfraction_effective)

Where if working in absolute units, then just as in v4.2,  scale = phi the volume fraction of particles, but now Svolfrac is only affects the shape of S(Q) and not the overall height of I(Q).

In the main fitting gui selecting volfraction_effective_mode=1, constrains Svolfrac = scale
                                                   volfraction_effective_mode=0,  allows Svolfrac to be fitted

In sum/multi plugin models then volfraction_effective_mode defaults to 0, exactly as v4.2

If loading a P(Q)S(Q) from sasview 4, then scale has to be set to scale_v4*volfraction

I think that this should be done before releasing 5.1 to the general public to avoid confusion later.

Would this work ? - your opinions please.

We could **instead** add volfraction explicitly to all P(Q) models that can take an S(Q), so that 

I(Q)=scale.volfraction.P(R,Q,volfraction).S(Q,Reff,Svolfrac)

But I  don't think that there any P(Q) that actually need volume fraction as a genuine parameter ? 

NOTES
[...]

Issue #1280, opened 2019-03-30
vesicle @ structure_factor incorrect (Trac #1257)


pkienzle:
Migrated from http://trac.sasview.org/ticket/1257
```json
{
    "status": "new",
    "changetime": "2019-03-25T21:00:59",
    "_ts": "2019-03-25 21:00:59.327232+00:00",
    "description": "Comparing vesicle@S with no core to sphere@S I get different answers.  Repeat using:\n{{{\n    $ sascomp vesicle@hardsphere,sphere@hardsphere radius=0,thickness thickness=40 -pars -midq -double\n}}}\n\nThis is not true for hollow_cylinder with no core vs cylinder:\n{{{\n    $ sascomp hollow_cylinder@hardsphere,cylinder@hardsphere radius=0,thickness thickness=40 -pars -double\n}}}\n\nI think the issue is that the vesicle model defines volfraction, and this is colliding with the definition of volfraction in the structure factor, which then gets turned into volfraction_S.\n\nVescicle and sphere match if volfraction_S is set to the same value as volfraction.\n{{{\n    $ sascomp vesicle@hardsphere,sphere@hardsphere volfraction_S=0.05 scale=20,1 radius=0,thickness thickness=40 -pars -midq -double\n}}}\n\nShould update product model to use volfraction if it exists in the model.\n\nThis will require changes to `sasmodels/product.py`, both when constructing the model and when splitting the parameter set into P and S parameters.\n",
    "reporter": "pkienzle",
    "cc": "",
    "resolution": "",
    "workpackage": "SasView Bug Fixing",
    "time": "2019-03-25T21:00:59",
    "component": "SasView",
    "summary": "vesicle @ structure_factor incorrect",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView 4.3.0",
    "owner": "",
    "type": "defect"
}
```

RichardHeenan:
after merging https://github.com/SasView/sasmodels/tree/ticket_822_v5_unit_tests  into master,
then vesicle@hardsphere still has issues due to collision of volfraction name?
In 5.0 the two volfraction are reporting same numbers, need to be able to keep them different and either of them fittable.

pkienzle:
You don't need to fit the two volfractions independently.  Instead, fit scale and the tied volfraction, and interpret it as S.volfraction = tied volfraction, and P.volfraction = scale*S.volfraction.

[...]

PR #1307, opened 2019-04-22
Only show volfraction once if it appears in both P and S models.

pkienzle:
Refs #1280, #1295, sasview/sasmodels#219, sasview/sasmodels#199, sasview/sasmodels#101.

Removes confusion when volfraction appears in both S and P models.  No longer supports volfraction renamed to volfraction_S or any other name collisions.   

This is a short-term fix.  Longer term address #1190, #1194 and sasview/sasmodels#231 to better decouple the GUI and the sasmodels library.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingLogic.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingLogic.py
@@ -215,6 +215,8 @@ class FittingLogic(object):
         """
         plots = []
         for name, result in return_data['intermediate_results'].items():
+            if isinstance(result, tuple) and len(result) > 1:
+                result = result[1]
             if not isinstance(result, np.ndarray):
                 continue
             plots.append(self._create1DPlot(tab_id, return_data['x'], result,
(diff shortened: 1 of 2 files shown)


rozyczko:
Behaves as described.
I moved the body of the "hack" to a separate method for easier maintenance.
Ready to merge.

RichardHeenan:
Building 5.0 locally in sasview branch ESS_GUI_1280 and sasmodels branch ticket-1257-vesicle-product I can confirm that 5.0 seems to give same results as 4.2 for vesicle alone and for vesicle with hardsphere (after making allowance for different starting parameters and the different scaling etc).

I also tried core_shell_sphere alone and that was OK, HOWEVER core_shell_shere times hardsphere did not give the correct results by default.

Had loaded the AOT microemulsion shell contrast, was then fitting radius & scale.

Though the radius_effective drop down in 5.0 was showing "unconstrained" a fit of scale & radius gave the results for "constrained".  Changing the drop down to "constrained", doing a fit, then back to "unconstrained"  got it working, so there seems to be an initialisation issue here. Ought to check this with some other models, but before I waste time here ... perhaps some other related changes all need to be merged with master??? ( I am rapidly getting confused with the various tickets here.)

rozyczko:
the combobox initialization issue fixed in #1308.

wpotrzebowski:
5.0 ready for testing on OSX

PR #1330, opened 2019-05-09
Release 4.2.2b

pkienzle:
New 4.2.2 release candidate branched directly from 4.2.1 tagged release.

***DO NOT MERGE***

Old release candidate in branch release-4.2.2 included changes to master since release, and should be ignored.

Includes new zenodo tag that was created for the release-4.2.2 branch. Don't know how to redirect it to the end of this branch.

### DIFF
--- a/docs/sphinx-docs/source/user/RELEASE.rst
+++ b/docs/sphinx-docs/source/user/RELEASE.rst
@@ -11,9 +11,9 @@ Features
 
 New in Version 4.2.2
 --------------------
-This release fixes two bugs: the inability to read in project files due to the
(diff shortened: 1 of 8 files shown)


pkienzle:
ready for testing on Win64

pkienzle:
ready for testing on OSX

pkienzle:
The triggered builds will use sasmodels/master rather than sasmodels/v0.99, which was the release tagged for 4.2.1.  There are 100+ changes since then, the biggest of which is the handling of volume fraction for the vesicle@structure_factor model.  Some things make work differently from the final release.

PR #1343, opened 2019-05-29
Copy sascalc changes from master to ESS GUI

pkienzle:
One half of the synchronization process for sascalc: copy changes from master to ESS_GUI branch.

* Simplified handling of the remaining C modules so there is no artificial "core" subpackage beside lib; changed run.py to access them.   Also supports the new buffer interface since the old interface was deprecated and leads to compiler warnings in python 3.7.

* Synchronized changes to support python 3 across both branches.

* Some code cleanup on corfunc transform.

* Some code cleanup on file extension registry, incorporating the fix from ESS GUI and repairing the py3 support changes from master.

* Pull forward fixes to HDF reader and the corresponding tests.

* Put shim in ESS GUI code to work with the sasview 4.x pagestate object.  The interface can be improved when we discontinue support for the wx GUI.

* Pull in changes to sascalc/fit/model.py model library.  ESS GUI has its own copy of the model library code in the fitting perspective folder, but this patch does not touch it.

* Minor change to invertor so that lstsq rcond uses the new numpy default value, and in a way that works across different versions of numpy.

* Report parameters include polydispersity information.


rozyczko:
The MSVC compiler doesn't support `typeof`, so sld2i_module.c and Cinvertor.c fail at the VECTOR definition.
This can be fixed by using `decltype` but I'm afraid it's MS-specific and also > C++11 only. We'll have to introduce some platform specific code there.
Interestingly, removing `typeof` entirely and using
``
(...)
    VEC_buf = VEC_view.buf; \
    VEC_len = VEC_view.len/sizeof(*VEC_buf); \
``
seems to work fine on windows.

src/sas/qtgui/Utilities/GuiUtils.py
```diff
@@ -1369,15 +1369,19 @@ def readProjectFromSVS(filepath):
...
-    state_reader = Reader()
-    data_svs, state_svs = state_reader.read(filepath)
 
-    output = []
+    # CRUFT: SasView 4.x uses a callback interface to register bits of state
+    state_svs = []
+    def collector(state=None, datainfo=None, format=None):
+        if state is not None:
+            state_svs.append(state)
+    state_reader = Reader(callback=collector)
```
rozyczko: This currently breaks 5.0
`ERROR: Error while reading the project file: __init__() got an unexpected keyword argument 'callback'`

rozyczko:
Thanks for the work! I've been trying to keep sascalc in reasonable closeness to master but of course there were still things not copied over.
Currently, the two issues I see are the `typeof` problem and the xml loader failures (on loading old .svs projects).
I will have a closer look at how to approach them but not before 5.0.1.
Also, I would like not to merge such a large codebase before 5.0.1 is out.

rozyczko:
In the light of the recent bug report, which this PR addresses, I think it will be prudent to include this code in 5.0.1 after all.
Feel free to merge.

pkienzle:
I pulled the changes in from ESS_GUI and checked that the sascalc tests passed.  Didn't check the GUI.

Issue #1351, opened 2019-06-12
5.0 Corfunc cannot save plot data


rprospero:
While the 5.0 corfunc can calculate the corfunc graph, there is no way for the user to export it to a file.  There needs to be a save button that will enable the user to write out the real space data.

rprospero:
Solved with pr 1352

Issue #1339, opened 2019-05-28
5.0 corfunc mac doesn't plot


pkienzle:
Going through the correlation function documentation (SasView 4.2.2 since dmg for mac doesn't provide 5.0 docs), I see the following:
* Name of the dataset is not appearing in the Name box.
* Range bars on graph are not draggable.
* Input dialog not scrollable (but I can access the buttons on the bottom by closing the log explorer)
* sigma value is slightly different (5.8e-10 rather than 7.7e-10)
* transform type selector not available
* plots failing (see below for traceback)

```
15:17:48 - ERROR: SasView threw exception: x and y must have same first dimension, but have shapes (9073,) and (14251,)
Traceback (most recent call last):
  File "/Users/pkienzle/Source/sasview/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py", line 268, in finish_transform
    self._realplot.draw_real_space()
  File "/Users/pkienzle/Source/sasview/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py", line 101, in draw_real_space
    self.axes.plot(data3.x, data3.y, label="3D Correlation")
  File "/Users/pkienzle/anaconda/envs/py36/lib/python3.6/site-packages/matplotlib/__init__.py", line 1855, in inner
    return func(ax, *args, **kwargs)
  File "/Users/pkienzle/anaconda/envs/py36/lib/python3.6/site-packages/matplotlib/axes/_axes.py", line 1527, in plot
    for line in self._get_lines(*args, **kwargs):
  File "/Users/pkienzle/anaconda/envs/py36/lib/python3.6/site-packages/matplotlib/axes/_base.py", line 406, in _grab_next_args
    for seg in self._plot_args(this, kwargs):
  File "/Users/pkienzle/anaconda/envs/py36/lib/python3.6/site-packages/matplotlib/axes/_base.py", line 383, in _plot_args
    x, y = self._xy_from_xy(x, y)
  File "/Users/pkienzle/anaconda/envs/py36/lib/python3.6/site-packages/matplotlib/axes/_base.py", line 242, in _xy_from_xy
    "have shapes {} and {}".format(x.shape, y.shape))
ValueError: x and y must have same first dimension, but have shapes (9073,) and (14251,)
```

wpotrzebowski:
[...]

PR #1352, opened 2019-06-12
ESS Corfunc Save Plot Data

rprospero:
This adds a save button, which resolves issue #1351.  It also fixes a plotting issue on Linux that likely mirrors the plotting issue on Mac in #1339.


src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
```diff
@@ -423,4 +427,22 @@ def title(self):
...
+        """
+        Save corfunc state into a file
+        """
+        f_name = QtWidgets.QFileDialog.getSaveFileName(
+            caption="Save As",
+            filter="Corfunc Text Output (*.crf)",
+            parent=None)[0]
+        if "." not in f_name:
+            f_name += ".crf"
+
```
rozyczko: When cancelled the dialog, this will create a file called ".crf" and save the data to it. Probably not what expected.

src/sas/qtgui/Perspectives/Corfunc/UI/CorfuncPanel.ui
```diff
@@ -393,6 +393,13 @@
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QPushButton" name="cmdSave">
+         <property name="text">
+          <string>Save</string>
```
rozyczko: Would be useful to be more descriptive - what is being saved? Maybe add a tooltip?

src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
```diff
@@ -423,4 +427,22 @@ def title(self):
...
+    def on_save(self):
+        """
+        Save corfunc state into a file
+        """
+        f_name = QtWidgets.QFileDialog.getSaveFileName(
+            caption="Save As",
+            filter="Corfunc Text Output (*.crf)",
+            parent=None)[0]
+        if "." not in f_name:
+            f_name += ".crf"
```
rozyczko: When cancelled the dialog, this will create a file called ".crf" and save the data to it. Probably not what expected.

src/sas/qtgui/Perspectives/Corfunc/UI/CorfuncPanel.ui
rprospero: I've added tooltips for the main corfunc buttons.

rprospero:
I've added support for scrolling the main main panel of the dialog.  This was really only a four line change to the code, but meant changing the indentation on a huge number of lines, so keep that in mind when looking at the commits.

PR #1356, opened 2019-06-13
Load in Mask Data column from NCNR 2D data.

krzywon:
The 2D data format from the VSANS instrument at the NCNR is now outputting a mask flag for each data point. This should read in the mask and format it so SasView can use it.

NB - I have no data sets with the mask column to test. The latest release of the Igor Macros (v7.93 - June 12, 2019) adds the mask data to the reduced data file. This PR may (should?) never get merged into master because we are actively working on writing VSANS data in NXcanSAS, which may supercede the .DAT format.

### DIFF
--- a/src/sas/sascalc/dataloader/file_reader_base_class.py
+++ b/src/sas/sascalc/dataloader/file_reader_base_class.py
@@ -229,6 +229,11 @@ class FileReader(object):
                     data.dqy_data = data.dqy_data.astype(np.float64)
                 if data.mask is not None:
                     data.mask = data.mask.astype(dtype=bool)
+                    # If all mask elements are False, give a warning to the user
+                    if not data.mask.any():
+                        error = "The entire dataset is masked and may not "
+                        error += "produce usable fits."
+                        data.errors.append(error)
 
                 if len(data.data.shape) == 2:
                     n_rows, n_cols = data.data.shape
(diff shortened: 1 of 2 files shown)


butlerpd:
ready for testing on Win64

krzywon:
Masks loaded in from example files appear jagged along the masking edge when a detector tube is masked anywhere along the tube. When a full tube is masked, the masking edge appears clean. The mask appears to be loading in as written to the file, so the issue is currently being looked into on the Igor side.

The image below shows masked fits for two panels of a single VSANS measurement. sans34736_ML.DAT has vertical detector tubes, and the left mask edge is evenly applied, but the top and bottom are not. sans34736_FB.DAT has horizontal detector tubes, with rectangular masks on both sides of the detector and a triangular mask in the upper left corner. No part of the mask is evenly applied.

JaggedMasking

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

krzywon:
Looking through the console log of the pull request build, the build server is using the python p(r) invertor, rather than the C invertor still used by this PR. The 3 failing tests are the same 3 failing in master that are being addressed with #1387. @wpotrzebowski can you explain this behavior?

pkienzle:
The failures should be resolved by #1402.  I kept the fix minimal so it can be quickly reviewed and merged.

butlerpd:
ready for testing on Win64

krzywon:
An error/warning is *supposed to* pop up when a fully masked data set is loaded, but that is tied to what is put into the console log. On the Windows build for this PR, nothing is being written to the console log. I'm not sure if this is a build server issue or a SasView issue, but will take some investigating.

PR #1357, opened 2019-06-14
Ess corfunc drag bars

rprospero:
This pull request should only be merged after #1352.  It allows the q-range in corfunc to be set via dragging the bounds on the plot, instead of only via the text boxes.


rprospero:
I've added the patch you submitted (with a mere cosmetic change).  The live update is a usability improvement.

Issue #1293, opened 2019-03-30
Cannot model Sesans Data in 5.0 gui


rprospero:
Process:

Load any SESANS file
Send the dataset to fitting
Pick any model
Plot the data and the model
Observe that the model appears as a flat line

The 5.0 GUI does not seem to perform the Hankel transform on the model before plotting it, resulting in a mixture of q-space and real space data.

rprospero:
I've confirmed that the model isn't being transformed.  There are two factors at play here.  First, the current code assumes that data with no uncertainty on the x coordinate cannot be transformed, but that does not apply to SESANS data.  Secondly, the SESANS data is no longer automatically having the transform applied and a smearing must be manually applied in the GUI.

I've found a solution to the first issue and am working on the second.

rozyczko:
Any chance of having this addressed this week, before 5.0 gets released on Monday?

butlerpd:
I think the comment in #1363 was meant to close this ticket but did not.  Can we close it?

wimbouwman:
I have just tested it with old Delft-data. There the new version works! I am very happy.

It would be good to test it also with some recent ISIS-data. Adam, could you also test it or send me a recent ISIS-file?
Best, Wim

wimbouwman:
It works for both ISIS and Delft measurements. It can be closed.

PR #1363, opened 2019-07-19
Ticket 1293 Sesans Plotting

rprospero:
The critical code looks like a no-op.  The self.xscale and self.yscale are properties that will clear the plot when the axis value is changed.  Even if the value itself doesn't change.  Thus, we need to ensure that we only update the axis scaling when the axes have actually changed in value.

This Fixes pr #1293


wimbouwman:
I have tested it with ISIS and Delft SESANS measurements and it works fine. The ticket can be closed.

Issue #1038, opened 2019-03-30
use Si from cephes (Trac #964)


pkienzle:
The cephes implementation of Si is more accurate than that which we are using in sasmodels.  Consider using it instead.

Migrated from http://trac.sasview.org/ticket/964
```json
{
    "status": "closed",
    "changetime": "2018-11-15T16:25:28",
    "_ts": "2018-11-15 16:25:28.673703+00:00",
    "description": "The cephes implementation of Si is more accurate than that which we are using in sasmodels.  Consider using it instead.",
    "reporter": "pkienzle",
    "cc": "",
    "resolution": "duplicate",
    "workpackage": "SasView Bug Fixing",
    "time": "2017-05-18T21:39:24",
    "component": "SasView",
    "summary": "use Si from cephes",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView Next Release +1",
    "owner": "",
    "type": "defect"
}
```

pkienzle:
Trac update at `2018/11/15 16:25:28`: 
* **pkienzle** commented:

See #1227.


* **pkienzle** changed resolution from "" to "duplicate"
* **pkienzle** changed status from "new" to "closed"

PR #1370, opened 2019-08-09
Converting cinvertor python

zattala:
Speed should be almost identical to C with Numba, without for the initial load about 2x/3x slower but for individual calculations still almost identical to C. 

Calculated results the same with or without Numba.

Accuracy, has a e-15 difference on the final output, also because of truncated pi used in invertor.c the outputs will be slightly different, e-05, e-04 relative error, but with truncated pi used in calc.py should revert to default error of e-15. 

Would also like more tests with smeared data if possible.


src/sas/sascalc/pr/calc.py
```diff
@@ -0,0 +1,368 @@
...
+    of r that is at least one sigma above 0.
+    """
+    nslice_d = np.float64(nslice)
+    i_r = np.arange(nslice)
+    r = d_max/nslice_d * i_r
+    pr_vals = pr_err(pars, err, d_max, r)
+
+    pr_val = pr_vals[0, :]
+    pr_val_err = pr_vals[1, :]
+
```
pkienzle: Can use `pr_val, pr_val_err = pr_err(pars, err, d_max, r)` even if `pr_err` returns a matrix rather than a tuple, but the above form won't work for a tuple.

src/sas/sascalc/pr/calc.py
```diff
@@ -0,0 +1,368 @@
...
+    total = np.sum(np.fabs(pr_val))
+
+    index = pr_val > pr_val_err
+    total_pos = np.sum(pr_val[index])
+
```
pkienzle: Don't need `index` as a separate variable in this case (but it doesn't do any harm, either).

src/sas/sascalc/pr/invertor.py
```diff
@@ -583,13 +587,17 @@ def estimate_alpha(self, nfunc):
...
+
             elapsed = time.time() - starttime
             initial_alpha = pr.alpha
             initial_peaks = pr.get_peaks(out)
+            #Inital Alpha correct, suggested not
 
             # Try the inversion with the estimated alpha
             pr.alpha = pr.suggested_alpha
             out, _ = pr.invert(nfunc)
+            #This invert different
```
pkienzle: 590, 595 and 600 are not useful comments.

src/sas/sascalc/pr/p_invertor.py
```diff
@@ -0,0 +1,663 @@
...
+        """
+        Function to set the y data.
+
+        :param data: Array of doubles to set y to.
+        :return: ny - Number of entries found.
+        """
+        data = np.float64(data)
+        ndata = len(data)
+        self.__dict__['y'] = np.zeros(ndata)
+
```
pkienzle: `y` gets overwritten on next line; don't need to initialize it.

src/sas/sascalc/pr/p_invertor.py
```diff
@@ -0,0 +1,663 @@
...
+        self.__dict__['y'] = np.array(data, 'd')
```
pkienzle: I think `data = np.float64(data)` above already forces data to an array of type 'd'.  This statement makes a copy. If it needs to be a copy, use `data = np.array(data, 'd')` above instead.  In any case, simply assign `self.__dict__['y'] = data` here.

src/sas/sascalc/pr/p_invertor.py
```diff
@@ -0,0 +1,663 @@
...
+
+        :param data: Array of doubles to set err to.
+        :return: nerr - Number of entries found.
+        """
+        data = np.float64(data)
+        ndata = len(data)
+        self.__dict__['err'] = np.zeros(ndata)
+
+        self.__dict__['err'] = np.array(data, 'd')
+
```
pkienzle: As above, for `set_y`.

src/sas/sascalc/pr/p_invertor.py
```diff
@@ -0,0 +1,663 @@
...
+        :param pars: c-parameters.
+        :return: oscillation figure of merit.
+        """
+        nslice = 100
+        pars = np.float64(pars)
+        pars = np.atleast_1d(pars)
+
+        oscill = calc.reg_term(pars, self.d_max, nslice)
+        norm = calc.int_pr_square(pars, self.d_max, nslice)
+        ret_val = np.sqrt(oscill/norm) / np.arccos(-1.0) * self.d_max
```
pkienzle: Use `np.pi` instead of `np.arccos(-1.0)`.

src/sas/sascalc/pr/invertor.py
zattala: Removed, they were just there during debugging, forgot they were left.

src/sas/sascalc/pr/calc.py
```diff
@@ -0,0 +1,368 @@
...
+    """
+    Orthogonal Functions:
+    B(r) = 2r sin(pi*nr/d)
+
+    :param d_max: d_max.
+    :param n: n.
+
+    :return: B(r).
+    """
+    return (2.0 * r) * np.sin(pi*(n*r)/d_max)
```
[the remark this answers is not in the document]
zattala: Done.

src/sas/sascalc/pr/calc.py
```diff
@@ -0,0 +1,368 @@
...
+import sys
+import math
+import time
+import copy
+import os
+import re
+import logging
+import time
+
+#from __future__ import division # at the top, for the 2.7 users
```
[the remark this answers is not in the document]
zattala: Done.

src/sas/sascalc/pr/calc.py
```diff
@@ -0,0 +1,368 @@
...
+def ortho_derived(d_max, n, r):
+    """
+    First derivative in of the orthogonal function dB(r)/dr.
+
+    :param d_max: d_max.
+    :param n: n.
+
+    :return: First derivative in dB(r)/dr.
+    """
+    pinr = pi * n * r/d_max
```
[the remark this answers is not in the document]
zattala: Done. I actually didn't see in the code where ortho_derived is used, not in calc or p_invertor.  maybe delete unless theres something I missed?

src/sas/sascalc/pr/calc.py
```diff
@@ -0,0 +1,368 @@
...
+    P(r) calculated from the expansion
+
+    :param pars: c-parameters.
+    :param d_max: d_max.
+    :param r: r-value to evaluate P(r).
+
+    :return: P(r).
+    """
+    total = np.zeros(len(r))
+    for i in range(len(pars)):
```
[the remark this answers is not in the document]
zattala: For some reason I was under the impression it doesn't work, but does, and changed it wherever was useful.

src/sas/sascalc/pr/p_invertor.py
zattala: Done.

src/sas/sascalc/pr/p_invertor.py
zattala: Done.

src/sas/sascalc/pr/p_invertor.py
zattala: Done.

src/sas/sascalc/pr/p_invertor.py
zattala: Done.

src/sas/qtgui/Perspectives/Inversion/InversionLogic.py
```diff
@@ -127,19 +129,12 @@ def newPRPlot(self, out, pr, cov=None):
...
-            if value > pmax:
-                pmax = value
-
-            y[i] = value
+        if cov2 is None:
+            y = pr.pr(out, x)
+        else:
+            (y, dy) = pr.pr_err(out, cov2, x)
+        total = np.sum(y * (pr.d_max / len(x)))
+        pmax = np.max(y)
```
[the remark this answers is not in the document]
pkienzle: total and pmax still being initialized but not used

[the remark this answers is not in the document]
zattala: Right, removed now.

zattala:
Just realised had to change setup.py, because when rebuilt without c extensions complains, this change works for me but maybe more needs to be done to run.py or somewhere else in setup.py not sure.

src/sas/qtgui/Perspectives/Inversion/InversionLogic.py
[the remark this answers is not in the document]
pkienzle: And there are a couple of initializations of y and err in new1DPlot (65-66, 91-92) that are no longer needed.

src/sas/sascalc/pr/calc.py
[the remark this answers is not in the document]
pkienzle: Agreed that it doesn't seem to be referenced in the code, but I suggest leaving it there for now, maybe adding "#TODO: unused?"

src/sas/qtgui/Perspectives/Inversion/InversionLogic.py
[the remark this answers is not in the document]
zattala: Removed now.

src/sas/sascalc/pr/calc.py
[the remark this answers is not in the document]
zattala: Good idea, added.

pkienzle:
Need to merge changes from ESS_GUI into converting_cinvertor_python branch.   You can see what these are by switching to  the ESS_GUI branch and diffing from the common branch point (which is 11296b9):
```
git diff 11296b9 run.py setup.py src/sas/sascalc/pr
```

The conflicts are mostly from changes in how we handle compiled models ported from the 4.x branch.  Since the C files no longer exist, you will obviously keep your changes.

In run.py you should keep the plugin model environment setup---it will be needed for headless fits once headless supports the new save project format from 5.x.

In invertor.py you should keep the changes to exception handling, as well as the change to the lstsq call.  I don't remember why add_errors was removed.

zattala:
I have kept ESS_GUIs version of everything, the only thing I changed from ESS_GUI and took converting_cinvertors value of was setup.py referencing pr inversion to this - 

srcdir = os.path.join("src", "sas", "sascalc", "pr")
package_dir["sas.sascalc.pr.core"] = srcdir
package_dir["sas.sascalc.pr"] = os.path.join("src", "sas", "sascalc", "pr")
packages.extend(["sas.sascalc.pr", "sas.sascalc.pr.core"])

And of course the removal of the C code itself in converting_cinvertor. 

Just removed the reference to the C code, it does appear to build and run correctly, and pass all the unit tests for me but may have missed something somewhere else if the C is referenced somewhere else.

rozyczko:
5.0 ready for testing on Win

zattala:
I am not sure, but this may be the problem - 
My initial change to setup.py not only removed the reference to c_extensions but kept a reference to a sascalc.pr.core, which does not appear to be in the current ESS_GUI branch, but must have been there when I first forked off of ESS_GUI as I did not add it, just removed c_extensions, 
so original - 
srcdir = os.path.join("src", "sas", "sascalc", "pr")
package_dir["sas.sascalc.pr.core"] = srcdir
package_dir["sas.sascalc.pr"] = os.path.join("src", "sas", "sascalc", "pr")
packages.extend(["sas.sascalc.pr", "sas.sascalc.pr.core"])

ESS_GUI - 
srcdir = os.path.join("src", "sas", "sascalc", "pr", "c_extensions")
package_dir["sas.sascalc.pr.core"] = srcdir
package_dir["sas.sascalc.pr"] = os.path.join("src", "sas", "sascalc", "pr")
packages.append("sas.sascalc.pr")
ext_modules.append(Extension("sas.sascalc.pr._pr_inversion",
                             sources=[os.path.join(srcdir, "Cinvertor.c"),
                                      os.path.join(srcdir, "invertor.c"),
                                      ],
                             include_dirs=[],
                             ))

So removing the reference to both pr.core and the extension_modules reference should leave - 
New:
package_dir["sas.sascalc.pr"] = os.path.join("src", "sas", "sascalc", "pr")
packages.append("sas.sascalc.pr")

Could this have caused the build to fail?

rozyczko:
The build failed at the code signing stage - something we are investigating currently. Your changes didn't cause the failure.

rozyczko:
5.0 ready for testing on Win

rozyczko:
The installer now gets created but the resulting installation (at least on windows) fails.
Will look at why later - most likely the pyinstaller script needs updating as well.

rozyczko:
5.0 ready for testing on Win

rozyczko:
5.0 ready for testing on Win

zattala:
sorry to commit again after the build has been tested but found small bug, just method being called wrong, when testing pr inversion with smearing. It does work properly now though.

pkienzle:
Before merging, we need ortho_transformed in calc.py to use sinc function so that the calculation works when q is zero.  There are other review comments, but these are for style and (minor) efficiency gains, not for bug fixes.

Also, can't merge because of conflicts.

pkienzle:
Ready for testing on Win64

butlerpd:
the 4.x version of this PR (#1038 ) was merged on October 1.  As agreed this PR should NOT be merged for 5.01 but leaving it here to remind us that we need to eventually bring it into 5.x

butlerpd:
One unit test is failing - 1.0000.....2 !=1 .  That seems like a rounding error?  Perhaps it should be an AssertAlmostEqual?

butlerpd:
The tests are all now passing and as per agreement on Tuesday calls going back a while this is now ready to merge.

Question for @wpotrzebowski or @rozyczko; given the plan for a 5.0.2 should we hold off for this week? or shall we just dump it in? the 4x version got merged a while back with no problems yet.  One caveat: we really **don't** want to wait longer than this week or we will once again have to worry about keeping this branch up to date and yada yada -- trying desperately to clear our PR backlog.

PR #1372, opened 2019-08-15
Simple bug reporting for the user

rozyczko:
Simple mechanism for quick bug reporting.
_Help_/_Report a bug_ menu item opens 
https://github.com/SasView/sasview/issues/new
in the default browser.


wpotrzebowski:
It seems to me that it requires some work/thinking and therefore I suggest moving it to 5.1.

butlerpd:
As per call - lets close this for now.  Once we have a good plan and implement in the branch we can open a new PR

PR #1378, opened 2019-08-22
Add long sesans file test example

rprospero:
This just adds another spin echo example file.  This is useful for simulating larger SESANS samples.


butlerpd:
What is the purpose of this file?  It is being added to the unit tests directory (which seems oddly structured anyway).  If this is meant to catch problems from changes to the reader this is appropriate but if it is an example file the user might want to load as suggested by the PR comment it is not. In that case it should be added to sasview/src/sas/sasview/test/sesans_data

rprospero:
I've moved the file to the location that you suggested, since its primary use is as an example file.

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

PR #1383, opened 2019-09-04
Converting bslloader to python 4x

pkienzle:
Use pure python implementation of BSL loader.  This PR is against master rather than ESS_GUI, so only applies to the 4.x release.  We will need to merge into ESS_GUI for the 5.1 release as a separate PR.


pkienzle:
ready for testing on Windows

pkienzle:
ready for testing on Win64

pkienzle:
4x and 5 branches are now consistent; either or both can be pulled

smk78:
I have done a side-by-side functionality comparison of the File Converter running in 4.2.2 and this converting_bslloader_python-4x branch for all 4 allowed input file types (using the test\convertible_files examples APS_X/APS_Y for 1D ASCII, SANS2D_100254_merged_ISIS2D for 2D ASCII, Z9800n.L1D/Z9800n.QAX for 1D BSL, and X2500n.L2D for 2D BSL). And I have tried reading the converted xml (for 1D) and h5 (for 2D) output files back into SasView 4.2.2.

The implementation of the File Converter in this branch reproduces the current behaviour in all bar one respect: it no longer throws an "iteration over 0-d array" warning when converting the 2D files!

_However_, the converted 2D BSL h5 output file does not load in SasView 4.2.2 or SasView run from this branch or SasView run from master. But it _does_load into SasView 4.2.1!!! Interestingly, the converted 2D ASCII h5 output file loads fine in all versions of SasView tried.

So, to summarise, this backport reproduces current behaviour, although one aspect of that current behaviour has changed since 4.1.2. Accordingly I approve this merge and will open a new issue to deal with that changed aspect.

PR #1384, opened 2019-09-05
pure python Pr invertor for 4.x

pkienzle:
Backport of Ziggy's pure python Pr invertor that runs in 4.x

This was done by rebasing commits from the ESS_GUI branch point onto master.   Changes unrelated to invertor were ignored.

The updates to qtgui InversionLogic.py were not backported.  These will need to be reapplied from the converting_cinvertor_python branch after sascalc is synchronized between master and ESS_GUI.  Similarly for the numba support and bumped numpy version in the yaml and spec files for the qt gui.  The following lists all files change on the 5.x branch:
```sh
git diff ESS_GUI..converting_cinvertor_python | grep ^diff
```

The GUI runs, but somebody else will need to assess whether it is working correctly.  The tests run.

We should probably apply only one of these PRs, but I'm not sure which.  We should synchronize sascalc between 4 and 5 before splitting it into its own package with its own release schedule, but that can be done from either master or ESS_GUI branch.  The main argument for doing everything on the 4.x branch is that similar changes for file_converter can't be tested in 5.x yet.


smk78:
Disclaimer: I have very limited expertise in doing P(r) analyses!

However, I have done a side-by-side functionality comparison of P(r) Inversion in 4.2.2 and this pr-4x branch by generating simulated I(q) data for a sphere and a cylinder, which should give very different results, and then inverting each.

The implementation of P(r) Inversion in this branch reproduces the current behaviour, and in each case both the fitted I(q) and the P(r) curve look sensible to me and the latter is almost identical to that generated by 4.2.2.

For the same input data and set of inversion parameters, this implementation of P(r) is fractionally slower (eg, by 0.01 seconds when inverting the cylinder data!) and gave a Chi2/dof value that differed by 0.00001 from that returned by 4.2.2.

Accordingly, I approve this merge, but will leave it to others to decide if the merge should be done (see first comment in this thread).

smk78:
pr_cylinder_comparison


pr_sphere_comparison

smk78:
[issue-1384.zip](https://github.com/SasView/sasview/files/3652567/issue-1384.zip)

PR #1387, opened 2019-10-01
Pr 4x

pkienzle:
Jenkins build for windows is now failing:

    https://jenkins.esss.dk/sasview/job/SasView_Win7/1454/

While trying to track it down, I did some code cleanup and found a bug:  pr_fit was not setting chi2 as expected.

The tests pass this machine so this doesn't resolve the error.  Leave the issue open until it builds properly on jenkins.


pkienzle:
Ready for testing on Win64.

pkienzle:
More code cleanup.  This includes the fix in #1402, so it should apply instead of or after that issue.

wpotrzebowski:
Ready for testing on Win64

src/sas/sascalc/pr/p_invertor.py
```diff
@@ -353,13 +349,10 @@ def get_iq_smeared(self, pars, q):
         q = np.atleast_1d(q)
         pars = np.float64(pars)
 
-
         npts = 21
```
wpotrzebowski: I know this hasn't been touched but magic numbers are always bit scary... Do we know why is it 21?

src/sas/sascalc/pr/p_invertor.py
```diff
@@ -556,59 +542,58 @@ def _get_matrix(self, nfunc, nr):
...
-                                                                      self.slit_height, self.slit_width, npts)/self.err[q_accept_x]
+                res = calc.ortho_transformed_smeared(
+                    x_use, self.d_max, j+offset,
+                    self.slit_height, self.slit_width, smear_npts)
             else:
-                a_use[q_accept_x, j] = calc.ortho_transformed(x_use, self.d_max, j+offset)/self.err[q_accept_x]
-
-        a_obj[0:self.npoints, :] = a_use
+                res = calc.ortho_transformed(x_use, self.d_max, j+offset)
+            a_view[q_accept_x, j] = res/err_use
```
wpotrzebowski: I think it makes sense that both smeared and un-smeared is divided by error (that wasn't the cas before). Just make sure this something intended.

wpotrzebowski:
ready for testing on OSX

src/sas/sascalc/pr/p_invertor.py
pkienzle: Number of points in the resolution integral.  Presumably chosen to balance speed and accuracy over the expected range of inputs.  I don't know how this particular number was chosen.

src/sas/sascalc/pr/p_invertor.py
pkienzle: This was the case before: all three branches used `.../self.err[q_accept_x]`.  I was just trying to make it more obvious.

src/sas/sascalc/pr/p_invertor.py
butlerpd: Should that be a comment added to the code for future developers?

PR #1394, opened 2019-10-24
Fix path to sascalc simulations in setup.py

llimeht:
The incorrect path leads to install errors when building binary packages.


pkienzle:
We should move realspace and simulation out of the src directory since they are not actually used anywhere.  Perhaps delete them.  Much of the geometry package is in sasmodels/explore/realspace.py.  

Anyway, changes to these directories should be harmless, so I have no problem merging them.

PR #1396, opened 2019-10-24
Fix image filename in docs

llimeht:
Image file was renamed but the rst was not updated.


butlerpd:
This looks to be new documentation for 5.x? Or was 4.x using extracting docs from bumps and 5.x is duplicating?

pkienzle:
Yes, qtgui docs.  The PR is against ESS_GUI.

llimeht:
Alternative patch (rebased onto current ESS_GUI HEAD) which renames the image file instead.

(I'm not sure I understand the duplication of the documentation between bumps and sasview.)

pkienzle:
The docs were copied to Qt because the bumps docs use screen shots from the bumps wx panels for the docs, not the qt panels that are implemented in the ESS GUI.

wpotrzebowski:
5.0 ready for testing on OSX

wpotrzebowski:
5.0 ready for testing on Win

PR #1395, opened 2019-10-24
Fix wrongly encoded quotes in changelog

llimeht:
sphinx running under Python 3 can't build the documentation with mixed encoding in the changelog.

PR #1397, opened 2019-10-24
Skip Qt imports when building docs

llimeht:
Mock-out Qt imports from the autodoc calls so that Qt is not imported; importing Qt adds lots of noise to the docs and also requires a display to be available.

### DIFF
--- a/docs/sphinx-docs/source/conf.py
+++ b/docs/sphinx-docs/source/conf.py
@@ -107,6 +107,8 @@ exclude_patterns = ["*sas.sasgui.perspectives.simulation.rst",
                     "*sas.sasgui.guiframe.custom_pstats.rst", # pstats not in standard library on Ubuntu out of the box
                     ]
 
+autodoc_mock_imports = ['sip', 'PyQt5']
+
 # The reST default role (used for this markup: `text`) to use for all documents.
 #default_role = None
 


docs/sphinx-docs/source/conf.py
```diff
@@ -107,6 +107,8 @@
                     "*sas.sasgui.guiframe.custom_pstats.rst", # pstats not in standard library on Ubuntu out of the box
                     ]
 
+autodoc_mock_imports = ['sip', 'PyQt5', 'PyQt5.QtGui', 'PyQt5.QtCore', 'PyQt5.QtWidgets']
+
```
pkienzle: From the [autodoc_mock_imports doc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports):

    You may only specify the root package of the dependencies themselves and omit the sub-modules

So `['sip', 'PyQt5']` should be enough.

llimeht: Potentially. The docs aren't actually built on CI so I'll have to manually test that.

(Perhaps an additional PR would be welcome to build the docs on CI too so that things like #1395 and #1396 get caught earlier.)

pkienzle: Yes, this should be a separate ticket.  We are already building them on the Jenkins server, but the doc errors aren't visible unless you dig into the console output from the build.

Maybe better [setuptools integration](https://www.sphinx-doc.org/en/master/usage/advanced/setuptools.html) would help?    Our doc builds are a little complicated since we are pulling in sasmodels and bumps documentation parts, so existing tools would require tweaking.

llimeht:
Yes, that seems to work fine. Updated (and rebased to current ESS_GUI HEAD).

llimeht:
In trying to build the docs on CI, I wonder if building with a display available (with xvfb) would be better than mocking the Qt imports.

pkienzle:
For the doc build it shouldn't matter that Qt import is mocked.  We are not trying to link to Qt documentation, or otherwise check that we are using the Qt APIs properly.

For Qt testing, sure.  Also when we add type signatures to the function and automate mypy tests.

llimeht:
The build log with Qt mocked out shows lots of additional errors from apidoc with the classes not being able to be imported when mocked.

[Example](https://travis-ci.org/llimeht/sasview/jobs/606275684#L3545)

```
WARNING: autodoc: failed to import module 'DensityPanel' from module 'sas.qtgui.Calculators'; the following exception was raised:
  File "/home/travis/miniconda/lib/python3.6/site-packages/sphinx/ext/autodoc/importer.py", line 32, in import_module
    return importlib.import_module(modname)
  File "/home/travis/miniconda/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/travis/build/llimeht/sasview/build/lib.linux-x86_64-3.6/sas/qtgui/Calculators/DensityPanel.py", line 11, in <module>
    from sas.qtgui.UI import main_resources_rc
  File "/home/travis/build/llimeht/sasview/build/lib.linux-x86_64-3.6/sas/qtgui/UI/main_resources_rc.py", line 29172, in <module>
    if qt_version < ['5', '8', '0']:
TypeError: '<' not supported between instances of 'split' and 'list'
```

Many of these look like they can be addressed by changing the SasView code; however, in the meantime, I would assume (but have not checked) that the apidoc for sas.qtgui are more or less missing from the built if Qt is mocked. This PR should be parked until I can either verify that the build output is indeed the same or we find ways to make the qtgui imports succeed when Qt is mocked.

pkienzle:
Thanks for the explanation.  

So does that mean that you don't need this PR if PR #1401 is successful?  

That is, your build server needs the qt libraries (and `services: xvfb` on travis-ci) to run the qt tests, so might as well assume they are there for the doc build as well, so no need to mock them?

llimeht:
On the surface, just building the docs inside xvfb with something like #1401 obviates the need for this PR, yes. (Note that the travis config item `services: xvfb` is linux-only.)

Two deeper questions:
 * is *requiring* a display for doc building acceptable? It's probably not an important restriction, but this requirement will be the sort of thing that keeps being forgotten and broken builds or broken releases will ensue.
 * are the errors that mocking causes actually symptomatic of underlying bugs that should be fixed in any case?

PR #1398, opened 2019-10-24
Don't build Qt GUI when setup.py is asked to clean

llimeht:
Calling "setup.py clean" should remove generated files and be fast, rather than creating lots of files and be slow. This patch at least stops "setup.py clean" from building the Qt GUI, although it doesn't clean up the generated files.

This is a bit of a crude hack; it's better than the current situation but far from perfect.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -196,7 +196,8 @@ class BuildSphinxCommand(Command):
         import build_sphinx
         build_sphinx.rebuild()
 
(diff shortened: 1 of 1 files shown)


wpotrzebowski:
5.0 ready for testing on OSX

wpotrzebowski:
5.0 ready for testing on Win

butlerpd:
Move to 5.1 and see if we can do more properly.

PR #1401, opened 2019-11-03
Test more on travis-ci

llimeht:
Recent work updating the Debian packages for sasview indicated that the 5.0.0 release went out with tests that failed and documentation that couldn't be built. Some of these things were fixed immediately following the release while others have been in subsequent PRs.

Building the documentation and running `GUITests.py` on travis so that failures are visible earlier will make it easier to spot regressions and help improve the quality of the SasView releases.

This PR enables:
 * building of the documentation on travis-ci; the documentation build throws a lot of warnings some of which are undoubtedly bugs to be fixed.
 * running of `GUITests.py` (test runner for the Qt GUI); these tests currently fail on CI. Some of these will be easy to fix while some will not.

While this PR will cause the test checks on GitHub to go from "passing" to "failing", it is only because it is running more tests. I think it's important to enable these additional tests, to fix either tests or code in subsequent PRs and to then gate future merging on the tests successfully passing.

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ addons:
     - liblapack-dev
     - gfortran
     - libhdf5-serial-dev
+services:
+  - xvfb
 before_install:
 - echo $TRAVIS_OS_NAME
 - wget $MINICONDA -O miniconda.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ script:
 - git clone --depth=50 --branch=master https://github.com/SasView/sasmodels.git sasmodels
 - git clone --depth=50 --branch=master https://github.com/bumps/bumps.git
 - ls -ltr
-- if [ ! -d "utils" ]; then mkdir utils; fi
+- mkdir -p utils
 - /bin/sh -xe sasview/build_tools/travis_build.sh
 - export LC_ALL=en_US.UTF-8
 - export LANG=en_US.UTF-8
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,9 +36,9 @@ before_install:
 - hash -r
 #- conda update --yes conda
 - conda info -a
-- conda install --yes python=$PY numpy scipy cython pylint wxpython matplotlib lxml h5py sphinx pyparsing html5lib reportlab pybind11 appdirs six numba mako
+- conda install --yes python=$PY numpy scipy cython pylint wxpython matplotlib lxml h5py sphinx pyparsing html5lib reportlab pybind11 appdirs six numba mako ipython qtconsole
 install:
-- pip install periodictable xhtml2pdf unittest-xml-reporting pylint
+- pip install periodictable xhtml2pdf unittest-xml-reporting pylint qt5reactor
 # Note: use "... || true" below to silently ignore failure
 - pip install pyopencl
 script:
--- a/docs/sphinx-docs/build_sphinx.py
+++ b/docs/sphinx-docs/build_sphinx.py
@@ -181,7 +181,12 @@ def apidoc():
         "-o", SASVIEW_API_TARGET, # Output dir.
         "-d", "8", # Max depth of TOC.
         "-H", "SasView", # Package header
-        SASVIEW_BUILD
+        SASVIEW_BUILD,
+        # omit the following documents from the API documentation
+        joinpath(SASVIEW_BUILD, "sas", "qtgui", "GUITests.py"),
+        joinpath(SASVIEW_BUILD, "sas", "qtgui", "convertUI.py"),
+        joinpath(SASVIEW_BUILD, "sas", "sasview", "welcome_panel.py"),
+        joinpath(SASVIEW_BUILD, "sas", "sasview", "wxcruft.py"),
     ])
 
     subprocess.check_call([
@@ -190,7 +195,8 @@ def apidoc():
         "-d", "8", # Max depth of TOC.
         "-H", "sasmodels", # Package header
         SASMODELS_BUILD,
-        joinpath(SASMODELS_BUILD, "sasmodels", "models"), # exclude
+        # omit the following documents from the API documentation
+        joinpath(SASMODELS_BUILD, "sasmodels", "models"),
     ])
 
 def build_pdf():
--- a/docs/sphinx-docs/build_sphinx.py
+++ b/docs/sphinx-docs/build_sphinx.py
@@ -235,8 +235,10 @@ def build():
     print("=== Build HTML Docs from ReST Files ===")
     subprocess.check_call([
         "sphinx-build",
+        "-v",
         "-b", "html", # Builder name. TODO: accept as arg to setup.py.
         "-d", joinpath(SPHINX_BUILD, "doctrees"),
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -31,19 +29,16 @@ before_install:
...
-- conda install --yes python=$PY numpy scipy cython pylint wxpython matplotlib lxml h5py sphinx pyparsing html5lib reportlab pybind11 appdirs six
+- conda install --yes python=$PY numpy scipy cython pylint wxpython matplotlib lxml h5py sphinx pyparsing html5lib reportlab pybind11 appdirs six ipython qtconsole
 install:
-- pip install periodictable xhtml2pdf pyopencl unittest-xml-reporting pylint
+- pip install periodictable xhtml2pdf pyopencl unittest-xml-reporting pylint qt5reactor
 script:
 - cd ..
 - export WORKSPACE=$(pwd)
-- git clone --depth=50 --branch=master https://github.com/SasView/sasmodels.git sasmodels
+- git clone --depth=50 --branch=tmp/doc-build https://github.com/llimeht/sasmodels.git sasmodels
```
pkienzle: Should reference SasView/sasmodels github.

llimeht: Of course. That's one of the reasons that this is a PR draft that GitHub won't let you merge, rather than a PR. That just working around a resolution to https://github.com/SasView/sasmodels/pull/373 .

llimeht:
Returning to this work on getting more CI testing done. The PR stack contains several sets of changes:

1. boring but needed misc fixes to `setup.py`, `rst` API syntax etc - I think these are all mostly uncontroversial although the `rst` changes are a minimal set to stop warnings and there are many more that could be made to improve the maths typesetting in the docs at a later date
2. adding doc building to travis-ci - this seems to work find and CI passes up to this point
3. adding `GUITests.py`to travis-ci - the tests run but fail in lots of ways; some of these look easy to solve and others might not be. There needs to be a plan on how to tackle these tests prior to turning them on in CI. (As an intermediate step, we could run the tests but ignore the result to track how they are going from the travis log files)

For the sake of simplicity, I can split them into separate the changes into (dependent) PRs if that makes it easier to digest.

docs/sphinx-docs/build_sphinx.py
```diff
@@ -176,30 +176,39 @@ def apidoc():
...
-                     "-H", "sasmodels", # Package header
-                     SASMODELS_BUILD,
-                     joinpath(SASMODELS_BUILD, "sasmodels", "models"), # exclude
-                     ])
+    subprocess.check_call([
+        "sphinx-apidoc",
+        "-o", SASVIEW_API_TARGET, # Output dir.
+        "-d", "8", # Max depth of TOC.
+        "-H", "SasView", # Package header
+        SASVIEW_BUILD,
```
pkienzle, line 184: Add comment saying that the following files are excluded from the docs.

src/sas/sascalc/invariant/invariant.py
```diff
@@ -817,10 +817,17 @@ def get_surface(self, contrast, porod_const, extrapolation=None):
         Historically, Sv was computed with the invariant and the Porod
         constant so as not to have to know the contrast in order to get the
         Sv as:
+
+        .. :math::
+
             surface = (pi * V * (1- V) * porod_const) / q_star
+
```
pkienzle, line 824: Use latex markup here and elsewhere, otherwise you get, e.g., `q subscript s t a r` all italic with math spacing rather than `q^*`.

pkienzle (changes requested):
Need to fix up the math and suppress GUI testing until it can pass without errors, then this can be merged. The GUI tests fixed and enabled in a separate PR.

src/sas/sascalc/invariant/invariant.py
llimeht: Given the style of the document is written in python variables rather than maths variables, I've turned these into literal blocks instead (plus a few other places in the same file that had similar problems, but weren't making sphinx emit warnings)

llimeht:
The GUI tests are still run (since they're quick compared to everything else in the test sequence) but the failure is suppressed for the time being:

`$PYTHON GUITests.py || true`

I've removed the remaining commit that was allowing me to test branches other than master & ESS_GUI, and so I think this is now ready to be merged.

src/sas/sascalc/invariant/invariant.py
```diff
@@ -849,7 +861,7 @@ def get_surface(self, contrast, porod_const, extrapolation=None):
 
     def get_volume_fraction(self, contrast, extrapolation=None):
         """
-        Compute volume fraction is deduced as follow: ::
+        Compute volume fraction is deduced as follows: ::
```
pkienzle, line 864: I think if you say `... follows::` then it will produce "... follows:" in markup followed by a literal block.

Similarly below.

src/sas/sascalc/invariant/invariant.py
```diff
@@ -964,13 +979,17 @@ def get_surface_with_error(self, contrast, porod_const, extrapolation=None):
         from the contrast and porod_constant wich are currently user inputs
         with no option for any uncertainty so no uncertainty can be calculated.
         However we include the uncertainty computation for future use if and
-        when these values get an uncertainty. This is given as:
+        when these values get an uncertainty. This is given as: ::
+
+            ds = sqrt[(s\'_cp)**2 * dcp**2 + (s\'_contrast)**2 * dcontrast**2]
+
+        where s'_x is the partial derivative of S with respect to x
```
pkienzle, line 986: Why the backslash in `s\'_cp` but not `s'_x`?

Do you need to escape the previous `'`? E.g., because there are two on the same line and RST is interpreting that as meaningful?

pkienzle (approved):
Some minor quibbles you may want to fix, but nothing serious.

src/sas/sascalc/invariant/invariant.py
llimeht: Yes indeed; I just left it in the style of the document which seemed to use ": ::" throughout.

src/sas/sascalc/invariant/invariant.py
llimeht: Yes, the '…' on the equation line was being interpreted as a quotation. (Actually, it still might be now I look at it).

The literal block is implicitly Python code and get syntax highlighted. Perhaps for this particular docstring, something other than prime is needed to indicate the derivative. Any thoughts on what?

pkienzle: `dS = sqrt[(dS/dcp)**2 * dcp**2 + (dS/dcontrast)**2 * dcontrast**2]`? 

Or just switch to math since right now it is neither math nor python:
```
\Delta S^2 = \left(\frac{\partial S}{\partial \mathrm{cp}}\right)^2 \Delta \mathrm{cp}^2 
+ \left(\frac{\partial S}{\partial \mathrm{contrast}}\right)^2 \Delta \mathrm{contrast}^2
```

pkienzle: Or `.. code-block:: none` for minimal change.

PR #1402, opened 2019-11-08
fix pr bug for older numpy: x[True] did not select all elements

pkienzle:
The OS/X build on Jenkins is failing because it has an old version of numpy.  This fixes the code so that old numpy will continue to work, and let the tests pass again.


wpotrzebowski:
ready for testing on OSX

Issue #1400, opened 2019-10-30
Generic scattering calculation is incorrect for orientation-averaged asymmetric particles


pkienzle:
The generic scattering calculator code has two forms for 1D curves: the full (slow!) calculation, which does a pair-wise sum over distance between points, and the "Debye full avg" (fast) calculation, which uses the distance of each point from the center.  The GUI only uses the Debye average, but the debye average only works when there is spherical symmetry.

The user interface should drop the Debye average calculation or provide three computation options instead of two.  Maybe:

    2D: Fixed particle orientation
    1D: Asymmetric particle avg.
    1D: Symmetric particle avg.

Compare the full calculation to the "average" calculation for the dna.pdb file in test/coordinate_data:

![dna](https://user-images.githubusercontent.com/473456/67896712-47527180-fb33-11e9-97a7-471475d2a752.png)
![dna_avg](https://user-images.githubusercontent.com/473456/67896753-618c4f80-fb33-11e9-914b-3356415afdaf.png)

Here is an example calculated for an ellipsoid (Rp=50, Re=125), compared to the sasmodels ellipsoid model:

![ellip](https://user-images.githubusercontent.com/473456/67896937-bfb93280-fb33-11e9-8272-66beeebeb2d6.png)

PR #1403, opened 2019-11-14
Generic scattering calculator C conversion 5.x

pkienzle:
Convert generic scattering calculator to C.

The calculator has been checked against the ellipsoid model, 1D, 2D and magnetic. 

The magnetic model will return different results since the top of the sample is not oriented toward the beam to correspond to the analytical calculations in sasmodels.  Furthermore, the C code had my and mz swapped when calling the calculator.

The is_avg flag does not work for particles without spherical symmetry.  This option will need to be removed in the GUI. See #1400.


pkienzle:
Note: oommf and debye avg. not yet tested in the qt gui, so not yet ready for review.

pkienzle:
Tested within GUI.  Ready for review.

wpotrzebowski:
5.0 ready for testing on OSX

PR #1404, opened 2019-11-14
fix wx4 problems with gen scattering panel

pkienzle:
* Fix infinite loop that appears in wxPhoenix.  Don't know why wx3 didn't complain.

* Make sure all widget handling is happening in the wx thread; the completion callback needs to trigger a call to the wx thread when it is convenient.


pkienzle:
These changes are included in PR #1405, so if that gets merged this PR can be deleted.

pkienzle:
#1405 includes #1404, so applying this patch before the other is fine.

PR #1405, opened 2019-11-14
Generic scattering calculator C conversion 4.x

pkienzle:
Like #1403 but applies to the 4.x branch.  This removes the last bit of C code.

The calculator has been checked against the ellipsoid model, 1D, 2D and magnetic.

The magnetic model will return different results since the top of the sample is not oriented toward the beam to correspond to the analytical calculations in sasmodels. Furthermore, the C code had my and mz swapped when calling the calculator.

The is_avg flag does not work for particles without spherical symmetry. This option will need to be removed in the GUI. See #1400.


pkienzle:
Note: oommf and debye avg. are failing in the wx gui, so not yet ready for review.

pkienzle:
oommf and debye avg are now fixed.  Ready for review.

pkienzle:
If this PR includes the changes from #1404.  If it is merged then #1404 can be deleted.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski (approved):

PR #1388, opened 2019-10-04
Properly load data without dI - Ticket 1386

krzywon:
The underlying issue of issue #1386 had nothing to do with the BSL reader or writer, but with the data not having dI data. Any 2D data set without dI had the same result of being unable to load (verified by removing columns 4 and above from a VSANS data set in 
[sans34736.zip](https://github.com/SasView/sasview/files/3691266/sans34736.zip). The plottable_2d object assumed all data would have dI included, but when they were not present, created an array of None, causing issues with the NaN removal procedure added to resolve issue #1292.

NB: This branch was built off of a point in master where unit tests for P(r) are failing.

Issue #1386, opened 2019-09-25
2D BSL files converted into h5 format no longer loading in SasView 4.2.2


smk78:
In testing issue #1383 it was discovered that 2D BSL files converted into h5 format were no longer loading into SasView 4.2.2, SasView running from master, or SasView running from the converting_bslloader_python_4x_branch. They _do_ however load in SasView 4.1.2.

More curiously, 2D ASCII files converted into h5 format load fine in all these versions.

smk78:
[issue-1386.zip](https://github.com/SasView/sasview/files/3651671/issue-1386.zip)

smk78:
Uploaded issue-1386.zip contains 7 files:

- X2500n.L2D , the original 2D BSL data (n=0: header, n=1: SANS, n=2: calibration (here blank), n=3: WANS)
- x25000-pr.h5 , converted file output by converting_bslloader_python_4x branch
- x25000-422.h5 , converted file output by SasView 4.2.2
- x25000-421.h5 , converted file output by SasView 4.2.1

smk78:
The above h5 files all load and display in SasView 4.1.2

The above files _do not_ load in SasView 4.2.2. The Console reports:

```
The Data file you selected could not be loaded.
Make sure the content of your file is properly formatted.
When contacting the SasView team, mention the following:
too many indices for arrayThe following issues were found whilst loading x25000-pr.h5:
```

The above files _do not_ load in SasView 4.x master.  The Command Window reports:

```
10:39:41 - INFO : sas.sascalc.dataloader.readers.xml_reader:  78: Start tag expected, '<' not found, line 1, column 1
10:39:41 - INFO : sas.sascalc.dataloader.readers.xml_reader:  78: Start tag expected, '<' not found, line 1, column 1
```

which is curious as they are h5 files...

smk78:
There are also differences in the size of the files (even though they were converted from the same input file with the same options and metadata entries):

- x25000-pr.h5 , 169 KB
- x25000-422.h5 , 130 KB
- x25000-421.h5 , 215 KB

smk78:
@krzywon is this something you could take a look at?

krzywon:
I have a fix for this issue in branch ticket-1386 and will submit a PR after checking a few more things, but I am more concerned about the data the BSL converter is producing. The data is in pixel space, not Q space. The NXcanSAS standard 
[...]

PR #1409, opened 2019-11-26
Ticket 1386: ESS GUI

krzywon:
This is a cherry-pick of the two commits from PR #1388 that were put into the 4.x branch for ticket #1386. Those changes have been approved and merged into master, but need to be incorporated into the ESS_GUI branch (for 5.0.1?).


src/sas/sascalc/dataloader/data_info.py
```diff
@@ -114,8 +114,12 @@ def __init__(self, data=None, err_data=None, qx_data=None,
...
         self.qy_data = np.asarray(qy_data)
         self.q_data = np.asarray(q_data)
-        self.mask = np.asarray(mask)
-        self.err_data = np.asarray(err_data)
+        if mask is not None:
+            self.mask = np.asarray(mask)
+        else:
+            self.mask = np.ones(self.data.shape, dtype=bool)
+        if err_data is not None:
+            self.err_data = np.asarray(err_data)
```
rozyczko: self.mask and self.err_data get defined only when mask/err_data is passed as arguments.
Subsequent calls to _validity_check (line 1055) will fail with attribute errors if those optional arguments are not passed.

krzywon: Good catch on this. self.err_data needs to be set to something for uncertainty propagation when combining data sets. I want to say self.err_data should be set to all zeros, but something about that doesn't seem quite right.

self.mask is set to an array of trues when no mask is passed, so that should be good already.

PR #1426, opened 2019-12-02
Unicode error

pkienzle:
Plugin model error handling was failing when the error message contained unicode.  As a result, bad models were not being trapped, and instead an error was raised when the fitting perspective was being initialized.  The result was no fitting perspective.

Reproduce by putting a spurious unicode character such as Σ in a plugin model so that it causes a syntax error.

This only seems to happen on Mac and only when clicking the app icon to start the program, but from the nature of the error, it ought to happen every time the program starts.

The sasview logs should now include the full traceback when other errors occur on startup.


pkienzle:
ready for testing on OSX

pkienzle:
Sasview 4.x is still being built against python 2.7, so sascalc should support it.  We will need to cherry-pick 08e6975 to master before moving it to python 3.x.

I do not understand the jenkins failure for OSX:
```
...
copying src/sas/qtgui/UI/convert_rc.sh -> build/lib.macosx-10.10-intel-2.7/sas/qtgui/UI
copying src/sas/qtgui/UI/convert_ui.sh -> build/lib.macosx-10.10-intel-2.7/sas/qtgui/UI
copying src/sas/qtgui/UI/main_resources.qrc -> build/lib.macosx-10.10-intel-2.7/sas/qtgui/UI
error: can't copy 'src/sas/qtgui/UI/res': doesn't exist or not a regular file
```

wpotrzebowski:
5.0 ready for testing on OSX

wpotrzebowski:
The code looks fine and works as expected. Related but separate issue is editing of custom model with a bad unicode character. Issue has been created #1431

PR #1428, opened 2019-12-08
Misc documentation fixes to squash sphinx warnings

llimeht:
Similar to https://github.com/SasView/sasmodels/pull/379, this PR sorts out a few rst formatting issues that generate warnings. Sometimes these problems actually cause sections of the documentation to be omitted from the output because sphinx just gives up.

This is the next step towards turning sphinx warnings into errors as discussed in #1401.


src/sas/qtgui/Perspectives/Fitting/media/fitting_help.rst
```diff
@@ -932,32 +927,30 @@ Fitting with the following caveats:
...
+.. image:: combine_batch_grid.png
 
-..
-  .. image:: combine_batch_grid.png
+In the example shown above the data is a time series with a shifting peak.
+The first part of the series was fitted using the *broad_peak* model, while
+the rest of the data were fit using the *gaussian_peak* model. Unfortunately the
+time is not listed in the file but the file name contains the information. As
+described in :ref:`Grid_Window`, a column can be added manually, in this case
+called time, and the peak position plotted against time.
```
pkienzle: Maybe this was intended to be a figure caption?  In which case changing `.. image::` above to `.. figure::` might be a better change?  Similarly for the next image.

llimeht: There are no other uses of 'figure' in that file, it being mostly screenshot walk-through. 

Looking at the current output and the git history more, there are several paragraphs and quite a few images that don't appear in the output at all, but I don't know if that is intentional https://github.com/SasView/sasview/commit/dbba13450dc275d14d77d7daaaaf60f96e3e5918.

PR #1429, opened 2019-12-08
Use same interpretor to build sasmodels

llimeht:
`setup.py` will build the sasmodels documentation if it is not present. The documentation build should be performed with the same interpreter as `setup.py` is using. If the interpreter is not specified when calling `make`, then the path will be searched for an executable called `python`; this can mean accidentally switching from Python 3 to Python 2, with required tools like sphinx perhaps not being available for that interpreter.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -186,7 +186,12 @@ class BuildSphinxCommand(Command):
             if os.path.isdir(SASMODELS_DOCPATH):
                 # if available, build sasmodels docs
                 print("============= Building sasmodels model documentation ===============")
-                smdocbuild = subprocess.call(["make", "-C", SASMODELS_DOCPATH, "html"])
+                smdocbuild = subprocess.call([
+                    "make",
+                    "PYTHON=%s" % sys.executable,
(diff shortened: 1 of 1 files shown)


pkienzle:
This won't work since the Makefile uses the `sphinx-build` command from the path, which uses whatever python was used to install it.  You will need to change the Makefile to use:
```make
SPHINXBUILD = $(PYTHON) -m sphinx
```
This is already done in sasmodels.  In bumps I use `$(PYTHON) -m sphinx.__init__`, but I don't think `.__init__` is needed anymore.

We should probably quote the python interpreter as `"$(PYTHON)" -m sphinx` in case somebody has spaces in their python path, but I haven't tried it.

llimeht:
This does work because this is about `sasview` calling `sasmodels` which, as you say, uses the specified PYTHON already.

When sasview builds its own documentation from setup.py, it does so by importing `build_sphinx` directly thereby staying with the same interpreter.

llimeht:
And yes, some quoting of `"$(PYTHON)"` is needed for those who put spaces in directory names.

Issue #1413, opened 2019-11-28
Corfunc is not working well in 5.0.1


smk78:
Testing 5.0.1RC on x64 in W10.

The Corfunc perspective is VERY sluggish in 5.0.1. It takes several seconds to graph any data, most notably when you send the selected dataset to the perspective. But sometimes, you click Transform and no transforms appear for almost a minute, or you have to click on the scattering data window to force the real space window to update.

Two other points:

I would not include the IDF on the correlation function graph. It is confusing, but also, with no way to add/remove datasets to the graph it obscures the CFs.

And I would definitely NOT auto populate the output parameters boxes. These ONLY have meaning if the data being analysed conforms to an idealised lamellar morphology. Implying otherwise is scientifically dangerous. Please bring back the Extract Parameters button (with a suitable warning alongside). And put the Transform button above the Output Parameters.

rprospero:
I'm going to tackle this today and tomorrow hopefully

smk78:
Testing 5.0.1 Build 321 on x64 in W10.
See comment on https://github.com/SasView/sasview/pull/1432

smk78:
PR https://github.com/SasView/sasview/pull/1432 seems to have addressed this issue.

PR #1432, opened 2019-12-10
Ticket1413 corfunc

rprospero:
This PR should handle #1413 

The primary complaint was the sluggishness of the Corfunc panel.  Looking through, panel was replotting the data every time the model changed.  Since the model had many parameters that were set sequentially, this caused several needless replots in a row.  The code now only plots when the model has been finalised and is noticeably snappier.

There has also been some tidying of the GUI.  The IDF hsa been removed and the Extract Parameters button has been re-instated.  The buttons have also been moved into a more logical grouping, as seen below.  (Ignore the idiosyncratic colour scheme)

2019-12-10-155716_1918x975_scrot

### DIFF
--- a/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
+++ b/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
@@ -238,6 +238,8 @@ class CorfuncWindow(QtWidgets.QDialog, Ui_CorfuncDialog):
         self.cmdExtrapolate.setEnabled(False)
         self.cmdTransform.clicked.connect(self.transform)
         self.cmdTransform.setEnabled(False)
+        self.cmdExtract.clicked.connect(self.extract)
+        self.cmdExtract.setEnabled(False)
         self.cmdSave.clicked.connect(self.on_save)
         self.cmdSave.setEnabled(False)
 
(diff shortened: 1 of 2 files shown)


src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
```diff
@@ -300,11 +300,15 @@ def extrapolate(self):
...
             params, extrapolation, _ = self._calculator.compute_extrapolation()
+            self.model.itemChanged.disconnect(self.model_changed)
+
             self.model.setItem(W.W_GUINIERA, QtGui.QStandardItem("{:.3g}".format(params['A'])))
             self.model.setItem(W.W_GUINIERB, QtGui.QStandardItem("{:.3g}".format(params['B'])))
             self.model.setItem(W.W_PORODK, QtGui.QStandardItem("{:.3g}".format(params['K'])))
             self.model.setItem(W.W_PORODSIGMA,
                                QtGui.QStandardItem("{:.4g}".format(params['sigma'])))
+            self.model.itemChanged.connect(self.model_changed)
+            self.model_changed(None)
```
pkienzle: `model_changed()` calls `draw_q_space()`, which clears the figure and draws.  Two lines below there is another call to `draw_q_space()` which again clears and draws.

The try/except only needs to cover `compute_extrapolation()`. 

Should the A, B, K and sigma parameters on the form be cleared if the calculation fails?

src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
```diff
@@ -456,26 +470,30 @@ def setData(self, data_item, is_batch=False):
...
+            self._canvas.extrap = None
+            self._canvas.draw_q_space()
+            self.cmdTransform.setEnabled(False)
+            self._path = data.name
+            self._realplot.data = None
+            self._realplot.draw_real_space()
+
+            self.model.setItem(W.W_FILENAME, QtGui.QStandardItem(self._path))
+        finally:
+            self.model.itemChanged.connect(self.model_changed)
```
pkienzle: If you move the update to `W.W_FILENAME` with the rest of the updates, then you don't need the try-finally block.

Rather than calling `draw_q_space()` directly, you could instead call the suppressed `model_changed()` as is done elsewhere.

rprospero:
I've made the changes that Paul suggested.  Also, failing to calculate an extrapolation now clears the extrapolation model parameters.

src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
rprospero: Good catch

rprospero:
> Code looks good. However, the corfunc dialog does not rescale well on windows.
> the left pane has a hardcoded width which is too large for even decently high resolutions.
> 
> This causes clipping on the charts, which also don't scale properly.
> 
> Additionally, the "Extract parameters" button is inconsistent wrt. style with the rest of the application - its width should be the same as the buttons above it ("Extrapolate", "Transform")

I've located the problem with the size of the left pane.  The width wasn't hard coded, but that particular layout combination was requiring that the Guinier and Porod regions were each individually at least as large as the three buttons below.  Hopefully this new version is a bit more usable.

2020-01-08-131206_1918x1040_scrot

smk78:
Testing 5.0.1 Build 321 on x64 in W10.
Loaded a dataset and sent it to Corfunc. But nothing displays, no data boxes are completed, and all buttons are greyed out. So Corfunc is currently broken in Master.
Will now check the branch https://github.com/rprospero/sasview/tree/Ticket1413-Corfunc

smk78:
On today's call it was greed to merge this PR and move the 'cherry's' to a new issue.

smk78:
Documentation now needs updating. See https://github.com/SasView/sasview/issues/1501

PR #1444, opened 2020-01-21
ESS GUI: Warn user when data set fully masked

krzywon:
This PR matches 5.x up to the 4.x branch by attaching a warning to any data set that is fully masked when it is loaded. The first of the two commits has already been merged into the ESS_GUI branch, but the 2nd commit (cherry-picked here) has not.

The PR for the 4.x branch: #1356 
The 1st PR for the 5.x branch: #1368

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -1265,7 +1265,7 @@ class DataExplorerWindow(DroppableDataLoadWidget):
                     if hasattr(item, 'errors'):
                         for error_data in item.errors:
                             data_error = True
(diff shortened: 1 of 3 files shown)


butlerpd:
Jeff to test GUI -- then approve .. .or not.

krzywon:
The 1st PR merge (#1368) that loaded in the mask was reverted along with a few other changes in a [recent commit](https://github.com/SasView/sasview/commit/48c44054f064cc1821f0caca5095eb679840a1eb). I've restored that change here. Also, I changed DataExplorer to better handle errors linked to loaded data sets. The current behavior reads the errors and creates a string out of them, but the string isn't used anywhere.

PR #1461, opened 2020-02-16
update/fix invariant perspective

butlerpd:
Updated the documentation for legibility and correctness and also fixed
factor of 2 error in s/v calculation. Finally changed default contrast
to be 1e-6 rather than 1 which was very confusing since sasmodels now
assumes the e-6.  Tested with theoretical data and all now works - note
that obtaining the correct Porod constant is not as straight forward as
id might seem for perfectly monodispers system with no resolution
smearing.

Still to do: add proper unit test


butlerpd:
The documentation on sld still needs updating.  I am actually unsure of the purpose of that section since the perspective does not currently allow computation of the contrast. Either contrast or vol fraction can be computed but not both ... it might be a feature request to make that an option.

As it is, this seems overly complex for this perspective and I fear could cause confusion by people thinking they need to understand it when it is mostly irrelevant to the use of the perspective? At a minimum rho* and eta need defining here. Since I'm not sure what they are I am hoping somebody else can weigh in?

smk78:
```
          V =  self.get_volume_fraction(contrast, extrapolation)
          Compute the surface given by:
            surface = (pi *V(1- V)*porod_const)/ q_star
```

Does V(1- V) actually execute properly if coded like this???

butlerpd:
That looks very odd -- yet it is computing correctly.  Hummm... needs a bit further study I guess?
I guess I'd argue that even though it seems to be being interpreted correctly now that may be a fluke of python 2.7 (4.x is still 2.7 I think?) and we should not expect it to continue to work in later version of python?

butlerpd:
also Travis CI is failing on unit tests it seems.  I foolishly assumed there were no unit tests for this (since obviously the s/v answer was wrong) .... but it may well be that there was one with an incorrect answer which is now failing? will need to look at that as well.

smk78:
```
Python 2.7.15 |Anaconda, Inc.| (default, May  1 2018, 18:37:09) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> print (0.1*(1-0.1))
0.09
>>> print (0.1(1-0.1))
  File "<stdin>", line 1, in <module>
TypeError: 'float' object is not callable
>>>
```

butlerpd:
Fascinating .. and yet it has been working for quite a long time it seems ... and still does when compiled - though I could do a ready for testing to be sure.  That said it would be a purely curiosity exercise... we should fix it regardless.  Can't to today but will do when I get a chance to also look at unit tests.

smk78:
> ```
>           V =  self.get_volume_fraction(contrast, extrapolation)
>           Compute the surface given by:
>             surface = (pi *V(1- V)*porod_const)/ q_star
> ```
> 
> Does V(1- V) actually execute properly if coded like this???

Turned out this was, ahem, in a comment! Doh!

RichardHeenan:
Started to read the new invariant docs ( at least I hope so as I built 4.x from the branch), must admit I've never previously used the invariant perspective in sasview.

The "slit smeared" version of the equation, also works for oriented fibres etc, see attached paper, but we don't have any options to use it ???

[Crawshaw_Vickers_Briggs_Heenan_Cameron_hyrdation_cellulose_fibres_Polymer_2000.pdf](https://github.com/SasView/sasview/files/4232149/Crawshaw_Vickers_Briggs_Heenan_Cameron_hyrdation_cellulose_fibres_Polymer_2000.pdf)

The attached also references Shioya M, Takaku A. J Appl Phys 1985;58:4074 as a source for the equation for 3 phase systems, though I don't have access to that paper.

It might be worth spelling out that the results for Q* also depends on having a very good flat background. (Should there be an option to force say Q^-4 at high Q and fit the background?)

Then, in typical fashion the invariant perspective stopped working, and was giving a lot of -

TypeError: float() argument must be a string or a number
  File "C:\sasview42\sasview\src\sas\sasgui\perspectives\invariant\invariant_panel.py", line 712, in compute_invariant
    msg = self.state.__str__()
  File "C:\sasview42\sasview\src\sas\sasgui\perspectives\invariant\invariant_state.py", line 230, in __str__
    value = float(value) * 100

Started a new session, but its still not working, and I need to go now.

smk78:
Thanks @RichardHeenan. If you've pulled on the 4.x master then you should be building the new docs. Alternatively just look at the online _web_ docs (not the pdf ones). Will revisit this at Code Camp.

How does one reproduce the error? I haven't seen it.

butlerpd:
? If you pulled master then you are NOT building these docs right? this is a PR to pull these changes from branch: Issue1434_Invariant_documentation.  Note there are significant differences between this and those in master.  Also note that the trying to build this will cause errors due to unit test failures at the moment.

smk78:
Oops! Hadn't spotted this was a PR!

RichardHeenan:
Again, building 4.x in this branch, whenI send the data to tinvariant perspective, the gitbash window reports errors as below, (but not in the sasview console)

17:24:31 - WARNING : sas.sasgui.guiframe.proxy: 156: Connection to http://www.sasview.org/latestversion.json failed...
17:24:51 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 854: Invariant state: String or Unicode type required
17:24:51 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 854: Invariant state: String or Unicode type required
17:24:51 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 854: Invariant state: String or Unicode type required
17:24:51 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 854: Invariant state: String or Unicode type required
17:24:51 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 854: Invariant state: String or Unicode type required
17:24:51 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 854: Invariant state: String or Unicode type required

RichardHeenan:
I have not checked the maths, but with the default contrast now at 1e-06, I can only get NaN for Volume Fraction and Specific Surface, suspect the e-06 is not needed ??? (as per the example in the docs) and the gui needs to say e-06/A^2 instead of 1/A^2 for the units.

butlerpd:
**Maths**

Actually it does have to be e-6 otherwise you are way off.  So to get the right answer I have to enter the contrast as a e-6.  However, for the test I was doing for which it was 3e-6, I also got NaN for the default 1e-6 which I agree is problematic .. but so is I think not having that in there.  I have not had a chance to see why 1 vs 3e-6 is ok but 1e-6 gives a NaN.. that did not make sense

**Other errors**

I do not understand what those mean nor did I get them.  Lets see if we can force a PR build despite the failing unit tests (I suspect we are too clever :-( but what the hey. If it works then we can get a better test?

butlerpd:
ready for testing on Win64

butlerpd:
OK ... looks like the build succeeded despite failing 10 unit tests.  Not sure how we managed to have 10 unit test with incorrect "correct" answers but I guess stuff happens?

smk78:
Just tried building & running this branch myself...

**Maths**
Bizarre! A contrast less than 2.363e-6 seems to give a NaN for the volume fraction!

**Errors**
Yeah, I get them too. They appear when the first dataset is sent to the perspective. Sending subsequent datasets doesn't cause them to repeat. They don't seem to be related to the file type loaded. Suspect it's something to do with the initialisation of the perspective.

smk78:
Afterthought. I think the volume fraction is derived from solving a quadratic isn't it? In which case the NaNs might be because the calculation is trying to take a negative square root?

butlerpd:
ah!  good point @smk78 will try to check into the idea of a negative square root.  Not sure I understand the other issues.

butlerpd:
After some more testing/sleuthing:
1) Errors are from 4.x and have clearly been there a while (possibly from some unicode  fix), so fixing it should not be part of this PR. To see the errors from this PR installable or 4.2.2 just send a data file to the invariant and look in the logfile. Since it does not seem to impact anything the user sees and we are bringing 4.x to EOL (after one last release hopefully) am thinking it is not worth fixing.  I note it does not seem to be an issue with 5.0.1
2) the NaN error is reported as "ERROR" by 5.0.1 and the error in the log file is more helpful and says "negative discriminant" ... so seems @smk78 is right.  That said not sure there is a fix.  That will only be true when the contrast is too far off in one direction I think?

butlerpd:
Good point.  We can fix the problem of a nan by making it abs() but then it would hide underlying problems and the hapless user would be happy but publish incorrect values?  I guess the choices then are to 
- check for negative before doing the sqrt and if negative send an error to the console that explains the above (i.e. that either rho or q* must be wrong and places to look) with a NaN/Error in the GUI, OR
- just expand the documentation to explain the error OR
- All of the above

Thoughts?

RichardHeenan:
Error popups are likely tricky, can we add line to the gui saying "if NaN appears here, see documentation for possible reasons".

butlerpd:
sorry yes .. that was what I meant in "send message to console" -- mostly because I know how to do that with the logging system.  Popups are harder 🙂   gota do gui stuff :-(

Am redoing the whole unit tests at the moment.  completely fake at the moment it seems.  On the other hand, am now getting worried about the slit smearing being performed in 5.0.1 ... need to check against 4.x

(NOTE to self: don't answer in the github slack channel :-)

RichardHeenan:
Sorry, I also meant, permanently add a line to the gui saying "if NaN appears here, see documentation for possible reasons", as the users don't always notice a console message, though a console message would also be good.

smk78:
Ok, I have made several modifications to invariant_help.html based on the above comments:

 - the meaning of rho* & eta have been clarified
 - a bit more explanation about the SLD fluctuation has been added
 - the reader is told that the perspective will try to apply the correct calculation based on the i/p file
 - but is also warned that the perspective defaults to the pinhole calculation
 - the reader is warned that the pinhole calculation does not apply any smearing
 - @RichardHeenan 's additional references have been worked in
 - a warning has been added about NaN appearing in the volume fraction box if A<0.25 

Did I miss anything?

butlerpd:
Still need to check the uncertainties on vol fraction. Also need to agree what else to do on the GUI to explain the root of negative number. Then will need a final pass on  docs to make sure they match the final changes.

butlerpd:
So looking at the documentation, I think it is in much better shape.  Comments on left to do:
- I will work to rewrite the first bit now that I understand the caveats and limitations a bit better
- I now understand what eta is and what sld section is saying ... Now my question is: Is this relevant anymore?  it seems a bit complicated for the average user and I don't see that it brings any new information? or does it. 
  - The equations all use del rho everywhere so why introduce eta which is never used but takes up a whole section to explain what it is?
  - I can see it would be IF we allowed eta to be returned in the absence of any knowledge of del rho ... or even allowed the option of getting del rho if vol fraction is specified .. but neither is the case.
  - I imagine if we expanded to more general cases this might also become important?
  - But so far see no purpose for it outside of confusing users given the current implementation?
- Under parameters probably need to modify the line about what Q* allows one to compute in our case? or is this the one and only reason to define an eta? to tease the user into what they COULD have if only we would give it to them :-)
- In the note about A<0.25, which is the math point we should probably add for the domain science user that this means something is wrong about applying the invariant to this data or the contrast term supplied is wrong.

butlerpd:
aha!! so that DID fix the problem.  Now it is failing because of a unit test I apparently missed. time to fix that.

smk78:
@butlerpd wrote above:

```
- I now understand what eta is and what sld section is saying ... 
  Now my question is: Is this relevant anymore?  it seems a bit complicated for the average
  user and I don't see that it brings any new information? or does it. 
- The equations all use del rho everywhere so why introduce eta which is never used but takes
  up a whole section to explain what it is?
- I can see it would be IF we allowed eta to be returned in the absence of any knowledge of
  del rho ... or even allowed the option of getting del rho if vol fraction is specified .. but
  neither is the case.
- I imagine if we expanded to more general cases this might also become important?
- But so far see no purpose for it outside of confusing users given the current implementation?
- Under parameters probably need to modify the line about what Q* allows one to compute in
  our case? or is this the one and only reason to define an eta? to tease the user into what
  they COULD have if only we would give it to them :-)
```

The way I look at it, we are providing a tool. What's the problem with hinting at what the tool can be used for? It might stimulate someone to do something they otherwise might not have considered? We don't compute the SSA unless the user is able to supply Cp. So I don't see it really matters that we don't spoon feed the SLD fluctuation.

I have added a couple of sentences to the bottom of the Parameters introduction.

```
- In the note about A<0.25, which is the math point we should probably add for the domain
  science user that this means something is wrong about applying the invariant to this data
  or the contrast term supplied is wrong.
```

Done.

src/sas/sascalc/invariant/invariant.py
```diff
@@ -810,23 +813,39 @@ def get_surface(self, contrast, porod_const, extrapolation=None):
...
         """
-        # Compute the volume
-        volume = self.get_volume_fraction(contrast, extrapolation)
-        return 2 * math.pi * volume * (1 - volume) * \
-            float(porod_const) / self._qstar
+        # convert porod_const to units of A^-5 instead of cm^-1 A^-4 so that
+        # s is returned in units of 1/A.
+        _porod_const = 1.0e-8 * porod_const
+        return float(_porod_const) / (2 * math.pi\
+                                     * math.fabs(float(contrast))**2)
```
pkienzle: No need for `\` to join lines when the expression is in parentheses.

Also, you don't need to convert from integer to float in python 3, or in python 2 if you have `from __future__ import division` at the top of your file, so:
```python
return _porod_const / (2 * math.pi * abs(contrast)**2)
```

src/sas/sascalc/invariant/invariant.py
```diff
@@ -921,45 +945,52 @@ def get_volume_fraction_with_error(self, contrast, extrapolation=None):
...
+        # until contrast and porod_constant are given with uncertainties set
+        # them to 0
+        dcontrast = 0
+        # If and when dporod_const exists remember it likely needs unit
+        # conversion from cm^-1 A^-4 to A^-5.
+        dporod_const = 0 * 1e-8
+        _porod_const = porod_const *1e-8
+
+        s = self.get_surface(contrast=contrast, porod_const=porod_const)
+        ds = math.sqrt((dporod_const)**2 + 2 * (_porod_const\
```
pkienzle: Extra parens around `dporod_const` don't do anything.  And again, the `\` isn't needed here,

src/sas/sascalc/invariant/invariant.py
```diff
@@ -537,9 +537,12 @@ def _get_qstar_uncertainty(self, data):
             msg += " and greater than 1; got x=%s, y=%s" % (len(data.x), len(data.y))
             raise ValueError(msg)
         else:
-            #Create error for data without dy error
+            # For reduced data sqrt(I) is incorrect! if the data has no dy
+            # then we should not invent it and then propogate that completely
+            # bogus value.  So changed behavoir on Mar 23, 2020 to return
```
pkienzle: `behavoir` => `behaviour`

src/sas/sascalc/invariant/invariant.py
```diff
@@ -537,9 +537,12 @@ def _get_qstar_uncertainty(self, data):
...
+            # -1 instead
             if data.dy is None:
-                dy = math.sqrt(data.y)
+                return -1
```
pkienzle: What happens to the -1 in the caller?    It seems like you are going to need to check for -1 specially since a variance of -1 makes no sense, and so you might as well return None instead to say that it is undefined.  You could also return zero to say there is no uncertainty, but don't try to compute χ² if you do that.

src/sas/sascalc/invariant/invariant.py
butlerpd: Thanks - was wondering why so many "float" statements were being used but not in other code. can I remove all "float(x)" statements in the file then?

src/sas/sascalc/invariant/invariant.py
butlerpd: fixing point one ... and point 2 but not sure that helps without violating lint rules? I either need a
third line which will require a \ apparently, or I need to move the start of the second line much farther forward that it should be given the parenthesis?  suggestions?

src/sas/sascalc/invariant/invariant.py
butlerpd: can this now be set as "resolved"

src/sas/sascalc/invariant/invariant.py
butlerpd: can this be set as resolved?

src/sas/sascalc/invariant/invariant.py
butlerpd: can this be set as "resolved"

src/sas/sascalc/invariant/invariant.py
butlerpd: I have used None.  I assume that should work then? can we treat this as resolved?

butlerpd:
Mostly there I think then.  Remaining tasks
- [ ] Write new ticket to look at redundancy in utests and whether or not we should be allowing low Q extrapolation using a power law (to be addressed in a future PR)
- [ ] Resolve @pkienzle  issues
- [ ] Finish documentation tweaks
- [ ] Resolve @RichardHeenan request for changes

src/sas/sascalc/invariant/invariant.py
pkienzle: float() will convert strings to double, so some of them may still be needed, depending on the code.

It will also convert numpy float32 to double, but that will happen with pretty much any arithmetic operation in python except if you are combining it with another float32.

src/sas/sascalc/invariant/invariant.py
pkienzle: Returning None should at least raise an error if you are trying to use it improperly, which is better than either -1 or NaN.

butlerpd:
ready for testing on Win64

butlerpd:
@RichardHeenan Im guessing a gross conflagration of different environments.  Have just launched a build (I hope) which you can download and test cleanly once it is done.  Though I am about to make some more changes to the docs (not related to your requests I dont think) so you may or may not want to install and test this version.

smk78:
@RichardHeenan , when you say you ' built the sasview v4 branch locally' you do mean you built the Issue1434_Invariant_Documentation branch for sasview? I just built the docs and tested the interface in Anaconda2 and the changes are good.

I _do_, however, also see the same messages as you in the console:
```
21:48:36 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 858: Invariant state: String or Unicode type required
21:48:36 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 858: Invariant state: String or Unicode type required
21:48:36 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 858: Invariant state: String or Unicode type required
21:48:36 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 393: Error occurred computing volume  fraction: Could not compute the volume fraction: negative discriminant
21:48:36 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 858: Invariant state: String or Unicode type required
21:48:36 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 858: Invariant state: String or Unicode type required
21:48:36 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 858: Invariant state: String or Unicode type required
21:48:36 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 393: Error occurred computing volume  fraction: Could not compute the volume fraction: negative discriminant
21:48:47 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 393: Error occurred computing volume  fraction: Could not compute the volume fraction: negative discriminant
21:48:58 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 393: Error occurred computing volume  fraction: Could not compute the volume fraction: negative discriminant
21:49:00 - ERROR : sas.sasgui.perspectives.invariant.invariant_panel: 393: Error occurred computing volume  fraction: Could not compute the volume fraction: negative discriminant
```
But it seems to work nonetheless...

butlerpd:
Yes ... the first set of errors @RichardHeenan as seen also by @smk78 those errors do not seem to matter but have been there for a very long time in 4.x releases.  So fixing this should not be part of this PR.

More importantly, since it does not cause any real problem and we are trying to sunset 4.x I would argue we should not try to fix later either (in other words am not writing a new issue for this). Thoughts?

smk78:
I wonder if the errors, at least some of them, come from the perspective initialisation; eg, with no data input the vol frac has to be ERROR. Then, even when you do pass it data, it's highly likely that the default SLD will still give a vol frac that is ERROR. And could the string/unicode errors be from the empty boxes like the one for the Porod Constant?

butlerpd:
ready for testing on Win64

butlerpd:
Yun's mailed in review:
Overall, it is a very well done and detailed description of the invariant. But I have a few minor suggestions.
1) There is a minor inconsistency regarding the calculated value for the
invariant in the introduction and the first equation. It is just a
prefactor difference, so not a big deal. But thought that it would be
better to be consistent.
2) There is still an assumption about the invariant, i.e., it may not
be universally true that it is a constant when its phase changes. The
assumption is that all component are incompressible, i.e., the
relative volume fraction of different components never changes.
Otherwise, obviously, the invariant is not a constant for compressible
liquid. This doesn't mean that all equations are not right. However,
it is just that invariant is not a constant any more.
3) For high Q extrapolation, not sure why the exponent has to be
between -3 and -4. Not sure wheter the estimation of the relative error
is needed as this depends on the power law function that can fit the
data.
4) For the volume fraction section, I assume there is a typo for " If
A<0.25 then the". I guess what it means to say is " If A>0.25 then
the"?
5) For "extension to three or four phases", this section basically
discusses the two phase systems with references to multiple phases
system. would it be better to change the section title to something
like "some details for two phase systems". Just a suggestion.

smk78:
I have corrected many typos in @butlerpd 's last revision of invariant_help. I have also word-smithed it a bit, moved a few lines around, and _removed_ the numbering in the How To Use section as list format was screwing up the formatting and impairing readability.

Addressing Yun's comments above:
1. Left for @butlerpd to clarify with Yun.
2. Have added a mention about system incompressibility to the Introduction.
3. Left for @butler to ponder as he added this wording! :-) Presumably the train of thought was that we are considering surface scattering and therefore the q-dependence must be >3 and <4? Which is true. But is that the same as the q-dependence of the function being used to extrapolate the data? Is any exponent ok as long as it closes the integral?
4. Whoops! That was my fault! Now corrected.
5. I have tried to make this a bit more 'here is the general formalism' and then 'look what happens in the two-phase case'. Is it any better?

smk78:
@butlerpd the Invariant analysis screenshot needs replacing!

smk78:
Ahead of the 5.0.2 release, invariant_help.rst from the ticket-1434 branch (the most current invariant docs) has been cherry-picked and copied into ESS_GUI master.

I have then added in the pre-prepared 5x invariant panel screenshots which also necessitated a small change the 'Using Invariant Analysis' text due to different behaviour to 4x

butlerpd:
The latest commit should address most of @yunliu01 remaining points:
1. Three things were done to address this issue
    - Added vector q in the beginning when describing the integral over all Q
    - Removed the misleading statement suggesting that to be equivalent to integrating over 4 pi in scattering angle (it is 4 Pi in q space and from o to inf in magnitude) 
    - Added a note explaining the lack of 4pi prefactor in the integral
2. Already addressed by @smk78 
3. Another two parter
    - Technically I believe that theoretically to be 100% correct one should be to high enough Q that only interfaces are visible wich for a pinhole camera would be somewhere between -3 for diffuse or highly surface fractal? and -4 for a sharp interface.  For slit smeared data however it will be less than -3 of course.  Also there may be cases where it is legitimate to have smaller power laws but believe that as a starting rule of thumb this is a useful idea.  Have so far only edited to be a less dogmatic.
    - Indeed the numerical values were totally  incorrect and don't remember putting them there. Agree that it was very poorly done overall but the aim was to point out the very frequently overlooked fact that the high Q data is quite important even though the intensity is small because it is being weighted in the pinhole case by q^2!  
       - Have removed the verbiage from the individual extrapolation sections
       - Have added some verbiage to the introduction to those sections stating the above hopefully more clearly and usefully.
4. Fixed by @smk78 
5. slightly added to @smk78  start at this by expanding his "In the case of a two-phase system:" title into a sentence and added some verbiage to the final derivation all to indicate that we are demonstrating that the general formalism reduces to the earlier equations in the special case of only 2 phases.

butlerpd:
ready for testing on Win64

butlerpd:
ready for testing on Win64

smk78:
Trying to build the branch locally, I note that at some point a pathlib dependency was added to sasmodels which is causing generate.py to fail. According to https://pypi.org/project/pathlib/ it was only shipped with Python from 3.4, so the 4.x build servers may need to have it installed for the full docs to compile correctly (because this branch compiles under Py2 but sasmodels now has a Py3 dependency).

smk78:
Have built this locally. Reads fine to me. I did make a grammar change. And I also reinstated (with slightly modified text) the warning about the Sv values having previously been wrong. I think we owe users that honesty?

rozyczko:
> Trying to build the branch locally, I note that at some point a pathlib dependency was added to sasmodels which is causing generate.py to fail. According to https://pypi.org/project/pathlib/ it was only shipped with Python from 3.4, so the 4.x build servers may need to have it installed for the full docs to compile correctly (because this branch compiles under Py2 but sasmodels now has a Py3 dependency).

Yes, we did have issues with this dependency on the 4.x build machines and wanted to fix it without having to modify the code. Wojtek found a hacky way of forcing python 2 to import pathlib properly and it is now "working" on the build servers.

RichardHeenan:
OK, I’ve only done a couple of hours work today, mostly Loki, so I still need to look at invariant ….
R

butlerpd:
ah! Thanks @rozyczko I did not realize it was fixed on the build servers!!  That is good news ... though it makes sense that would be why the builds aren't failing even though Travis is.

I have several notes in 4.x tickets and sasmodels PR I think about this problem @smk78 ... the issue was introduced into sasmodels as I recall by @toqduj to fix string issues nicely. It works well for 5.x because it is running on python 3 but you need some dependency work as noted by @pkienzle in that PR when it was merged into sasmodels.  I can now stop worrying about how we can release the final 4.3 then .. and ignore Travis warning - Excellent

RichardHeenan:
Running build 100 from PR, the gui looks good, minor issues in docs -

mentions m = -3 or -4 instead of +3 or +4 in the context, else give equation as q^m

special case of m=-3 is for _infinite_ slit smearing

no mention of need to remove correct flat (incoherent and/or inelastic) background from SANS data, as the method here does not allow for this (perhaps we should have a box to put in optional flat bkg ?) 


refs [3] & [4] at end have weird (1,2) with circular links back to same page, formatting issue?

butlerpd:
Thanks for the corrections @smk78 my English is not so good.  Please don't tell my mother who was an English teacher 😃

As for the Sv being wrong in previous versions, I agree 100% on transparency but not at all convinced this is the right way to do it.  I have a few reasons:
1. I thought that was in fact what the release notes were for (and for important things notes on the release web page). Otherwise what is the point for those? Isn't that in fact how professional products work?
2. If one takes this to its logical conclusion (the philosophical ad absurdum method - showing off my classical philosophy training even if my English is off) you end up with the documentation about what a feature is and how it should work and how to use it with a copy of the looong list of release note fixes.  Alternatively as you write it **here** as a dire red warning for "all previous versions" somebody has to remember when putting out the next next release to remove the warning.  Something very likely to be overlooked if nobody has worked on that part of the code, particularly if it takes some time to get there.  This again would be why one would instead use release notes which are automatically preserved by release.
4. From the perspective of a user, I normally do NOT go looking a the feature documentation to see if what I did before was wrong. If I have a concern I will most likely look in the release notes or forum emails etc.  I will go to the docs when I am trying to use a feature for the first time and want some guidance on when to use, how to use, caveats on its use etc .. things to help me figure out what I'm supposed to do or keep me from misusing it.  Finding some statement about earlier versions of this program would only confuse me and for myself I would find it somewhat "unprofessional" in the sense of not a professional job of writing documentation which would turn me off from the product.. but maybe that is just me.

so I would argue that this should be moved to the release notes (Now in fact not later) and try to remember to add it to the top of the list of fixes listed on the web page.  This is also I would argue why we need to work on getting a real forum going that is easily searchable.

I'd be interested in what others think.

butlerpd:
* good catch @RichardHeenan on the sign of m!! ... always confusing that.
* With respect to background that is what the input box of "background"  is. Its use I think is documented in the second half which goes through the process step by step. There is I believe a ticket (possibly in the wishlist milestone) to offer an option to have the software automatically try to find a background the way the PR perspective does but that is a lot of work that cannot happen as part of this PR.  So not sure what you are asking for here. Could you please clarify?
* Regarding the -3 special case...again not clear to me exactly what are you asking for?  It is documented in a note that a sharp interface which would give a -4 slope would give a -3 for an infinite slit smeared. What is left unsaid, but hopefully is ok, would be that diffuse interfaces, with say a surface fractal structure, which would usually give slopes between -3 and -4 for pinhole would give -2 to -3 for slit smeared.  Yun and I have also discussed vigorously and he finally came up with a case he thinks of an interface whose density changes perpendicular to the surface is some fractal like function could mathematically yield a flatter slope for pinhole even (-2?) but we are not sure how common it would be to find such systems in real life. There is also the question of polymer chains. 
 etc. Hence the "it may be other than the guiding rule of thumb but you should think about what it means and weather that is reasonable" statement
* Didn't notice the odd reference numbering. Is that because they are hyperlinks back into the text to where it is referenced.  Those I find ugly looking but they do work I must say. I will check

smk78:
> ah! Thanks @rozyczko I did not realize it was fixed on the build servers!! That is good news ... though it makes sense that would be why the builds aren't failing even though Travis is.
> 
> I have several notes in 4.x tickets and sasmodels PR I think about this problem @smk78 ... the issue was introduced into sasmodels as I recall by @toqduj to fix string issues nicely. It works well for 5.x because it is running on python 3 but you need some dependency work as noted by @pkienzle in that PR when it was merged into sasmodels. I can now stop worrying about how we can release the final 4.3 then .. and ignore Travis warning - Excellent

I added pathlib to my Anaconda2 environment!

smk78:
@butlerpd I guess I figure my users are more likely to see a dire warning in the docs than in the release notes... Though I would actually, how do you say, double-down, and do both.

Changing the text to say all 4x versions upto and including 4.2.2 and all 5x versions upto and including 5.0.2 are wrong would deal with your point 2.

But I'll go with a majority verdict!

butlerpd:
But wouldn't the users seing it in the docs be the ones that don't actually need to know? if they are looking at those docs then their version will give the right answer. The people who need to know are those who used it successfully before.  Why would they be looking at the docs again? most people don't look at them even the first time -- they email help@sasview 😃  

At least that would change would indeed deal with it becoming incorrect but the question remains. are we going down the path of putting all the but fixes (that cause or could possibly cause an erroneous result) into the docs as well as in the release notes? and if so how far back should we go? there are quite a few we have found over the years I think.

But I would like to hear what others think as you said

RichardHeenan:
My vote - Just leave the factor 2 mention in the release notes for now, see how long before anyone notices, as I doubt that many people use the invariant perpsective.

butlerpd:
@RichardHeenan I think I understand your comment 3 which was in fact quite straight forward and correct - my bad.  In the latest commit I have fixed 3 of the 4 concerns:
* Corrected the sign (everywhere it occurred)
* Added a (long) parenthetical phrase at the end of the first sentence under calculation which now reads: 
```
Assuming isotropic scattering, acquired on a typical 'pinhole geometry'
instrument, the invariant integral can be computed from the 1D reduced data
(assuming the reduced data has removed all background from sample
holders,incoherent scattering in the case of neutrons, etc.) as:
```
* added the word *infinite* as you appropriately suggest

The problem with item 4 is that it is a Sphinx uglyness.  References are two way: clicking in the text will get you to the reference while clicking in the ref section takes you to the location in the text that the citation applies to (see refs 1,2 and 5).  The problem comes when a citation is used twice in the document.  At that point, the single ref number is no longer adequate so sphinx adds them to the beginning of the reference.  I see 3 options:
* Leave as it is. This allows accessing the different points in the text that refer to this partticular citation
* Only use the ref once and delete if from the second one (or maybe be real ugly and "hard code" a ref with no link in the second case
* remove all citations within the text and just leave as a bibliography.

Note that there are many places throughout the documentation that have this uglyness ☹️

butlerpd:
Note that I am also still gathering a bit more input on whether (and/or when) previous bug warnings should be noted directly in the documentation rather than simply in the release notes before merging.

smk78:
Merging this branch should mean that 4.x will then compute the same as 5.x once PR https://github.com/SasView/sasview/pull/1590 is merged. _AND_ both will have the latest help text.

smk78:
@smk78 wrote above:
> Merging this branch should mean that 4.x will then compute the same as 5.x once PR #1590 is merged. _AND_ both will have the latest help text.

NOT QUITE! @butlerpd notes in PR https://github.com/SasView/sasview/pull/1590 that:
>Finally a bit moot but spent most of my time rederiving ds which was clearly still wrong -- that needs backporting to 4.x PR

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski (approved):

PR #1482, opened 2020-03-19
4.x: Check packages comparison to YAML files

krzywon:
As suggested yesterday, check_packages.py in master now compares the appropriate yaml file against what is currently installed on the users system. I've added two new dependencies that might need to be incorporated into the existing yaml files, but I only have a Windows system right now. The 5.x fix is a little more complex due to the number of yaml files in the branch so I haven't finished it yet.

Features:
- Checks the installed packages against the yaml file, including version numbers
- Flags packages that aren't installed as a major issue
- Flags packages with unexpected or unknown versions as a minor issue

Possible TODOs:
- Add PyYaml and colorama to the dependency files

### DIFF
--- a/check_packages.py
+++ b/check_packages.py
@@ -4,10 +4,16 @@ Check the existing environment against expected values
 from __future__ import print_function
 
 import sys
-import yaml
 import subprocess
-
-from colorama import Fore, Style
+try:
+    import yaml
+    from colorama import Fore, Style
+except ImportError:
+    print("The PyYaml and colorama packages are required for this module. To " +
+          "install them, please try one of the following:")
+    print("\tAnaconda: conda install pyyaml colorama")
+    print("\tBase python: pip install pyyaml colorama")
+    quit()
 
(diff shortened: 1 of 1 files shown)


pkienzle:
On Mac conda, I needed to install PyYAML via pip since the conda yaml package didn't work on python 2.7:
```pip install pyyaml```
I also needed to install the `colorama` package.

smk78:
Once I installed pyyaml in my Anaconda2/Win10/x64 environment and rebuilt this branch check_packages worked a dream.

pkienzle:
Now that I remember that you need to `conda search "*yaml*"' I see that PyYAML **is** available within conda using `conda install pyyaml`.  This works for me as well.

butlerpd:
We know why tests are failing and that fix is already in master (problem with travis.yml). So what else is needed to approve this one? seems like it is similar to 5.x which was approved? or is it more complicated than that?

PR #1485, opened 2020-03-19
Ess gui 1484 bg colors

rozyczko:


gonzalezma:
Unfortunately, this does not solve the problem. I guess this is a Qt issue in Ubuntu. The problem is that on the first selection, the selected line changes (automatically?) bg to blue and fg to white. But once that Data1 and Data2 are selected and are compatible, the bg is changed to white. Up to that point, things are OK. But if then one tries to change the selection, the selection box remains with a white bg and I guess that the fg is automatically set to white, so the text becomes invisible.
This is in contrast with the behaviour on Windows, where the line selected has always a blue background with white foreground.

gonzalezma:
BTW, the changes in FittingWidget.py should go into another branch, no?

rozyczko:
Thanks for looking at this. Yes. the FittingWidget changes showed up because I based my new branch off the branch with those Fitting changes, rather than off ESS_GUI directly... a bit clumsy.
I think your suggestion of hard coding the colour to off-white is probably the best.
We would need to set the comboboxes to this new colour on initialization though.

Issue #1476, opened 2020-03-18
Quick Plot in Data Explorer not working for 1D data


gonzalezma:
An empty plot window appears, with limits X=[0,1] and Y=[0,1]. But it works fine for 2D data.
Tested on Windows 8 (SasView 5.0.1) and building from ESS-GUI in Ubuntu 18.

rozyczko:
Hmm nice find. We should probably start doing proper GUI testing before release. Whether automated or manual.

gonzalezma:
I'm doing a bit of manual testing now. I guess Richard is going to do the same. To try to do something practical as well, I could try to fix this, unless you already know what is going on.

rozyczko:
Go ahead, Miguel. Let me know if you need any help.

gonzalezma:
Piotr, I found that what is happening is that there is a call to Plotter.plot with None data, so the function returns immediately without doing anything (just refreshing the canvas). I tried the following:

    def plot(self, data=None, color=None, marker=None, hide_error=False, transform=True):
        """
        Add a new plot of self._data to the chart.
        """
        if data is None:
            if self.data[0] is not None:
                # for quick plot
                data = self.data[0]
            else:
                # just refresh
                self.canvas.draw_idle()
                return

This works to fix the Quick Plot and it does not seem to break the usual fit plots, but as I don't know 'who' else could be using this method, I did not want to commit anything before getting your opinion.

rozyczko:
This is a bit more clumsy. At some point I rewrote the Plotter.plot() behaviour, such that it now expects `data` to be in the argument list.
The call in DataExplorer.py does not put data in when calling `plot`, so no plot is generated.
All that is needed is to add `data=data` to arguments of the `plot()` call in lines 1520 and 1542.

Your fix does work but it's trying to mask the actual issue of QuickPlot not sending the stuff it should :)

rozyczko:
fixed in ESS_GUI_1476_quick_plot.

PR #1486, opened 2020-03-19
Fixed Plotter.plot() call for 1D plots in Quick Plotting. #1476

rozyczko:


gonzalezma:
Fixes the issue and changes are minimal, so it can be merged.

Issue #1431, opened 2019-12-09
When loading plugin model with a spurious unicode character plugin editor crashes


wpotrzebowski:
This a follow up to #1426. There is currently no handling for the a spurious unicode character in plugin model editor (Edit custom model), so it crashes when one tries to edit it.

wpotrzebowski:
To be decided if it is 5.0.1 or beyond

rozyczko:
At the least, this shouldn't crash the module. Maybe a simple try..catch for the relevant exception is sufficient at this stage?

rozyczko:
Added encoding directive to the plugin log writer. This helps and SV opens nicely now.

butlerpd:
was this committed to ESS_GUI (i.e. now in 5.0.1) and waiting for testing? or waiting to be made into a pull request?

rozyczko:
This is in ESS_GUI and awaits testing, please.

butlerpd:
5.0.1 is now out so this needs to move to 5.0.2 or be closed.  Could someone who knows how to create this original error please check that the problem no longer exists in the 5.0.1 release?

Otherwise I recommend we close this.

wpotrzebowski:
@rozyczko: does it mean that it is fixed already in ESS_GUI branch?

rozyczko:
The fix is already in ESS_GUI.

wpotrzebowski:
@rozyczko: just to confirm the expected output should be something like this: 
  File "site-packages/sasview-5.0.1-py3.6-macosx-10.9-x86_64.egg/sas/qtgui/Utilities/TabbedModelEditor.py", line 141, in onLoad
  File "site-packages/sasview-5.0.1-py3.6-macosx-10.9-x86_64.egg/sas/qtgui/Utilities/TabbedModelEditor.py", line 149, in loadFile
  File "/Users/sasview/miniconda3/envs/qt5_osx/lib/python3.6/encodings/ascii.py", line 26, in decode
UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 52: ordinal not in range(128)

And model with spurious character shouldn't load?

rozyczko:
No, you should be able to load the plugin into the editor. At least this is what sasview does on windows and Ubuntu...

wpotrzebowski:
@rozyczko It seems, it needs some more investigation then. Which unicode character you use for testing?

rozyczko:
[...]

PR #1489, opened 2020-03-21
Added utf-8 specs to open plugin file. Addressing #1431

wpotrzebowski:
When loading plugin model with a spurious unicode character plugin editor crashes #1431


wpotrzebowski:
5.0 ready for testing on OSX

PR #1499, opened 2020-03-31
Ess gui 1355 file converter

rozyczko:


src/sas/qtgui/Utilities/FileConverter.py
```diff
@@ -0,0 +1,487 @@
...
+        source.thickness = Utilities.toFloat(self.txtMSa_Thickness.text())
+        source.transmission = Utilities.toFloat(self.txtMSa_Transmission.text())
+        source.temperature = Utilities.toFloat(self.txtMSa_Temperature.text())
+        source.position = Vector(x=Utilities.toFloat(self.txtMSa_PositionX.text()),
+                                 y=Utilities.toFloat(self.txtMSa_PositionY.text()))
+        source.orientation = Vector(x=Utilities.toFloat(self.txtMSa_OrientR.text()),
+                                    y=Utilities.toFloat(self.txtMSa_OrientP.text()),
+                                    z=Utilities.toFloat(self.txtMSa_OrientY.text()))
+
+        source.details = self.txtMSa_Details.toPlainText()
```
krzywon: Like my other comment, shouldn't all of this be in getSampleMetadata()?

src/sas/qtgui/Utilities/FileConverter.py
```diff
@@ -0,0 +1,487 @@
...
+        """
+        sample = Sample()
+        sample.name = self.txtMSa_Name.text()
+        sample.beam_size = Vector(x=Utilities.toFloat(self.txtMSo_BeamSizeX.text()),
+                                  y=Utilities.toFloat(self.txtMSo_BeamSizeY.text()))
+        sample.beam_shape = self.txtMSo_BeamShape.text()
+        sample.wavelength = Utilities.toFloat(self.txtMSo_BeamWavelength.text())
+        sample.wavelength_min = Utilities.toFloat(self.txtMSo_MinWavelength.text())
+        sample.wavelength_max = Utilities.toFloat(self.txtMSo_MaxWavelength.text())
+        sample.wavelength_spread = Utilities.toFloat(self.txtMSo_Spread.text())
```
krzywon: Shouldn't all of this be in getSourceMetadata()?

src/sas/qtgui/Utilities/FileConverter.py
rozyczko: oh my.. embarrassing mistake ;)

rozyczko:
> I believe some of the source and sample meta data are getting placed into the wrong classes, so those elements are never saved. See my comments. I consider this a blocker.

Good catch. fixed.

> 1. The file selector button reads 'Save File' which might suggest the conversion happens when the button is pressed. Different wording ('Select File'?) might be helpful.

This can't be easily achieved with the standard `QtFileLoad` widget. One can overwrite the labels, but this would require reimplementing the class and all of the functionality of `getSaveFileName` static function. By no means trivial. Unless you think this is indeed a serious issue, I'd like to avoid that.

> 2. Most other tool windows are separate window elements, but this one acts as a modal window. Even worse the window stays in front of other windows when focus is lost.

Fixed

> 3. Most other tool windows have a close button in addition to the normal window close.
Added the `Close` button.

> 4. For files that are saved but not immediately loaded, success isn't obvious, with only a small line in the log at the bottom.
I also added a message bar text update. It is visible when no load takes place, since the file load immediately replaces the conversion message, but at least now we have this information shown in two places (for when the checkbox is off).

src/sas/qtgui/Utilities/FileConverter.py
rozyczko: yes, it should. fixed

smk78:
Just testing the changes...

image

If you try and convert a file without adding any metadata you get:

```
12:33:05 - ERROR: Traceback (most recent call last):
  File "D:\SasViewCodeDev\ticket1355-file-converter\src\sas\qtgui\Utilities\FileConverter.py", line 119, in onConvert
    self.readMetadata()
  File "D:\SasViewCodeDev\ticket1355-file-converter\src\sas\qtgui\Utilities\FileConverter.py", line 281, in readMetadata
    'detector': [self.getDetectorMetadata()],
  File "D:\SasViewCodeDev\ticket1355-file-converter\src\sas\qtgui\Utilities\FileConverter.py", line 204, in getDetectorMetadata
    detector.distance = Utilities.toFloat(self.txtMD_Distance.text())
AttributeError: module 'sas.sascalc.file_converter.FileConverterUtilities' has no attribute 'toFloat'
```

Tried a 1D ASCII text file too and got the same thing.

rozyczko:
> Just testing the changes...
> 
> image
> 
> If you try and convert a file without adding any metadata you get:
> 
> ```
> 12:33:05 - ERROR: Traceback (most recent call last):
>   File "D:\SasViewCodeDev\ticket1355-file-converter\src\sas\qtgui\Utilities\FileConverter.py", line 119, in onConvert
>     self.readMetadata()
>   File "D:\SasViewCodeDev\ticket1355-file-converter\src\sas\qtgui\Utilities\FileConverter.py", line 281, in readMetadata
>     'detector': [self.getDetectorMetadata()],
>   File "D:\SasViewCodeDev\ticket1355-file-converter\src\sas\qtgui\Utilities\FileConverter.py", line 204, in getDetectorMetadata
>     detector.distance = Utilities.toFloat(self.txtMD_Distance.text())
> AttributeError: module 'sas.sascalc.file_converter.FileConverterUtilities' has no attribute 'toFloat'
> ```
> 
> Tried a 1D ASCII text file too and got the same thing.

Looks like you didn't do `python setup.py build install` after pulling another branch?

krzywon:
I made a small change to catch a couple more meta data lines getting put in the wrong place. Otherwise this now looks fine. @smk78 feel free to merge this when you agree.

smk78:
Yes, pulling and building made a difference! Sorry! But... if I try and load the 2D BSL up pops the little dialog to select frames (good!), so I select 1 and 40 with increment 1 and Ok, click convert, and now I get:

```
18:01:27 - ERROR: Traceback (most recent call last):
  File "D:\SasViewCodeDev\ticket1355-file-converter\src\sas\qtgui\Utilities\FileConverter.py", line 136, in onConvert
    dataset = self.extractBSLdata(self.ifile)
  File "D:\SasViewCodeDev\ticket1355-file-converter\src\sas\qtgui\Utilities\FileConverter.py", line 414, in extractBSLdata
    frame_data = loader.load_frames(frames)
  File "D:\SasViewCodeDev\ticket1355-file-converter\build\lib.win-amd64-3.6\sas\sascalc\file_converter\bsl_loader.py", line 153, in load_frames
    x = self.n_rasters * range(1, self.n_pixels+1)
TypeError: unsupported operand type(s) for *: 'int' and 'range'
```

No file is produced.

rozyczko:
Hmm I remember fixing that very issue but somehow this didn't get pushed...
Maybe because it is in sascalc? Fixed now.

**However**, loading this file in 4.2.2 gives me 
`2020-04-01 20:44:12 : iteration over a 0-d array`
error and no conversion is done.

In the current branch I end up with `ERROR: Unable to calculate dimensions of 2D data` coming from the cansas writer and upon debugging ( sas\sascalc\file_converter\nxcansas_writer.py line 336 )
it seems data sizes are inconsistent (`n_rows * n_columns != data.size`)

Loading single frames seems to work, though.

smk78:
Yes, I now see that same error. But as you say you can convert and load 1 frame. So I am inclined to say for now that we accept this as it stands and revisit if someone actually complains that they can't load multiple frames? Unless @krzywon has any thoughts about the error?

Is there a reason that the frame selection popup always makes the default for the first and last frames to be n_frames - 1? Would defaulting the first frame to 1 not be more sensible?

And if we can't get the loading of multiple frames working easily, can we change the error message to say, eg, Unable to calculate dimensions of 2D data; loading a single frame. (Or, better still, tell them which frame!)?

rozyczko:
It probably would make more sense to use the first and last frame indeed :)

krzywon:
I agree with @smk78 that this is ready but with the caveat that a ticket should be added regarding the multi-frame 2D loading issue. I don't think it's a straightforward and likely needs to be fixed in both master and ESS_GUI. My only question is do we merge now and put it in 5.0.2 or do we wait? My inclination is now.

butlerpd:
given 5.0.1 has absolutely nothing it seems that this is better than that? and we are about to release 5.0.2 so if we merge now this would make it in. My 2 cents is it should definitely go in now.

smk78:
Just been testing again. The only thing not working properly is multi-frame 2D BSL; ordinary 2D BSL is ok.

So I vote we merge and note this in the Release Notes. And if I can find @butlerpd 's new ESS_GUI documentation branch I might pop a note in the help...

butlerpd:
that would be ESS_documentation @smk78 :-)

smk78:
Thank you, Sir! Got it!
Actually, turns out it is slightly more involved because the File Converter is not in the TOC's of course... it being a 'new' feature an awl.

smk78:
Ok, documentation updated in https://github.com/SasView/sasview/pull/1506

Someone merge this, I want my dinner!

PR #1512, opened 2020-04-14
Fix merge conflict in the windows YML file

rozyczko:
Quick fix for conflict merge after large PR merge.


rozyczko:
explicit versioning was added to address a buggy IPython+pyinstaller combo. This, it seems, has been resolved and the most recent versions give correct behaviour. This is what's committed now for 5.0.2.
We need to revisit the fix and probably freeze the versioning to the current number yet again, so we are on the safe side, until the next series of module updates. Probably before 5.0.3 or 5.1, whichever comes first.

Issue #1490, opened 2020-03-23
Problem using constraints in 5.x


wpotrzebowski:
Reported by a user: 

"When you add constraints, it is a pain and it is not working.
you can not define all M1= M2 and M3=M2  (with M1, M2 and M3 having 10 parameters)
also you can not remove more than one constraint

there are some bugs in this part.
The old version was easier to handle.
I had 3 curves and it only displays M3 twice and M1 if I want to add constraints. M2 does not exist for some reasons.

Also how can you add constraint by yourself?
once you filled the lines, you have to pass by the add constraint bottom to add constraints and not being able to type them… or copy paste..."

butlerpd:
As per today's conversation am moving this to a 5.0.3 milestone in preparation for a 5.0.2 release.

butlerpd:
As per today's conversation am moving this to a 5.0.3 milestone in preparation for a 5.0.2 release.

gonzalezma:
Tests using a standard case and comparing 5.0.1 with the new ESS_GUI_1490 branch:
[SasView5_constraints_UI.pdf](https://github.com/SasView/sasview/files/4540058/SasView5_constraints_UI.pdf)

gonzalezma:
Updated ESS_GUI_1490 branch works reasonably well:
[SasView5_constraints_UI_2.pdf](https://github.com/SasView/sasview/files/4540070/SasView5_constraints_UI_2.pdf)

RichardHeenan:
My views re questions asked in the pdf above.
Q1.  all or only adjustable params on the left side ? Difficult one this as lots of users will start by wanting to constrain all in one FitPage to all in another, then edit some of the constraints.  If you later "fix" the parameter, the constraint on it may  still exist in the maths but is not actually being used (thus ought to change back to the normal font). Of course being able to save and reload projects with multiple data and multiple constraints is good.
What we now still need is a way to switch data set associated with a FitPage. Having set up my complex constraints across several data sets I have to enter all the constraints over again for different data. i.e. send data to overwrite that for any current FitPage. Yes there is a copy fitPage, but it won't replicate the constraints.]
Q2 - OK
Q3 as per Q1, I think need to add everything, but you might convince me otherwise.
Q4.1 & 4.2, agree
Q5 I need to try 5.0.2
Q6 yes a mechanism to delete a constraint would be useful
Q7 yes we ought to check for bad syntax, non-existent parameters before accepting the constraint

wpotrzebowski:
[...]

PR #1533, opened 2020-04-29
Ess gui 1490 constraints

m2cci-NMZ:
Changes the user experience when  using constraint. See #1490.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
@@ -105,10 +105,10 @@ class ComplexConstraint(QtWidgets.QDialog, Ui_ComplexConstraintUI):
         items1 = [param for param in self.tabs[tab_index1].main_params_to_fit]
         self.cbParam1.addItems(items1)
 
-        # M2 doesn't have to be non-constrained
+        # M2 has to be non-constrained
         self.cbParam2.clear()
         tab_index2 = self.cbModel2.currentIndex()
-        items2 = [param for param in self.params[tab_index2]]
+        items2 = [param for param in self.params[tab_index2] if not self.tabs[tab_index2].paramHasConstraint(param)]
         self.cbParam2.addItems(items2)
 
         self.txtParam.setText(self.tab_names[tab_index1] + ":" + self.cbParam1.currentText())
--- a/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
@@ -101,7 +101,7 @@ class ComplexConstraint(QtWidgets.QDialog, Ui_ComplexConstraintUI):
         """
         self.cbParam1.clear()
         tab_index1 = self.cbModel1.currentIndex()
-        items1 = [param for param in self.params[tab_index1] if not self.tabs[tab_index1].paramHasConstraint(param)]
+        items1 = [param for param in self.params[tab_index1]]
         self.cbParam1.addItems(items1)
 
         # M2 doesn't have to be non-constrained
--- a/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
@@ -101,7 +101,7 @@ class ComplexConstraint(QtWidgets.QDialog, Ui_ComplexConstraintUI):
         """
         self.cbParam1.clear()
         tab_index1 = self.cbModel1.currentIndex()
-        items1 = [param for param in self.params[tab_index1]]
+        items1 = [param for param in self.tabs[tab_index1].main_params_to_fit]
         self.cbParam1.addItems(items1)
 
         # M2 doesn't have to be non-constrained
--- a/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
@@ -288,7 +288,7 @@ class ComplexConstraint(QtWidgets.QDialog, Ui_ComplexConstraintUI):
             self.lblWarning.setText(txt)
         self.constraintReadySignal.emit(cons_tuple)
         # reload the comboboxes
-        #self.setupParamWidgets()
+        self.setupParamWidgets()
 
     def onSetAll(self):
         """
--- a/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
@@ -42,6 +42,7 @@ class ComplexConstraint(QtWidgets.QDialog, Ui_ComplexConstraintUI):
         self._constraint = Constraint()
         self.all_menu = None
         self.parent = parent
+        self.redefining_warning = ""
 
         self.warning = self.lblWarning.text()
         self.setupData()
@@ -125,12 +126,11 @@ class ComplexConstraint(QtWidgets.QDialog, Ui_ComplexConstraintUI):
             txt = "No parameters in model "+self.tab_names[0] +\
                 " are available for constraining.\n"+\
                 "Please select at least one parameter for fitting when adding a constraint."
-            self.lblWarning.setText(txt)
         else:
+            txt = self.redefining_warning
             self.cmdOK.setEnabled(True)
             self.cmdAddAll.setEnabled(True)
-            txt = ""
-            self.lblWarning.setText(txt)
+        self.lblWarning.setText(txt)
 
     def setupTooltip(self):
         """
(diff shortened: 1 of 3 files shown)


wpotrzebowski:
5.0 ready for testing on OSX

src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
```diff
@@ -258,15 +259,36 @@ def constraint(self):
...
+        #check if constraint has been redefined
+        if self.constraintIsRedefined(cons_tuple):
+            txt = "Warning: parameter " + \
+                  cons_tuple[0] + "." + cons_tuple[1].param +\
+                  " has been redefined."
+            self.lblWarning.setText(txt)
         self.constraintReadySignal.emit(cons_tuple)
         # reload the comboboxes
-        self.setupParamWidgets()
+        #self.setupParamWidgets()
```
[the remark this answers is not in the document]
m2cci-NMZ: Yes indeed, I forgot to uncomment!

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -435,6 +435,17 @@ def onConstraintChange(self, row, column):
             # Update the tabs for fitting list
             constraint = self.available_constraints[row]
             constraint.active = (item.checkState() == QtCore.Qt.Checked)
+            param = item.data(0)[item.data(0).index(':') + 1:item.data(0).index('=')].strip()
+            model = item.data(0)[:item.data(0).index(':')].strip()
+            tab = self.available_tabs[model]
```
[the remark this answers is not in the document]
m2cci-NMZ: A condition to be sure we we get a tab object befor calling the method was added

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -2623,14 +2636,17 @@ def setParamEditableByRow(self, row, editable=True):
     def isCheckable(self, row):
         return self._model_model.item(row, 0).isCheckable()
 
-    def selectCheckbox(self, row):
+    def selectCheckbox(self, row, select):
```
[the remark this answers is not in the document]
m2cci-NMZ: I needed to extend the beahaviour! Method is renamed to `changeCheckboxStatus` and `select` parameter was renamed to `checkbox_status`

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1055,6 +1055,19 @@ def getComplexConstraintsForModel(self):
                     for s in range(param_number) if self.rowHasActiveComplexConstraint(s)]
         return params
 
+    def getAllComplexConstraintsForModel(self):
+        """
```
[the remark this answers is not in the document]
m2cci-NMZ: Renamed the method to `getFullConstraintNameListForModel`, hope it's explicit enough!

src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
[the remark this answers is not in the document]
m2cci-NMZ: My bad, I got confused... From what I understand this was repopulating the comboboxes with parameters that did not have a constraint, but if we remove this limitation by allowing the user to redefine a constraint, will this still be necessary?

src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
```diff
@@ -258,15 +259,36 @@ def constraint(self):
...
         Respond to Add constraint action.
         Send a signal that the constraint is ready to be applied
         """
         cons_tuple = self.constraint()
+        #check if constraint has been redefined
+        if self.constraintIsRedefined(cons_tuple):
+            txt = "Warning: parameter " + \
+                  cons_tuple[0] + "." + cons_tuple[1].param +\
+                  " has been redefined."
+            self.lblWarning.setText(txt)
```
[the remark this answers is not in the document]
m2cci-NMZ: @RichardHeenan any thoughts? Is redefining a constraint a common task?

src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
```diff
@@ -101,7 +102,7 @@ def setupParamWidgets(self):
         """
         self.cbParam1.clear()
         tab_index1 = self.cbModel1.currentIndex()
-        items1 = [param for param in self.params[tab_index1] if not self.tabs[tab_index1].paramHasConstraint(param)]
+        items1 = [param for param in self.tabs[tab_index1].main_params_to_fit]
```
rozyczko, line 105: This will allow the same constraint to be used against two separate, varying parameters.
E.g.
`M1.scale = M2.scale` and `M1.scale = M3.scale` to be fitted at the same time.
Is this what we want? I just wonder what the optimization process would look like for this case.
I have explicitly removed this case but maybe there was some discussion which I missed that made it a requested feature?

src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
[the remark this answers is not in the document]
rozyczko: This might indeed not require a disruptive message box. It's difficult to accidentally redefine constraint so I guess a message is enough.

src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
rozyczko: Actually, 4.2 does even worse. if you set the constraints as above, the fitting completes, but the first constraint is silently overwritten and the fitted parameters have `M1.scale` equal to `M3.scale` only. `M2.scale` varies independently.

gonzalezma: I don't think that we want this, because as pointed by Piotr it would be a source of confusion. For me, it should be clear that the constraint definition is slave_parameter = master_parameter, so the slave parameter will follow the master one, which is the one varied on the fit. Probably this needs to be made clear in the documentation, as I guess for many users (probably including Lionel) this is not clear. For example, from a mathematical point of view, a user could expect that the 2 definitions above imply M1.scale = M2. scale = M3. scale (and that it does not matter much which of the 3 is chosen as master). If I remember right, there was a ticket already on this. I don't think that it is easy to handle this situation as soon as the expressions are more complex than M1.param = M2.param = ... = MN.param, so I'd suggest to keep things simple for now and only allow a unique definition for a slave parameter.
Nico, did you change this behaviour on purpose or it was a mistake?

butlerpd: Sorry folks I have been covered over and unable to keep up much but will try to look at all this soon as well as constraints are important for us also.  But to @gonzalezma point, yes there is a ticket about that (quite old I think).  If I understand the arguments (and need to look more carefully to be sure) I think you are making the same argument as @pkienzle made to me at the time. From the user perspective however (like foolish me) it looks obvious that in fact it shouldn't matter which is master.  I understand that it may be behind the scenes and certainly the last thing we need is for the cure to be worse than the disease :-)  I will try to find that ticket and link it here when I get a chance - it may be in sasmodels though.

m2cci-NMZ: Yes indeed, having this wouldn't make sense and would most likeley cause an error in the optimization engine, but from what I understand each time a constraint is added, it is stored in the child of the parameter at row 0, so adding `M1.scale = M2.scale` and then adding `M1.scale = M3.scale` should overwrite the first constraint, just like in 4.2 (hence the warning informing the user he is overwriting an existing constraint)?  I removed this limitation as it was mentioned in #1490

gonzalezma: If I remember right, Paul K made the point that the computer = is not the mathematical =. Here doing A=B followed by A=C will make A=C (but not A=B and A=C). As Piotr noted, doing this in 4.2 will just overwrite the constraint definition and ignore the first definition. The only way to go beyond this would be to do a "syntactic" analysis of the constraints defined to find that we have A=B=C, but I think this is far from trivial as the user starts to write something a little bit more complex. I'm afraid that this could be the cause of some obscure errors (so the cure being worse than the disease), so personally I would avoid it and I'd try to think how things can be made clear in the interface and the documentation. But I let the real software experts, such as Piotr, Paul K and Nico decide what can be done or not.

butlerpd: yep .. now that you say it that is what I remember as well (about computer vs mat = not being = 😄 ) so indeed if too complicated to address under the hood best not to, but anything in the interface that makes it harder for the poor user to do the wrong thing is certainly a welcome improvement.

smk78: Ohhhh... so if I want to tie three parameters together in a situation where we will call A 'master', you mean I'm supposed to do B=A and then C=B???

gonzalezma: This should work. But personally I find clearer to select one master parameter, let's say A and define the constraints as B=A and C=A.

m2cci-NMZ: What could be done to force the selection of one master parameter would be to exclude slave parameters from the combobox

rozyczko: Yes, I think this makes sense.In the above example by @smk78, you shouldn't be able to constrain C to B, because B has already been constrained to another parameter. You should constrain C directly to A, and B should not be in the list of parameters available for constraining (the left combobox).

PR #1550, opened 2020-05-27
Ess gui 1522 poly check

### LINKED ISSUE #1522: Incorrect behavior of "fittable" checkbox in polydispersity tab
Once polydispersity is turned on, one can visit the polydispersity tab and choose the function and all the other parameters including the value of the PD parameter itself.  One can also choose to keep it fixed to that value during the fitting or let the fitting fit the polydispersity as well by checking or unchecking the checkbox.

However, checking and unchecking the box seems to be recognized as a change that need recomputing.  if one checks the box and then computes the curve is computed with something like a PD of more than 1. If it is unchecked and compute hit again it computes as if PD has been set to zero even if that is not true.

Simple example to demonstrate:
open a new fit panel and choose sphere and sphere.  Change the fit options to only go to 0.005 min q and log space 150 points. Now back to Model tab and compute then show plot.  We have the typical sphere form factor with no polydispersity smearing. Now check polydispersity and go to that tab
make PD=0.3 and compute. the oscillations get significantly damped as they should. Now check the fittable box (which means nothing for a theory curve but I can so I try it of course :-) and the Show plot turns to compute. So hit compute.  the cuve now goes almost to a straight line which from what I can tell is applying a polydispersity of between 1 and 2 .... .even though the PD is STILL set at 0.3.  type 0.3 and return and nothing happens because no change is detected. Now uncheck the box and again compute option appears.  Hit it and the original form factor with all the oscillations appears (i.e. as if PD=0 ... even though it is still 0.3).  The only way to get 0.3 back is to type in 0.29 o 0.31 or something close but different.  Then compute appears and the correct curve will be computed.

For fun you can now add a data set (say polyspheres.txt) the behavior is essentially the same. But worse now, checking the box again and fitting does seem to compute the proper polydispersity (using DREAM you can see where the center is that it calculates) but that does not get passed back somehow and the curve is using either the very large distribution (when checked) or zero when unchecked regardless of the PD value.... until a small change is made to that value to reset it. Using LM, the Chi^2 seems to be correct though using DREAM, while the DREAM distribution plots show the correct PD, the Chi^2 seem to be huge ... at least sometimes?

Bottom line is that I would say that polydispersity seems to be completely unusable at the moment?

m2cci-NMZ:
fixes #1522 
Wrong behaviour was caused by a the PD value attributed to the checkbox state. Also added a test that will fail if the bug happens.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
@@ -137,7 +137,7 @@ class ComplexConstraint(QtWidgets.QDialog, Ui_ComplexConstraintUI):
         Tooltip for txtConstraint
         """
         p1 = self.tab_names[0] + ":" + self.cbParam1.currentText()
-        p2 = self.tab_names[1] if len(self.tab_names) != 1 else self.tab_names[0] +"."+self.cbParam2.currentText()
+        p2 = self.tab_names[1] +"."+self.cbParam2.currentText() if len(self.tab_names) != 1 else self.tab_names[0] +"."+self.cbParam2.currentText()
         tooltip = "E.g.\n%s = 2.0 * (%s)\n" %(p1, p2)
         tooltip += "%s = sqrt(%s) + 5"%(p1, p2)
         self.txtConstraint.setToolTip(tooltip)
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1405,10 +1405,8 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
                 if parameter_name_w in self.poly_params_to_fit:
                     self.poly_params_to_fit.remove(parameter_name_w)
             self.cmdFit.setEnabled(self.haveParamsToFit())
-            # force data update
-            key = parameter_name + '.' + delegate.columnDict()[delegate.poly_pd]
-            self.poly_params[key] = value
-            self.updateData()
+            # Update state stack
+            self.updateUndo()
 
         elif model_column in [delegate.poly_min, delegate.poly_max]:
             try:
--- a/src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py
+++ b/src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py
@@ -442,6 +442,8 @@ class FittingWidgetTest(unittest.TestCase):
         self.widget._poly_model.item(0,0).setCheckState(2)
         # Assure the parameter is added
         self.assertEqual(self.widget.poly_params_to_fit, ['radius_bell.width'])
+        # Check that it's value has not changed (reproduce the polydispersity checkbox bug)
+        self.assertEqual(self.widget.poly_params['radius_bell.width'], 0.0)
 
         # Add another parameter
         self.widget._poly_model.item(2,0).setCheckState(2)


src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
```diff
@@ -137,7 +137,7 @@ def setupTooltip(self):
         Tooltip for txtConstraint
         """
         p1 = self.tab_names[0] + ":" + self.cbParam1.currentText()
-        p2 = self.tab_names[1] if len(self.tab_names) != 1 else self.tab_names[0] +"."+self.cbParam2.currentText()
+        p2 = self.tab_names[1] +"."+self.cbParam2.currentText() if len(self.tab_names) != 1 else self.tab_names[0] +"."+self.cbParam2.currentText()
```
rozyczko, line 140: This is probably a part of another commit?

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1405,10 +1405,8 @@ def onPolyModelChange(self, top, bottom):
                 if parameter_name_w in self.poly_params_to_fit:
                     self.poly_params_to_fit.remove(parameter_name_w)
             self.cmdFit.setEnabled(self.haveParamsToFit())
-            # force data update
-            key = parameter_name + '.' + delegate.columnDict()[delegate.poly_pd]
-            self.poly_params[key] = value
-            self.updateData()
```
rozyczko, line 1411: The point of the above assignment was to redo the model/chart update on **selecting** a polydisperse parameter.
SasView 5 is trying to be "interactive", meaning chart/model updates happen on any parameter change.
So, if you enter a PD value and switch focus from the cell, the chart updates properly. 
However I can see that this does not correspond to what we do with main parameters. As such, this is probably a better solution in terms of functionality.

src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py
```diff
@@ -442,6 +442,8 @@ def testPolyModelChange(self):
         self.widget._poly_model.item(0,0).setCheckState(2)
         # Assure the parameter is added
         self.assertEqual(self.widget.poly_params_to_fit, ['radius_bell.width'])
+        # Check that it's value has not changed (reproduce the polydispersity checkbox bug)
+        self.assertEqual(self.widget.poly_params['radius_bell.width'], 0.0)
```
rozyczko, line 446: good test!

src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
m2cci-NMZ: Yes, this fixes a failing unit test, which was causing the interpreter to exit and not perform the rest of the unit tests

src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py
m2cci-NMZ: Thx!

src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
rozyczko: Interesting. On Windows it just marks the test fail without crash. This used to run, so must've been something I introduced (semi) recently... thanks for spotting this issue and fixing

butlerpd:
5.0 ready for testing on Win

butlerpd:
5.0 ready for testing on Win

wpotrzebowski:
5.0 ready for testing on OSX

Issue #1414, opened 2019-11-29
No pan function in plot windows in 5.0.1


smk78:
In an email thread @rozyczko says:

> The matplotlib toolbar is non-resizable in the new version and it was taking away way too much room from the charts. Also, one cannot subclass the toolbar in such a way that only certain icons/shortcuts are present - it is either the whole toolbar or nothing. So we chose no toolbar. Most of its functionality is however implemented in the right-click context menus, and zoom being triggered by the mouse wheel. Not so for panning 

I would argue the pan function is essential. Otherwise, unless your ROI just happens to be right in the middle of the graph, the only way to examine a ROI in detail is to right-click and use Set Graph Range, which is a lot more work.

My vote would be to use left-click-and-hold to enable panning, but then I'm a Windows user!

butlerpd:
I had also noted this lack of toolbar and zoomability but figured it would have to be a 5.0.2 or later issue.  I would also argue that only the scroll button is not really acceptable for zooming.  In fact the history behind the toobar in 4.x is  **largely** because of that very issue: for most people the scrolling for zoom is a) obscure and b) awkward even when you discover it.   Indeed, matplotlib does not really support all we want which is also why, in 4.x, we overwrote/augmented the matplotlib default toolbar.  I don't see why that code cannot be used here as well - it is not wx code but matplotlib code.  That said, 4.x I will admit that one of our 4.x plot problems is that they do eat up an inordinate amount of the plot space with area around the actual graph, such as the vast area left for labels and to some extent the toolbar.

So I would add to @smk78 comment that actually we need another zoom function.  His point about the ROI is exactly the issue.  For most of us (certainly me) when we zoom (which is relatively frequently) it is to see a very specific area, which means I want to have precise control of the box (top, bottom, left and right).  I find the scroll button mostly useless ... and then no pan makes it completely useless, but even with a pan is makes it hard work to get exactly to where I want to be.  Again I point out that we did not do all that work in 4.x GUI to get around the shortcomings of matplotlib just for laughs :-)  Of course how we provide zoom and pan seamlessly without a toolbar may require a bit more thought since @smk78  pan would be my zoom :-)

So while I realize this may not be a 5.0.1 given that a) it is not different from 5.0 and b) probably a lot more work, it seems to me that this is one of the critical issues for 5.x to be really usable and for many of us to feel comfortable migrating to 5.x for real work?

rozyczko:
[...]

PR #1552, opened 2020-05-28
Enable mpl toolbar

rozyczko:
Addresses #1414 by adding a way to enable/disable the default matplotlib toolbar on 1D and 2D plots.
This is available in the right click context menu of a plot as `Toggle Navigation Menu` item.


butlerpd:
5.0 ready for testing on Win

PR #1553, opened 2020-05-28
Ess gui 1547 smearing

gonzalezma:
Corrected bad behaviour for dQ data smearing and made interface and behaviour for the different cases equivalent to that of SasView 4.2.2

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
@@ -49,6 +49,7 @@ DEFAULT_PINHOLE_DOWN=0.0
 
 class SmearingWidget(QtWidgets.QWidget, Ui_SmearingWidgetUI):
     smearingChangedSignal = QtCore.pyqtSignal()
+
     def __init__(self, parent=None):
         super(SmearingWidget, self).__init__()
 
@@ -71,6 +72,9 @@ class SmearingWidget(QtWidgets.QWidget, Ui_SmearingWidgetUI):
         self.pinhole = 0.0
         self.slit_height = 0.0
         self.slit_width = 0.0
+
+        # current accuracy option
+        self.accuracy = ""
 
         # Let only floats in the line edits
         self.txtSmearDown.setValidator(GuiUtils.DoubleValidator())
@@ -198,10 +202,15 @@ class SmearingWidget(QtWidgets.QWidget, Ui_SmearingWidgetUI):
         # don't save the state if dQ Data
         if smearing == "Custom Pinhole Smear":
             self.pinhole = d_up
-            self.accuracy = accuracy
         elif smearing == 'Custom Slit Smear':
-            self.slit_height = d_down
-            self.slit_width = d_up
+            self.slit_height = d_up
+            self.slit_width = d_down
+        # check changes in accuracy
+        if self.accuracy != accuracy:
+            self.accuracy = accuracy
+            if accuracy == 'High' or accuracy == 'Extra high':
+                QtWidgets.QMessageBox.information(self, "Accuracy Warning",
+                  "Higher accuracy is very expensive, \nso fitting can be very slow!")      
 
         self.onIndexChange(index)
 
@@ -248,8 +257,8 @@ class SmearingWidget(QtWidgets.QWidget, Ui_SmearingWidgetUI):
         self.lblSmearDown.setText('Slit width')
         self.lblUnitUp.setText('<html><head/><body><p>Å<span style=" vertical-align:super;">-1</span></p></body></html>')
         self.lblUnitDown.setText('<html><head/><body><p>Å<span style=" vertical-align:super;">-1</span></p></body></html>')
-        self.txtSmearDown.setText(str(self.slit_height))
-        self.txtSmearUp.setText(str(self.slit_width))
+        self.txtSmearUp.setText(str(self.slit_height))
+        self.txtSmearDown.setText(str(self.slit_width))
         self.txtSmearDown.setEnabled(True)
         self.txtSmearUp.setEnabled(True)
 
--- a/src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
@@ -176,10 +176,7 @@ class SmearingWidget(QtWidgets.QWidget, Ui_SmearingWidgetUI):
         elif text == "Use dQ Data":
             self.setElementsVisibility(True)
             self.setDQLabels()
-            if self.smear_type == "Pinhole":
-                self.onPinholeSmear()
-            else:
-                self.onSlitSmear()
+            self.onDQSmear()
         elif text == "Custom Pinhole Smear":
             self.setElementsVisibility(True)
             self.setPinholeLabels()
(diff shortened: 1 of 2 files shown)


src/sas/qtgui/Perspectives/Fitting/UI/SmearingWidgetUI.ui
```diff
@@ -101,6 +101,9 @@
           <property name="title">
            <string>Accuracy</string>
           </property>
+          <property name="toolTip">
+             <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Higher accuracy is very expensive. Use it with care!&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+          </property>
```
rozyczko, line 106: Maybe we can have a custom label (there's plenty of room on this control!) which will display the above message explicitly rather than in a tooltip, whenever users choose higher accuracy?
This is enhancement rather than issue with the current code, so feel free to make it into a separate GH issue.

src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
```diff
@@ -338,10 +348,8 @@ def onPinholeSmear(self):
             len_data = len(data.data)
             data.dqx_data = np.zeros(len_data)
             data.dqy_data = np.zeros(len_data)
-            data.dqx_data[data.dqx_data == 0] = percent * data.qx_data
-            data.dqy_data[data.dqy_data == 0] = percent * data.qy_data
             q = np.sqrt(data.qx_data**2 + data.qy_data**2)
-            data.dx_data = data.dqy_data = percent*q
+            data.dqx_data = data.dqy_data = percent*q
```
rozyczko, line 352: That's an overdue fix - thanks for applying it here.

src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
```diff
@@ -355,7 +363,9 @@ def onPinholeSmear(self):
             backend_accuracy = ACCURACY_DICT.get(accuracy)
             if backend_accuracy:
                 self.current_smearer.set_accuracy(accuracy=backend_accuracy)
-
+            else:
+                self.current_smearer.set_accuracy(accuracy='low')
+
```
rozyczko, line 368: Why is this necessary? ACCURACY_DICT contains all possible keys, and in case we don't find the right key, I think it is better to just stay with the previously defined accuracy level, rather than reset it quietly?

src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
```diff
@@ -321,6 +314,23 @@ def setState(self, smearing, accuracy, d_down, d_up):
...
+        # data.dqx_data & data.dqy_data (if 2D),
+        # so only need to set accuracy for 2D
+        _, accuracy, _, _ = self.state()
+        self.current_smearer = smear_selection(self.data, self.kernel_model)
+        if isinstance(self.data, Data2D):
+            backend_accuracy = ACCURACY_DICT.get(accuracy)
+            if backend_accuracy:
+                self.current_smearer.set_accuracy(accuracy=backend_accuracy)
+            else:
+                self.current_smearer.set_accuracy(accuracy='low')
```
rozyczko, line 332: Why is this necessary? ACCURACY_DICT contains all possible keys, and in case we don't find the right key, I think it is better to just stay with the previously defined accuracy level, rather than reset it quietly?
(Also, below similar suggestion)

src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
gonzalezma: I just reused the same code already present in def onPinholeSmear. But note that when a data set is loaded or a model selected, getSmearInfo is called without the smearing interface being "activated", so at this point backend_accuracy is None and the if ... else is needed, I think.

src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
gonzalezma: See above.

src/sas/qtgui/Perspectives/Fitting/UI/SmearingWidgetUI.ui
gonzalezma: Why not? I think that a user will see the tooltip whenever changing the accuracy, but I don't know if he/she will read it or just ignore it :-) So I'll let Paul B decide what's best.

butlerpd: Sorry am looking at your code changes vs what I had to do in 4.x (which was apparently only a year and a bit ago -- seems like forever.

Not sure I should be the decider.  Was going to do some functionality testing after I compare the codes and see but my first guess would be that @rozyczko suggestion of making it a more permanent warning when high is chosen would be good.  I'd still leave the tool tip on that case though since there are 4 levels as I recall that get progressively more costly and @rozyczko suggestion would only be for the worst case?

What does Lionel think? he is testing right? 😄  speaking of which maybe we can make an installable from this PR.. I cannot remember the 5.x magic runes need to look them up.

butlerpd:
5.0 ready for testing on Win

butlerpd:
OK so I see where the code was completely refactored for this GUI so cannot easily translate what was done for 4.x.  but suspect the divide by zero fix in 4.x is not here?  I have copied the major things that might be relevant to 5.0 into the issue.  Will now try to test funtionality.

gonzalezma:
Yes, I've seen the strange 'swapping'. It's curious that when you have a data set loaded, this does not seem to happen. I'll give it a look and try to find what's going on. 
As for the dramatic different behaviour of height/width smearing, I had already noticed it. But I looked into the equations given in the help and saw that u and v are not equivalent, so I assumed that this was OK. The actual calculations are done in sasmodels, so if needed perhaps is best to open another issue to check the slit smearing.

butlerpd:
u vs v is probably another issue indeed. interesting the math is different. I'll need to look at that and understand.  Anyway it is lucky that they are different or I would never have caught on that they are swapping.

Am now trying to do check with fitting and data and have run into a different rather nasty but it seems. As soon as I figure that out I'll finish testing that part

Suspect the swapping is part of the same issue we had with the pd? ... something about how data in the tabs gets carried over to the main panel when you go there? maybe it is a more fundamental design issue?

butlerpd:
I have checked the commutativity of u and v in 4.x and it behaves similarly. It seems odd to me but that is another ticket (it may in fact be correct so I need to check that first). Otherwise everything else checks out including 1D data that I put a few 0s in dq to see what happened.  The loader loads it but the fitpage decides that there is no dq data so does not even give that option.  2D seems correct as well. So if we can stop the switching (which dramatically affects the results at present) then we should merge this.

gonzalezma:
The swap of the values should be solved now. It was due to an inconsistent assignment between height/width and up/down variables, so the inputs got swapped. I still don't understand why this will not happen when a data set is present, but I checked that this still works in that case. I added an explicit warning message that will pop up when a user selects high or extra-high accuracy. But I let Piotr decide if the way I did this is OK or if it should be in a different way.

butlerpd:
cool .. will try to test this.  I guess that maybe if it is in the data it is read correctly so it does not depend on the GUI to assign the value from the entry box to the correct value since it is at that point just displaying?

butlerpd:
5.0 ready for testing on Win

wpotrzebowski:
5.0 ready for testing on OSX

Issue #1529, opened 2020-04-27
ESS GUI Corfunc: Input area is not scrollable in Corfunc Perspective in 5.0.2


smk78:
This issue is carried over from https://github.com/SasView/sasview/issues/1339.

The corfunc window in 5.0.2 is not currently scrollable. One consequence of this is that the Matplotlib toolbar for the correlation function graph window is unaccessible on some platforms depending on the screen size/resolution, or requires minimisation of the Log Explorer window.

Issue #1543, opened 2020-05-17
SasView application window needs to be scrollable 


butlerpd:
I suspect it was a conscious decision to remove the scrollable nature of the application window as well as most all windows within the application that were implemented in 4.x.  However, despite a lot of work to reduce the infinitely long panels (by adding tabs etc) and some attempts, I think, at clever estimation of real estate, the fact remains, that the number and variety of hardware, and more importantly screen form factors, that people will use is too large to guess so that two problem occur:
1) a new perspective will likely get released unusable by many until enough different users, with different hardware, have tried it (see issue #1332).  One can of course try to have a list of rules for handling the sizing of the perspective under various scenarios but would require quite a lot more effort and discipline that is likely given current resources on the project?
2) In some cases it will always remain unusable. For example running 5.0.2 on my laptop through a browser (I am using Paperspace but our summer schools use amazon's version of virtual desktops and I expect this approach to only grow with time). 4.x can be a bit painful with the scrolling but perfectly functional for quick checks/fits/modeling whereas 5.0.2 is useless: Aside from corfunc not being accessible, even removing the footer bar (never mind the console log) I cannot see anything below the title to the plot box nor can I see any of the buttons on the bottom of the fit page (so cannot show plot, fit or get help). Even inversion I can only reach the calculate button by removing the footer bar. Note that using the same virtual box on my 27" monitor things were perfectly... except of course corfunc.

NOTE that this was not a problem in mac through 4.x because all panels were free and independent. However, now that the mac is also enforcing the entire SasView workflow to remains inside the main SasView window I believe this will also be a problem for the mac

butlerpd:
Have had another, generally very happy user note the problem of not being able to reach the critical buttons (like fit 😄 ) when using their laptop.  As this renders the program literally unusable, and in view of what else we are labeling critical, am raising this to critical.

rozyczko:
What needs to be done (most likely) is adding QScrollArea element between the main canvas and the QDockWidget. This should allow scrolling of the whole main area.

butlerpd:
Will try a quick fix to prevent accidental zooming of graphs

PR #1562, opened 2020-05-29
Added vertical and horizontal scrollbars to the main canvas. #1543

rozyczko:
it's scrollable now!

### DIFF
--- a/src/sas/qtgui/MainWindow/MainWindow.py
+++ b/src/sas/qtgui/MainWindow/MainWindow.py
@@ -22,8 +22,8 @@ class MainSasViewWindow(QMainWindow, Ui_SasView):
         self.workspace = QMdiArea(self)
         # some perspectives are fixed size.
         # the two scrollbars will help managing the workspace.
-        self.workspace.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOn)
-        self.workspace.setHorizontalScrollBarPolicy(Qt.ScrollBarAlwaysOn)
+        self.workspace.setVerticalScrollBarPolicy(Qt.ScrollBarAsNeeded)
+        self.workspace.setHorizontalScrollBarPolicy(Qt.ScrollBarAsNeeded)
         self.screen_width =  screen_resolution.width()
         self.screen_height = screen_resolution.height()
         self.setCentralWidget(self.workspace)
--- a/src/sas/qtgui/MainWindow/MainWindow.py
+++ b/src/sas/qtgui/MainWindow/MainWindow.py
@@ -22,8 +22,8 @@ class MainSasViewWindow(QMainWindow, Ui_SasView):
         self.workspace = QMdiArea(self)
         # some perspectives are fixed size.
         # the two scrollbars will help managing the workspace.
-        self.workspace.setVerticalScrollBarPolicy(Qt.ScrollBarAsNeeded)
-        self.workspace.setHorizontalScrollBarPolicy(Qt.ScrollBarAsNeeded)
+        self.workspace.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOn)
+        self.workspace.setHorizontalScrollBarPolicy(Qt.ScrollBarAlwaysOn)
         self.screen_width =  screen_resolution.width()
         self.screen_height = screen_resolution.height()
         self.setCentralWidget(self.workspace)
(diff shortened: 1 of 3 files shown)


butlerpd:
From the screenshot this looks like it would fix the corfunc issue as well. Can somebody test that? and then link that ticket?  ... though the auto closing does not seem to be working so well

krzywon:
@butlerpd, I'll gladly test, but what corfunc issue are you referring to?

butlerpd:
@krzywon - issue #1529 on Kanban board

butlerpd:
5.0 ready for testing on Win

krzywon:
All perspectives can be scrolled vertically and horizontally. Here is an image of the Corfunc perspective, with scroll bars, so this should also fix #1529.

sasview_5x_corfunc_with_scrollbars

butlerpd:
cool. @smk78, can we now close this ticket as well?

smk78:
Yes! Close https://github.com/SasView/sasview/issues/1529 too!

pkienzle:
The main window scrolling works on mac.

However, there is conflict between scrolling the main canvas and scrolling within a window on the canvas.  With the trackpad motion, some mouse wheel events are given to the main canvas and some are given to the window under the mouse pointer. With plot windows the result is a graph that is so badly zoomed that you can't tell where the data is. It would be best to ignore wheel events on the window unless that window is selected. Maybe need a way to unselect a window to enable scrolling the canvas when the window is large.

Some weird interactions between scrolling and maximizing. May want to eat all events generated between maximize start and maximize complete.  The result is only glitchy, not unusable.  There is occasional glitchiness between maximize, scrolling and arrow keys as well.

rozyczko:
As discussed during the call, I modified the scrollbars to be always visible in a bid to help with functionality.

pkienzle:
The request to show/hide scroll bars is ignored on the mac.  They are visible if you have "Show scroll bars: Always" turned on in the General Systems Preferences, but only appear when scrolling otherwise.  I don't think this is a battle worth fighting.

If it is easy to suspend scroll wheel zoom in the plots, and instead feed the events to the underlying canvas, that might still be useful.

krzywon:
Would adding the following to PlotterBase.py help with @pkienzle's suggestion?
`def wheelEvent(self, event): event.ignore()`

Where I found the idea:
https://doc.qt.io/qt-5/qwidget.html#wheelEvent

In regards to the scroll bars, always showing them detracts more than it adds, in my opinion.

RichardHeenan:
Tried this branch, local build on Win10, great step forward in use of screen real estate.

(If I drag vertical scroll bar in a FitPage, then mouse wheel will move that scroll until it gets to end of travel, then the main canvas moves, but that is fine once you realise what is going on.)

butlerpd:
Sounds like we can merge this PR - but waiting for response from Piotr on @krzywon suggestion. Would it be better? would it be a trivial change here? if yes to both is it something that should be in a new ticket rather than tackle here?

Unless it really is trivial and obviously a much better solution I would vote we get this out as is for 5.0.3 and put the suggested change as a future issue to consider.

rozyczko:
> Would adding the following to PlotterBase.py help with @pkienzle's suggestion?
> `def wheelEvent(self, event): event.ignore()`

We probably don't want to remove wheel events from plots. These events are used for zooming control. Admittedly, we can now zoom with actual MPL zoom tool in the MPL toolbar, so removing this functionality might have some merit.

pkienzle:
The other question is whether the permanent scrollbar are a net benefit.  I side with @krzywon on this one, and have them turned off until needed on Mac. I suggest reverting the last commit (or not) and merging.

RichardHeenan:
On Windows we need the scroll bars as corfunc is unusable on average size screen, and many more perspectives are unusable on smaller screens and via other interfaces.  We all also gain much more real estate to park all those polydispersity and residuals plots that keep popping up even if you have minimised them.

rozyczko:
@RichardHeenan I don't think the issue is with the scrollbars themselves. They stay. What PK and JK argue is whether we need the scrollbars always present on the canvas as opposed to present only when there is hidden content in widgets displayed on the canvas.

I tend to agree with them and show them (automatically) only when needed. That is, only when a widget (perspective, chart, plot etc) requires it.

PR #1565, opened 2020-05-29
(ESS_GUI) Report dialog enhancements

### LINKED ISSUE #1564: Enhancements to Report Results
Report Results should also contain:

- the SasView version number
- the optimiser used
- and any distribution functions used (at present the _value_ of any dispersity is returned, but not which function it relates to).

krzywon:
When generating a fit report, the SasView version, sasmodels version, bumps optimizer, and functions used for polydispersity are now included in the dialog.

fixes #1564

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/ReportPageLogic.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ReportPageLogic.py
@@ -10,8 +10,12 @@ from io import BytesIO
 import urllib.parse
 
 from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
+from bumps import options
 
 import sas.qtgui.Utilities.GuiUtils as GuiUtils
+from sas.sasview import __version__ as SASVIEW_VERSION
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Perspectives/Fitting/ReportPageLogic.py
```diff
@@ -10,8 +10,12 @@
 import urllib.parse
 
 from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
+from bumps import options
 
 import sas.qtgui.Utilities.GuiUtils as GuiUtils
+from sas.sasview import __version__ as SASVIEW_VERSION
+from sasmodels import __version__ as SASMODELS_VERSION
```
rozyczko, line 17: I would probably pedantically change the import order and put `sasmodels` after `bumps`, since `sasmodels` are higher in hierarchy than `sas`, on level with `bumps` but this is just a general comment.

smk78:
Example output.
[fit_report.pdf](https://github.com/SasView/sasview/files/4704249/fit_report.pdf)

Issue #1554, opened 2020-05-28
ESS_GUI: annulus slicer not opening


murphyryanp:
I tried opening the annulus slicer tool from ESS_GUI, but it fails to open and gives the error in Log Explorer: 

> File "C:\Users\murph\sasview\src\sas\qtgui\Plotting\Slicers\AnnulusSlicer.py", line 32, in __init__
    self.xmaxd = self.xmax
> AttributeError: 'AnnulusInteractor' object has no attribute 'xmax'

After looking through the branch history, this error (as well as batch assignment and slicer parameters) was already addressed by Piotr in the branch called "ESS_GUI_1381_slicer" regarding #1381, but this branch was never merged into ESS_GUI.

Should this lost branch be tested and merged for the next release?

butlerpd:
Nice catch @murphyryanp.  I suspect that @rozyczko had bigger plans for this branch? @rozyczko what do you think? should we merge this? or should we make a new branch just for this issue? would be good to fix this bug in 5.0.3.

rozyczko:
We *must* merge this, preferably creating a new branch and cherrypicking relevant changes from `ESS_GUI_1381_slicer`.
This branch was supposed to go into 5.0.2 but at some point we decided to postpone the batch slicer functionality and I missed the fact that the branch contained a critical fix. 
Thank you @murphyryanp for finding it!
This could actually be a good first issue for you, if you don't mind looking at it closer?

murphyryanp:
@rozyczko  Sounds good. I'll create a branch from ESS_GUI and add your relevant changes from the ESS_GUI_1381_slicer branch

PR #1567, opened 2020-05-29
Ess gui 1554 slicer

murphyryanp:
This is mostly a "merge rescue operation" from @rozyczko branch 'ESS_GUI_1381_slicer'. I pulled in all non-conflicting changes from the 'ESS_GUI_1381_slicer' branch into the 'ESS_GUI' branch, and then removed some features that I think require further development. Key changes regarding issue #1554 are summarized below:

- Fixes the critical issue regarding the annulus slicer not opening. Also, the parameters can be set and are updated in the SlicerParameters window.
- Removed the batch slicing tab in the SlicerParameter widget and the associated widgets (which are commented out) since it seems this needs more tweaking. However, the batch slicing is a really nice feature (that I would like to use!) but the auto-saving and fitting selection features are currently inactive. Should we save the batch slicing feature for a later release?
- Minimum limits to the annulus slicer changed from 0.01 to dqmin, where dqmin is defined by the input 2D data. This is for VSANS or other low-q 2D data that may require annulus slices with widths less than 0.01 A^-1.
- Maximum limits to the annulus slicer changed to the maximum of |qx| (instead of the minimum), which is mainly an issue for offset 2D data. If the 2D data has an offset, then the user won't be limited to setting the annulus to the edge closest to the offset center. Instead, the limit becomes the edge farthest from the offset center. (We may also want to consider adding an evaluation of ymax to handle the possible cases where data is offset in the qy direction instead of the qx direction.)


murphyryanp:
So should I merge this Pull Request as is, but also create a new issue for Paul's suggestion to integrate context manager for all the slicers?

pkienzle:
I turned my comment into a separate issue.  Still needs verification that it functions as expected on windows before merging, unless @rozyczko has already done so?

butlerpd:
5.0 ready for testing on Win

src/sas/qtgui/Plotting/SlicerParameters.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, parent=None,
         self.active_plots = active_plots
 
         # Initially, Apply is disabled
-        self.cmdApply.setEnabled(False)
+        #self.cmdApply.setEnabled(False)
```
murphyryanp, line 44: @butlerpd Here's one example of the command commented out related to the batch slicer tab. The batch slicing tab was removed in the .ui file.

butlerpd:
LOL  missed that.  but this should be merged I think.

Eventually, depending on whether @rozyczko thinks the batch slicing should be restarted from master or continued from the previous branch he may want to merge ESS_GUI master back into his batch slicer working branch I guess.

rozyczko:
I'll most likely reapply changes from my old branch to ESS_GUI and finish off the slicer. Soon.

PR #1568, opened 2020-05-29
Set SAS_DLL_PATH on sasview startup

### LINKED ISSUE #1560: set the sasmodels dll cache directory in sasview startup
As noted in #1534, sasmodels is  files into the ~/.sasmodels directory (which is fine --- sasmodels may be used for applications outside of sasview), but this is confusing for sasview users.  Sasview should set the SAS_DLL_PATH environment to ~/.sasview/compiled_models on startup so that users have a single directory tree associated with the sasview application.

pkienzle:
Fixes #1560: SAS_DLL_PATH set on sasview startup if not already set in the environment

### DIFF
--- a/src/sas/qtgui/MainWindow/MainWindow.py
+++ b/src/sas/qtgui/MainWindow/MainWindow.py
@@ -1,13 +1,15 @@
 # UNLESS EXEPTIONALLY REQUIRED TRY TO AVOID IMPORTING ANY MODULES HERE
 # ESPECIALLY ANYTHING IN SAS, SASMODELS NAMESPACE
+import os
+import sys
+
 from PyQt5.QtWidgets import QMainWindow
 from PyQt5.QtWidgets import QMdiArea
 from PyQt5.QtWidgets import QSplashScreen
 from PyQt5.QtWidgets import QApplication
 from PyQt5.QtGui import QPixmap
 from PyQt5.QtCore import Qt
-import os
-import sys
+
 # Local UI
 from sas.qtgui.UI import main_resources_rc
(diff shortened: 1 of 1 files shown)


wpotrzebowski:
5.0 ready for testing on OSX

wpotrzebowski:
5.0 ready for testing on Win

src/sas/qtgui/MainWindow/MainWindow.py
```diff
@@ -61,8 +63,11 @@ def run_sasview():
     from sas.logger_config import SetupLogger
     SetupLogger(__name__).config_development()
 
-    # initialize OpenCL setting
-    from sas import get_custom_config
+    # initialize sasmodels settings
+    from sas import get_custom_config, get_user_dir
+    if "SAS_DLL_PATH" not in os.environ:
```
rozyczko, line 68: Would it make sense to add `SAS_DLL_PATH` to the custom config file, like `SAS_OPENCL` does below?

pkienzle: It's not really an option that the user needs to change, and there is no support for changing it in the gui, so no need to have it in the user config file.

pkienzle:
> The code looks good but I am wondering where "compiled_models" folder should created. I cannot see this in sasview repo but I may be missing something

It's in sasmodels [kerneldll.py](https://github.com/SasView/sasmodels/blob/8dde7330cf3f28c29efc411f9bfd55c86c2a92c4/sasmodels/kerneldll.py#L271-L272).

Issue #1535, opened 2020-05-11
ESS_GUI: Existing common parameters not preserved between models in 5.x


smk78:
Reported by user William Sharrat. Verified by @smk78 .

In 4.x, if you have a set of parameters in a FitPage and then change model (eg, sphere > vesicle), the values of those parameters in the new model that are the same as those in the old model (eg, scale, background & radius in the example above) are _preserved_. This can be very useful.

But in 5.x, when you change model, all parameters are reset to the model default. The 4.x behaviour should be reimplemented.

rozyczko:
This was deliberate on my part - keeping the same parameter values between models can be confusing, since there is not `Reset parameters` functionality. If I start playing with parameters and decide to go back to the defaults, I just switch to another parameter and back. Also, are we certain names of the parameters should be equated between models? What about custom plugins?

Probably the best situation would be to keep all `changed` parameters for models used in state, so when reverted, they are set back, e.g. when I set `scale` to `2.0` in `sphere` and change model to `cylinder`, the value is the default 1.0, but when switched back to `sphere` the GUI remembers `sphere`'s 2.0.

smk78:
But we are talking here about switching model in the _same_ FitPage. Basically it's a handy way of scoping the form of similar models without having to create multiple FitPages for different models and re-type the parameters (because copy/paste won't work with different models). If one wants to get the model back with default parameters in 4.x then all you need do is open a new FitPage and re-select the model, as you say.

Keeping the changed parameters in state doesn't help in this instance because you want parameters of the same name updated in the new model.

You are right that the limitation of this is the parameter names being the same between models. But, I suggest, there is actually quite a lot of commonality for this to be useful, particularly within model categories, without the need for a master lookup table.

butlerpd:
I agree with @smk78.  This is an extremely useful functionality.  When working on a real project you don't know what you are up against the way you do in a tutorial (the classic cooking show : I just happened to have already prepared xx so we don't have to wait 😄 

So you will try a sphere then maybe a cylinder then a core shell cylinder then an ellipsoid etc. Each time playing with parameters.  A few things to note:
1. for a lot of materials the parameters with the same name are going to be of the same order, partly because the overall size of the object is clear from the data anyway but also things like self assembled systems generally have a fixed smallest dimension so the radius of the sphere would not be far from the radius of a rod made of the same stuff. Further background and volume fraction are most likely not changing. Having to reset these parameters every time can get very annoying.
[...]

PR #1570, opened 2020-05-30
Added state awareness for changed parameters. #1535

rozyczko:
Both for same category and different categories.


src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py
```diff
@@ -1525,6 +1525,39 @@ def testReplaceConstraintName(self):
...
+        """
+        # select model: cylinder / cylinder
+        category_index = self.widget.cbCategory.findText("Cylinder")
+        self.widget.cbCategory.setCurrentIndex(category_index)
+
+        model_index = self.widget.cbModel.findText("cylinder")
+        self.widget.cbModel.setCurrentIndex(model_index)
+
+        # modify the initial value of radius (different from default)
+        new_value = "333.0"
```
krzywon, line 1540: You define the value here but don't use it in later tests.

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -3740,6 +3748,24 @@ def loadAnalysisFile(self):
         filename = QtWidgets.QFileDialog.getOpenFileName(**kwargs)[0]
         return filename
 
+    def getParameterDict(self):
+        """
+        Gather current fitting parameters as dict
+        """
+        param_list = self.getFitParameters()
+        param_list = self.getFitPage()
```
krzywon, line 3756: Anything returned from self.getFitParameters() is immediately overwritten.

gonzalezma:
Nice functionality. The main limitation is that "memory" is restricted to the current model, so some parameter values can be lost if several models are tried.
Example: Starting with a set of values for the sphere, select polymer_micelle (scale, background, and sld_solvent are passed; sld and radius not, as they do not exist in the polymer_micelle model, so this is OK). If then one comes back to the sphere the initial values for sld and radius are lost and the default values are presented instead.
Note also that the passing of the parameters does not work with models allowing for a variable interface (or number of parameters), such as onion or core_multi_shell. In this case, nothing is passed, not even scale and background and then if one comes back to the sphere model, all the parameters are reset to the default values. 
I don't know how easy/feasible is to maintain a list with the last user defined value for each parameter name and use this instead of just the list of the last model. If it requires much refactoring, this PR could be merged as already an improvement over the present behaviour, and create a new enhancement issue to address this limitation.

rozyczko:
> Nice functionality. The main limitation is that "memory" is restricted to the current model, so some parameter values can be lost if several models are tried.
> Example: Starting with a set of values for the sphere, select polymer_micelle (scale, background, and sld_solvent are passed; sld and radius not, as they do not exist in the polymer_micelle model, so this is OK). If then one comes back to the sphere the initial values for sld and radius are lost and the default values are presented instead.
> Note also that the passing of the parameters does not work with models allowing for a variable interface (or number of parameters), such as onion or core_multi_shell. In this case, nothing is passed, not even scale and background and then if one comes back to the sphere model, all the parameters are reset to the default values.
> I don't know how easy/feasible is to maintain a list with the last user defined value for each parameter name and use this instead of just the list of the last model. If it requires much refactoring, this PR could be merged as already an improvement over the present behaviour, and create a new enhancement issue to address this limitation.

I am working on most of those issues now, will commit to this branch

rozyczko:
To make this work properly (changes preserved on category swap, retain data on polydispersity and various checkboxes) more involving changes need to be made. Let's postpone this to 5.0.4

Issue #1558, opened 2020-05-28
wx4: constraints dialog not rendering properly


pkienzle:
The constraints dialog is not rendering properly in wx 4 on the mac.

Should probably be fixed if releasing sasview 4.3 with wx4.

butlerpd:
@pkienzle - I believe this was fixed by PR #1573 which was merged in September 1 2020.  Can we close this issue?

lucas-wilkins:
Not using wx anymore

PR #1573, opened 2020-06-01
wx4: fix handling of constraints tab

pkienzle:
See #1558.

### DIFF
--- a/src/sas/logger_config.py
+++ b/src/sas/logger_config.py
@@ -26,6 +26,7 @@ class SetupLogger(object):
             self._read_config_file()
             logging.captureWarnings(True)
             logger = logging.getLogger(self.name)
+        logging.getLogger('matplotlib').setLevel(logging.WARN)
         return logger
 
     def config_development(self):
@@ -35,6 +36,7 @@ class SetupLogger(object):
         logger = logging.getLogger(self.name)
         self._update_all_logs_to_debug(logger)
         logging.captureWarnings(True)
+        logging.getLogger('matplotlib').setLevel(logging.WARN)
         return logger
 
     def _read_config_file(self):
--- a/src/sas/sasgui/perspectives/fitting/simfitpage.py
+++ b/src/sas/sasgui/perspectives/fitting/simfitpage.py
@@ -777,8 +777,7 @@ class SimultaneousFitPage(ScrolledPanel, PanelBase):
         box_sizer.Add((20, -1))
         box_sizer.Add(self.set_button, flag=wx.RIGHT | wx.EXPAND, border=5)
         sizer_constraint.Add(box_sizer, flag=wx.RIGHT | wx.EXPAND, border=5)
-        self.sizer_all_constraints.Insert(before=0,
-                             item=sizer_constraint,
+        self.sizer_all_constraints.Insert(0, sizer_constraint,
                              flag=wx.TOP | wx.BOTTOM | wx.EXPAND, border=5)
         self.FitInside()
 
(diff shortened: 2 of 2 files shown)


pkienzle:
Travis is failing because python 2.7 is not longer supported for sasmodels master.  Possible solutions are to drop python 2.7 support for sasview, pin the sasmodels for sasview to an older version, or use backported pathlib in sasview travis.

butlerpd:
😃  you got their first... Was going to say that since the build system has a patch and Steve now has pathlib in his working environment, it makes sense that it should just be added to travis.

pkienzle:
Thing is that pathlib is out of date and it needs pathlib2, but pathlib2 installs under the wrong name (pathlib2), so really, need to install pathlib2 then rename the directory in site packages to pathlib, or just grab the pathlib2/__init__.py from the pathlib2 repo (which specific release hash) and drop it as pathlib.py in the site packages directory.  Hence, travis is still failing.

pkienzle:
It would be easy enough to modify sasview so that it doesn't depend on pathlib 3.5. However, if you are going to go through the exercise of verifying that sasview 4.3 works with wx 4 (which means testing every piece of functionality) you can do so with python 3.6, thus dealing with the python 2.7 end-of-life issue at the same time. Therefore, I will push a change to travis ci as part of this ticket which removes the python 2.7 tests.

If you want to continue supporting 2.7, push the following patch to sasmodels and update jenkins to include the pathlib2 package in the build environment:
```
$ git diff
diff --git a/sasmodels/generate.py b/sasmodels/generate.py
index a0e62b2..4033b88 100644
--- a/sasmodels/generate.py
+++ b/sasmodels/generate.py
@@ -182,7 +182,11 @@ except ImportError:
 # pylint: enable=unused-import
 
 # BRP added to avoid open files:
-from pathlib import Path
+# CRUFT: python < 3.5 needs pathlib2
+try:
+    from pathlib2 import Path
+except ImportError:
+    from pathlib import Path
 
diff --git a/setup.py b/setup.py
index 019151b..a60ac98 100644
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,11 @@ import sys
 from setuptools import setup
 from setuptools.command.test import test as TestCommand
 # BRP added to avoid open files:
-from pathlib import Path
+# CRUFT: python < 3.5 needs pathlib2
+try:
+    from pathlib2 import Path
+except ImportError:
+    from pathlib import Path
 
 class PyTest(TestCommand):
     user_options = [('pytest-args=', 'a', "Arguments to pass to pytest")]
```

pkienzle:
Maybe add an installation note that 2.7 can still be used if the following is saved as `pathlib.py` in site-packages:

https://github.com/mcmtroffaes/pathlib2/blob/e0e9fc32530cfa5fe6d8da38fee8efb8959b07c2/pathlib2/__init__.py

butlerpd:
Given that 4.3 is supposed to be the EOL version for this branch I would be very loath to add cruft to sasmodels just to support a 2.7 python. I think that the above installation note would however be very useful. At this point, the hope that I/w can test the wx4 and python 3.6 mode  thoroughly enough is at more aspirational than even optimistic I fear.

pkienzle:
When https://github.com/SasView/sasmodels/pull/404 is merged we can reinstate python 2.7 tests.

butlerpd:
ready for testing on Win64

butlerpd:
So the build "failed" but looking at the Jenkins log it seems that the code compiled properly and even created an exe 
```
Successful compile (140,719 sec). Resulting Setup program filename is:
C:\jenkins\workspace\SasView_Win7_64bit_PullRequest\sasview\installers\Output\setupSasView.exe
```
But then fails on some kind of testing?
```
C:\jenkins\workspace\SasView_Win7_64bit_PullRequest\sasview>cd C:\jenkins\workspace\SasView_Win7_64bit_PullRequest 
Archiving artifacts
Recording test results
ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?
Setting status of 2c58964f44ad477e1d21c86acc84432e979d3393 to FAILURE with url http://jenkins.esss.dk/sasview/job/SasView_Win7_64bit_PullRequest/101/ and message: 'Build finished. '
Using context: PullRequestBuilder_Win64
Finished: FAILURE
```

@pkienzle or @wpotrzebowski ... any ideas? is that an issue with the wx4. stuff Paul did (I note that the previous pull request worked ok) or is this a build machine issue? or both?

wpotrzebowski:
@rozyczko has been looking at it and it seems that he fixed Win_64bit issue. Is it testing issue?

butlerpd:
it does seem to be at testing issue given it says "no test report found?"

butlerpd:
Piotr looking at this issue. Hold till we figure out why build is failing.

Issue #1521, opened 2020-04-20
Incorrect polydispersity reported when fitting without polydispersity on


butlerpd:
Testing of 5.0.2RC2 on widows10

Load a file and send to fitting. Choose any form factor (easiest to leave form factor off).  Note that on the Model tab you can fold down the "volume" parameters to see the polydispersity which is set to 0 and Gaussian (even though polydispersity has not been checked).

Now turn polydispersity on and go to the Polydispersity tab. Set a polydispersity to 0.3 and compute.
Go back to the Model tab and fold down the parameter with polydispersity. It correctly shows 0.3. Now turn polydispersity off and fit again.  The fit (usually) seems to do the right thing but now fold down the parameter again and the polydispersity still says 0.3. You cannot access the polydispersity tab to see what it says because it is grayed out and of course you cannot change it from the Model tab

The net effect is that if I take all the parameters in the Model tab as results I have the wrong polydispersity. Yes I can figure it out if I'm clever but that is not the assumption that should be made :-)

when the polydispersity is turned off, ideally, in my opinion, the "volume' parameters should not be able to roll down to polydispersity but at a minimum, all polydispersity parameters should be set to 0.

butlerpd:
I strongly suspect this has been the case since 5.0.0.  Can we move it to 5.0.3?

butlerpd:
This needs to be checked that it has not been fixed. If not I suspect is would make a "good first issue" and thus labeling it as such

butlerpd:
Indeed it has not been addressed so moving to 5.0.5

PR #1577, opened 2020-06-03
Ess gui 1538 poly constraints

### LINKED ISSUE #1538: Constrained fitting doesn't work when setting min/max for polydispersity
The issue occurs when using Constrained fit (with set up constraints) and polydispersity set up. If one introduces then reasonable min/max values in Polydispersity tab and then do "Fit" , the fitting results is definitely not correct (flat line). This behaviour is not observed with setting min/max for polydispersity in "Single Fit". This can be reproduced by running "Constrained and SImultaneous Fitting" example from http://www.sasview.org/downloads/simultaneous_1d_fitting_in_sasview_v5.pdf and setting min/max when tackling polydispersity

m2cci-NMZ:
Fixes #1538

Corrects some bugs on the polydispersity and adds some unit tests, updates `sasmodel` polydispersity with the GUI values, and also adds a mechanism that sets the `sasmodel` polydispersity to 0 whenever the polydispersity is unchecked in the GUI.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1439,6 +1439,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
                 return
             key = parameter_name + '.' + delegate.columnDict()[model_column]
             self.poly_params[key] = value
+            self.kernel_module.setParam(key, value)
 
             # Update plot
             self.updateData()
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1423,7 +1423,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
                 # Can't be converted properly, bring back the old value and exit
                 return
 
-            current_details = self.kernel_module.details[parameter_name]
+            current_details = self.kernel_module.details[parameter_name + '.width']
             if self.has_poly_error_column:
                 # err column changes the indexing
                 current_details[model_column-2] = value
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -508,11 +508,9 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         # Set sasmodel polydispersity to 0 if polydispersity is unchecked, if not use Qmodel values
         if self._poly_model.rowCount() > 0:
             for key, value in self.poly_params.items():
-                if key[-5:] == 'width':
-                    if isChecked:
-                        self.kernel_module.setParam(key, value)
-                    else:
-                        self.kernel_module.setParam(key, 0)
+                if key[-6:] == '.width':
+                    self.kernel_module.setParam(key, (value if isChecked else 0))
+
 
     def toggleMagnetism(self, isChecked):
         """ Enable/disable the magnetism tab """
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -505,6 +505,14 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         self.tabFitting.setTabEnabled(TAB_POLY, isChecked)
         # Check if any parameters are ready for fitting
         self.cmdFit.setEnabled(self.haveParamsToFit())
+        # Set sasmodel polydispersity to 0 if polydispersity is unchecked, if not use Qmodel values
+        if self._poly_model.rowCount() > 0:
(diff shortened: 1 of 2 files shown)


m2cci-NMZ:
5.0 ready for testing on OSX

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -505,6 +505,14 @@ def togglePoly(self, isChecked):
...
         # Check if any parameters are ready for fitting
         self.cmdFit.setEnabled(self.haveParamsToFit())
+        # Set sasmodel polydispersity to 0 if polydispersity is unchecked, if not use Qmodel values
+        if self._poly_model.rowCount() > 0:
+            for key, value in self.poly_params.items():
+                if key[-5:] == 'width':
+                    if isChecked:
+                        self.kernel_module.setParam(key, value)
+                    else:
+                        self.kernel_module.setParam(key, 0)
```
pkienzle, line 515: You need to protect against parameters with 'width' in the name, so use `key[-6:] == '.width'`.

You might want to use an if-expression here: `self.kernel_model.setParam(key, (value if isChecked else 0))`

butlerpd:
I believe this would have been the bug #1521 I reported. When toggled off the calculation behaves correctly however the GUI on the main tab (if ou unroll the polydisperse parameter) displays the last set polydispersity even though none was used ... it should either be blank (preferable) or zero as it suggests to the user that polydispersity is being used - also I did not test what happens to the report, but guessing it would also indicate a polydispersity had been applied when in fact it had not.

m2cci-NMZ:
@butlerpd This PR doesn't fix #1521, you will still see the value in the main tab when unrolling even if polydispersity is turned off. 
@pkienzle In the current ESS_GUI the plot looks fine when toggling polydispersity off because it is made with a copy of the `sasmodel` object (which is updated). 
This PR fixes mainly two things:
1. In the current ESS_GUI branch, changing polydispersity in the GUI does not update the `sasmodel` object so if you try to fit it will start fitting with a default value. 
2. There is also a bug that sets the min/max values of the parameter itself when changing min/max polydispersity values, which is related to issue #1538. For instance if you set the `radius.width` max in the GUI to 0.5, it will set the `radius` max to 0.5.

butlerpd:
😞   and I got so excited seeing the comment!  I was thinking @m2cci-NMZ was fixing two problems in one go 😃

pkienzle:
Tying this behaviour to the toggle seems like it will miss cases.  What happens when loading a saved model? What happens when pasting parameters?

For model-view-controller (MVC) you should have an underlying model object M containing the complete state that could be rendered to the GUI view, or alternatively to a save file, or sent to a calculation engine, or whereever.  Looking at the code, the poly_params dict holding the GUI values for pd lives in the controller (FittingWidget) rather than the model (FitPage) so you are going to have to work hard to make things consistent across views.

The radical fix is to pull the SasView model out of sasmodels, dump it into sas.sascalc.fitting, and add the pd toggle to that class.  Then the SasviewModel object (which already holds the model parameters) can preserve the pd values even when pd is off, and ignore them when running the calculation.  The pd check box can be saved/restored along with the rest of the fit description. You can rerender the parameter table on toggle rather than update the individual items so that you do not need to repeat the logic in too many places.

Maybe too much of a change for the next release? In which case  could still move the poly_params to FitPage, but set the pd values to the model on model actions (calc, fit, save) rather trying to manage it within the controller.  And again, rerender the parameter table on toggle to keep down the repeated logic.

gonzalezma:
Tried this and I think that it fixes successfully issues #1522 and #1538. #1521 remains to be done.  
As noted by @pkienzle , it is not possible to apply constraints between the PD values. This is possible with SasView 4, so I guess that it is a needed functionality and should be added. Is there an issue already for that? 
I let @m2cci-NMZ  decide about the MVC model suggested by @pkienzle, but from my ignorance, it seems like a large non-trivial work to do, so perhaps this should go to a new enhancement issue.

rozyczko:
Constraints between PD parameters are a separate story, already getting close to being done.

Our model (`poly_model`) indeed resides in the same object as the underlying dictionary used to store parameters for passing to sasmodels (`poly_params`). These two models should be refactored to separate layers, and this is indeed technical debt waiting to get fixed.
Note, that `poly_model` is used for presenting `poly_params` in more than one view simultaneously, so it is working as expected.

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1415,7 +1421,7 @@ def onPolyModelChange(self, top, bottom):
                 # Can't be converted properly, bring back the old value and exit
                 return
 
-            current_details = self.kernel_module.details[parameter_name]
+            current_details = self.kernel_module.details[parameter_name + '.width']
```
rozyczko, line 1424: You could use the already defined `parameter_name_w`

m2cci-NMZ:
Agreed with @gonzalezma, refactoring the model so it sticks to the MVC pattern seems to require non-trivial changes.

wpotrzebowski:
5.0 ready for testing on OSX

wpotrzebowski:
My bad - I've been checking against outdated code. All introduced tests passed.

Issue #1526, opened 2020-04-22
Project Save/Load functionality of 4.x needs to be restored


smk78:
Currently, and unlike 4.x, 5.x will only save and re-load data used for fitting analyses (ie, if a SasView session also involved correlation function analysis, invariant analysis or P(r) inversion, those data and windows are forgotten).

As of release 5.0.2 the only mitigation we are providing is a Known Issue in the Release Notes.

krzywon:
This isn't 100% complete. Inversion and Corfunc states need to be serialized and able to save/load their states.

rozyczko:
Oops, sorry. I'm a simple man. I see `Merged` I click `Close`. Thanks for being vigilant.

butlerpd:
No worries ... this was confusing.  We agreed at last call to move this to the done column for the purposes of the hackathon as part of the functionality requested (Pr) was added even though all was never something we would have time do to. However the ticket was to remain open to have the rest of the perspectives added at a future date (in a different hackathon?)

PR #1579, opened 2020-06-05
Save inversion state in project and analysis files (ESS_GUI)

### LINKED ISSUE #1578: Changing perspective closes the perspective (ESS_GUI)
While trying to work through #1526 I came across an issue when changing perspectives. In 5.x when changing between perspectives the current perspective is closed and a new one is opened. This causes any information currently in the perspective to be thrown away. Only the charts are retained. In 4.x, the perspectives are hidden and always present in the background, retaining fit information when going back and forth.

The use case against closing perspectives: A user wants to run a complex, constrained fit on a series of data sets and then, after running the fit, wants to find the invariant for a data set or two. As soon as the user switches to the invariant state, the complex constrained fit is thrown out and won't return when returning to the fitting perspective.

I would suggest in 5.0.3, at the very least, when switching between perspectives, warn the user the information will not be retained.

krzywon:
The Inversion perspective can now be saved and loaded in project and analysis files. Also, all perspectives retain their information when changing between perspectives allowing project files to save state of all serializable perspectives at one time.

Closes #1578 
Works toward #1526 - Remaining: save/load Corfunc and Invariant perspectives

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -502,6 +502,7 @@ class DataExplorerWindow(DroppableDataLoadWidget):
                     logging.error("Project load failed with " + str(ex))
                     return
         cs_keys = []
+        visible_perspective = DEFAULT_PERSPECTIVE
         for key, value in all_data.items():
             if key == 'is_batch':
                 self.chkBatch.setChecked(value == 'True')
@@ -514,15 +515,24 @@ class DataExplorerWindow(DroppableDataLoadWidget):
             if 'cs_tab' in key:
                 cs_keys.append(key)
                 continue
+            if 'visible_perspective' in key:
+                visible_perspective = value
             # send newly created items to the perspective
             self.updatePerspectiveWithProperties(key, value)
 
+        # Set to fitting perspective and load in Batch and C&S Pages
+        self.cbFitting.setCurrentIndex(
+            self.cbFitting.findText(DEFAULT_PERSPECTIVE))
         # See if there are any batch pages defined and create them, if so
         self.updateWithBatchPages(all_data)
 
         # Only now can we create/assign C&S pages.
         for key in cs_keys:
             self.updatePerspectiveWithProperties(key, all_data[key])
+
+        # Set to perspective shown when project was saved
+        self.cbFitting.setCurrentIndex(
+                self.cbFitting.findText(visible_perspective))
 
     def updateWithBatchPages(self, all_data):
         """
@@ -558,12 +568,12 @@ class DataExplorerWindow(DroppableDataLoadWidget):
     def updatePerspectiveWithProperties(self, key, value):
         """
         """
-        if value.get('fit_data'):
+        if 'fit_data' in value:
             data_dict = {key:value['fit_data']}
             # Create new model items in the data explorer
             items = self.updateModelFromData(data_dict)
 
-        if value.get('fit_params'):
+        if 'fit_params' in value:
             self.cbFitting.setCurrentIndex(self.cbFitting.findText(DEFAULT_PERSPECTIVE))
             params = value['fit_params']
             # Make the perspective read the rest of the read data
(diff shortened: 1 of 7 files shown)


rozyczko:
attempt to save project with P(r) gives me
`data cannot be serialized to json: <class 'method'>`

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -118,32 +119,21 @@ def onExcelCopy(self):
...
-                        params[i].append(tab_data)
-                    else:
-                        params[i] = [tab_data]
-            else:
-                if id in params:
-                    params[id].append(tab_data)
-                else:
-                    params[id] = [tab_data]
-        return params
+            all_data = self.getSerializedFitpage(tab, all_data)
```
rozyczko, line 129: This is a bad way of calling a method. You both modify the passed argument and send it back as return value.
Wouldn't it make sense to just pass `tab`? `getSerializedFitpage` doesn't need to know the whole `all_data` content, only the tab name it needs to serialize. The result can be then just appended to `all_param` in the caller.

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -118,32 +119,21 @@ def onExcelCopy(self):
...
+        return all_data
 
-    def serializeCurrentFitpage(self):
+    def serializeCurrentPage(self, all_data):
```
rozyczko, line 132: See my comment above - let's not modify arguments passed to a method but rather return the modifications.

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -118,32 +119,21 @@ def onExcelCopy(self):
...
         # serialize current(active) fitpage
-        return self.getSerializedFitpage(self.currentTab)
+        return self.getSerializedFitpage(self.currentTab, all_data)
 
-    def getSerializedFitpage(self, tab):
+    def getSerializedFitpage(self, tab, all_data):
         """
```
rozyczko, line 137: as above

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
```diff
@@ -494,6 +501,41 @@ def updateDataList(self, dataRef):
...
+            'q_max': self._calculator.q_max,
+            'q_min': self._calculator.q_min,
+            'rg': self._calculator.rg,
+            'slit_height': self._calculator.slit_height,
+            'slit_width': self._calculator.slit_width,
+            'suggested_alpha': self._calculator.suggested_alpha,
+            'x': self._calculator.x,
+            'y': self._calculator.y,
+        }
+        return params
```
rozyczko, line 537: apart of ease of debugging, there's no reason to introduce a separate variable `params`. You could just create the dictionary as the argument of `return`.

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
```diff
@@ -590,7 +632,8 @@ def removeData(self, data_list=None):
             data_list = [self._data]
         self.closeDMax()
         for data in data_list:
-            self._dataList.pop(data)
+            if data in self._dataList.keys():
```
rozyczko, line 635: yes, we should check the key, but a quicker way to assure the op doesn't raise is to just pass `None`:

`self._dataList.pop(data, None)`

This will return the value if key exists in the dictionary, and None otherwise (no raise)

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -553,12 +558,13 @@ def updateWithBatchPages(self, all_data):
     def updatePerspectiveWithProperties(self, key, value):
         """
         """
-        if 'fit_data' in value:
+        if value.get('fit_data'):
```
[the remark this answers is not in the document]
krzywon: This was a change I thought I had reverted back to its original state in my last commit. Thanks for catching this.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -125,6 +125,13 @@ def addWidgets(self):
         """
         Populate the main window with widgets
         """
+        # Preload all perspectives
+        loaded_dict = {}
+        for name, perspective in Perspectives.PERSPECTIVES.items():
+            loaded_perspective = perspective(parent=self)
+            loaded_dict[name] = loaded_perspective
+        self.loadedPerspectives = loaded_dict
```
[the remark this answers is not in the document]
krzywon: All perspectives are pre-loaded in wxPython and I have never had memory issues with any of the 4.x releases.

Oddly enough, and completely unexpected to me, it seems running SasView from my branch uses less memory than running SasView from the ESS_GUI branch. According to Windows task manager, at idle, the ESS_GUI branch uses 137.8 MB of RAM and my branch uses 136.0 MB.

PR #1580, opened 2020-06-08
Ess gui 1545 copy paste

### LINKED ISSUE #1545: Copy/pasting fitting parameters causes the model name to be copied
When copy pasting parameters from the `FittingWidget`, you also copy the model name. If you then set-up constraints, you will see the model (e.g. M1) appear twice in the Fitpage list.

m2cci-NMZ:
Fixes #1545 

Removes tab names in the dict used for updating the fit page when pasting. Also adds unit tests for PR #1533

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
@@ -136,10 +136,8 @@ class ComplexConstraint(QtWidgets.QDialog, Ui_ComplexConstraintUI):
         """
         Tooltip for txtConstraint
         """
-        p1 = self.tab_names[0] + ":" + self.cbParam1.currentText()
-        p2 = self.tab_names[1] +"."+self.cbParam2.currentText() if len(self.tab_names) != 1 else self.tab_names[0] +"."+self.cbParam2.currentText()
-        tooltip = "E.g.\n%s = 2.0 * (%s)\n" %(p1, p2)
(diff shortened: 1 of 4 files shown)


src/sas/qtgui/Perspectives/Fitting/ComplexConstraint.py
```diff
@@ -137,7 +137,7 @@ def setupTooltip(self):
         Tooltip for txtConstraint
         """
         p1 = self.tab_names[0] + ":" + self.cbParam1.currentText()
-        p2 = self.tab_names[1] if len(self.tab_names) != 1 else self.tab_names[0] +"."+self.cbParam2.currentText()
```
rozyczko, line 140: I am not sure we need to even modify this tooltip - it seems to be generated just once, at the widget instantiation and doesn't update on model change. We would probably hardcode a simple text and just present it.
Alternatively we could try updating this tooltip on model change...
What do you think?

gonzalezma: This is needed to show the same behaviour as for the standard case (when the parameters are set manually without the copy/paste), but it is true that a standard harcoded text will be enough.

m2cci-NMZ: Yes, a hardcoded text seems enough, I don't think that having a tooltip that changes upon the models selected would be usefull as it is just an example. In fact, I had to do this change to satisfy a failing unit test, as my interpreter has the bad habit of exiting every time a unit test fails :-(

Issue #1574, opened 2020-06-01
Invariant perspective fixes need to be ported to 5.x


butlerpd:
The invariant perspective had a factor of 2 error in the calculation of the specific surface area.  This was corrected along with a lot of unit test fixes and upgrades and a massiver rewrite of the documentation.  All these changes need to be ported to 5.x.

butlerpd:
This was fixed for 5.0.3 not sure why the issue was not closed. closing it now

PR #1590, opened 2020-06-14
Ess gui 1574 invariant

butlerpd:
This addresses issue #1574 and is basically a forward port of the changes worked out for 4.3.

From the user perspective the most important changes are:
* It changes the way Sv is calculated (it no longer uses the invariant but only contrast and the porod constant -- it also returns the true Sv rather than 2*Sv as previous versions do)
* The documentation has been completely rewritten and is now a fairly good introduction to the Invariant as well as a tutorial for usage.

Other fixes include
* changed the unit test to match the new behavior but more importantly, in most cases it now relies on an independently calculable value for the return value rather than just being the value that the current algorithm returns.
* Added the new test data sets to the example data sets the user can access
* Fixed many of the uncertainty calculations though they could be improved with better GUI inputs (e.g. input Porod constant or \rho with uncertainty)


src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -318,6 +318,9 @@ def calculateThread(self, extrapolation):
             try:
                 surface, surface_error = \
                     inv.get_surface_with_error(self._contrast, self._porod)
+                if surface_error == 0:
```
rozyczko, line 321: `surface_error` is a float. Comparing floats is a tricky thing, maybe we should do the comparison against some very small value?
`if abs(surface_error) < 0.00001` or something like `if math.isclose(surface_error, 0)`

pkienzle: The test against zero happens to not a problem in this case because the `get_surface_with_error` function *always* returns zero for the surface error. If it were actually calculated, you probably want to use `surface_error < 1e-7*surface` or some such rather than an absolute test so you don't have to worry about the units used for the surface calculation.

There are bigger problems, though.  The setting of the `surface_error` field is overridden on [line 417](https://github.com/SasView/sasview/blob/af992c426120c8ba6dbeee85b6cac535af375961/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py#L417-L420), at least for the case when `bool(surface) is True`.  That means that when surface is non-zero, then a zero will be entered into the surface error field overriding the setting "NONE" being set here.

That `if surface:` test also means that when the `get_surface_with_error` function returns zero the value in the surface field will not be updated, even if it is currently not zero.

A further problem is if an exception is raised here then `surface` is undefined, and the `if surface:` test will throw an exception.

Still another question/concern is that this code is calling `setItem` directly, whereas line 417 is using `reactor.callFromThread` to call `setitem` — why the difference?

So the conditions you want seems to be:
```python
if surface present:
    if get_surface_with_error throws an exception:
        display error in surface and surface error fields
    else:
        display surface value in surface field
        if surface error is zero:
           display none in surface error field
        else:
           display surface error value in surface error field
else:
    display none in surface and surface error fields
```

Maybe hijack `updateModelFromThread` with:
```python
    if value is None:
        item = QtGui.QStandardItem("NONE")
    elif np.isnan(value):
        item = QtGui.QStandardItem("ERROR")
    else:
        item = QtGui.QStandardItem(str(float('%.3g' % value)))
    self.model.setItem(widget, item)
```

Then your logic here would be:
```python
if self._porod:
    try:
        surface, surface_error = ...
        if surface_error < 1e-7*surface:
            surface_error = None
    except Exception as ex:
        calculation_failed = True
        msg += str(ex)
        surface = surface_error = np.NaN
else:
    surface = surface_error = None
```
and below call `updateModelFromThread` for surface and surface_error without the `if surface:` test.

You can do something similar for the qstar logic.

pkienzle: Re "... the if surface: test will throw an exception.":  This is not an issue because line 335 checks for `calculation_failed`.

You still may want to use `reactor.callFromThread` to update the fields within the exception code, or better yet, use `calculation_failed` condition to set the extrapolation values to `None` so that the form update logic is done together as a block at the bottom.

butlerpd: Thanks guys. This was in fact the one bit of code I was hoping to get feedback on as there was no copy paste from 4.x possible here and while we dealt with in wx, this code is beyond my monkey see monkey do approach (having not seen before 😃  so what we are trying to do is this:  currently the calculation was disabled because it was totally bogus and moreover there is no information with which to compute the uncertainty.  That should be another ticket (I need to check).  The ideas was that if not calculated the GUI should show "none" whereas if it is calculated (at some point in the future when the GUI allows the necessary input AND the user chooses to provide it), no matter how small it is, it should show the numeric value.

Is what you suggest above .. which I've only partially understood so far I'll admit (not had time to really think about it yet) or am I totally barking up the wrong tree here?

pkienzle: If you assign zero to a value you can reliably test whether it is zero, but you won't be able to distinguish that from something that is accidentally zero because of floating point math.  

If it is possible that the calculated value is zero and you want to display 0.0, then you should return something else (None?) from the sascalc function when the value is not calculated.

Re `reactor.callFromThread`: I do not know if it is required. In wx you need to use the same thread for all GUI operations or weird things can happen, but I don't know what happens in qt. Maybe @rozyczko can answer?

butlerpd: but if the return value has to be a float it cannot be "None" right? maybe nan? but that suggest and error. Or are you allowed to return a "None" object instead of a float? I guess I can try..

pkienzle: You can return anything you want . All Python values are objects that carry their type along with them, even floats, and python does nothing to enforce the type of the argument received or the returned value.  The caller has to sort it out by using code like `if value is None:`.

We can hint what we think the type should be, which is handy for the maintainer and can let automated tools (e.g., *mypy*) check that you are using types correctly.  In that case we sould define the function as follows:
```python
from typing import Optional, Tuple

def get_surface_with_error(
    self, 
    contrast: float, 
    porod_const: float, 
    extrapolation: Optional[str]=None,
) -> Tuple[float, Optional[float]]:
    ...
```
or in python2.7 style as used by sasmodels:
```python
def get_surface_with_error(self, contrast, porod_const, extrapolation):
    # type: (float, float, Optional[str]) -> Tuple[float, Optional[float]]
    ...
```

butlerpd:
5.0 ready for testing on Win

butlerpd:
In testing these changes to address the review a number of other issues have arisen that cannot be addressed here.  These will be documented in separate tickets.

wpotrzebowski:
5.0 ready for testing on OSX

test/sasinvariant/test/utest_data_handling.py
```diff
@@ -112,164 +111,195 @@ def test_initial_data_processing(self):
...
+        # and finite number of points we need quite a lot of points to ensure
+        # that the fit averages close to 4. SasView estimates 3.92 at the
+        # moment 
+        self.assertTrue(math.fabs(inv._high_extrapolation_function.power-4)<0.1)
+
         self.assertEqual(qs_extr, _qstar+delta_qs_extr)
-        self.assertAlmostEqual(dqs_extr, math.sqrt(dqstar*dqstar + delta_dqs_extr*delta_dqs_extr), 10)
-        
+        self.assertAlmostEqual(dqs_extr, math.sqrt(dqstar*dqstar \
+                                        + delta_dqs_extr*delta_dqs_extr), 10)
```
pkienzle, line 277: PEP8 discourages the use of `\` as a line extender (python gets confused sometimes if there is a space afterward, or if there is some problem with CRLF endings).  You do not need it within parentheses.

Better would be to break at a higher level semantic boundary:
```
self.assertAlmostEqual(
    dqs_extr, math.sqrt(dqstar*dqstar + delta_dqs_extr*delta_dqs_extr), 10)
```
or
```
self.assertAlmostEqual(
    dqs_extr, 
    math.sqrt(dqstar*dqstar + delta_dqs_extr*delta_dqs_extr), 
    10)
```

This appears in a number of places, not all of them noted in this review.

test/sasinvariant/test/utest_data_handling.py
```diff
@@ -473,10 +540,11 @@ def setUp(self):
         self.scale = 1.5
         self.rg = 30.0
         x = np.arange(0.0001, 0.1, 0.0001)
-        y = np.asarray([self.scale * math.exp( -(q*self.rg)**2 / 3.0 ) for q in x])
+        y = np.asarray([self.scale *
+                        math.exp( -(q*self.rg)**2 / 3.0 ) for q in x])
```
pkienzle, line 544: Break at higher level semantic boundaries. Also, lint will complain about space after open or before close parentheses:
```
y = np.asarray([self.scale * math.exp(-(q*self.rg)**2 / 3.0)
                for q in x])
```

test/sasinvariant/test/utest_data_handling.py
```diff
@@ -521,89 +598,103 @@ def setUp(self):
...
```
pkienzle, line 602: as above

test/sasinvariant/test/utest_data_handling.py
```diff
@@ -348,10 +397,11 @@ def setUp(self):
...
```
pkienzle, line 401: as below

src/sas/sascalc/invariant/invariant.py
```diff
@@ -537,9 +538,12 @@ def _get_qstar_uncertainty(self, data):
...
             raise ValueError(msg)
         else:
-            #Create error for data without dy error
+            # For reduced data sqrt(I) is incorrect! if the data has no dy
+            # then we should not invent it and then propogate that completely
+            # bogus value.  So changed behaviour on Mar 23, 2020 to return
+            # None instead
             if data.dy is None:
-                dy = math.sqrt(data.y)
+                return None
```
pkienzle, line 546: Lint doesn't like `else` blocks when the `if` block ends in `return`.

wpotrzebowski:
5.0 ready for testing on OSX

wpotrzebowski:
I've been testing PR on mac. When invariant cannot be calculated ("ERROR" appears), clicking on "Status" button generates 
ERROR: Traceback (most recent call last):
  File "site-packages/sasview-5.0.2-py3.6.egg/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py", line 443, in onStatus
  File "site-packages/sasview-5.0.2-py3.6.egg/sas/qtgui/Perspectives/Invariant/InvariantDetails.py", line 78, in showDialog
One potential solution to this issue is to disable "Status" button when proper "ERROR" is generated.

wpotrzebowski:
5.0 ready for testing on OSX

smk78:
Ok, if someone could change the units on the contrast box

image

to say x10<sup>-6</sup> Å<sup>-2</sup> I think this will be good to merge.

wpotrzebowski:
5.0 ready for testing on OSX

butlerpd:
Sorry about the confusion.  The "error" was not an actual code error.  It may be that a better algorithm could be implemented but it would have to be figured out, then implemented and tested. for now a value of "error" when the contrast term is not exactly right is in fact the correct answer.

We need to revert the UI changes and the associated changes in the documentation before this is ready to merge

A quick fix to avoid almost ever seeing the vol frac show "ERROR" would be to set delta rho default judiciously.  Looking at the vol frac equation that means that 4*Q*<2*pi*(delrho)^2

From this it is clear that as long as delrho is set to a value greater than the actual value (the one in Q*) the determinant will remain positive.  while not 100% fool proof a value of 8e-6 would probably cover almost all practical cases.  So I suggest we do that. Of course 1 is also much bigger but by 6 orders of magnitude and quite problematic in its own right.

butlerpd:
An interesting side comment: there is a weakness in the unit testing that has been exposed. the unit tests should have failed when changing to say you are taking sld in units of e^6. Of course it didn't because the computational code does its own unit tests and continues to send 2e-6 as the value (which **is** correct for the data set used in the test) and so continues to return the correct vol fraction and surface area (which are independently calculated so known correct).  The Gui unit tests were changed to match the change in the GUI but that seems ... wrong? but right?  A larger and much more philosophical discussion is involved here but thought I'd flag anyway 😃

smk78:
Unfortunately I have just spotted a couple more things that trouble me:

1) In the Invariant tab of the Invariant Perspective you can change the Min and Max Q values. My working assumption is that this is so that you can exclude points at either end of the data prior to the extrapolations? _However_, if you try this you will find that the extrapolations _always use the full range of the data_ (ie, if you had a dataset with 50 points and opted to extrapolate the low-Q end with 10 points it uses points 1 to 10. If you were to then set a Qmin to exclude the first 10 points you would expect that it would extrapolate on points 11 to 20. It doesn't, it still uses points 1 to 10. The same behaviour happens with the high-Q extrapolations). So, is this by design? If so, why are the Total Q Range boxes editable?

2) Here is an invariant calculation on one of our test datasets
image
I notice that the Invariant Total returned in the Perspective window is _only_ the Q<sup>\*</sup> from Data shown in the Status Dialog window; it doesn't include the contributions to Q<sup>*</sup> from the area under the extrapolations. Surely it should?

smk78:
If you _do_ now generate a negative discriminant you get this in the console:

```
QObject::connect: Cannot queue arguments of type 'QVector<int>'
(Make sure 'QVector<int>' is registered using qRegisterMetaType().)
QObject::connect: Cannot queue arguments of type 'QVector<int>'
(Make sure 'QVector<int>' is registered using qRegisterMetaType().)
11:27:42 - WARNING: Calculation failed: Could not compute the volume fraction: negative discriminant
```

The last line also gets displayed in the Log Explorer. Which is good!

smk78:
3) If you _Enable High-Q extrapolation_, select _**Fix**_, enter a _Power_, and then click _Calculate_, an extrapolation is generated, but the radio button jumps back to **_Fit_**. This does not happen in 4.x and could catch out the unwary?

4) It also appears that no matter what data you load, or what low-Q extrapolation you perform on it, the **_percent_** Q<sup>\*</sup> from Low-Q in the chart in the _Status_ > Dialog window is always exactly zero! In 4.x, even fractions of a percent to 2 dp are shown. Interestingly, the Dialog window does show a numerical value for the contribution to Q<sup>\*</sup> from Low-Q.

wpotrzebowski:
It seems that this perspective requires more attention but I guess it's improved from 5.0.2? If this is the case I would argue to merge it now and work on remaining issues later (post 5.0.3). @butlerpd  already created few tickets.

smk78:
So the most troubling is 2) above, ticketed, I've just noticed, by @butlerpd in https://github.com/SasView/sasview/issues/1606. Because it means the value of the Invariant Total we are returning in the Invariant Analysis window is wrong. In my screenshot above, wrong by over ~30 %!

We could argue, I suppose, that 30 % is a big improvement on the 100+ % error that all previous releases of SasView were returning... But I kind of feel it is incumbent on us to give people a truthful answer this time?

smk78:
I've also found some differences in the behaviour of 4x and 5x doing the same operations in each. Now, I've built _master_ on the assumption that the latest version of the Invariant got merged into it. But I'm struggling to satisfy myself that is the case...

butlerpd:
I would not worry about 4.x.  We **did** do a merge of the 4.x PR as I recall but before it was really ready? - so it should be part way towards the latest.  The latest should reside in 5.x PR while the 4.x PR should have the next to latest, modulo the last change on the uncertainty in s (ds - which may or may not be appropriate to port back to 4.x) and some inevitable "look and feel" differences.

butlerpd:
> Unfortunately I have just spotted a couple more things that trouble me:
> 
> 1. In the Invariant tab of the Invariant Perspective you can change the Min and Max Q values. My working assumption is that this is so that you can exclude points at either end of the data prior to the extrapolations? _However_, if you try this you will find that the extrapolations _always use the full range of the data_ (ie, if you had a dataset with 50 points and opted to extrapolate the low-Q end with 10 points it uses points 1 to 10. If you were to then set a Qmin to exclude the first 10 points you would expect that it would extrapolate on points 11 to 20. It doesn't, it still uses points 1 to 10. The same behaviour happens with the high-Q extrapolations). So, is this by design? If so, why are the Total Q Range boxes editable?
>    I notice these boxes are _not_ editable in 4.x.
>    Also see #1608

Yep I think that is another of the "many issues with invariant" - not sure I'd caught that yet. Can you please ticket?

smk78:
The default _Contrast_ in 4.x needs updating to be the same as in 5.x
Ticketed as https://github.com/SasView/sasview/issues/1612

smk78:
The uncertainty on the computed volume fraction in 4.x significantly differs from that in 5.x
Ticketed as https://github.com/SasView/sasview/issues/1613

smk78:
> So the most troubling is 2) above, ticketed, I've just noticed, by @butlerpd in #1606. Because it means the value of the Invariant Total we are returning in the Invariant Analysis window is wrong. In my screenshot above, wrong by over ~30 %!

>We could argue, I suppose, that 30 % is a big improvement on the 100+ % error that all previous releases of SasView were returning... But I kind of feel it is incumbent on us to give people a truthful answer this time? (Though I am reminded of the exchange between the Jack Nicholson and Tom Cruise in A Few Good Men: "You want answers?" | "I want the truth!" | "You can't handle the truth!" :-) )

4.x _master_ **is** correctly returning the Invariant Total (ie, the sum of the low-Q integral + data integral + high-Q integral)

smk78:
4.x and 5.x are generating rather different high-Q extrapolations for the same data/parameters/settings. Visually, however, the 5.x looks the more believeable.
Ticketed as 4.x issue https://github.com/SasView/sasview/issues/1614

smk78:
I have brought the 4.x **Issue1434_Invariant_Documentation** branch up to the same default contrast and brought its invariant_help up to the same as in the **ESS_GUI_1574_Invariant** branch (with due respect for the differences in UIs and their operation). Those changes are no waiting in the PR https://github.com/SasView/sasview/pull/1461 which I've asked @wpotrzebowski  to review.

butlerpd:
I have now looked through all the committed code and agree that all is correct.  Only needs testing of the latest "total invariant" fix code which looks right but ... (though it may not fix issue #1610 - it should fix #1606 however).  Since I initiated this pull request I cannot approve it but for the record, I approve of all the code changes.  I will do a final read of the docs. Then if @smk78 can test the latest fix and approve this can be merged.... finally

smk78:
Testing @wpotrzebowski 's last commit (https://github.com/SasView/sasview/pull/1590/commits/5e0592cbabecd7b4b3caf78d4a623de83dbeed94) above. I have Good news and Bad news...

Building the 4.x **ticket1434-invariant-documentation** branch:
image

Building the 5.x **ticket1574-invariant** branch:
image

The Good news:
The _Invariant Total_ and its uncertainty now agree across both branches!!!

The Bad news:
The _Q* from Data_ values (incl the uncertainty) in the 5.x _Status_ Dialog have **also** been replaced by the _Invariant Total_ values (and uncertainty). Unfortunately this matters, because 5.x is not computing the correct _Volume Fraction_. What seems to be happening is that 5.x is computing the _Volume Fraction_ from the _true_ value of _Q* from Data_ (currently not appearing in the Dialog) rather than from the _Invariant Total_. If I put that in to numbers from the above screen shots:
```
__________________4.x is giving:______________5.x is giving:
Q* from Low-Q:    7.84e-06 +/- 3.12e-07_______7.84e-06 +/- 3.12e-07
Q* from Data:     0.00426  +/- 6.21e-05_______0.00525  +/- 8.05e-05 (this has been over-written)
Q* from High-Q:   0.000983 +/- 5.12e-05_______0.000983 +/- 5.12e-05
Invariant Total:  0.00525  +/- 8.05e-05_______0.00525  +/- 8.05e-05
Volume Fraction:  0.0676   +/- 0.00112________0.054    +/- 0.000834
A:                0.0630______________________0.051
```
So if you add the Q* for the 4.x column you get the _Invariant Total_, but not if you add them in the 5.x column.

Checking the _Volume Fraction_, phi1, is easier if you do it as phi1.phi2=A, as _Invariant Total_ = 1e8 . 2 .pi^2 . contrast^2 . A = 0.083 . A for the parameters above. So A(4.x) gives the correct _Invariant Total_ of 0.00525, but A(5.x) gives an _Invariant Total_ of 0.00426 which is what the _Invariant Total_ was (just the _Q* from Data_) **before it was overwritten**.

butlerpd:
Documentation Review:
Overall looks really good now.  I see three minor fixes that could/should be made:
1. Small error in the first sentence (I'll fix that)
2. I think the new warning about needing to properly set the SLD should move to the usage section
3. the current screen shot shows the default to be 2e-6 even though the warning specifically says it is 8e-6. this might be confusing.

butlerpd:
Regarding the problem reported by @smk78 I think the first part (putting total in Q* from data) is probably because the UI is grabbing the number from the same place as the Q* total from the front panel (further suggesting this was actually done by design because the usage was not understood by whomever was coding it)

I am concerned however about the volume fraction since that is being computed in sascalc along with all the Q* values and should be independent of where the GUI puts what.  speaks to some cross talk between the computation and the GUI?

butlerpd:
I think this is now ready to go.  I fixed item 1 and 2 in documentation review and changed in such a way that I don't think 3 (a new image) is necessary.

I have also fixed the issue with qtotal vs qdata.  I believe the vol frac calculation is correct (based on unit test data) and ascertained that the low q chart is in fact behaving as correctly as it can (did I mention that low q extrapolation doesn't generally matter 😃 

Also the problem with extrapolation not being able to be turned back off is partly solved.  The numerical values were not being reset like the chart values were.  However the extrapolation graph items clearly are also not getting reset.... but ran out of time trying to figure that out. At this point it is probably time to stop and merge (assuming the latest reviews ok) so we can get a release out would be my vote.

wpotrzebowski:
5.0 ready for testing on OSX

wpotrzebowski:
5.0 ready for testing on Win

wpotrzebowski:
5.0 ready for testing on Linux

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -412,13 +414,15 @@ def calculateThread(self, extrapolation):
             reactor.callFromThread(self.updateModelFromThread, WIDGETS.W_SPECIFIC_SURFACE_ERR,
                                    surface_error)
 
-        qstar_total += qstar_low + qstar_high
+        qstar_total = qstar_data + qstar_low + qstar_high
         qstar_total_error = np.sqrt(
-            qstar_total_error * qstar_total_error
+            qstar_data_err * qstar_total_error
```
wpotrzebowski, line 419: Shouldn't be "qstar_data_err * qstar_data_err" ?

src/sas/qtgui/Perspectives/Invariant/InvariantDetails.py
```diff
@@ -202,4 +205,7 @@ def checkValues(self):
             msg += "The sum of all extrapolated contributions is higher " \
                    "than 5% of the invariant.\n"
 
+        if msg == '':
```
wpotrzebowski, line 208: Do we want to output it or it was set for debugging purposes?

smk78: > Do we want to output it or it was set for debugging purposes?

No. That can stay in. Similar messaging is in 4.x.

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
smk78: Hmm. Will need @butlerpd to check this one. Where these things are calculated and passed from is a bit of a can of worms as you no doubt discovered.

src/sas/qtgui/Perspectives/Invariant/InvariantDetails.py
```diff
@@ -77,13 +78,20 @@ def showDialog(self):
...
-        self.txtQDataErr.setText(self._model.item(WIDGETS.W_INVARIANT_ERR).text())
-
         # Reset progress counters
         self.progress_low_qstar = 0.0
         self.progress_high_qstar = 0.0
-        self.progress_qstar = 100.0
+        self.progress_data_qstar = 100.0
+
+        # Q* from data
+        self.qdata = float(self._model.item(WIDGETS.D_DATA_QSTAR).text())
```
rozyczko, line 87: This will fail with `ValueError` when the Q* calculations fail leaving the field `D_DATA_QSTAR` empty. Would be good to surround it with `try..except`

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -331,8 +326,11 @@ def calculateThread(self, extrapolation):
             surface = None
 
         if (calculation_failed):
+            self.cmdStatus.setEnabled(False)
             logging.warning('Calculation failed: {}'.format(msg))
             return self.model
+        else:
```
rozyczko, line 332: This `else` is redundant, since the above condition exits the method. Remove `else` and reset the indentation on line 333 (enabling cmdStatus)

src/sas/qtgui/Perspectives/Invariant/InvariantDetails.py
butlerpd: Yes this was very deliberate (and took me a while to figure out). Currently, without this, the default is message in the details warnings box is: `No Details on calculations available.`  Which is misleading since the implication to me (and confused me when I was using it at first) is that something went wrong.  

NOTE: this is set in the panel initialization as the default entry at which point it is true that no details are avalialbe since the code running the calculation has not run so decided that changing it there was not correct?

Also, oddly line 59 sets the same message but is never used anywhere that I can see! (and changing it has no effect I could see).  Was going to delete but decided I didn't know enough?  Should we delete?
```python
        self.warning_msg = "No Details on calculations available...\n"
```

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
butlerpd: hummm... don't remember doing anything with cmdStatus but agreed. Will do.

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
butlerpd: Thanks @smk78 for letting me take a look - sometimes weird things happen in this invariant code indeed.... but in this case I am thoroughly embarrassed. @wpotrzebowski is correct. Here we are just adding the errors in quadrature not doine sophisticated error propagation through a function (where something like this might be used).  My bad (must have been in a hurry to get to something else ☹️  really good catch on the review!

src/sas/qtgui/Perspectives/Invariant/InvariantDetails.py
butlerpd: Hummm... so I just copied this from the low q and hi q code, neither of which uses a try except.  Should there be a try except on all three?

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
rozyczko: This code was marked as changed/added in this PR, which is why I looked at it :) This is just a stylistic issue, so feel free to just click on `Resolve`.

src/sas/qtgui/Perspectives/Invariant/InvariantDetails.py
rozyczko: OK. These fields can be left without `try...except` but should probably be made read-only. Currently they can be modified but modifications don't do anything since these values are output only. This is more of a 'would be nice and less confusing to the user' than 'it will wreak havoc if users enter junk' category, hence can safely be taken out of the PR and made a separate issue to address in 5.0.4/5.1

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
butlerpd: and easy to fix .. I see that Paul K. has spotted another instance (mine in that case) and labelled "lint" I will fix both shortly

butlerpd:
Ok ... I have addressed all comments now I think, including making all (I think) of the output only boxes that were editable now only read-only.  I note that in 4.x we made the equivalent of "disabled" rather than read only so it was dead obvious that they were output only.  There is the subtlety though that in Qt, disable also greys the font besides the background.

I also apparently had failed to push the partial fix to #1607 which now clears all the detail values before a new calculation. It should be almost as simple to fix the other part which is the persistence of the extrapolation curves but I cannot see how to do it yet.  I can see where they are calculated IF extrapolation is true - but they then persist forever until another extrapolation=true calc.  What we need is to have them cleared every time a calculation is done so that if extrapolation=false they don't show up on the plot.

Finally a bit moot but spent most of my time rederiving ds which was clearly still wrong -- that needs backporting to 4.x PR

Otherwise I am good with this being merged unless y'all find something new.

smk78:
Having pulled the branch and built it, when I tried to run it I got this error:
```
  File "run.py", line 138, in <module>
    run_sasview()
  File "D:\SasViewCodeDev\ticket1574-invariant\src\sas\qtgui\MainWindow\MainWindow.py", line 106, in run_sasview
    mainwindow = MainSasViewWindow(screen_resolution)
  File "D:\SasViewCodeDev\ticket1574-invariant\src\sas\qtgui\MainWindow\MainWindow.py", line 38, in __init__
    from .GuiManager import GuiManager
  File "D:\SasViewCodeDev\ticket1574-invariant\src\sas\qtgui\MainWindow\GuiManager.py", line 50, in <module>
    import sas.qtgui.Perspectives as Perspectives
  File "D:\SasViewCodeDev\ticket1574-invariant\src\sas\qtgui\Perspectives\__init__.py", line 5, in <module>
    from .Invariant.InvariantPerspective import InvariantWindow
  File "D:\SasViewCodeDev\ticket1574-invariant\src\sas\qtgui\Perspectives\Invariant\InvariantPerspective.py", line 13, in <module>
    from sas.sascalc.invariant import invariant
  File "D:\SasViewCodeDev\ticket1574-invariant\src\sas\sascalc\invariant\invariant.py", line 19
    from __future__ import division
    ^
SyntaxError: from __future__ imports must occur at the beginning of the file
```
I have resolved it by moving the from \_\_future\_\_ import from line 19 to line 9

smk78:
Building the 5.x **ticket1574-invariant** branch:

More Good news:
The _Total Q range_ boxes are no longer editable.
The displayed _Invariant Total_ (and its uncertainty) agree with 4.x.
The _Q* from_ values (and their uncertainties) in the _Status_ Dialog window now agree with 4.x.
Help doc is now good.

The Bad news:
The computed _Volume fraction_ in 5.x is **not** the same as in 4.x, and I think it is 5.x that is wrong. :-(
Now the _Status_ Dialog has been corrected it is clearer what is happening; in 5.x the _Volume fraction_ is being computed from the _Q* from Data_ **and not** from the _Invariant Total_.

butlerpd:
Thanks @smk78.... but I just looked at the code to remind myself and in fact the GUI code does not choose which invariant to use in calculating the vol_fraction. It only sends the contrast. It does have to tell the computation which extrapolations to use (high, low or both) but I think that is being done correctly?  Also my tests show that vol fraction is giving the correct result. can you tell me which data set you are using (or post here if it is not one in the test suite) along with the different values you are seeing?

smk78:
Just using the \test\1d_data\AOT_Microemulsion-Drop_Contrast.xml
Chose a background of 0.13 and a contrast of 6.5e-06 with Guinier and fixed extrapolations.
As I said yesterday:
> Checking the Volume Fraction, phi1, is easier if you do it as phi1.phi2=A, as Invariant Total = 1e8 . 2 .pi^2 . contrast^2

Which works out as Invariant Total = 0.083 . A

butlerpd:
Sorry @smk78 ... I think I'd figured out on Friday that was your data set from the figure posted above.  My memory is not so good ☹️  will try to see what I can figure out. Thanks.

src/sas/qtgui/Perspectives/Invariant/UI/InvariantDetailsUI.ui
```diff
@@ -89,6 +89,9 @@
         <property name="toolTip">
          <string>Extrapolated invariant from low-Q range.</string>
         </property>
+        <property name="readOnly">
+         <bool>true</bool>
+        </property>
```
pkienzle, line 94: On Mac there is no visual indication that a field is read-only.  You click and either a cursor does or does not appear. When it doesn't you click a bunch of times to make sure you didn't miss the entry field when clicking.

Maybe set the background for read-only fields to the same color as the form so they look like the labels that they are?

butlerpd:
I think I see the problem @smk78 and you would be correct (my examples did not have much contribution from the extrapolations 😦 It seems that just setting the state of the extrapolation is not sufficient -- you must also call the volume fraction method not only with the contrast but also the extrapolation you want to use (low, high,or both)  if no extrapolation string is passed the code assumes extrapolation = None 🤕. Will try to fix asap but am rapidly becoming a pumpkin as far as SasView is concerned I'm afraid.

src/sas/qtgui/Perspectives/Invariant/InvariantDetails.py
```diff
@@ -60,13 +60,14 @@ def __init__(self, parent):
...
         self.qstar_total = None
+        self.qdata = None
         self.qhigh = None
         self.qlow = None
         self._model = None
 
         self.progress_low_qstar = 0.0
         self.progress_high_qstar = 0.0
-        self.progress_qstar = 100.0
+        self.progress_data_qstar = 100.0
```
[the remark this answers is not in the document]
butlerpd: Agreed it is .... not ideal.  However I am going to ask to defer this to the next effort.  Indeed part of the naming comes from the fact that the widget chosen for this **IS** in fact a progress bar which is not ideal either and part of @smk78 issue with why low q is always reporting 0% when 4.x was reporting a value of 0.6%.  Also I am not yet familiar enough with the code base much less Qt to make such wholesale changes and testing for unintended consequences (e.g. did I really catch all the calls using it) to feel comfortable doing this before a release with zero time to look at it.

For now that bit is just copied/pasted and keeping exactly the same style as the entire rest of the invariant GUI code.

src/sas/qtgui/Perspectives/Invariant/InvariantDetails.py
```diff
@@ -77,13 +78,29 @@ def showDialog(self):
...
+
+
+        # Q* from data
+        self.qdata = float(self._model.item(WIDGETS.D_DATA_QSTAR).text())
+
+        self.txtQData.setText(str(self.qdata))
+        self.txtQDataErr.setText(self._model.item(WIDGETS.D_DATA_QSTAR_ERR).text())
+        try:
+            self.progress_data_qstar = (self.qdata/self.qstar_total)*100.0
+        except:
```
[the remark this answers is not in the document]
butlerpd: Again am going to ask to defer this.  This would be a  change to the original structure of the code.  The change here is another copy paste from the other identical sections below it. I would rather keep the style the same until we change all of them. Otherwise it becomes hard to follow (at least for me who is not a python ninja 😃  when each related bit is written with a different style.

src/sas/qtgui/Perspectives/Invariant/UI/InvariantDetailsUI.ui
butlerpd: Sadly that is true for PC also.  This is the quick improvement over all those fields being editable making the user think they were changing something when in fact the code was ignoring them.

I can also disable the box which will make it gray like the background but then the text can no longer be copied to the clipboard and more annoyingly it is grey as well instead of black which I personally find more annoying that the current.  As I stated somewhere (maybe in the commit message? if so that would be pointless since nobody reads those 😃  I do not know yet how to achieve what we had in 4.x.  Moreover I think this is a general thing in the ESS_GUI and standardizing output boxes would be I think the right answer - but that requires a discussion I think.

wpotrzebowski:
5.0 ready for testing on OSX

smk78:
Have pulled and re-built the 5.x **ticket1574-invariant** branch and re-compared it against the 4.x **ticket1434-invariant-documentation** branch, and...

...I am delighted to report that both branches are now returning the same numbers for the same parameters!

The one thing that I do notice is that if you specify a _Porod Constant_, though both branches return a value for the _Specific Surface_ (and the same value!), 4.x gives NONE for the uncertainty whilst 5.x leaves the uncertainty box blank. But I think @butlerpd was aware of this. However, it is not a feature that should delay merging this branch any longer.

smk78:
Finally, just to add that this is not the end of matters: as our Issues list will testify, both the Invariant (and Correlation Function) Perspectives need a lot more TLC!

PR #1593, opened 2020-06-15
ESS_GUI: silence matplotlib debug/info warnings in log files

pkienzle:


pkienzle:
76000 out of 82000 lines in my sasview.log contain matplotlib.

rozyczko:
Maybe by chance you have `FILTER_DEBUG_LOGS` set to `False` in your `custom_config.py`?
By default it should filter MPL silliness and it's set to `True`

pkienzle:
It is set to `True`.  

Regardless, when I turn on debug mode for sasview/sasmodels I don't want to be flooded with MPL stuff, so the PR still stands.

rozyczko:
fine

PR #1598, opened 2020-06-18
ESS_GUI: allow data replacement on a fit page

### LINKED ISSUE #1537: Allow for replacing data in a Fit Page
Follow up to the discussion in #1490 and our biweekly call.

The first step to copy/paste constraints project is to allow for replacing existing data in a Fit Page.
This could be done similarly to Append to new plot, with the "Send To" button allowing for sending data to an existing page (but by default creating a new one).

m2cci-NMZ:
Allows sending a data set to the active fit page via a checkbox in the `DataExplorer`, which will replace the current dataset. Performing the replacement will not delete any children of the replaced data set, as described in the linked issue.
Fixes #1537

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -732,7 +732,7 @@ class DataExplorerWindow(DroppableDataLoadWidget):
                 title = self._perspective().title()
             else:
                 title = self._perspective().windowTitle()
-            msg = title + " does not allow replacing multiple data. Please select only one data."
+            msg = title + " does not allow replacing multiple data."
             msgbox = QtWidgets.QMessageBox()
             msgbox.setIcon(QtWidgets.QMessageBox.Critical)
             msgbox.setText(msg)
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -726,6 +726,19 @@ class DataExplorerWindow(DroppableDataLoadWidget):
 
         if len(selected_items) < 1:
             return
+        #Check that you have only one box item checked when swaping data
+        if len(selected_items) > 1 and self.chkSwap.isChecked():
+            if hasattr(self._perspective(), 'title'):
+                title = self._perspective().title()
+            else:
+                title = self._perspective().windowTitle()
+            msg = title + " does not allow replacing multiple data. Please select only one data."
+            msgbox = QtWidgets.QMessageBox()
+            msgbox.setIcon(QtWidgets.QMessageBox.Critical)
+            msgbox.setText(msg)
+            msgbox.setStandardButtons(QtWidgets.QMessageBox.Ok)
+            retval = msgbox.exec_()
+            return
 
         # Which perspective has been selected?
         if len(selected_items) > 1 and not self._perspective().allowBatch():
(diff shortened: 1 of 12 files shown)


m2cci-NMZ:
> In my local build, there are error messages when the new data is longer than the old (but not when it is same length or shorter).
> 
> Perhaps you need to populate the child arrays after the first plot or fit, as I noticed they appeared immediately - as a copy of the previous ones ?
> 
> Else looks good, (Though the branch still has plotting issues if same data is on more than one FitPage tab.)
> 
> File "C:\sasview42\sasview\src\sas\sascalc\data_util\calcthread.py", line 269, in _run
> self.compute(*args, **kwargs)
> File "C:\sasview42\sasview\src\sas\qtgui\Perspectives\Fitting\ModelThread.py", line 257, in compute
> res = dict(x = self.data.x[index], y = output[index],
> IndexError: boolean index did not match indexed array along dimension 0; dimension is 92 but corresponding boolean dimension is 124

Nice catch! Were you using instrumental smearing? Seems like the problem comes from the smearer

RichardHeenan:
No, was not using smearing.

butlerpd:
It looks like the residual calculation rather than smearing no?

RichardHeenan:
Was using the test data sets in table below, sorry, the 10000A sphere and RT2  do  have the smearing turned on by default, but the AOT does not. 
I’m getting slightly confused, but I now think the errors appear when the length of the smeared data changes either up or down.  
Swapping smeared into non-smeared FitPage works, and turns the smearing on.
Whether smearing is included is a feature of the data not the model, so "swap" will need to allow for this.
  
                                                                Size                smearing
AOT_microemulsion-core ….                 137             none

10000A_sphere….                                    75       Use dq data

ISIS_Polymer_blend_RT2                       124       Use dq data

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -726,14 +726,13 @@ def isItemReady(index):
...
-        # Which perspective has been selected?
-        if len(selected_items) > 1 and not self._perspective().allowBatch():
+        #Check that you have only one box item checked when swaping data
+        if len(selected_items) > 1 and (self.chkSwap.isChecked() or not self._perspective().allowBatch()):
             if hasattr(self._perspective(), 'title'):
                 title = self._perspective().title()
             else:
                 title = self._perspective().windowTitle()
-            msg = title + " does not allow multiple data."
+            msg = title + " does not allow replacing multiple data."
```
rozyczko, line 735: You probably want to use the perspective name rather than the `title()`  method, since `title()` returns the full string including the fit algorithm, so we get a pretty confusing message
`Fit panel - Active Fitting Optimizer: Levenberg-Marquardt does not allow replacing multiple data`.

m2cci-NMZ:
Problem was caused from the `smearer` not being updated when data was swapped. The data `setter` now forces a reset of the `smearer`, which should solve the problem. Minor issues on error messages were also adressed

butlerpd:
actually I think it makes sense?  I'm guessing that swapping data behaves exactly like sending data to fitpage with parameters on it  but no fit? If that is the case then Nparams=0.  Once you choose what parameters to fit (however many that may be) and click fit then an actual fit is performed with Nparams>0 (if it is not you get a warning error).  If that is the case (as opposed to instantly doing a new fit when you swap data) then the behavior is correct (assuming the change is commensurate with that normalization difference).  Not sure if that is what is being done?

m2cci-NMZ:
> actually I think it makes sense? I'm guessing that swapping data behaves exactly like sending data to fitpage with parameters on it but no fit? If that is the case then Nparams=0. Once you choose what parameters to fit (however many that may be) and click fit then an actual fit is performed with Nparams>0 (if it is not you get a warning error). If that is the case (as opposed to instantly doing a new fit when you swap data) then the behavior is correct (assuming the change is commensurate with that normalization difference). Not sure if that is what is being done?

Yes, when swapping data (or changing the model), the method that computes the chi square does not take into acount degrees of freedom.

PR #1603, opened 2020-06-25
ESS_GUI travis opencl error

pkienzle:


.travis.yml
```diff
@@ -2,17 +2,23 @@ matrix:
...
-    env: PY=3.6
+    env:
+      - MINICONDA=https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
+      - PY=3.6
   - os: osx
     language: generic
-    env: PY=3.6
+    env:
+      - MINICONDA=https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
+      - PY=3.6
```
rozyczko, line 12: We aren't building Windows version on Travis?
(just a discussion point for future, not something to address now, obviously)

pkienzle: With Jenkins doing the official builds our travis coverage didn't not need to be extensive.

Replacing travis and jenkins with github actions is a possibility, which Andrew is exploring in https://github.com/SasView/sasmodels/pull/406.  This activity is incomplete, since the travis build is currently doing things like slack notification that are not part of the github actions build. Somebody with admin privileges on sasmodels needs to review and update the build system so that we only have to maintain one. Please continue this discussion in that PR.

PR #1628, opened 2020-07-15
User feedback when fitting fails because of inconsistent constraint

### LINKED ISSUE #1532: Add a constraint checking mechanism
When fitting with constraints, no check is performed, so the user can define unfeasible constraints. If he makes a mistake when defining a constraint and tries to fit, he will not be prompted that something is wrong, there will only be the stack trace printed in the log console, which gives him no clue on what he is doing wrong. This can be particularly frustrating when then number of constraints is high, as the user is more likely to make mistakes that are hard to spot. A validation/checking mechanism giving the user feedback whenever constraints are unfeasible would greatly improve user experience.

m2cci-NMZ:
Fixes #1532 
This PR attemps giving the user some feedback if he types an unconsistent constraint e.g. he  mispells a parameter in a constraint by typing `M1.radiu` instead of `M1.radius`. If this happens, the `BumpsFitting` module will add the exception and the traceback to the result object, and the GUI will analyze it and warn the user by showing a message box telling the user that the fitting has failed. The message box will tell him the number of the constraint that is faulty and highlight the mispelled parameter. 
Also adds some unit tests for the `ComplexConstraint` module.


src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -492,6 +492,43 @@ def fitComplete(self, result):
             msg = "Fitting failed. Please ensure correctness of chosen constraints."
             self.parent.communicate.statusBarUpdateSignal.emit(msg)
             return
+        # Warn the user if fitting has been unsuccessfull
+        if not result[0][0][0].success:
```
krzywon, line 496: A few lines after this if statement the line `results = result[0][0]` appears. Using this definition here could help with the code readability.

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -571,6 +608,26 @@ def fitFailed(self, reason):
...
+        return None
+
+    def findNameErrorInConstraint(self, constraint, expression, name_error):
+        """
+        Finds the unconsistent name in a constraint
+        Takes a faulty constraint, the translated constraint (by expression.py), and the NameError symbol as arguments
+        """
+        for name in constraint.split():
+            if name in expression and name_error in name:
+                return name
```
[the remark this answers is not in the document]
m2cci-NMZ: Unfortunately the exception is raised at the first invalid constraint. Catching all of them at once would require non trivial changes, as we would need to remove the invalid constraint, recompile `functiondef` in `expression.py` execute it and catch the next exception raised.

m2cci-NMZ:
Thanks for the review! Something I forgot to point out: this PR will not catch `SyntaxError` such as missing paranthesis.

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -353,9 +355,12 @@ def abort_test():
     try:
         best, fbest = fitdriver.fit()
         errors = []
+        stack_trace = None
     except Exception as exc:
         best, fbest = None, np.NaN
-        errors = [str(exc), traceback.format_exc()]
+        exc_type, exc_value, exc_tb = sys.exc_info()
```
rozyczko, line 361: why use `sys.exc_info` when you already have all the information in `traceback.format_exc`?

src/sas/sascalc/fit/expression.py
```diff
@@ -212,7 +212,7 @@ def eval_expressions():
     # CRUFT: python < 3.0;  doc builder isn't allowing the following exec
     # https://stackoverflow.com/questions/4484872/why-doesnt-exec-work-in-a-function-with-a-subfunction/41368813#comment73790496_41368813
     #exec(functiondef, global_context, local_context)
-    eval(compile(functiondef, '<string>', 'exec'), global_context, local_context)
+    eval(compile(functiondef, functiondef, 'exec'), global_context, local_context)
```
rozyczko, line 215: `<string>` is the default argument for when the code is not taken from a file, as the docs describe
https://docs.python.org/3/library/functions.html#compile
Is there a reason to use `functiondef` here?

m2cci-NMZ: Since the code of the function is not in a file, setting `filename` to `functiondef` allows to get the code of the function via the `f_code.co_filename` attribute of the traceback object, which is then analzed to retrieve the faulty constraint.

src/sas/sascalc/fit/BumpsFitting.py
m2cci-NMZ: `traceback.format_exc()` returns a string, whereas sys.exc_info() returns a traceback object, so instead of parsing a string, the traceback methods and attributes can be used in order to get information on failing code.

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -493,12 +493,46 @@ def fitComplete(self, result):
...
+                # warn the user of which constraint is failing and highlight the error
+                error_text = 'Fit failed because constraint <b>%s</b> is inconsistent:<br>%s'%(
+                    constraint_no + 1,
+                    self.tblConstraints.item(constraint_no, 0).text().replace(name_error, "<b>"+name_error+"</b>"))
+            else:
+                error_text = 'Fit failed'
+            QtWidgets.QMessageBox.warning(self,
+                                          'Warning',
+                                          error_text,
+                                          QtWidgets.QMessageBox.Ok)
```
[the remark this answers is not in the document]
m2cci-NMZ: Good point. Not a very good idea to change the workflow. I believe that batch fitting would cause a separate box to pop up for each fit, which is a problem. I changed the behaviour so it show the message in the status bar

pkienzle:
Parsing the stack trace is liable to be brittle.  It will be worse when sascalc is broken out into a separate package with a different release schedule. It would be better to send valid constraints to the fitting engine rather than trying to parse and interpret the errors that are returned.  Best would be to check the expression when it is entered into the GUI so the user has immediate feedback. This won't catch all the problems, so still need to catch the errors from the fit.

In the process of investigating expression.py I noticed that the constraints are not properly ordered and cycles are not properly identified.  Even though the 5.x GUI is hobbled for constraints (you can't directly assign a derived parameter to another parameter for example), this is easy to defeat by entering an expression:
```python
M1.sld = M1.sld_solvent + 2
M1.sld_solvent = M1.sld + 4
```

I have some code which detects name and cycle errors (diff attached).  I can post it to this PR but you will still need to do some work to integrate it with the GUI.  For example, a direct call to check the syntax that happens in the GUI thread rather than going through the fit thread.  We could define our own exception classes which can list the parameters that are unknown rather than passing them a string that needs to be parsed if we a more minimal change. Let me know.

[update.txt](https://github.com/SasView/sasview/files/4931827/update.txt)

m2cci-NMZ:
At some point I was thinking about writing a function that removes recursively all known symbols from a constraint and have the constraint validated whenever it returned an empty string, but I thought this would create  more problems that it would solve... I could try and give it a shot though

pkienzle:
After checking for syntax errors using compile() and making sure all symbols are known you can do the final step of running the compiled function.  This will cover more cases than you can reasonably handle with static analysis.

The function call might fail due to divide by zero or similar, depending on the parameter values.  This may not be a problem— the fit ought to reject those parameter values and move on to a feasible region. I guess displaying the error to the user the same way we report unknown symbols and circular references is the best we can do. We could use this to force them to start the fit at a feasible point.

It is worth checking that the fit engine does handle constraints which raise an error, and patch bumps if it doesn't work properly.

BTW,  I modified my patch to allow python keywords `is`, `and`, `or`, `not`, etc. in expressions:
```python
from keyword import iskeyword
...
def _check_free_variables(target, expr, symbol_table):
    undefined = [sym for sym in _symbols(expr)
                 if sym not in symbol_table and not iskeyword(sym)]
    ...
```

pkienzle:
I updated expression.py which has the function `check_constraints` that returns a list of errors if any (see attached).  To call this you will need to walk the models and extract all the fitting parameters, putting them into a symbol table with their current values.

Something like:
```python
symtab = {}
exprs = {}
for model in models:
   for parameter in model.parameters:
        symtab[f"{model.name}.{parameter.name}"] = parameter.value
duplicates = []
for parameter_name, expression in constraints:
    if parameter_name in exprs:
        duplicates.append(parameter_name)
    exprs[parameter_name] = expressions
errors = sas.sascalc.fit.expressions.check_expressions(symtab, exprs)
if duplicates:
    errors.append("Duplicate parameter definitions for "+", ".join(duplicates))
if errors:
    display(errors)
```

Similar logic is already occurring in sascalc/fit/BumpsFitting.  It would be good to refactor it so that the same code can be used in the GUI and in sascalc.

[expression.txt](https://github.com/SasView/sasview/files/4933804/expression.txt)

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -550,6 +576,35 @@ def batchComplete(self, result):
...
+        ConstraintSyntaxError: syntax problem such as missing parentheses
+        """
+        assert(isinstance(exception, Exception))
+        if isinstance(exception, CyclicDefinitionError):
+            error_message = "Cyclic dependency with parameter: <b>%s</b>"%(exception.parameter)
+
+        elif isinstance(exception, UnknownParameterNameError):
+            constraint = exception.constraint
+            for symbol in exception.unknown_symbol:
+                constraint = constraint.replace(symbol, "<b>%s</b>" % (symbol))
```
pkienzle, line 594: `unknown_symbol` is a string containing a comma separated list of symbols, so this loops over the individual characters.

`replace` does not respect word boundaries, so if the unknown symbol is 'radius', `replace` will replace it inside 'M1.radius_polar'.

You can use `re.sub` instead.  Unfortunately, you cannot use the `\b` word boundary marker since symbols can contain periods in addition to alphanumeric characters, so you need to use the more awkward:
```python
# Identify the symbol for replacement as everything between a word boundary. Since symbols
# can contain '.', we need to use negative lookbehind and negative lookahead on a character set
# containing '.' to check for the boundaries.  Also, if there is a '.' in the symbol it could match any
# character, so it needs to be turned into a regular expression that matches '.'.
pattern = f"(?<![a-zA-Z0-9_.]){symbol.replace('.', '[.]')}(?![a-zA-Z0-9_.])"
re.sub(pattern, f"<b>{symbol}</b>")
```

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -550,6 +576,35 @@ def batchComplete(self, result):
...
+            error_message = "Unknown symbol: %s" % (constraint)
+
+        elif isinstance(exception, ConstraintSyntaxError):
+            constraint = exception.constraint
+            syntax_problem = exception.parameter
+            highlighted_syntax = syntax_problem.replace(syntax_problem[exception.offset-1], "<b>%s</b>" % (syntax_problem[exception.offset-1]))
```
pkienzle, line 600: If the error is at the letter 'a' for example, then this will bold-face every 'a'. If the error is at a blank space, then nothing will appear boldfaced.  Maybe the following, which highlights everything after the offset?
```python
highlighted_syntax = syntax_problem[:exception.offset]+"<b>"+syntax_problem[excpetion.offset:]+"</b>"
```

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -550,6 +576,35 @@ def batchComplete(self, result):
...
+        """
+        Returns an error string highlighting the error in the constraint.
+        The argument is an exception that was raised by expression.py, which are interpreted as following:
+        CyclicDefinitionError: parameter has cyclic dependencies
+        UnknownParameterNameError: unknown symbol found in constraint
+        ConstraintSyntaxError: syntax problem such as missing parentheses
+        """
+        assert(isinstance(exception, Exception))
+        if isinstance(exception, CyclicDefinitionError):
+            error_message = "Cyclic dependency with parameter: <b>%s</b>"%(exception.parameter)
```
pkienzle, line 589: Cyclic dependencies can have multiple parameters.

pkienzle:
BTW, you don't need to go to extraordinary lengths to parse and markup errors returned from fitting since all the errors that you know about should have been reported before the fit started.  Still need to return and display them since there may be errors that you don't know about.

src/sas/sascalc/fit/expression.py
```diff
@@ -171,48 +197,141 @@ def compile_constraints(symtab, exprs, context={}):
     Use help(fn) to see the code generated for the returned function fn.
     dis.dis(fn) will show the corresponding python vm instructions.
     """
+    retfn, errors = _compile_constraints(symtab, exprs, context=context)
+    if errors:
+        raise SyntaxError(error for error in errors)
```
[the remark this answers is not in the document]
m2cci-NMZ: I raised a `RuntimeError`, with a list containing the errors as argument (I had forgotten the brackets)

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -4311,3 +4311,14 @@ def getCurrentFitState(self, state=None):
...
+
+    def getSymbolDict(self):
+        """
+        Return a dict containing a list of all the symbols used for fitting and their values,
+        e.g. {'M1.scale':1, 'M1.background': 0.001}
+        """
+        sym_dict = {}
+        model_name = self.kernel_module.name
+        for param in self.getParamNames():
+            sym_dict[f"{model_name}.{param}"] = self._model_model.item(self.getRowFromName(param), 1).text()
```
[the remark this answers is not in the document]
m2cci-NMZ: I forgot to cast the string into a double. This was creating problems when using operators that dont work on strings, such as multiplication

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
```diff
@@ -889,3 +891,15 @@ def isParamPolydisperse(param_name, kernel_params, is2D=False):
...
+def compileConstraints(symtab, constraints, check_cyclic = False):
+    exprs = {}
+    duplicates = []
+    for parameter_name, expression in constraints:
+        if parameter_name in exprs:
+            duplicates.append(parameter_name)
+        exprs[parameter_name] = expression
+    errors = check_constraints(symtab, exprs)
+    if duplicates and check_cyclic:
+        errors.append("Duplicate parameter definitions for " + ", ".join(duplicates))
```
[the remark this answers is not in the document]
m2cci-NMZ: Function has been renamed. Duplicate definitions should not be a problem since the data structure of the model does not allow them

m2cci-NMZ:
As suggested, the user-defined exceptions were removed. An html flag was added to `check_constraints` so it would return an html string, which is then wrapped up and returned to the controller that displays it. If the user enters an inconsistent constraint, a messagebox showing all the errors will pop, interrupting flow. Failing fits will show a message in the status bar.

src/sas/sascalc/fit/expression.py
```diff
@@ -44,40 +44,79 @@ def model_update():
...
+def _check_syntax(target, expr, html = False):
+    try:
+        compile(expr, expr, "exec")
+    except SyntaxError as exc:
+        if html:
+            constraint = "% s = % s" % (target, expr)
+            text = exc.text.strip("\n")
+            highlighted_text = text[:exc.offset-1] + "<b>" + \
+                               text[exc.offset-1:] + "</b>"
+            constraint = constraint.replace(text, highlighted_text)
```
pkienzle, line 70: This is confusing since it is not obvious that `text` is a substring of `expr`.  Some comments are in order.

If the error is a missing close parenthesis, then it will show up after the end of text, and no error will be indicated.

Multiline syntax will not be formatted correctly without `<pre>`.

The `% s` syntax in the format string is valid but unusual. Should be just `%s`.

So maybe:
```python
if "\n" in expr:
    # Multiline expression. Be lazy and just show line, col for syntax error since this should never happen.
    return [f"Syntax error on line {exc.lineno} column {exc.offset} for {target}:\n<pre>\n{expr}\n</pre>"]

if exc.offset > len(expr):
     # Single line expression with error after the expression. Probably missing a closing paren, but it could
     # also be that the expression ends with an operator such as "3+4+"
    return [f"Expression `{target} = {expr}` is not complete."]

# Single line expression. Wrap the tail of the expr after the syntax error in <b>...</b>.
# exc.lineno=1, exc.text = expr+"\n", and exc.offset = location of the syntax error in expr.
return [f"Syntax error in expression '{target} = {expr[:exc.offset-1]}<b>{expr[exc.offset-1:]}</b>'"]
```

m2cci-NMZ:
> Testing within the GUI I was able to add constraints with errors on the constraints tab without any warnings.
> 
> When running the syntax error appeared on the terminal.
> 
> A red "Stop Fit" button was available even though the fit was already aborted. On clicking "Stop Fit" the program hung with the Mac beach ball spinning.

Could this come from some lock that is never released? I guess that clicking on "Stop fit" is in fact launching another fit that is waiting for the lock.
The `RuntimeException` raised by `compile_constraints` is never caught by sascalc, the fit thread catches it and outputs it to the log. 
The `ParameterExpression` call should be sourrounded by `try..except` blocks in order to avoid any problems.

pkienzle:
With a little help I've managed to set up constraints on a fit tab. Same issue: incorrect constraints do not raise an error before the fit.

Is there a reason why the GUI uses model:parameter but the constraint expressions use model.parameter? Seems like that might cause some user confusion.

m2cci-NMZ:
What is the incorrect constraint you are setting up? I've had some problems with colons or commas, or any other valid symbol, as long as the expression is syntaxically correct.



> Is there a reason why the GUI uses model:parameter but the constraint expressions use model.parameter? Seems like that might cause some user confusion.

I don't know if there is a particular reason for this.

pkienzle:
Here are a couple of nonsense constraints: `M1.scale = M1.scale` and `M1.sld = M1.radius +`.  Neither of them triggered an error before submitting to fit. I set them in the fit tab by editing the constraint on the parameter rather than the constraints tab.

m2cci-NMZ:
The error check is now perfomed when the constraint is added to the model, so adding incorrect constraints directly in the fit tab should display errors. `try...except` blocks were also added so that exceptions raised by `ParameterExpression` can be caught and displayed by the GUI

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -447,3 +449,43 @@ def currentTab(self):
...
+        Returns a list of the constraints for all fitting tabs. Constraints
+        are a tuple of strings (parameter, expression) e.g. ('M1.scale',
+        'M2.scale + 2')
+        """
+        constraints = []
+        for tab in self.getFitTabs():
+            tab_name = tab.modelName()
+            tab_constraints = tab.getConstraintsForModel()
+            constraints.extend((tab_name + "." + par, expr) for par, expr in
+                               tab_constraints)
```
pkienzle, line 469: Prefer line breaks at higher level boundaries:
```
constraints.extend((tab_name + "." + par, expr)
                                for par, expr in tab_constraints)
```

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -447,3 +449,43 @@ def currentTab(self):
...
+            symbol_dict.update(tab.getSymbolDict())
+        return symbol_dict
+
+    def getConstraintTab(self):
+        """
+        Returns the constraint tab, or None if no constraint tab is active
+        """
+        if any(isinstance(tab, ConstraintWidget) for tab in self.tabs):
+            constraint_tab = next(tab for tab in self.tabs if isinstance(tab,
+                                                                     ConstraintWidget))
```
pkienzle, line 488: Break at higher level:
```
constraint_tab = next(
    tab for tab in self.tabs if isinstance(tab, ConstraintWidget))
```
or
```
constraint_tab = next(tab
                      for tab in self.tabs
                      if isinstance(tab, ConstraintWidget))
```

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -791,13 +791,36 @@ def modifyViewOnRow(self, row, font=None, brush=None):
...
+        constraint_list.append((self.modelName() + '.' + constraint.param,
+                                constraint.func))
+        # Call the error checking function
+        errors = FittingUtilities.checkConstraints(symbol_dict, constraint_list)
+        if errors:
+            # Display the message box
+            QtWidgets.QMessageBox.critical(self,
+                                           "Inconsistent constraint",
+                                           errors,
+                                           QtWidgets.QMessageBox.Ok)
```
pkienzle, line 816: Maybe break after paren and indent by 4.

src/sas/qtgui/Perspectives/Fitting/UnitTesting/ConstraintWidgetTest.py
```diff
@@ -82,38 +84,233 @@ def testOnFitTypeChange(self):
...
+        self.widget.updateFitLine("test_tab")
+        # We should have 2 tabs in the model tab
+        self.assertEqual(self.widget.tblTabList.rowCount(), 2)
+        # One constraint in the constraint tab
+        self.assertEqual(self.widget.tblConstraints.rowCount(), 1)
+        # Constraint should be active
+        self.assertEqual(self.widget.tblConstraints.item(0, 0).checkState(), 2)
+        # Check the text
+        self.assertEqual(self.widget.tblConstraints.item(0, 0).text(),
+                         test_tab.kernel_module.name +
```
pkienzle, line 163: PEP8: prefer operator on the start of the next line... this makes it easier to see that it is not an additional function argument.

src/sas/qtgui/Perspectives/Fitting/UnitTesting/ConstraintWidgetTest.py
```diff
@@ -82,38 +84,233 @@ def testOnFitTypeChange(self):
...
+                         ":" +
+                         "scale" +
+                         " = " +
```
pkienzle, line 166: Why not `+ ":scale = "`?

m2cci-NMZ:
Thanks! Formatting was corrected, and unit tests were added to get coverage on the new methods.

PR #1635, opened 2020-07-20
Allow .DAT reader to read rectangular data sets (wxPython)

krzywon:
This is a parallel PR for #1625 for py2.7.

The branch for the PR is built off branch wx_reorg_tests which is based off wx_loader_error_handling. The PRs for all of these branches are chained off one another.


src/sas/sascalc/dataloader/readers/red2d_reader.py
```diff
@@ -236,30 +236,19 @@ def get_file_contents(self):
...
+        x_bins = np.unique(qx_data)
+        y_bins = np.unique(qy_data)
+        # For non-uniform Qx and Qy arrays
+        if round(len(x_bins) * len(y_bins) / len(qx_data)) >= 2:
+            x_bins = [qx_data[i] for i in takewhile(
+                lambda i: qx_data[i] < qx_data[i + 1], range(len(qx_data)))]
+            # Above list comprehension rejects last item
+            x_bins.append((qx_data[len(x_bins)]))
+            qy = np.reshape(qy_data, (len(qx_data)//len(x_bins), len(x_bins)))
+            y_bins = np.transpose(qy)[0].tolist()
```
pkienzle, line 249: This code would benefit from some comments describing the data structure that it is operating over and what it is doing.

Nearest I can tell it is assuming that qx and qy are rectilinear row-major arrays, with qx increasing along the rows and qy increasing along the columns. It is using the first row of qx and first column of qy as the bin centers for the entire array.  So you ought to be able to use:
```python
x_bins = qx_data.reshape(num_qy, num_qx)[0, :]
y_bins = qy_data.reshape(num_qy, num_qx)[:, 0]
```
The problems is that you do not know how many qx and qy.

Using `np.unique` is a little risky since the data writer may have introduced spurious differences in the stored values for qx and qy, even though they are nominally the same.  It may be that you end up with a situation where the size of `x_bins * y_bins` does not match the array.  I'm guessing this won't cause runtime errors since sasview is only using x_bins and y_bins as suggested rectilinear bin centers, and will rebin the data onto those centers later, but it does mean that you may end up with some extremely narrow rows/columns.

It took some time to understand what takewhile is doing.

How about you instead set the number of x_bins as the first k such that `qx_data[k] < qx_data[k-1]`, then take the first row:
```
# Assume qx_data and qy_data are flattened rectilinear arrays in row major order with qx_data increasing
# along each row. Set x_bins to the first row of qx and y_bins to the first column of qy_data. You can
# find the end of each row by comparing qx_data[k] < qx_data[k-1].  With argmax you can find
# the index of the first true value, and add one to convert to row length.  Set the last element to True in
# case there is only one row.
num_qx = np.argmax(np.hstack((qx_data[1:] < qx_data[:-1], True)) + 1
num_qy = len(qy_data) // num_qx
assert num_qx * num_qy == len(qy_data), "q data is not rectilinear"
x_bins = qx_data.reshape(num_qy, num_qx)[0, :]
y_bins = qy_data.reshape(num_qy, num_qx)[:, 0]
```

This code will fail if you have a single column qy != 0 with qx values computed for a flat detector, or due to numerical jitter on a spherical detector leading to slightly different qx values in each bin.

If add the additional assumption that `num_qx > 1` and `num_qy > 1` you can remove the `hstack`.

src/sas/sascalc/dataloader/data_info.py
```diff
@@ -37,11 +41,11 @@ class plottable_1D(object):
     y = None
     dx = None
     dy = None
-    ## Slit smearing length
+    # Slit smearing length
```
pkienzle, line 44: Autodoc can recognize lines starting with #: as docstrings for the class attributes that follow.

See for example the ModelInfo [definition](http://www.sasview.org/docs/_modules/sasmodels/modelinfo.html#ModelInfo) and [docs](http://www.sasview.org/docs/dev/sasmodels-api/sasmodels.html#model-info-and-parameter-tables).

krzywon: That's new to me! I'll add this to another of my upcoming updates.

src/sas/sascalc/dataloader/readers/red2d_reader.py
pkienzle: Just to be clear, I was proposing that you replace the entire block, from line 240-252, including the np.unique and the conditional test. This will be more robust than using np.unique. It will also be faster since it won't need to do a sort, but that is less important.

src/sas/sascalc/dataloader/readers/red2d_reader.py
```diff
@@ -292,22 +284,18 @@ def get_file_contents(self):
...
-                                                 + (dqy_data * sin_th) * \
-                                                  (dqy_data * sin_th))
-                    self.current_dataset.dqy_data = np.sqrt((dqx_data * sin_th) * \
-                                                 (dqx_data * sin_th) \
-                                                 + (dqy_data * cos_th) * \
-                                                  (dqy_data * cos_th))
+                    self.current_dataset.dqx_data = np.sqrt(
+                        (dqx_data * cos_th)**2 + (dqy_data * sin_th)**2)
+                    self.current_dataset.dqy_data = np.sqrt(
+                        (dqx_data * sin_th)**2 + (dqy_data * cos_th)**2)
```
pkienzle, line 298: `np.hypot(a, b)` computes `sqrt(a**2 + b**2)` element-wise. Apparently this is a C99 math function, so it should be pretty common everywhere.

PR #1658, opened 2020-08-10
Load constraints after fit pages when loading project

### LINKED ISSUE #1653: 5.0.4 new constraints checks over zealous on load project
In my local build of v5, constrained projects, such as the examples attached here https://github.com/SasView/sasview/issues/1645 do not now load properly, as the checks are being made on Fitpage1 before pages 2 & 3 are loaded, so params mentioned in the constraints do not exist yet, a couple of error boxes pop up, thus the final constraints list only has M3:param = M2 or M1 params but nothing else.

m2cci-NMZ:
Fixes #1653 
This PR aims at having constraints applied when all fit pages are loaded when loading a project, instead of having them applied immediately after each fit page loading. This was causing problems as the GUI would complain that there where unknown symbols in the constraint if they were referring to parameters of a yet to be loaded fit page.

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -519,14 +519,14 @@ class DataExplorerWindow(DroppableDataLoadWidget):
                 visible_perspective = value
             # send newly created items to the perspective
             self.updatePerspectiveWithProperties(key, value)
-        constraint_dict = self.readConstraintsFromProject(all_data)
-        self._perspective().updateFromConstraints(constraint_dict)
         # Set to fitting perspective and load in Batch and C&S Pages
         self.cbFitting.setCurrentIndex(
             self.cbFitting.findText(DEFAULT_PERSPECTIVE))
         # See if there are any batch pages defined and create them, if so
         self.updateWithBatchPages(all_data)
-
+        # Get the constraint dict and apply it
+        constraint_dict = GuiUtils.getConstraints(all_data)
+        self._perspective().updateFromConstraints(constraint_dict)
         # Only now can we create/assign C&S pages.
         for key in cs_keys:
             self.updatePerspectiveWithProperties(key, all_data[key])
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -535,47 +535,6 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         self.cbFitting.setCurrentIndex(
                 self.cbFitting.findText(visible_perspective))
 
-    def readConstraintsFromProject(self, all_data):
-        """
-        Extracts constraints from *all_data* dict and returns a dict where keys
-        are the tab name and values are a list of constraints on that tab. The
-        dict can then be passed to the updateFromConstraints method from the
-        fitting perspective to apply the constraints with error checking
-        mechanism
-        """
-        constraint_dict = {}
-        for key, value in all_data.items():
-            # make sure we dealing with params
-            if 'fit_params' not in value:
-                continue
-            params = value['fit_params']
-            for page in params:
-                if not isinstance(page, dict):
-                    continue
-
-                tab_name = None
-                constraint_list = []
-                for param_name, param_value in page.items():
-                    # make sure we are dealing with lists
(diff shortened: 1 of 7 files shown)


src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -534,6 +535,47 @@ def readProject(self, filename):
...
+                        cons = param_value[4+ioffset]
+                        # get the constraint
+                        if cons is not None and len(cons) == 5:
+                            constraint_list.append(cons)
+                    else:
+                        continue
+
+                if tab_name and constraint_list:
+                    constraint_dict.update({tab_name: constraint_list})
+        return constraint_dict
```
pkienzle, line 577: The `readConstraintsFromProject` function belongs with the data reader (currently in `GuiUtil`) since it will also be useful when running fits headless. It does not need to be a method in this class since it makes no reference to `self`. 

It may make sense to have it as a method on the project data structure, but this is a style preference: `data.get_constraints()` vs. `get_constraints(data)`. Don't make it an attribute like `data.constraints`: it does more work than people expect when accessing an attribute, and it won't be used frequently enough to justify writing cacheing code to make it efficient. Don't call it `read` since it isn't actually reading the file.

You may want to move some of the checking directly into the project reader so that the data structures are clean and this code can be made simpler.  In general cleaning up data on input saves trouble in the rest of the code (for example, I'm guessing that the length greater than 5 test on line 567 shows up elsewhere), but if you hide the internals through method calls like `get_constraints()` there won't be much difference.

There may be some advantage to collecting and reporting errors and warnings on load before the user does too much work with the loaded data, but this particular code doesn't do any error reporting (whether it should be reporting errors is a separate question).

pkienzle: ... make that `fit_project.get_constraints()` or `get_constraints(fit_project)` since a fit project may contain multiple data sets.

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -519,7 +519,8 @@ def readProject(self, filename):
                 visible_perspective = value
             # send newly created items to the perspective
             self.updatePerspectiveWithProperties(key, value)
-
+        constraint_dict = self.readConstraintsFromProject(all_data)
+        self._perspective().updateFromConstraints(constraint_dict)
```
rozyczko, line 523: `updateFromConstraints` is not available for other perspectives. Please move these 2 lines it a few lines below, after the explicit change to Fitting perspective.

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -187,6 +188,33 @@ def updateFromParameters(self, parameters):
...
+                    param = constraint_param[1]
+                    value_ex = constraint_param[2]
+                    validate = constraint_param[3]
+                    function = constraint_param[4]
+                    constraint = Constraint()
+                    constraint.value = value
+                    constraint.func = function
+                    constraint.param = param
+                    constraint.value_ex = value_ex
+                    constraint.validate = validate
```
rozyczko, line 213: we can probably leave out creating intermediate variables and assign list values directly to constraint attributes
`constraint.value = constraint_param[0]` etc.

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -525,3 +553,15 @@ def getConstraintTab(self):
...
+        """
+        Returns the tab with with attribute name *name*
+        """
+        assert(name, str)
+        for tab in self.tabs:
+            if tab.modelName() == name:
+                return tab
+            else:
+                continue
+        return None
```
rozyczko, line 567: Add empty line

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -525,3 +553,15 @@ def getConstraintTab(self):
...
+    def getTabByName(self, name):
+        """
+        Returns the tab with with attribute name *name*
+        """
+        assert(name, str)
+        for tab in self.tabs:
+            if tab.modelName() == name:
+                return tab
+            else:
+                continue
```
rozyczko, line 566: redundant `else`

Issue #1467, opened 2020-02-19
better handling of qt ui generator in setup.py


pkienzle:
`python setup.py --help` triggers the Qt UI generator.  Instead the UI generator should only be triggered when needed.

Maybe something like [setup_qt](https://github.com/xmikos/setup_qt) or [pyqt_distutils](https://github.com/ColinDuquesnoy/pyqt_distutils) would help?  Don't know if either will automatically trigger as part of the build if you don't specifically ask for them.

See also PR #1398.

rozyczko:
I would prefer just spawning a subprocess for UI conversion script. This allows me to be agnostic wrt. underlying conversion command (e.g. PySide vs PyQt, PyQt4.x vs PyQt5.x etc) and allows me to run the conversion from the command line if I choose to.
Instead, I can look at the setup.py arguments and choose the ones which should not trigger the UI build.

pkienzle:
Both packages support the various backends, but it looks like setup_qt is a bit better match to your needs: it lets you override the configuration options on the build command.

I considered doing what you proposed and triggering on the basis of setup.py options, but decided that there were too many of them, and we should try an existing package before rolling our own.

lucas-wilkins:
Can possibly roll this into #2140

lucas-wilkins:
Looks like its actually dealt with, closing. @rozyczko can reopen if I'm wrong

PR #1661, opened 2020-08-13
Minor improvements to setup.py

rozyczko:
Addresses #1467 by specifying command arguments which should not trigger the UI build.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -109,7 +109,8 @@ class BuildSphinxCommand(Command):
 
 
 # _standard_ commands which should trigger the Qt build
-build_commands = ['build', 'build_py', 'develop', 'test']
+build_commands = ['install', 'build', 'build_py', 'bdist', 'bdist_rpm',
+                  'bdist_wheel', 'develop', 'test']
 # determine if this run requires building of Qt GUI ui->py
 build_qt = any(c in sys.argv for c in build_commands)
 
--- a/setup.py
+++ b/setup.py
@@ -107,15 +107,11 @@ class BuildSphinxCommand(Command):
         import build_sphinx
         build_sphinx.rebuild()
 
-# determine if this run requires building of Qt GUI ui->py
-build_qt = True
-# _standard_ commands which shouldn't trigger the Qt build
-no_build_commands = ['clean', 'sdist', 'bdist', 'check', 'upload', 'register', '--help']
 
-for c in no_build_commands:
-    if any([c in b for b in sys.argv]):
-        build_qt = False
-        break
+# _standard_ commands which should trigger the Qt build
+build_commands = ['build', 'build_py', 'develop', 'test']
+# determine if this run requires building of Qt GUI ui->py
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -107,7 +107,17 @@ def run(self):
         import build_sphinx
         build_sphinx.rebuild()
 
-if 'clean' not in sys.argv:
+# determine if this run requires building of Qt GUI ui->py
+build_qt = True
+# _standard_ commands which shouldn't trigger the Qt build
+no_build_commands = ['clean', 'sdist', 'bdist', 'check', 'upload', 'register', '--help']
```
pkienzle, line 113: Use a whitelist rather than a blacklist.  First, there are a lot of commands available to setup that you have not covered (e.g., --author). Second, you can issue multiple commands in the same line, such as `python setup.py clean build`.

setup.py
```diff
@@ -107,7 +107,17 @@ def run(self):
...
+
+for c in no_build_commands:
+    if any([c in b for b in sys.argv]):
+        build_qt = False
+        break
```
pkienzle, line 118: Do you need substring matching for your checks?

With a whitelist and no substring matching this could be simplified to:
```
build_qt = any(c in sys.argv for c in build_commands)
```

If you do need substring checks, probably better to force them to the beginning of the string:
```
is_build_command = lambda arg: any(arg.startswith(cmd) for cmd in build_commands)
build_qt = any(is_build_command(arg) for arg in sys.argv)
```
or even switch to re checks if you need more control.

setup.py
```diff
@@ -107,7 +107,17 @@ def run(self):
...
+
+if build_qt:
     _ = subprocess.call([sys.executable, "src/sas/qtgui/convertUI.py"])
```
pkienzle, line 121: Not sure why there is the spurious `_ = exp` here.

rozyczko: > Not sure why there is the spurious `_ = exp` here.

`subprocess.call` returns a `returncode` and I want to be explicit that I'm ignoring it.

setup.py
rozyczko: I don't need substrings. The only whitelisted commands will be easily spelled out.

pkienzle: On reflection the whitelist question may be moot: if you do `python setup.py clean build` then setup.py will call convertUI then process the clean and build. If clean removes the UI then this is a bug, but I don't think we care that much.

setup.py
```diff
@@ -107,7 +107,13 @@ def run(self):
         import build_sphinx
         build_sphinx.rebuild()
 
-if 'clean' not in sys.argv:
+
+# _standard_ commands which should trigger the Qt build
+build_commands = ['build', 'build_py', 'develop', 'test']
```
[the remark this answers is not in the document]
rozyczko: I thought `install` just installs whatever is in build/ but apparently it does build the application if necessary. TIL.
Added some more commands.

PR #1667, opened 2020-09-01
Disable rather than remove constraint when fitting a single fit page

### LINKED ISSUE #1654: 5.0.4 disable rather than remove constraints if do a fit on a single FitPage
In local build of ESS_GUI. set up constraints between FitPages (e.g. use the examples attached to https://github.com/SasView/sasview/issues/1645 )

If you hit fit on a single page that has constraints, a box pops up asking either to cancel the fit or remove the constraints.

Is there a good reason why the constraints have to be removed rather than simply disabled in the constraints tab?
(I know that I did mention this previously, but we do like to run a few fits on a single FitPage,then go back to the constrained version, and it is annoying to have to re-create all the constraints.)

m2cci-NMZ:
Fixes #1654 : when fitting a single fit page with a constraint relying on another fit page, the constraint will just be disabled instead of being removed. The corresponding line in the constraint widget will be unchecked, so one can recheck if needed.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
@@ -1055,7 +1055,7 @@ class ConstraintWidget(QtWidgets.QWidget, Ui_ConstraintWidgetUI):
     def uncheckConstraint(self, name):
         """
         Unchecks the constraint in tblConstraint with *name* slave
-        parameter.
+        parameter and deactivates the constraint.
         """
         for row in range(self.tblConstraints.rowCount()):
             constraint = self.tblConstraints.item(row, 0).data(0)
@@ -1064,3 +1064,7 @@ class ConstraintWidget(QtWidgets.QWidget, Ui_ConstraintWidgetUI):
             # string.
             if name in constraint:
                 self.tblConstraints.item(row, 0).setCheckState(0)
+                # deactivate the constraint
+                tab = self.parent.getTabByName(name[:name.index(":")])
+                row = tab.getRowFromName(name[name.index(":") + 1:])
+                tab.getConstraintForRow(row).active = False
--- a/src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
@@ -1060,7 +1060,7 @@ class ConstraintWidget(QtWidgets.QWidget, Ui_ConstraintWidgetUI):
         for row in range(self.tblConstraints.rowCount()):
             constraint = self.tblConstraints.item(row, 0).data(0)
             # slave parameter has model name and parameter separated
-            # by semicolon e.g `M1:scale` so no need to parse the constraint
+            # by colon e.g `M1:scale` so no need to parse the constraint
             # string.
             if name in constraint:
                 self.tblConstraints.item(row, 0).setCheckState(0)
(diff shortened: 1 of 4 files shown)


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1131,13 +1131,14 @@ def getConstraintsForFitting(self):
             msg = "The current fit contains constraints relying on other fit pages.\n"
             msg += "Parameters with those constraints are:\n" +\
                 '\n'.join([cons[0] for cons in multi_constraints])
-            msg += "\n\nWould you like to remove these constraints or cancel fitting?"
+            msg += "\n\nWould you like to deactivate these constraints or " \
+                   "cancel fitting?"
```
pkienzle, line 1135: [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) recommends parentheses instead of line continuation character for breaking expressions over multiple lines..

RichardHeenan (commented):
Tested the code from this branch, behaves as expected, so I could approve this.
However, there is another "feature" to improve please. When I re-activate the previously turned off constraint, the constrained parameter in the model no longer has its check box "on", thus defeating our new convention that constrained parameters must be "checked as adjusting".  Fixing this could be a new ticket, but perhaps can be done here?

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -1051,3 +1051,16 @@ def getReport(self):
         msgbox.setWindowTitle("Fit Report")
         _ = msgbox.exec_()
         return
+
+    def uncheckConstraint(self, name):
```
rozyczko, line 1055: This should probably also include changing the `.active` attribute of corresponding Constraint object.
Otherwise, calls like `rowHasActiveConstraint` and `rowHasActiveComplexConstraint` will behave incorrectly.

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1146,9 +1147,15 @@ def getConstraintsForFitting(self):
...
+                constraint_tab = self.parent.perspective().getConstraintTab()
                 for cons in multi_constraints:
-                    self.deleteConstraintOnParameter(param=cons[0])
+                    # deactivate the constraint
+                    row = self.getRowFromName(cons[0])
+                    self.getConstraintForRow(row).active = False
+                    # uncheck in the constraint tab
+                    if constraint_tab:
+                        constraint_tab.uncheckConstraint(
+                            self.kernel_module.name + ':' + cons[0])
```
rozyczko, line 1158: As above, this should also change the constraint object attribute

rozyczko (changes requested):
Minor changes requested

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
m2cci-NMZ: Now the method changes the constraint attribute, so it can be used safely.

m2cci-NMZ:
> However, there is another "feature" to improve please. When I re-activate the previously turned off constraint, the constrained parameter in the model no longer has its check box "on", thus defeating our new convention that constrained parameters must be "checked as adjusting". Fixing this could be a new ticket, but perhaps can be done here?

I'm working on it, should be in an upcoming PR :-)

rozyczko (approved):

Issue #1542, opened 2020-05-17
Crosstalk between Corfunc and Invariant perspectives


butlerpd:
As of 5.0.2, if one sends data to corfunc and plots it, then sends the same data to the invariant all behaves as expected.  However, if in fact the data is sent to Corfunc and the correlation function plotted (hitting `extrapolate` and then `transform`) so that the second plot (the `Real space data`) before then sending the same data set to the invariant, there are two things wrong:
1. the axis labels have now changed from I(Q) vs Q to gamma vs x(A)
2. The plot not only contains the I(Q) data but also 3 lines whose labels suggest they are the `1D Correlation` the `3D Correlation` and the `Interface Distribution Function` which now, as described in issue #1541 get replicated in every plot every time one updates and recomputed the invariant.

NOTE: This can be fixed by sending the data to some other pesrpective and then sending back to the Invariant.

Seen on both Windows  and Mac

smk78:
Quoting @butlerpd above:
>NOTE: This can be fixed by sending the data to some other pesrpective and then sending back to the Invariant.

Actually, not quite. If you need to use the _Append to_ button you will find _every dataset created in your session gets appended_ to the graph!

butlerpd:
? I think that is another bug @smk78  ... or feature request some might say, which I forgot to enter but was meaning to. Or did I misunderstand your comment?  basically, if you try to plot any dataset that has been used for fitting of various kinds you have to tick the box then create (or append) to plot will plot that data set and ALL the children of that dataset.  That is because when you tick that box the default is to ALSO tick the **hidden** boxes of all the child plots.
Another problem I have is that the number of child plots becomes very unwieldy and not sure we need them all there but that is probably a much bigger and deeper design/logic issue

I will try to create this issue soon. is this the issue? or is it really the cross talk specifically between invariant and corfunc?

smk78:
No it's a general issue. If you use, say, corfunc, then invariant, and then fitting, and then find you want to append to the last graph window, _everything_ SasView seems to know about gets appended, I(q) data, models, correlation functions, IDF's,...

butlerpd:
Good -- that is the other known issue but independent of this one.  I will make it now.  This issue is only a problem with invariant and corfunc and related to fitting in each.  It does not cross talk with fitting in invariant or fitting from my tests.

wpotrzebowski:
ready for testing on OSX

Issue #1541, opened 2020-05-17
Invariant and the infinite multiplication of plots


butlerpd:
When using the invariant perspective in 5.0.2, every time the invariant is recomputed a new plot is created (and all the old plots are there as well).  Not sure if the old plots are still linked to the invariant panel or become orphaned.  Thus changing the q limits, the contrast, the background, the extrapolation, the porod constant etc etc etc., each generates a new plot so that quickly I can have a dozen plots floating about.  this should behave as the other perspectives and update the appropriate plot not create a new one (as do inversion and corfunc etc).

Problem noted on both mac and pc

PR #1675, opened 2020-09-22
ESS_GUI: Save/Load Corfunc and Invariant Perspectives

### LINKED ISSUE #1526: Project Save/Load functionality of 4.x needs to be restored

### LINKED ISSUE #1539: Corfunc requires two shots to populate the data name box
as of 5.0.2, for some reason, when sending to data to corfunc, the dataset name does not get picked up by the GUI unless it is sent a second time (so clicking twice on `Send data to` button). Note that the data itself does seem to be sent and plots appropriately. This is true on both Mac and PC This is not the case for the other perspectives.

Observerd both on mac and PC

### LINKED ISSUE #1655: Corfunc and Invariant Perspectives not able to remove/swap data (5.x)
The Corfunc and Invariant perspectives have no ability to add, swap, or remove the existing data set. This will make loading projects and analysis files impossible (#1526) if data is already present in either perspective.

### LINKED ISSUE #1657: Loading project without fit_params entry causes empty fit window (v5.0.3)
In the latest 5.x release, it is possible to generate a saved project with no fit data (only P(r), e.g.). When loading a project of this type, the fitting perspective has no tabs and a second, small fit page is generated next to it with a single tab. Sending data to fit can crash the GUI.

### LINKED ISSUE #1674: Reloading a project in the same session duplicates the model/residuals in the Data Explorer
When reloading a project in the same session, the data explorer will duplicate the model and residuals children of the dataset (see screenshot attached)
dataexplorer_glitch
)
The problem seems to come from the `self.data_name_dict` in the `DataManager` which is not reset when loading a project, so it stores previous dataset names and thus renames the loaded data by adding a number between brackets and incrementing it each time the dataset is reloaded.

krzywon:
This allows data from the Corfunc and Invariant perspectives to be saved and loaded in project and analysis files. This also fixes a few other issues associated with loading projects. Unit tests added for all serialization methods.

Closes #1526 : All perspectives can now be saved and loaded
Fixes #1657 : Loading project files without fitting perspective data now load properly
Fixes #1655 : Corfunc and Invariant perspectives are now able to delete and swap data as needed
Fixes #1674 : The data names are now reset when loading a new project
Fixes #1539 (Unintended fix) : The name now appears the first time you send data to the Corfunc perspective


src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -221,8 +221,9 @@ def closeEvent(self, event):
         if self._allow_close:
             # reset the closability flag
             self.setClosable(value=False)
-            # Tell the MdiArea to close the container
-            self.parentWidget().close()
+            # Tell the MdiArea to close the container if it is visible
+            if self.parentWidget():
```
pkienzle, line 225: Somewhat suspicious: the code requesting close gets here so the widget exists, but it is not bound to a parent so it doesn't get removed.

Is this a dangling pointer, where the widget was already removed, but some other function is saying to remove it again? Or is this a zombie window which might reappear later even though you told all windows to close?

The same code is in invariant and inversion, so update/ignore in all places.

src/sas/qtgui/Utilities/GuiUtils.py
```diff
@@ -1239,7 +1239,7 @@ def jdefault(o):
             return add_type(content, type(o))
 
         # ndarray
-        if isinstance(o, np.ndarray):
+        if isinstance(o, (np.ndarray, np.float)):
```
pkienzle, line 1242: `np.float` is just the python `float` type, which does not have a "tolist" attribute:
```
>>> __builtin__.float == np.float
True
>>> np.float(10.).tolist()
AttributeError: 'float' object has no attribute 'tolist'
```
I'm not sure it is meaningful to turn a scalar into list, but `np.float32/np.float64` both define `tolist` to return a scalar builtin floating point value.

So maybe you mean `if isinstance(o, (np.ndarray, np.float32, np.float64):`?

I haven't looked at the context to see what happens if the type condition returns false.

pkienzle (commented):
Code looks okay but I didn't test.

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
krzywon: The parentWidget for a perspective will be a widget in the main window, but only if that perspective is active. All non-active perspectives are removed from the main window and held in the `self.loadedPerspectives` variable of GuiManager.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -190,19 +190,23 @@ def loadAllPerspectives(self):
         # Load all perspectives
         loaded_dict = {}
         for name, perspective in Perspectives.PERSPECTIVES.items():
-            loaded_perspective = perspective(parent=self)
-            loaded_dict[name] = loaded_perspective
+            try:
+                loaded_perspective = perspective(parent=self)
+                loaded_dict[name] = loaded_perspective
+            except Exception as e:
+                logger.log(f"Unable to loading {name} perpesctive.\n{e}")
```
pkienzle, line 197: `f"Unable to load {name} perspective.\n{e}"`

krzywon: I no talk English good...

rozyczko (approved):
Works as expected.
Code is ready, too.

smk78:
Built and tested this branch on W10/x64.

**Performed correlation function analysis** on \test\1d data\ISIS_98929.TXT (load data, send to Corfunc, Calculate background, Extrapolate, Transform & Extract Parameters). Then did Save Project. Restarted SasView and did Load Project.

Everything comes back _except_ the Real Space Correlations plot. See below.
image
Which is a bit odd because it has re-drawn the Scattering Data plot.

However, it _does_ know about the transformed data, because if you do Create New it plots it:
image

smk78:
**Performed Invariant analysis** on \test\1d data\Ludox_silica.xml (load data, send to Invariant, changed background, Calculate). Then did Save Project. Restarted SasView and did Load Project.

Seemed to work as expected.

smk78:
After performing the test of Invariant project saving/loading above, I then, without closing the session (ie, SasView contained the reloaded Invariant project), loaded ISIS98929.TXT and performed Correlation Function analysis. I then saved the session to a new project. This took a while but no cause was immediately obvious. But when I closed SasView I found my Command Prompt window filled with the following messages (so many they filled the window buffer):
```
INFO: data cannot be serialized to json: <class 'numpy.float32'>
```
Closing SasView, restarting it, then re-loading the saved invariant-then-corfunc project file _did_ work, and I could toggle between the two analyses using the Analysis menu option.

I then went 'the whole 9 yards', loading AOT_Microemulsion-Drop_Contrast.xml and fitting it with 10% Lognormal PD aswell. I then saved the session to a new project, closed SasView, restarted it, and re-loaded the invariant-then-corfunc-then-fitting project file. This also worked, though none of the fitting plots (data, residuals or size distribution) appeared until I did Show Plot in the FitPage. Interestingly the 'Create New' plot I did after doing invariant-then-corfunc _was_ displayed.

smk78 (approved):
So, in summary: this PR does reinstate the ability to save correlation function or invariant analyses to projects, either singly, or in combination with one another, or in combination with fitting analyses without any apparent ill effects. For this reason I will approve the PR.

However, as my comments above demonstrate, there is some weirdness regarding what plots are redrawn and which are not.

butlerpd:
So the question now seems to be whether @krzywon thinks we should merge and create new tickets based on the testing or whether some or all of these issues are "obvious" and should be fixed as part of the PR.  To summarize my understanding of the issues (@smk78 can correct)
1. Real Space Correlation plot does not populate properly on loading project (but can be manually forced to plot)
2. saving PR and Invariant (at least in the sequence described) is a bit slow to save but otherwise works perfectly fine "functionally." However there are a ton of "data cannot be serialized" info messages (probably from a loop?). As info they are not errors I guess but perhaps this at least can be fixed by moving the info statement outside the loop?
3. Fitting plots do not show on loading fitting project at least when part of a Invariant+corfunc+fitting saved project. Question:What is the current behavior of fitting save? is this part of the problem in not seeing plots in fitting till doing a fit or asking to show plot?

krzywon:
1. The Corfunc correlation plot is now updated and the buttons are enabled properly with my latest push.
2. The 'data cannot be serialized' only happens when data loaded from a project is saved again and only effects the data in plots. I wasn't able to solve it here and opened a new issue #1706.
3. Invariant plots are now plotted, but the plots are still effected by issues #1541 and #1542.

PR #1683, opened 2020-10-07
ESS_GUI: Fix file decoding issue

krzywon:
This fixes an issue loading windows-1252 encoded text files that have special characters (e.g. Å) in the file contents. Unit tests for files with different encoding added.


pkienzle:
Tests failing:
```
  File "utest_ascii.py", line 135, in test_encoding
    ansi = self.loader.load(find("encoding_ANSI.txt"))
  File "/Users/travis/build/SasView/sasview/src/sas/sascalc/dataloader/loader.py", line 390, in load
    return self.__registry.load(file, format)
  File "/Users/travis/build/SasView/sasview/src/sas/sascalc/dataloader/loader.py", line 85, in load
    return self.load_using_generic_loaders(path)
  File "/Users/travis/build/SasView/sasview/src/sas/sascalc/dataloader/loader.py", line 118, in load_using_generic_loaders
    "Generic readers failed to load %s" % path)
sas.sascalc.dataloader.loader_exceptions.NoKnownLoaderException: Generic readers failed to load data/encoding_ANSI.txt
```

src/sas/sascalc/dataloader/file_reader_base_class.py
```diff
@@ -80,8 +80,14 @@ def read(self, filepath):
...
-                    self.f_open = open(filepath, 'rb')
-                    self.get_file_contents()
+                    try:
+                        self.f_open = open(filepath, 'rb')
+                        self.get_file_contents()
+                    except ValueError:
+                        # Issue(s) found while reading file in binary mode
+                        # Attempt to open in text-only mode
+                        self.f_open = open(filepath, 'r')
+                        self.get_file_contents()
```
pkienzle, line 90: If `get_file_contents` fails on `'rb'` then the file isn't closed automatically, but instead relies on the garbage collector to close it.  Better to use a context manager:
```python
try:
    with open(filepath, 'rb) as self.f_open:
        self.get_file_contents()
except ValueError:
    with open(filepath, 'r') as self.f_open:
        self.get_file_contents()
```

I don't think you need the `finally` block if you do this. The file will be closed by the context manager, and the other operations (the extension deprecation method and the data sorting) didn't ever need to be in `finally`. Since the only exception bypass is for the `FileContentsException`, as well as external control exceptions like the keyboard interrupt and the quit signal, the other operations can occur after the exception block is complete.

Probably too much for this PR, but it would be better to avoid adding temporary state to the object. Can `get_file_content` be modified to accept a file handle? Then you can use:
```
with open(filepath, 'rb') as f_open:
    self.get_file_contents(f_open)
```

smk78:
Seeing @pkienzle 's comment I'll hold off road-testing this PR for the moment then...

src/sas/sascalc/dataloader/file_reader_base_class.py
krzywon: I have implemented the `with open()` changes on my local machine. I did not make the suggested change to pass f_open to get_file_contents - that would require changes in almost every reader.

Fixing the ANSI file loading issue in Linux (works in Windows) is trickier, so I am closing this PR until I have a solution.

smk78:
Update from AdrianR on how the failing ANSI-encoded file was generated:

>Data were written originally on a Linux system but then further combined on two different Windows computers.  Transfers were made in part by USB memory stick, and in part by attachments as e-mail.  The attachments were sometimes made as zipped files.

krzywon:
5.0 ready for testing on Win

smk78 (approved):
Others have commented on the code, so I will only address functionality. I tested an installer version with this branch on W10/x64, and also got the User who reported the original issue to test the same installer.
We both find that this PR fixes the original issue (I only had the one file to test, but the User has tested several others).
I have also test loaded, and checked by creating plots, most of the data in our \test folder just in case this PR had adversely impacted the loaders in some way. I did not find any problems.
The User has reported
```
I have made some tests with a few other files that I have (e.g. xml). There is an anomaly in not showing the first part of the <Title> field for at least some of those files.
```
However, I have not seen this in any xml file I've loaded from the \test folder. If the User can provide more information we can take it from there. But for now I am going to approve and merge this PR.

Issue #1676, opened 2020-09-25
Checkbox of highlighted row is checked also when clicking another checkbox


toqduj:
I noticed this when fitting on a Mac: When you have a list of parameters with one (random) row highlighted, and you click on another checkbox, the checkbox of the highlighted row is toggled as well. 

![Screen Shot 2020-09-25 at 10 12 52](https://user-images.githubusercontent.com/5449929/94252506-4e8d2580-ff24-11ea-836f-f8440dd60e2b.png)

RichardHeenan:
It is the same on Windows, have had some discussion about it with Piotr, but seems it is a "feature" that will be hard to get rid of ????

rozyczko:
Yes, this has been implemented as a "feature" - select several rows and click any of the selected checkboxes. This sets the state on all selected.
The problem appears when you don't `select` the row but click on the checkbox only. This action will be understood as 'change the selection status of the clicked checkbox and all the other selected rows', which in this case is clearly incorrect.

I think there was an existing GH issue for this, but can't find it now?

toqduj:
Hm, I don't think this works as a feature, it's very confusing when the highlighted row is no longer visible and you're clicking a checkbox somewhere else. I'm not sure if there's much of a use case selecting many checkboxes.. 
Maybe if the selection is larger than 1 row, it would make more sense?

PR #1690, opened 2020-10-21
A bit "smarter" extended selection checkbox toggle. #1676

rozyczko:
- If the checkbox is clicked in a row which is NOT selected (dark background), only this row gets toggled
- If the checkbox is clicked in a row which is selected (one or many rows can be selected) all rows get toggled based on the content of the checkbox clicked.

This is hopefully less confusing for single row checks while maintaining extended selection applicability.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1061,11 +1061,19 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         return [s.row() for s in self.lstParams.selectionModel().selectedRows()
                 if self.isCheckable(s.row())]
 
-    def setParameterSelection(self, status=QtCore.Qt.Unchecked):
+    def setParameterSelection(self, status=QtCore.Qt.Unchecked, item=None):
         """
         Selected parameters are chosen for fitting
         """
         # Convert to proper indices and set requested enablement
+        if item is None:
+            return
+        # We only want to select/deselect all items if
+        # `item` is also selected!
+        # Otherwise things get confusing.
+        # https://github.com/SasView/sasview/issues/1676
+        if item.row() not in self.selectedParameters():
+            return
         for row in self.selectedParameters():
             self._model_model.item(row, 0).setCheckState(status)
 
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1061,11 +1061,19 @@ def selectedParameters(self):
         return [s.row() for s in self.lstParams.selectionModel().selectedRows()
                 if self.isCheckable(s.row())]
 
-    def setParameterSelection(self, status=QtCore.Qt.Unchecked):
+    def setParameterSelection(self, status=QtCore.Qt.Unchecked, item=None):
         """
         Selected parameters are chosen for fitting
         """
         # Convert to proper indices and set requested enablement
+        if item is None:
```
krzywon, line 1069: This breaks the '(De-)Select parameters for fitting' options when right-clicking on parameters. The `selectParameters()` and `deselectParameters()` methods will need to pass `item` somehow.

krzywon (changes requested):
The code looks fine and the functionality works as described, but this breaks something (see my comment).

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
rozyczko: Great catch! Fixed.

RichardHeenan:
In my current local build of this branch, the mutual constrain does not work, is there a commit of fix above needed still?  I get error
11:59:36 - ERROR: Traceback (most recent call last):
  File "C:\sasview42\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 757, in showMultiConstraint
    self.addConstraintToRow(constraint=constraint, row=row)
  File "C:\sasview42\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 810, in addConstraintToRow
    symbol_dict = self.parent.perspective().getSymbolDictForConstraints()
  File "C:\sasview42\sasview\src\sas\qtgui\Perspectives\Fitting\FittingPerspective.py", line 537, in getSymbolDictForConstraints
    symbol_dict.update(tab.getSymbolDict())
  File "C:\sasview42\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 4387, in getSymbolDict
    model_name = self.kernel_module.name
AttributeError: 'NoneType' object has no attribute 'name'

rozyczko:
I don't see this failure - what are you constraining, @RichardHeenan ?

krzywon (approved):

krzywon:
The issues noted by @RichardHeenan might need to be addressed before merging, but this should otherwise be ready.

rozyczko:
I am not able to see the issue @RichardHeenan is reporting and looking at the code it doesn't seem the current change could cause the above error.
I'll create another issue for the observed bug and close this PR.

Issue #1699, opened 2020-10-30
Allow check/uncheck of sub-selected data in Data Explorer


smk78:
In 4.x & 5.x, the Data Explorer currently allows one to select (check)/deselect (uncheck) ALL loaded data.

However, in 5.x, the GUI manager _also_ allows you to highlight a sub-selection according to normal OS rules (ie, by holding down the shift or ctrl keys whilst selecting) so that selections like the following are possible:
![image](https://user-images.githubusercontent.com/10938679/97692628-ecca5900-1a97-11eb-9103-72371ec8bf06.png)
But what you cannot then do is check/uncheck just the data in that sub-selection in one go.
Could we implement that?

krzywon:
This was fixed when #1700 was merged.

PR #1700, opened 2020-10-30
Allow setting checkbox state on item selection in DataManager. #1699

rozyczko:


smk78 (approved):
I've not reviewed the code, but have done a functionality test building the branch locally on W10/x64.

It works as expected.

It is ever so slightly unintuitive, in that having selected your data sets by shift-click or ctrl-click you then have to realise you need to check a box for any selected data set in order for it to check them all (I initially right-clicked looking for a 'select' on the context menu).

rozyczko:
> I initially right-clicked looking for a 'select' on the context menu).

Yes, this functionality should be available also from the right click menu. 
Added in the most recent commit.

smk78:
Pulled and tested latest commit.
The context menu now shows 'Select'/'Unselect' options and these do do something with the checkboxes of any selected data sets.
_However_ when you choose 'Select' what it does is turn the checkbox from an outline square to a filled-in square. Whereas, when you select data sets individually the checkbox gets ticked.
What I see is that SasView will only action data sets that are ticked, not those with filled-in squares!!!

rozyczko:
> What I see is that SasView will only action data sets that are ticked, not those with filled-in squares!!!

Thanks. This is fixed now and the code got simplified in the process.

smk78:
Pulled and tested latest commit.
Glitch is now fixed!

butlerpd:
Do we know why Travis is barfing here? otherwise I guess we can merge?

rozyczko:
Looks like a random fluke? Let's try on Jenkins.

rozyczko:
5.0 ready for testing on Win

rozyczko:
Right, succeeded on Jenkins, so most likely a random barf.

PR #1709, opened 2020-11-27
Various Invariant Plotting Fixes

### LINKED ISSUE #1541: Invariant and the infinite multiplication of plots

### LINKED ISSUE #1542: Crosstalk between Corfunc and Invariant perspectives

### LINKED ISSUE #1605: Problem loading canSAS data into Invariant
The `latex_smeared.xml` data set in the test folder includes 2 separate curves: a slit smeared low Q USANS data set and a high Q pinhole smeared SANS data set.  These are loaded as two separate files into the data manager. Checking only one of the data sets allows only that data to be sent to be plotted or to fitting and once fitted and plotted with the fit only that data set is shown.  however, when sent to the invariant both the checked AND unchecked data sets are plotted in the invariant plot though it appears the Invariant is only calculated from the chosen data set? At least the default qmin and qmax come from that data set and the extrapolations appear to be made from that data set.
NOTE:  This is true in 5.0.2 and in the local build of the 5.x invariant branch.  The dataloader work may or may not have changed this behaviour

krzywon:
This PR fixes a few issues related to plotting, most of which have to do with the Invariant Perspective.
- Fixes #1541: The invariant perspective now updates existing plots so a new window is not generated every time the calculator is run.
- Fixes #1542 and Fixes #1605: The invariant perspective no longer pulls the plot state from the DataExplorer so plots with similar names as Invariant plots are no longer plotted alongside the Invariant plots (same data name or same theory name like Corfunc).

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -1058,24 +1058,44 @@ class DataExplorerWindow(DroppableDataLoadWidget):
                 else:
                     self.plotData([(plot_item, main_data)])
 
+        append = False
+        plot_to_append_to = None
         for plot_to_show in plots_to_show:
-            # Check if this is merely a plot update
-            if self.updatePlot(plot_to_show):
-                continue
+            # Check if this plot already exists
+            shown = self.updatePlot(plot_to_show)
+            # Retain append status throughout loop
+            append = shown if shown else append
 
-            # Residuals get their own plot
-            if plot_to_show.plot_role == Data1D.ROLE_RESIDUAL:
+            plot_name = plot_to_show.name
+            role = plot_to_show.plot_role
+
+            if (role == Data1D.ROLE_RESIDUAL and shown) or role == Data1D.ROLE_DELETABLE:
+                # Nothing to do if separate plot already shown or to be deleted
+                continue
+            elif role == Data1D.ROLE_RESIDUAL:
+                # Residual plots should always be separate
                 plot_to_show.yscale='linear'
                 self.plotData([(plot_item, plot_to_show)])
-            elif plot_to_show.plot_role == Data1D.ROLE_DELETABLE:
-                # No plot
-                continue
+            elif append:
+                # Assume all other plots sent together should be on the same chart if a previous plot exists
+                if not plot_to_append_to:
+                    plot_to_append_to = self.active_plots[plot_name]
+                # Remove existing if already exists
+                self.appendOrUpdatePlot(self, plot_to_show, plot_to_append_to)
             else:
                 # Plots with main data points on the same chart
                 # Get the main data plot
                 if main_data is not None and not self.isPlotShown(main_data):
                     new_plots.append((plot_item, main_data))
                 new_plots.append((plot_item, plot_to_show))
+
+        if append:
+            # Append any plots handled in loop before an existing plot was found
+            for _, plot_set in new_plots:
+                self.appendOrUpdatePlot(self, plot_set, plot_to_append_to)
+            # Clear list of any potential new plots now that they're appended to the existing plot
+            new_plots = []
+
         if new_plots:
             self.plotData(new_plots)
 
@@ -1203,6 +1223,14 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         for _, plot_set in new_plots:
             if type(plot_set) is type(old_plot._data[0]):
                 old_plot.plot(plot_set)
+
+    @staticmethod
+    def appendOrUpdatePlot(self, data, plot):
+        name = data.name
+        if name in plot.plot_dict.keys():
+            plot.replacePlot(name, data)
+        else:
+            plot.plot(data)
 
     def updatePlot(self, data):
         """
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -995,7 +995,7 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         item = GuiUtils.itemFromDisplayName(name, self.model)
         return item
 
-    def displayData(self, name=None, is_data=True, id=None):
+    def displayDataByName(self, name=None, is_data=True, id=None):
         """
         Forces display of charts for the given name
         """
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -1035,7 +1035,7 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         Forces display of charts for the given data set
         """
         # data_list = [QStandardItem, Data1D/Data2D]
-        plot_to_show = data_list[1]
+        plots_to_show = data_list[1:]
         plot_item = data_list[0]
 
         # plots to show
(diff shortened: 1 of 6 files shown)


krzywon:
5.0 ready for testing on Win

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -1035,7 +1035,7 @@ def displayData(self, data_list, id=None):
         Forces display of charts for the given data set
         """
         # data_list = [QStandardItem, Data1D/Data2D]
-        plot_to_show = data_list[1]
+        plots_to_show = data_list[1:]
```
rozyczko, line 1038: This is a very good fix not just for Invariant but for plotting in general.

rozyczko (commented):
Code looks good.

butlerpd:
- #1541
  - indeed no plots are generated at all if there is no extrapolation.  4.x and before do display the data with no extrapolation as would be expected.
  - If the FIRST extrapolation attempt is only one of the two possiblities, the low Q and the High Q extrapolation seem to generate separate plots, each showing the data and the respective extrapolation.
     -  e.g. If a low q extrapolation alone is done a plot is generated
     -  if then the high Q extrapolation is attempted, either leaving the low q extrapolation on or turning it off generates a second plot showing the low Q extrapolation only.
     - If the high q is left on and the low q extrapolation either turned back on or turned off (depending on which version of the previous step was done) the high Q extrapoltion plot goes blank and does not seem to be recoverable. 
     - The first plot with the low q extrapolation continues to work (seen by making sure the low q extrapolation is on and changing the #points to extrapolate).
  - IF however the first invariant extrapolation attempt uses BOTH low and high then only one plot is generated and only one seems to ever be generated from there on out.
  - There are a variety of variations not all described
- #1542  seems to be fixed from a quick couple of tests
- #1605 more or less fixed
  - only one data set is loaded on graph as expected
  - if both sets in the one file are chosen an error is thrown
  - BUT.. if one choose set a then chooses an extrapolation (in order to generate a plot) and then deselects a and selects b and sends to the invariant (still with the same extrapolation turned on) the data set is NOT updated.  However the extrapolation indicates it is operating on the correct data set.  If one can force a new plot by selecting the other extrapolation, or kills the plot and forces a recalculation, then the new data set does indeed display.  This is probably linked to the effort to suppress the constant generation of new plots?

krzywon:
> * #1541
>   
>   * If the FIRST extrapolation attempt is only one of the two possiblities, the low Q and the High Q extrapolation seem to generate separate plots, each showing the data and the respective extrapolation.

I had also noticed this but thought 2 plots at most is better than 1 for every fit. I partly fix this in PR #1717 (where extrapolation plots are thrown out when checkboxes are unticked) but it's still not completely solved. I can work on this if you think people will find this as (or more) irritating than the current situation. The work around I found is to close the existing plots and recalculate.

> * #1542  seems to be fixed from a quick couple of tests

Excellent!

> * #1605 more or less fixed
>   
>   * BUT.. if one choose set a then chooses an extrapolation (in order to generate a plot) and then deselects a and selects b and sends to the invariant (still with the same extrapolation turned on) the data set is NOT updated.  However the extrapolation indicates it is operating on the correct data set.  If one can force a new plot by selecting the other extrapolation, or kills the plot and forces a recalculation, then the new data set does indeed display.  This is probably linked to the effort to suppress the constant generation of new plots?

I had to test this to see what you're talking about and this is not good... The extrapolations are recalculated, but the plot window is not updated with the new data set. I'll look into this.

krzywon:
My latest commit takes care of the plotting issue related to switching files in the invariant perspective. Each data set gets its own plot when switching data until that plot is closed.

This does not fix the issue of separate plots when extrapolations are calculated separately. I hope to fix that in #1717

krzywon:
My latest set of commits handle the case where the extrapolations are plotted separately. This PR will need a new review at this point.

rozyczko (approved):
looks good!

wpotrzebowski:
ready for testing on OSX

judithhouston:
ready for testing on OSX

smk78:
@butlerpd commented:
> * #1541
>   * indeed no plots are generated at all if there is no extrapolation.  4.x and before do display the data with no extrapolation as would be expected.

Is this dealt with in any of the PRs?

smk78:
> My latest set of commits handle the case where the extrapolations are plotted separately. This PR will need a new review at this point.

Building this branch locally I find:
* Only enabling low-Q ext > Calculate, only generates one plot
* Then enabling high-Q ext > Calculate, adds that to the existing plot
* Unchecking either ext > Calculate, does nothing but that is not part of this PR
* Only enabling high-Q ext > Calculate, only generates one plot
* Then enabling low-Q ext > Calculate, adds that to the existing plot
* Unchecking either ext > Calculate, does nothing again

So functionality would seem to be as intended.

smk78:
@butlerpd commented:
> > * #1605 more or less fixed
> >   
> >   * BUT.. if one choose set a then chooses an extrapolation (in order to generate a plot) and then deselects a and selects b and sends to the invariant (still with the same extrapolation turned on) the data set is NOT updated.  However the extrapolation indicates it is operating on the correct data set.  If one can force a new plot by selecting the other extrapolation, or kills the plot and forces a recalculation, then the new data set does indeed display.  This is probably linked to the effort to suppress the constant generation of new plots?

@krzywon replied:
> > I had to test this to see what you're talking about and this is not good... The extrapolations are recalculated, but the plot window is not updated with the new data set. I'll look into this.
>
> My latest commit takes care of the plotting issue related to switching files in the invariant perspective. Each data set gets its own plot when switching data until that plot is closed.

Building this branch locally I find:

* Load set A > Invariant > select an Ext > Calculate, generates one plot as above
* Load Set B > _deselect Set A_ > Invariant > select an Ext > Calculate, generates a second plot and switches the dataset name in the Invariant panel
* Load Set C with plots from Set A and Set B still up > _deselect Set A and Set B_ > Invariant > select an Ext > Calculate, generates a third plot

So functionality would seem to be as intended.

ALTHOUGH:
* Deselect Set B > reselect Set A > Invariant > select an Ext > Calculate, switches the dataset name back in the Invariant panel and does nothing else, or adds the data and the calculated extrapolation to the plot for Set B, not Set A. _However at the moment this behaviour is quite tricky to reproduce because this branch doesn't have enough functionality to help you force a redraw!_

smk78 (approved):
Have reviewed for functionality and I think is good to merge at this stage.

PR #1712, opened 2020-11-30
Ticket 1710: Remove deprecated method call from NXcanSAS reader

### LINKED ISSUE #1710: sasview fails to open .h5 files using h5py 3.1
h5py v3.1 was recently released.  It's causing some regression in sasview, with test .h5 files no longer able to be opened.
No clear error or Exception is given.  `load_using_generic_loaders` from sascalc/dataloader/loader.py reports `NoKnownLoaderException: Generic readers failed to load`, meaning simply that `cansas_reader_HDF5.read(file)` returns an empty list.

It's not clear to me why cansas_reader_HDF5.read is returning an empty list. Running through each step in `read_children()` in the reader by hand, the .h5 files opens successfully and the elements get read successfully one by one.  I didn't check `add_intermediate()` explicitly.  At some point self.output must be getting blanked, or not assigned values.  The blank value of self.output (returned by `cansas_reader_HDF5.get_file_contents`) is the problem.

The error was reported for Debian at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=974181
Debian test logs using h5py 3.1 can be found at https://ci.debian.net/packages/s/sasview/unstable/amd64/
e.g. https://ci.debian.net/data/autopkgtest/unstable/amd64/s/sasview/8477673/log.gz

Intermediate errors are reported in the log:
```
WARNING  sas.sascalc.dataloader.file_reader_base_class:file_reader_base_class.py:160 SasView cannot load /tmp/autopkgtest-lxc.179sxrt1/downtmp/autopkgtest_tmp/test/sasdataloader/test/test_data/simpleexamplefile.h5.
                                                                                     Invalid XML syntax
WARNING  sas.sascalc.dataloader.file_reader_base_class:file_reader_base_class.py:160 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
```

The reference to XML and 'utf-8' codec sounds as if sasview is trying to handle the .h5 file as an xml file, which of course would be wrong.  I'm not sure this is the real problem though.  The error `Generic readers failed to load` makes sense, if cansas_reader_HDF5.read is returning an empty data list.  The question is why is it doing that?  Why does its `self.output` remain empty for a valid test file like simpleexamplefile.h5?

krzywon:
This fixes #1710. The h5py package v3.0 removed a previously deprecated method that was still present in the NXcanSAS reader. I tested this locally with h5py v2.8, v2.9, and v3.1. I did not update the version for h5py in the build and conda files but I'm not opposed to doing this.

### DIFF
--- a/src/sas/sascalc/dataloader/readers/cansas_reader_HDF5.py
+++ b/src/sas/sascalc/dataloader/readers/cansas_reader_HDF5.py
@@ -182,7 +182,7 @@ class Reader(FileReader):
 
             elif isinstance(value, h5py.Dataset):
                 # If this is a dataset, store the data appropriately
-                data_set = value.value
+                data_set = value[()]
                 unit = self._get_unit(value)
                 # Put scalars into lists to be sure they are iterable
                 if np.isscalar(data_set):


drew-parsons:
Works for me.

drew-parsons (approved):
Fixes handling under h5py 3.1 when patched to sasview 5.0.3 on Debian.

Issue #1616, opened 2020-07-06
ESS_GUI: Model label on plot keeps being reset


smk78:
Noticed by User EmilyE on Mac. Verified by @smk78 on Windows. Using 5.0.2.

Matplotlib really, really, doesn't like long dataset names. If you display such a dataset the actual graph gets squashed up in favour of displaying the legend! Like this:
![image](https://user-images.githubusercontent.com/10938679/86573418-49938800-bf6c-11ea-80dd-b1e7809ae969.png)
You can use Modify Plot Property for each dataset on the plot to change the label for that data to something more manageable, for example
![image](https://user-images.githubusercontent.com/10938679/86573633-91b2aa80-bf6c-11ea-8d10-2c925e18bcc3.png)
But, if you then change any parameter in the model, forcing a re-draw, the _model label_ gets reset, like this:
![image](https://user-images.githubusercontent.com/10938679/86573851-dcccbd80-bf6c-11ea-9271-e5329e7235d5.png)
Which is very irritating!

Modify Plot Property doesn't seem to give any control of the font size of legend text?

As a more general comment, I think the default font size of legend text could be reduced. Maybe by two font sizes?

smk78:
This foible, together with points 5) & 6) in https://github.com/SasView/sasview/issues/1584, have been driving me mad this week!

RichardHeenan:
> This foible, together with points 5) & 6) in #1584, have been driving me mad this week!

The work-around for the residuals re-plot is to drag the residuals plot down or sideways in the new "expanding screen".

butlerpd:
can @krzywon or @rozyczko verify that the second problem has been fixed? My test with one of our hackathon PR shows that once I change the label from the file name to something else, that new value sticks no matter how many replots are done including a full fit.  The first issue is still an issue but maybe that is also addressed with some file name work @krzywon?

wpotrzebowski:
@butlerpd On OSX it still seems to be resetting whenever I change parameter value or fit. This may be because code for OSX is branched from Windows in the few places (something that we did when we tried to address huge legends issue on Mac). I will take a look at this.

PR #1713, opened 2020-12-03
Allow display name changes in data explorer (1702)

### LINKED ISSUE #1701: Issue with slashes in data titles in CanSAS1D (and probably NXcanSAS)
Up to and including release 5.0.3, the Data Explorer displayed the name of the _file_ that had been loaded. Since the release, several improvements and bug fixes have been made to the data loader. One improvement being that if it can recognise the name of the _data_ in the file, it will display this in the Data Explorer instead of the file name. The CanSAS1D and NXcanSAS formats will benefit most from this.

In testing the fix to https://github.com/SasView/sasview/pull/1683, User Adrian R has spotted that if the string between <title> </title> contains a forward slash or a back slash, none of the string before the slash gets displayed as the data name in the Data Explorer. The full data name is, however, displayed when you expand Info for that data set. Thus it would seem that the code translating the title string into a data set name is interpreting the slashes.

Characters like percent or vertical bar are unaffected.

The following zip file contains a screenshot illustrating the issue and some example datasets.
[TitleAnomaly.zip](https://github.com/SasView/sasview/files/5465790/TitleAnomaly.zip)

**NB: You must run a recent developer build to see this behaviour! For example:**

https://jenkins.esss.dk/sasview-beta/job/SasView-Windows10-ghprb/58/artifact/sasview/dist/setupSasView.exe after Oct 27.

### LINKED ISSUE #1702: Allow for a choice of how data is named in the Data Explorer
See also https://github.com/SasView/sasview/issues/1701.

Up to and including release 5.0.3, the Data Explorer displayed the name of the file that had been loaded. Since the release, several improvements and bug fixes have been made to the data loader. One improvement being that if it can recognise the name of the data in the file, it will display this in the Data Explorer instead of the file name. The CanSAS1D and NXcanSAS formats will benefit most from this.

User Adrian R makes the following comment on this switch:
```
Perhaps I can make a suggestion about this enhancement in version 5: while I realise that often people like the sample 
description as provided in the title of a data file to appear in plots or in the table when selecting data, there are 
equally cases where one does not find that appropriate.   A choice would be useful.

For example, there can be repeated runs for the same sample, either as a time sequence with varying properties/sample 
conditions or simply made as checks that data are reproduced.  In this case the title may be identical for several data 
sets but the measurements are distinguished by file name or run/sequence number.  It could then be convenient to 
choose what is presented both in the data panel and on the plot/fit output.

The data files that I sent [see #1701 ] happened to be some where the displaying by title could be appropriate but with 
part of the information missing, this became much less helpful.  There are quite a number of data sets where I have 
several measurements with the same title that would only be distinguished by other metadata (instrument configuration, 
wavelength, ....).
```

krzywon:
A new option was added to the data explorer context menu allowing users to change the display name for a data set. The option opens a window offering 4 options: Keep the existing name, change to the data tile, change to the data file name, or change to any name the user would like. Unit tests were added to check base functionality and catch a few odd edge cases.

This fixes #1701 and closes #1702.

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -1495,9 +1495,10 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         self.actionSelect.setEnabled(True)
 
         # Name Changing
-        # Disallow name changes after the data has been sent to any perspective to prevent orphaned plots
-        plots = GuiUtils.plotsFromModel(str(model_item.text()), model_item)
-        self.actionChangeName.setEnabled(len(plots) == 1)
+        # Disallow name changes after the data has been assigned any plots to prevent orphans
+        children = list(GuiUtils.getChildrenFromItem(model_item))
+        hashables = [child for child in children if isinstance(child, GuiUtils.HashableStandardItem)]
+        self.actionChangeName.setEnabled(len(hashables) <= 1)
         # Do not allow name change for lower level plots
         self.actionChangeName.setVisible(model_item.parent() is None)
 
@@ -1521,10 +1522,11 @@ class DataExplorerWindow(DroppableDataLoadWidget):
 
         # Get the model item and update the name change box
         model_item = model.itemFromIndex(proxy.mapToSource(index))
-        plots = GuiUtils.plotsFromModel(str(model_item.text()), model_item)
 
-        # Do not allow name changes then the data is in a perspective
-        if len(plots) == 1:
+        # Do not allow name changes after the data has plots assigned
+        children = list(GuiUtils.getChildrenFromItem(model_item))
+        hashables = [child for child in children if isinstance(child, GuiUtils.HashableStandardItem)]
+        if len(hashables) <= 1:
             self.nameChangeBox.model_item = model_item
             # Open the window
             self.nameChangeBox.show()
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -130,7 +130,7 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         self.data_proxy.setFilterRegExp(r"[^()]")
 
         # Create a window to allow the display name to change
-        self.nameChangeBox = ChangeName()
+        self.nameChangeBox = ChangeName(self)
 
         # The Data viewer is QTreeView showing the proxy model
         self.treeView.setModel(self.data_proxy)
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -1449,7 +1449,6 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         self.context_menu.addAction(self.actionDeselect)
         self.context_menu.addSeparator()
         self.context_menu.addAction(self.actionChangeName)
-        self.context_menu.addSeparator()
         self.context_menu.addAction(self.actionDataInfo)
         self.context_menu.addAction(self.actionSaveAs)
         self.context_menu.addAction(self.actionQuickPlot)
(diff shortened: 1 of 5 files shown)


src/sas/qtgui/MainWindow/NameChanger.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        Find the radio button that is selected and find its associated textbox
+        """
+        buttonStates = [self.rbExisting.isChecked(), self.rbDataName.isChecked(),
+                   self.rbFileName.isChecked(), self.rbNew.isChecked()]
+        textValues = [self.txtCurrentName.text(), self.txtDataName.text(),
+                      self.txtFileName.text(), self.txtNewCategory.text()]
+        newValues = [textValues[i] for i, value in enumerate(textValues) if buttonStates[i]]
+        name = newValues[0] if len(newValues) == 1 and newValues[0] else self.txtCurrentName.text()
+        self._model_item.setText(name)
```
rozyczko, line 72: There is no check on name collision - do we want to allow multiple items with the same name?
Currently, when a dataset is loaded with the same name, it gets the standard extension [1], [2] etc.

rozyczko (changes requested):
Code looks good, minor functionality related issues raised.

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -1498,6 +1505,21 @@ def onCustomContextMenu(self, position):
         # Fire up the menu
         self.context_menu.exec_(self.current_view.mapToGlobal(position))
 
+    def changeName(self):
+        """
+        Open a modal window that can change the display name of the selected data
```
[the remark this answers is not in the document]
krzywon: This was supposed to be a modal window. I should add a data deletion signal to close the window when the model item is deleted. The second case isn't likely if the window is modal, but I would prefer to cover edge cases.

src/sas/qtgui/MainWindow/NameChanger.py
krzywon: Good catch - I'll pass the new name through the DataManager rename routine

krzywon:
5.0 ready for testing on Win

src/sas/qtgui/MainWindow/NameChanger.py
```diff
@@ -71,6 +81,16 @@ def getNewText(self):
...
+    def removeData(self, data_list=None):
+        """
+        Safely remove data from the window in the unlikely event a data deletion signal is sent to the modal window
+        """
+        if not data_list or self._model_item not in data_list:
+            return
+        # Reset model_item and data to None and close the window
+        self.model_item = None
+        self.close()
+
```
rozyczko, line 100: Very nice feature! We should have these safeguards in more places.

rozyczko (approved):

rozyczko:
Changes look good, all issues addressed. Ready to merge

krzywon:
Here are a few comments I've received from Adrian Rennie on the functionality. I'm going to try to implement his suggestions but I'm not sure I will be able to in the PR.

> (a)  One of the main reasons to use this option is when there are several files that appear, with for example, the same title.  In this case, it would be useful to select a number of runs and choose an option for update automatically to e.g. the filename if that is distinctive and the etitle is always the same.
>
>(b)  A further comment is that if one changes the name and then changes it back again to what it was previously a '(1)' appears after the name although there is not any extra data set showing in the Data Explorer.  Is this intentional? 
>
>(c)  In terms of future use, I could imagine that people might like to use some automatic selection as the name of other information that might be available.  At the moment that is usually rather little but might include the 'Run' number but one might envisage loading other metadata as parameters from a canSAS or nxs file.

rozyczko:
> > (b)  A further comment is that if one changes the name and then changes it back again to what it was previously a '(1)' appears after the name although there is not any extra data set showing in the Data Explorer.  Is this intentional?

This looks like a bug, the state should know about data filenames at current time. 

> > (c)  In terms of future use, I could imagine that people might like to use some automatic selection as the name of other information that might be available.  At the moment that is usually rather little but might include the 'Run' number but one might envisage loading other metadata as parameters from a canSAS or nxs file.

This is something we have thought about before and even created a ticket (here or locally in Jira, don't recall).
Modifying the visibility of datasets (not just the check status) based on various criteria (datatype, string in name etc)
Definitely (c) is a separate, large project.

smk78:
Have just been testing this branch.
I note that when you plot data, the plot legend always displays the 'data name' irrespective of the 'filename' or any name assigned. However, fitting models use the name assigned, not the data name. For example:
image
image
Here I left the name of the core contrast unmodified, but called the drop contrast 'fred'.
I think this is likely to be very confusing?

butlerpd:
The point raised by @smk78 has been addressed in another ticket #1616 and a PR #1745 while c) needs to be turned into another ticket. a) looks like an extra feature request. Is that a quick fix or should it be a separate ticket?

That leaves b) as a bug to address?

Issue #1705, opened 2020-11-10
3D magnetic SANS


dehoni:
Implement a generalized 3D description of magnetic SANS with an arbitrary field/polarisation axis. This is needed to describe crystals with strong magnetic anisotropy (somewhere in space), or where only a small magnetic field is used to set polarisation scattering, or for AC field configuration as used for time-resolved SANS (TISANE).

PR #1714, opened 2020-12-13
3 dpolarisation

dehoni:
This branch addresses issue #1705 to improve magnetic SANS allowing for an arbitrary polarization/field direction, which is not restricted to the plane perpendicular to the beam. Tightly linked to the SASmodels issue https://github.com/SasView/sasmodels/issues/432.


wpotrzebowski:
ready for testing on Linux

src/sas/sascalc/calculator/geni.py
```diff
@@ -240,25 +221,37 @@ def _calc_Iqxy_magnetic_helper(
...
+        pz = cos_spin  
+
+
+        qvector = [qxk*norm, qyk*norm, 0]
+        Mvector = [mx, my, mz]
+        Pvector = [px, py, pz]
+
+        
+        Mperp = Mvector-norm*np.dot(Mvector, qvector)*qvector 
+        MperpP = Mperp-np.dot(Mperp, Pvector)*Pvector
```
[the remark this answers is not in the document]
dehoni: I have changes the matrix multiplication where it appeared in the file.

dehoni:
The regularisation of the Fouriertransform to get rid of termination effects and ripple structures might be needed for the test data set "A_raw_example-1.omf". For other sets like [fe_40nm_saturation.zip](https://github.com/SasView/sasview/files/6289700/fe_40nm_saturation.zip) this issue is not apparent. Needs further investigation.

toqduj:
Is this defined using Stokes' parameters?

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -262,12 +267,21 @@ def check_value(self):
         if text_edit.text():
             value = float(str(text_edit.text()))
             if text_edit == self.txtQxMax:
-                if value <= 0 or value > 1000:
+                try:
+                    max_q = numpy.pi / (max(self.sld_data.xstepsize, self.sld_data.ystepsize, self.sld_data.zstepsize) )                   
+                except:
+                    max_q = 1000
```
pkienzle, line 273: You don't want to use bare except statements (here and elsewhere): they block program shutdown signals such as Ctrl-C.

At minimum use "except Exception". Better still, only trap exceptions that you anticipate and can deal with. In this case the expression should only fail is if sld_data is not available or step sizes are not defined, both of which will raise `AttributeError`.

Wouldn't it be better to trap missing sld_data, etc., earlier? Conversely, if max_q/max_step are failing because the information isn't available yet, do they then get updated when the information becomes available? Maybe the solution is to avoid assigning max_q/max_step until they are needed in the calculation?

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -359,6 +374,9 @@ def write_new_values_from_gui(self):
         if self.txtUpTheta.isModified():
             self.model.params['Up_theta'] = float(self.txtUpTheta.text())
 
+        if self.txtUpPhi.isModified():
+            self.model.params['Up_phi'] = float(self.txtUpPhi.text())
```
pkienzle, line 378: Can the text contain non-float values? If so, what happens when an exception is thrown? Verify that a reasonable message is shown to the user when bad values are entered.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
dehoni: The reader now catches missing values in the data, such that the input should be reasonable. All other (unavoidable) instances of except have been changes to "except Exception".

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
dehoni: There are now validators for all the inputs to allow only reasonable values.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
pkienzle: This `except:` is still unchanged? Yes the chances of an interrupt within this try block are vanishingly small, but still non-zero.

Also, unnecessary parens around max().

pkienzle (approved):
I only looked at the code, I did not test.

It still needs one trivial change before merging.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
dehoni: This is resolved with the last commit c949983 "catch except cases already at the inputs".

dehoni:
> 
> 
> Is this defined using Stokes' parameters?
The formalism can be set in relation to the representation of polarised light in terms of Stokes parameters. While in the case of neutrons the polarisation P and the Pauli spin matrix are true axial vectors in real space (with associated polar and azimuthal angle), in the case of light the corresponding quantities are vectors in an abstract space (with one real angle giving the direction of the plane of polarisation and a phase between the orthogonal x and y component producing circular polarisation).
For further details see A G Klein and S A Werner, Neutron Optics, Rep. Prog. Phys. 46, 259 (1983).

dehoni:
mag_sld initialises xstepsize, xnodes etc . There is no need to catch it in the reader and creating a potential string/float conflict. The utest_sasview passes on my local machine now...

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Linux

src/sas/sascalc/calculator/geni.py
```diff
@@ -224,41 +205,51 @@ def _calc_Iqxy_magnetic(
...
-        qsq = qxk**2 + qyk**2
-        norm = 1./qsq if qsq > 1e-16 else 0.
-        perp = norm*(qyk*mx - qxk*my)
-        px = perp*(qyk*cos_spin + qxk*sin_spin)
-        py = perp*(qyk*sin_spin - qxk*cos_spin)
-        pz = mz
-
-        ephase = vol*np.exp(1j*(qxk*x + qyk*y))
+        p_hat = np.array([sin_spin * cos_phi, sin_spin * sin_phi, cos_spin ])
+        q_hat = np.array([qxk, qyk, 0]) * norm
```
pkienzle, line 226: `q=0` will result in NaN since `1/|q| → ∞` and so `0 ⋅ 1/|q|` is undefined.

Is there a neutral value that you can use at `q=0`, such as the average for `q` around the unit circle?

pkienzle:
I got numba working for [mx, my, mz] and did some minor code formatting changes.

I suppressed some prints that were being displayed in the omf reader.

Numba logging was being very noisy so I suppressed it in geni.py. This is not the right place for it since numba is also used in pr, but there is currently no right place.

src/sas/sascalc/calculator/geni.py
```diff
@@ -221,44 +206,60 @@ def _calc_Iqxy_magnetic(
...
 
-        ephase = vol*np.exp(1j*(qxk*x + qyk*y))
+        if abs(qxk) > 1.e-16 or abs(qyk) > 1.e-16:
+            norm = 1/np.sqrt(qxk**2 + qyk**2)
+            q_hat = np.array([qxk, qyk, 0]) * norm
+        else:
+            # For homogeneously magnetised disc Mperp can be associated to the
+            # magnetsation corrected for demag factorfield q->0, i.e. M-Nij M
+            # with Nij the demagnetisation tensor (Belleggia JMMM 263, L1, 2003).
+            q_hat = np.array([0.5, 0.5, 0])
```
pkienzle, line 236: `q_hat` is not a unit vector. Is this intended?

dehoni: Right. The trace of np.outer in the function orth should be always 1.

PR #1715, opened 2020-12-17
Various Invariant GUI Fixes

### LINKED ISSUE #1604: The Invariant low Q extrapolation choice is not honored
In 5.x series, changing the low Q extrapolation from Guinier to power law seems to work on the GUI until one clicks compute.  At this point the GUI unchecks the power law button and rechceks the Guinier button and produces a Guinier extrapolation anyway.

While it may be true that a low Q power law should not be allowed, as long as it is there it should be honored

NOTE: This was tested in the 5.x invariant branch on windows.

### LINKED ISSUE #1607: Once extrapolation is turned on in the invariant it cannot be turned off
Sending data to the invariant and clicking on compute the invariant under the actual data is computed and a plot of the data shown.  If one then chooses, for example, a low Q Guinier extrapolation the Invariant is computed under bot the data and the extrapolated curves and the plot shows the data and the low q extrapolated curve.  Unchecking the extrapolation and recomputing continues to show the same data AND the extrapolated curve and appears to be calculating it as well.

Since the data under the extrapolated curve is currently never included in the invariant on the front panel (see issue #1606) this is currently moot but will become a problem as soon as that issue is fixed. Further more the plot should no longer be showing the extrapolated curves

### LINKED ISSUE #1610: Chart in Invariant Perspective Status Dialog not displaying the low-Q contribution to Q*
It appears that no matter what data you load, or what low-Q extrapolation you perform on it, the percent Q* from Low-Q in the **chart** in the Status > Dialog window is always exactly zero!

In 4.x, even fractions of a percent to 2 dp are shown.

Interestingly, the Dialog window does show a numerical value for the contribution to Q* from Low-Q.

### LINKED ISSUE #1611: Inconsistent behaviour of extrapolation Fit/Fix radio buttons in Invariant Perspective
If you Enable High-Q extrapolation, select **_Fix_**, enter a _Power_, and then click _Calculate_, an extrapolation is generated, but the radio button jumps back to **_Fit_**.

This does not happen in 4.x and could catch out the unwary?

krzywon:
The work in this PR ensures the invariant GUI is no longer reset when the invariant is calculated. The Low-Q, and High-Q extrapolation selections are retained and honored, and the fit values are properly updated once the calculation is finished. Many of the invariant GUI unit tests are updated to be sure the state is retained after a calculation is run. InvariantDetailsTests.py are now enabled in GUITests.py and new test suites were created for each of the individual perspectives.

This now has two higher pull requests merged into it. I've updated the linked tickets.

This fixes #1604, fixes #1607, fixes #1610, and fixes #1611.

### DIFF
--- a/src/sas/qtgui/Perspectives/Invariant/UnitTesting/InvariantPerspectiveTest.py
+++ b/src/sas/qtgui/Perspectives/Invariant/UnitTesting/InvariantPerspectiveTest.py
@@ -34,6 +34,9 @@ class InvariantPerspectiveTest(unittest.TestCase):
             def __init__(self):
                 self.model = QtGui.QStandardItemModel()
 
+            def plotData(self, data_to_plot):
+                pass
+
         class dummy_manager(object):
             def __init__(self):
                 self.filesWidget = MainWindow()
@@ -45,12 +48,13 @@ class InvariantPerspectiveTest(unittest.TestCase):
                 return GuiUtils.Communicate()
 
         self.widget = InvariantWindow(dummy_manager())
-        data = Data1D(x=[1, 2], y=[1, 2])
+        data = Data1D(x=[1, 2], y=[1, 2], dy=[0.01, 0.01])
         GuiUtils.dataFromItem = MagicMock(return_value=data)
         self.fakeData = QtGui.QStandardItem("test")
 
     def tearDown(self):
         """Destroy the DataOperationUtility"""
+        self.widget.setClosable(True)
         self.widget.close()
         self.widget = None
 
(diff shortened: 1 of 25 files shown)


smk78:
With the caveat that I'm testing the git_commit installer (Build 66) triggered in https://github.com/SasView/sasview/pull/1718 and which therefore won't have all the Invariant GUI/Plotting fixes mentioned in this issue or https://github.com/SasView/sasview/pull/1709, there is something (else) odd going on with the extrapolations. So I'll mention it here and someone can tell me if it is known about or not.

The way I would have assumed an extrapolation would work is that it fits the specified function to the specified number of experimental data points (Npts) and then uses the function parameters from that fit to extend the experimental data in the direction specified as far as the specified limit in that direction.

What the Invariant routine seems to do is splice a fixed number (20) of extrapolation points into the region between the specified limit and the 'innermost' of the Npts experimental data points. So here is a low-Q Guinier extrapolation with Npts=3:
image
Arguably this would probably do the job. But now look what happens if I increase Npts=25:
image
The low-Q contribution to Q* for the former case is 0.15%, for the latter case it is 0.13%.
Are we happy the extrapolations are working as we expect?

butlerpd:
That is very weird.... Of course I don't remember trying to use just 3 points to extrapolate but have used 40 ... how would that work.  It actually worked OK for my testing now that I think about it so maybe this is fixed in one of the chained PRs?  I'm guessing the slight difference in low Q contribution to Q* comes from numerical roundoffs on a very low number (0.1%)?

smk78:
Thinking about it some more, I think what we are labelling the 'Low-Q extrapolation' is not. It is the _fit_, right?

Can we display the actual extrapolated data that is used to compute the Invariant?

butlerpd:
No sure what you mean @smk78.  The fit to the data is what is used to calculate the Invariant is it not? On the other hand I just realized that you must have been doing a power law extrapolation rather than Guiner?  The second plot looks wrong for a Guinier extrapolation?

I see @krzywon has merged all the downstream commits back into this PR so maybe best to create a new build to test?

krzywon:
I triggered a build in #1718 after merging the changes that @smk78 is using (correct me if I'm wrong here, Steve).

smk78:
Yes, I was using the second build triggered on https://github.com/SasView/sasview/pull/1718.

smk78:
What I mean @butlerpd , if I was programming the Invariant (which I am surely glad I am not and anyways am incapable of) I would fit each end of the data with my chosen function, then use those fits to extend the range of the data to low and high Q, possibly run a spline through the three contributions to smooth things, and then integrate under that.

Is that, more or less, what SasView is doing, or is it doing something else?

I don't think what is appearing on the plots with the label 'extrapolation' are the extrapolations, I think those are the fits used to generate the extrapolations. If those really _are_ the extrapolations then clearly something weird is going on...

smk78:
Ok, interesting. I've just pulled and built the invariant_plots branch, and this seems to be much better behaved:
image
But I think this confirms that we are looking at a fit?

butlerpd:
I think my understanding is somewhat different .. but could be wrong.  Basically the "extrapolation" is done via a fit to the data points selected (i.e. the last, or first, n points as specified).  Then the parameters extracted are used to create whatever points are needed.  The integration is numerical under the actual data and then points chosen out of the extrapolated fit curve ... if that makes sense.  So we are seeing both. the fit curve IS the extrapolated curve?  I think ....

src/sas/qtgui/Perspectives/Invariant/UnitTesting/InvariantPerspectiveTest.py
```diff
@@ -121,46 +93,59 @@ def testDefaults(self):
...
+            # enabled text boxes
+            self.widget.txtVolFract.isReadOnly(), self.widget.txtVolFractErr.isReadOnly(),
+            self.widget.txtSpecSurf.isReadOnly(), self.widget.txtSpecSurfErr.isReadOnly(),
+            self.widget.txtInvariantTot.isReadOnly(), self.widget.txtInvariantTotErr.isReadOnly(),
+            # radio buttons exclusivity
+            self.widget.rbFixLowQ.autoExclusive(), self.widget.rbFitLowQ.autoExclusive(),
+            self.widget.rbFixHighQ.autoExclusive(), self.widget.rbFitHighQ.autoExclusive()
+            ]
+        self.assertTrue(all(v is False for v in false_list))
+        self.assertTrue(all(v is True for v in true_list))
```
rozyczko, line 148: Those tests are sort of silly, the default state is tested with the GUI testing with Squish.

src/sas/sascalc/invariant/invariant.py
```diff
@@ -424,6 +424,25 @@ def __init__(self, data, background=0, scale=1):
...
+    def scale(self):
+        return self._scale
+
+    @scale.setter
+    def scale(self, value):
+        self._scale = value
+
+    def set_data(self, data):
+        self._data = self._get_data(data)
+
```
rozyczko, line 445: I find mixing explicit setters/getters (`set_data`, `get_data`) and implicit (through decorators) confusing, we should only have one consistent API for the given calculator.

rozyczko (approved):
Great effort, splitting perspective tests is a very good thing. 
Minor questions raised, not big enough to request a change though.

PR #1716, opened 2020-12-17
Floating point percentages in Invariant details

### LINKED ISSUE #1610: Chart in Invariant Perspective Status Dialog not displaying the low-Q contribution to Q*

krzywon:
Show the percent contribution of the individual Q* values as floating point value (2 decimal places) in the Invariant details window.

This is the 3rd PR in a chain of PRs for the invariant perspective: This -> #1715 -> #1709

Fixes #1610

### DIFF
--- a/src/sas/qtgui/GUITests.py
+++ b/src/sas/qtgui/GUITests.py
@@ -17,7 +17,7 @@ The tests can be run with ``python GUITests.py``, or
 To get more verbose console output (recommended), use ``python GUITests.py -v``
 """
 
-# Llist of all suite names. Every time a new suite is added, its name should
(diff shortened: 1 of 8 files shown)


smk78:
This appears to be working as described in the git_commit installer (Build 66) triggered in https://github.com/SasView/sasview/pull/1718.

PR #1717, opened 2020-12-21
Remove Invariant plots when extrapolations turned off

### LINKED ISSUE #1607: Once extrapolation is turned on in the invariant it cannot be turned off

krzywon:
This forces the deletion of the high-q and/or low-q Invariant extrapolation plots when one has been previously calculated, but the user runs another calculation without that particular extrapolation selected.

This is the 4th PR in a chain of PRs for the invariant perspective: This -> #1716 -> #1715 -> #1709

fixes #1607

### DIFF
--- a/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
+++ b/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
@@ -398,7 +398,7 @@ class InvariantWindow(QtWidgets.QDialog, Ui_tabbedInvariantUI):
             power_low = self._calculator.get_extrapolation_power(range='low')
 
             # Plot the chart
-            title = "Low-Q extrapolation"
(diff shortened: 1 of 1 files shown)


smk78:
This appears to be working as described in the git_commit installer (Build 66) triggered in https://github.com/SasView/sasview/pull/1718.

PR #1718, opened 2020-12-29
Invariant: Allow user to change extrapolated Q limits

### LINKED ISSUE #1609: Changing Q-range limits in Invariant Perspective has no effect on extrapolations
In the Invariant tab of the Invariant Perspective you can change the Min and Max Q values. My working assumption is that this is so that you can exclude points at either end of the data prior to the extrapolations? However, if you try this you will find that the extrapolations always use the full range of the data (ie, if you had a dataset with 50 points and opted to extrapolate the low-Q end with 10 points it uses points 1 to 10. If you were to then set a Qmin to exclude the first 10 points you would expect that it would extrapolate on points 11 to 20. It doesn't, it still uses points 1 to 10. The same behaviour happens with the high-Q extrapolations). So, is this by design? If so, why are the Total Q Range boxes editable?

I notice these boxes are not editable in 4.x.

Also see https://github.com/SasView/sasview/issues/1608

krzywon:
This PR allows a user to change the extrapolation Q limits in the invariant perspective.

fixes #1609 

This is the 5th (and likely final) PR in the invariant chain: This -> #1717 -> #1716 -> #1715 -> #1709

### DIFF
--- a/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
+++ b/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
@@ -304,7 +304,9 @@ class InvariantWindow(QtWidgets.QDialog, Ui_tabbedInvariantUI):
                 range="low", npts=int(self._low_points),
                 function=function_low, power=self._low_power_value)
             try:
-                qstar_low, qstar_low_err = self._calculator.get_qstar_low()
+                qmin_ext = float(self.txtExtrapolQMin.text())
+                qmin = None if qmin_ext < self._data.x[0] else qmin_ext
+                qstar_low, qstar_low_err = self._calculator.get_qstar_low(qmin)
                 low_calculation_pass = True
             except Exception as ex:
                 logging.warning('Low-q calculation failed: {}'.format(str(ex)))
@@ -331,7 +333,9 @@ class InvariantWindow(QtWidgets.QDialog, Ui_tabbedInvariantUI):
                 range="high", npts=int(self._high_points),
                 function=function_high, power=self._high_power_value)
             try:
-                qstar_high, qstar_high_err = self._calculator.get_qstar_high()
+                qmax_ext = float(self.txtExtrapolQMax.text())
+                qmax = None if qmax_ext > self._data.x[int(len(self._data.x) - 1)] else qmax_ext
+                qstar_high, qstar_high_err = self._calculator.get_qstar_high(qmax)
                 high_calculation_pass = True
             except Exception as ex:
                 logging.warning('High-q calculation failed: {}'.format(str(ex)))
@@ -421,7 +425,6 @@ class InvariantWindow(QtWidgets.QDialog, Ui_tabbedInvariantUI):
             # for presentation in InvariantDetails
             qmax_plot = Q_MAXIMUM_PLOT * max(temp_data.x)
             qmax_input = float(self.txtExtrapolQMax.text())
-
             qmax_plot = qmax_input if qmax_plot > qmax_input else qmax_plot
 
             power_high = self._calculator.get_extrapolation_power(range='high')
@@ -540,6 +543,10 @@ class InvariantWindow(QtWidgets.QDialog, Ui_tabbedInvariantUI):
 
         self.txtExtrapolQMax.textChanged.connect(self.checkQRange)
 
+        self.txtNptsLowQ.textChanged.connect(self.checkQRange)
+
+        self.txtNptsHighQ.textChanged.connect(self.checkQRange)
+
     def stateChanged(self):
         """
         Catch modifications from low- and high-Q extrapolation check boxes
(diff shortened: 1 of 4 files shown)


krzywon:
5.0 ready for testing on Win

smk78:
Have just tested the git_commit installer for this PR. Functionally it seems to work. I could edit the extrapolation limits, and clicking Calculate caused the plot with the extrapolations to update accordingly.

krzywon:
ready for testing on Win

smk78:
Retested new git_commit installer (now build 66). Functionality still there. But, and I didn't check this earlier, I notice that there seems to be an upper Q extrapolation limit of 0.843 /Ang. Entering any number larger than this and clicking Calculate has no effect!

krzywon:
One thing to note - the number of data points used in the calculation start from the first and last point of the data set regardless of the Q values selected. The upper and lower q bounds are limited to the data q-values however many points you select from each end. Also, there is a check that to be sure the extrapolations don't overlap, though I'm not convinced this is the correct behavior.

My next round of fixes on the invariant were going to allow the user to select the starting and ending points of the data for each extrapolation independent of the extrapolation limit (and add estimators for background and other values). I just haven't made it there yet.

> there seems to be an upper Q extrapolation limit of 0.843 /Ang. Entering any number larger than this and clicking Calculate has no effect! This is using a Fixed extrapolation power.

krzywon:
I'm rereading your comment and realize what I said has nothing to do with your issue. Any value above a certain point fails, not below a certain point. @phgilbert saw something similar if I'm not mistaken. Let me look into this.

krzywon:
ready for testing on Win

krzywon:
ready for testing on Win

krzywon:
@smk78 and @phgilbert - I've applied a couple fixes to this branch and extrapolations seem to work better now. Can you both give the latest build a test when you get a chance?

phgilbert:
I took a look at the limits of the extrapolation. There still seems to be an upper limit to the high-q extrapolation beyond which the plot will not update (for both fit and fix at any slope). For me, this limit is ~ 0.45 /Å. The data set I was using ends at ~ 0.17 /Å (in case this limit is proportional to the data based on the observation of a different limit by @smk78). 

Also, the value of the invariant under the status tab for each extrapolation is invariant (does not change) with the q-range chosen. In other words, only the plot is modified by the q-range chosen. The actual returned calculated value seems to remain the same regardless of range.

butlerpd:
Just realized that the extensive documentation overhaul that @smk78 and I undertook not that long ago is going to need to be revisited 😒  The extrapolation was originally fixed at 1E-5 to 10 1/A. On the other hand I'm pretty sure the plot never extended out to 10 1/A? so maybe there is as @phgilbert says some fraction of Q_max that it plots to.  As for calculation I hope it is that the change is so small that it is not picked up in the 2decimal places reported?  One hopes?

krzywon:
The extrapolation calculation not changing with Q-limits is not good. I made some changes to the calculator and may set the wrong q values for the fit vs. the extrapolated points. I'll look into it.

@butlerpd sadly, yes, the documentation will need an update, mostly for Q-limits and the shape of the high-Q extrapolation.

krzywon:
ready for testing on Win

krzywon:
ready for testing on Win

smk78:
Just tested the git_commit installer (Build 75) for this PR.
On sending the \test\1d_data\AOT_Microemulsion-Drop_Contrast.xml (a dataset with a Qmax of 0.281) to the Invariant, and setting
image
this happened
image
Two things to note:
1) The plot shows the input data in the legend twice, although the context menu only shows it once, and if I remove the data in the context menu then both appearances in the legend vanish (and I'm just left with the extrapolations). If I close the plot window and click Calculate the same thing happens (4 things in the legend), and
2) Though the extrapolation Qmax is 10, the high-Q extrapolation stops at 0.843. This is the same behaviour as reported above.

If I load the 10000A_sphere_dsm.xml which has a Qmax of 0.0052 but leave the extrapolation Q-range at 10 the actual extrapolation stops at 0.0158.

What seems to happen is that the high-Q extrapolation is a fixed number of points (499) with a constant step (1.16e-3 in the first case, and 2.606e-5 in the second case). I've no idea how it decides the step size.

butlerpd:
That looks like a smoking gun to me 😄

krzywon:
Out of curiosity, I fired up release v4.2.2, loaded the same data file as @smk78 and... Got the same result. The high Q extrapolation ends at 0.843 Angstrom<sup>-1</sup>. This has been around for quite a while (and not introduced here). I'm willing to try and fix the issue in this PR, but I will add a ticket documenting the issue, regardless.

Invariant_HighQ_4_2_2

butlerpd:
Agreed.  @smk78 sleuthing I think points to something in the sascalc part that does the extrapolation.  Given we need to turn this around quickly I would vote for making this new ticket become the focus of a new PR .. particularly since this is functionally equivalent to 4.x and @krzywon has a few other issues to finish addressing before the release of an RC next week?  Anyway that is my vote.  I all else is good I vote for a merge

Issue #1644, opened 2020-07-30
Modernize CI and CD


wpotrzebowski:
Current Jenkins setup requires a lot of maintenance and in parts is prone to single point of failure. We want to explore possibilities of Travis and GIthub Actions in order to replace/complement  current Jenkins setup. More detailed description about current CI and CD to come. 

Current setup on Jenkins involves following steps:

1.  Git pull sasview, sasmodels and bumps
2.  Create conda enviroment from yml file 
3.  Build sasmodels and sasmodels docs
4. Install sasmodels in conda env
5. Build sasview UI and docs
6. Install sasview
7. Perform tests
8. Run pyinstaller to generate installers
9. Sign installer file
10. Save conda environment file in case there is broken dependency
10. Store artifacts (intsallers, archives, env files).

Ad. 1. Either master/ESS_GUI or pull request branch is used. Pull request trigger requires "special" comment in PR
Ad. 2. Initially the plan was to have minimalistic yml file ideally one for all platforms but this quickly proved to be impossible to maintain. Now files are very specific and per OS.

andyfaff:
I can help work on this.

wpotrzebowski:
The current setup requires a lot of maintenance particularly handling environments, maintaining slaves and issuing certificates. Not sure if moving to GH actions or Travis will resolve these issues?

rozyczko:
It can to a large extent. The EasyScience/EasyDiffraction project at DMSC is using GitHub actions for their software, including crossplatform packaging, certificates and building pull requests. We can start from there.

One thing to find out is the potential limits on CPU, memory and bandwidth that GH/Travis impose on builds for non-paying users.
If these are too restrictive we might need to stay with our Jenkins...

krzywon:
In the meeting, there was a discussion about how to perform the magic PR comments to trigger a build. We might be able to use the [pull_request_review_comment](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#pull_request_review_comment) Webhook in some way.

wpotrzebowski:
The example in sasmodels looks encouraging but I guess devil in the details. 
It seems possible to use conda https://github.com/marketplace/actions/conda-action and pyinstaller https://github.com/marketplace/actions/pyinstaller-windows with github actions and I guess it should be possible pull different repositories (e.g. sasmodels when building sasview)?
 
Will it be enough to use gitnub-hosted service or should we rather look for self-hosted solutions:
[...]

PR #1730, opened 2021-01-13
Run tests using GitHub Actions

llimeht:
As a starting point for #1644, here's a reimplementation of the existing .travis.yml as GitHub Actions. I started this as a proof of concept as part of the Hackathon... however, it is at feature parity with the travis testing at this point so it's probably ready for us to switch from Travis-CI to GitHub Actions right now. 

What it does:

- builds SasView, builds the documentation and runs tests on Ubuntu, Windows and macOS.
- install packages via pip (conda is not used); pip wheels are cached 
- runs GUI tests under Linux

What it does not yet do that travis does:
- report status to Slack - this looks easy enough to add but needs a slack workspace admin to do set up the web hook secrets

What it does not _yet_ do but travis doesn't either:
- build the Qt GUI under MacOS (wheels for pyuic5 are not currently available it seems)
- enforce that the GUI tests pass; they don't currently pass (but we obviously should have that as a goal)
- run GUI tests under MacOS and Windows
(shouldn't be blockers for merging)

As noted in #1644, future development could extend these actions/add additional actions to build/test installers and even publish installers and wheels for the packages.

Limitations that need fixing prior to merging:
- the known errors in building the docs are currently suppressed as the SasView documentation doesn't currently build (see https://github.com/bumps/bumps/pull/45 and https://github.com/SasView/sasmodels/pull/441). Prior to merging, we should remove the `|| true` where the docs are built. 

Are there any other thoughts, comments, desires or suggestions?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,144 @@
+name: Test
+
+on:
+  [push, pull_request]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  test:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        python-version: [3.6, 3.7, 3.8]
+
+    steps:
+
+    - name: Obtain SasView source from git
+      uses: actions/checkout@v1
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    ### Caching of pip downloads and local wheel builds
+
+    - name: Obtain pip cache (Linux)
+      uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


wpotrzebowski:
@llimeht great progress! I guess we need to ask @ajj to add slack webhook. 
How about pyinstaller that we use for producing final execulables. Any thoughts about it?
Code/script looks good to me and I think it is mergable now.

wpotrzebowski (approved):

llimeht:
@wpotrzebowski, I think pyinstaller stuff would be quite possible to add in to this, with two jobs:
1. regular testing of PRs
2. building and uploading when releases are made

I know nothing about pyinstaller though as I normally make Debian packages :) I can look at that later in a separate PR.

I'll remove the 'draft' annotation from this and future changes like removing the error suppression on doc building we can do later.

wpotrzebowski (approved):

PR #1736, opened 2021-01-14
Minor improvements to GitHub actions

llimeht:
A few minor tweaks to the new GitHub Actions 'test.yml' 

-  make it easier to spot problems in the documentation building (splitting the sasmodels and bumps doc builds from the sasview doc build); sphinx output is so long that being able to use the Actions bookmarks in it to view just the desired part is helpful
- make sure DISPLAY is set when building the SasView docs so that importing PyQt5 works

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -130,7 +130,7 @@ jobs:
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
         cd docs/sphinx-docs/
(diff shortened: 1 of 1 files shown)


wpotrzebowski (commented):
Currently we are not building bumps docs because they are not pulled into sasview (at least that's my understanding). I think it is totally fine to do this for testing purposes but maybe it is better to do it in bumps repo?

wpotrzebowski (approved):

Issue #1584, opened 2020-06-11
Legend + MPL Toolbar


sl-saxs:
I've just tried running SasView, downloaded from the ESS_GUI branch.
The context menu (right-click into the plot window) now shows the options "Toggle Navigation Menu" and "Toggle Legend".
"Toggle Navigation Menu" gives the MPL toolbar. Zoom, Pan, Home, Back and Forward do work fine. However, choosing "Edit axis, curves and image parameters" from the MPL toolbar opens a window with two tabs "Axes" and "Curves" where one can edit the plot style, legend, etc. So what I did is:
    1) Changed the legend the label from "M1 [flexible_cylinder]" to "mm". This changes don't appear in the legend and they don't appear in the context menu "M1 [flexible_cylinder]" -> "Modify Plot Properties"
    2) Changed the plot style from "Solid line" to "Octagon" -> these changes don't appear in the context menu "M1 [flexible_cylinder]" -> "Modify Plot Properties" where it still says that the curve is plotted as a solid line whereas in the plot window the curve now appears as "Octagon".
    3) Changed plot style in context menu "M1 [flexible_cylinder]" -> "Modify Plot Properties" to be "Red" and "Circles". The curve is now displayed as red circles. But choosing "Edit axis, curves and image parameters" from the MPL toolbar now opens a window with only one tab "Axes" where before it showed two tabs "Axes" and "Curves". It was the "Curves" tab where one could modify the plot style and the legend. But the comboBox in the "Curves" tab only showed one curve to be edited, in that case "M1 [flexible_cylinder]". But more curves were shown in the plot window.
   4) If one goes from the context menu "M1 [flexible_cylinder]" -> "Modify Plot Properties" and deletes the "Legend label" completely, the legend remains unchanged in the plot window. If the "Legend label" is just modified instead of deleted completely, the modifications appear on the plot window.
   5) from the context menu -> "Toggle Legend". Makes the legend disappear if it is visible and makes it appear if it is invisible. If the legend is invisible and another curve is added to the plot or if parameters of the currently plotted curves are changed, the legend pops up again. It seems that the legend invisibility doesn't persist.

rozyczko:
Meh, yeah, that's one issue with the toolbar - the `Edit` functionality has been implemented separately, as right click context menu operations. Toolbar `Edit` options are a bit wonky so I might need to remove the Edit icon after all.

sl-saxs:
[...]

PR #1745, opened 2021-01-15
Ticket 1616

wpotrzebowski:
Preserving label and custom plot settings from being reset every time fit or comput has been executed.
A lot of GUI tests has been failing but mostly due to slicer changes

### DIFF
--- a/src/sas/qtgui/Plotting/PlotProperties.py
+++ b/src/sas/qtgui/Plotting/PlotProperties.py
@@ -1,10 +1,7 @@
 from PyQt5 import QtCore
-from PyQt5 import QtGui
 from PyQt5 import QtWidgets
 
 from sas.qtgui.Plotting.PlotUtilities import COLORS, SHAPES
-
-from sas.qtgui.UI import main_resources_rc
 from sas.qtgui.Plotting.UI.PlotPropertiesUI import Ui_PlotPropertiesUI
 
 class PlotProperties(QtWidgets.QDialog, Ui_PlotPropertiesUI):
--- a/src/sas/qtgui/Plotting/PlotUtilities.py
+++ b/src/sas/qtgui/Plotting/PlotUtilities.py
@@ -38,7 +38,7 @@ COLORS = OrderedDict([
         ('Cyan', '#17becf'),
         ('Magenta', '#e377c2'),
         ('Yellow', '#bcbd22'),
-        ('Black', 'k'),
+        ('Black', '#000000'),
         ('Custom', 'x'),
 ])
 
--- a/src/sas/qtgui/Plotting/PlotUtilities.py
+++ b/src/sas/qtgui/Plotting/PlotUtilities.py
@@ -32,12 +32,12 @@ SHAPES = OrderedDict([
 
 # MPL Colors dictionary. Ordered for consistent display
 COLORS = OrderedDict([
-        ('Blue', 'b'),
-        ('Green', 'g'),
-        ('Red', 'r'),
-        ('Cyan', 'c'),
-        ('Magenta', 'm'),
-        ('Yellow', 'y'),
+        ('Blue', '#1f77b4'),
+        ('Green', '#2ca02c'),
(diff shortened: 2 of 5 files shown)


wpotrzebowski:
ready for testing on Win

wpotrzebowski:
ready for testing on OSX

smk78:
Pulled this branch and built locally.
Loaded in the three AOT_Microemulsion datasets from \test\1d_data, checked them all and did Create New to get a plot.
Used the context menu to change the name of a dataset in the legend. This does seem to have been preserved no matter what other changes I applied.
Similarly, my change of title for the whole plot window seems to have stuck.
BUT
I changed the symbol and colour of a dataset and at some point the colour was randomly changed, I think after I brought up the plot navigation menu, went in to settings, changed a couple of values and exited.
I changed the X-axis label from the Q default to q, and at some point that got changed back to Q!
And whether I use the context menu Set Graph Range or the Navigation Settings to change the axis limits, the change is applied, but as soon as you do something else to the plot they automatically get reset!
As far as I can tell it is actually impossible to use any of the provided functionality to display a plot with more than one dataset in the colours, symbols, and on the axes, you want. :-(

wpotrzebowski:
https://user-images.githubusercontent.com/4189922/104799041-114aed00-57cc-11eb-9d4a-916e974e3f84.mov

wpotrzebowski:
@smk78 thanks for testing it. I would expect that at least some of your tests should work... Can you please check what happens when you follow steps from attached screen recording? (sorry for low quality but files cannot be bigger than 10MB).

rozyczko:
> I think after I brought up the plot navigation menu, went in to settings, changed a couple of values and exited.
> I changed the X-axis label from the Q default to q, and at some point that got changed back to Q!
> And whether I use the context menu Set Graph Range or the Navigation Settings to change the axis limits

Ah, I see what the issue is. Anything that we can change from the plot context menu is retained. Anything that is applied from the MPL toolbar `Figure Options` is not stored, since it belongs to a different object which we don't track... I believe this was the main reason I didn't originally want the MPL toolbar to be there, since its options clash with our own chart options exposed in the context menu and tracked independently.
The zoom/pan/home buttons in the toolbar are the reason we have reenabled it, but little was done to the chart options in the toolbar.
Unfortunately, we can't easily remove the options button which would be the best solution.

I propose we treat these issues separately. This ticket addresses the properties we can easily control and this is what Wojtek did.
We may close it after code review.

We may open another ticket to address the MPL toolbar properties lifetime issues and I can start looking at it, unless Wojtek prefers to continue the effort?

rozyczko (approved):
Code looks good, extra points for cleaning up the syntax!

butlerpd:
ready for testing on Win

butlerpd (approved):
The PR fixes what it advertises to fix on windows.  I vote this be merged.  The other issue with plot properties and separate and much larger and has its own issue I believe.

smk78:
> @smk78 thanks for testing it. I would expect that at least some of your tests should work... Can you please check what happens when you follow steps from attached screen recording? (sorry for low quality but files cannot be bigger than 10MB).

@wpotrzebowski : if I follow your video I can reproduce the behaviour you demonstrate in W10.

smk78:
Question: when you go into Modify Plot Property on the Residuals, change the label and font size and exit, the plot colour changes. Why is that? This is one of the 'quirks' I find irritating. With multiple plots you can end up with red and green data, which is obviously not ideal!

wpotrzebowski:
Last commit should prevent annoying color change

rozyczko:
> Last commit should prevent annoying color change

It's either this or using the `classic` style directive, but `classic` will sooner or later get deprecated so let's just switch to the "modern" way and use explicit hex codes for colours

wpotrzebowski:
Merging this PR. Issues 5) and 6) from https://github.com/SasView/sasview/issues/1584 mentioned in the original ticket https://github.com/SasView/sasview/issues/1616 are still unresolved.

PR #1744, opened 2021-01-15
Doc building fixes

llimeht:
The major reason for the current failure to build the sasview docs is sasmodels being monkeypatched into `sas.models`. Having loaded all the classes, sphinx knows that `sas.models` should exist but then can't actually load any source for them. That results in the rather odd errors that we have been seeing:

```
WARNING: autodoc: failed to import class 'sas.models.FlexCylEllipXModel' from module 'sas.sascalc.fit'; the following exception was raised:
…
AttributeError: module 'sas.sascalc.fit' has no attribute 'sas'                                                                                                                                                                         
```

Note that this PR is not enough to actually make the documentation build cleanly, with the following 19 warnings/errors still emitted:

```
…/sasmodels/sasmodels/resolution.py:docstring of sasmodels.resolution.eval_form:3: WARNING: Unknown interpreted text role "fun".
…/sasview/src/sas/qtgui/Calculators/GenericScatteringCalculator.py:docstring of sas.qtgui.Calculators.GenericScatteringCalculator.GenericScatteringCalculator._create_default_1d_data:4: WARNING: Unexpected indentation.                                                                                                                                                                                                
…/sasview/src/sas/qtgui/Calculators/GenericScatteringCalculator.py:docstring of sas.qtgui.Calculators.GenericScatteringCalculator.GenericScatteringCalculator._create_default_1d_data:5: WARNING: Block quote ends without a blank line; unexpected unindent.                                                                                                                                                            
…/sasview/src/sas/qtgui/MainWindow/DataManager.py:docstring of sas.qtgui.MainWindow.DataManager.DataManager.__init__:3: WARNING: Unexpected indentation.
…/sasview/src/sas/qtgui/MainWindow/DataManager.py:docstring of sas.qtgui.MainWindow.DataManager.DataManager.__init__:4: WARNING: Block quote ends without a blank line; unexpected unindent.                                                                                                                                                                                                                             
…/sasview/src/sas/qtgui/MainWindow/DataManager.py:docstring of sas.qtgui.MainWindow.DataManager.DataManager.__init__:5: WARNING: Field list ends without a blank line; unexpected unindent.                                                                                                                                                                                                                              
…/sasview/src/sas/qtgui/Plotting/Arrow3D.py:docstring of sas.qtgui.Plotting.Arrow3D.Arrow3D.__init__:7: WARNING: Field list ends without a blank line; unexpected unindent.
[...]

PR #1747, opened 2021-01-16
Build installer for windows using GitHub actions

llimeht:
This is a first attempt at building an installer for windows using GitHub actions. 

The installer that has been generated is available for download from the [summary of the action](https://github.com/llimeht/sasview/actions/runs/490201987). The artifact is called SasView-Installer-{OS}-{PYVER}-{GIT_HASH} (e.g. SasView-Installer-windows-latest-3.8-bbd5cf8edd62cb978e9970ae16d4c9df8d717832) and is a zip file containing the setupSasView.exe installer. (A build triggered by making a release on GitHub should have a nicer artifact name!)

I've only got as far making this build pass, not actually testing that the installer works. Testers needed!

Comments, suggestions and testing most welcome.

(This is Work-In-Progress and not ready to be merged just yet)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Docs
 
 on:
   [push, pull_request]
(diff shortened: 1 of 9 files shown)


wpotrzebowski:
ready for testing on Win

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Linux

wpotrzebowski:
Code seens fine but Jenskins builds stuck on docs building therefore I would suggest that we wait with merging it until we have a release candidates out.

butlerpd:
Agreed. Probably the most prudent at this point.

butlerpd:
Not bad!! I'm not sure what the other zip files are about, but downloading the one with the name following the format listed above.  It contained only the sasiewsetup.exe file which ran the installation as expected including installing a sasview.exe.  It was not clear to me what version it was built off of (I was too lazy to track the commit hash :-) - but it is before 5.0.4 it seems.

However it seemed to be clear the .sasview/config directory (not sure if that was due to this install but it happened). Also it insisted in placing .sasview into OneDrive. Moreover it insists on giving the dreaded `cannot execute script sasview` no matter what I do. The first step of copying the custom_config.py gets rid of the "can't copy ..." error and now replaces it with an INFO only from line 144 in GuiUtils.py.  No error but it still shows the same error.  Trying each of the 4 steps in the FAQ did not seem to help.

I doubt this is due to github actions and this could use some more testing, but I cannot demonstrate it is fully operational yet.  Still this is a huge step forward it would seem?

rozyczko:
At some point during the code camp I was successful in installing the GH generated installer

> this installer seems to work... at least to the point where I can load a sphere_80.txt and fit it. https://github.com/llimeht/sasview/suites/1844672867/artifacts/35718498 There's noise from logging that says it's not quite right, but it's progress. pyinstaller has come a long way and so I started afresh with the .spec file to see what it could manage on its own without additional config.

Since it no longer works properly, according to @butlerpd , something went wrong between the two versions.
Should be reasonably possible to figure out where.

llimeht:
I can rebase this onto the current `main` to rebuild the installer with the most recent code might be useful. 

This PR is tangled up with some other doc fixes (#1744) that were preventing building from completing at all during the hackathon. I suspect I need to tease the two apart first... I was content to let them get mixed up while prototyping this, but a cleaner set of changes for merging would be better.

butlerpd:
That is probably what is needed. It would be nice to get these changes merged into main but not clear how to do that at the moment. If you could disentangle and rebase that might at least get this one taken care of?

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

llimeht:
Rebased and rebuilt installer available at: 
https://github.com/llimeht/sasview/actions/runs/719553614

(This PR branch is still based on #1744, so let's get that merged as a first step)

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win

wpotrzebowski:
ready for testing on Linux

Issue #1566, opened 2020-05-29
Default Checkboxes in data manager need changing


butlerpd:
The biggest problem here relates to the fact that every action that spawns "children" (the two plots from corfunc, the several ones from fitting like P and S and function etc) are **correctly** hidden in the data tree (i.e. the data tree does not unfold automatically).  However checking the top level data also checks **ALL** the children (but that action is completely hidden from the unsuspecting user) so that when the user next clicks on create graph, or append to graph, they are not only sending the plot they meant to send but the sometimes large number of associated plots. 

There could be a few different approaches, but given that most often one does not want to send all the children to a plot (actually almost never is it all of them anyway) I think the correct answer is that the children do not inherit the checking automatically but that the user has to unfold the tree and manually pick those children plots they want to plot.

Separate but on a similar default checkbox issue, I would argue that when loading a series of data, either they are **not** all selected by default (my number one choice) OR the default menu item on the selection drop down si `unselect all` rather than `select all` since that is pointless if they are all already selected by default.

rozyczko:
How about the `tristate`? 
![001-threestate](https://user-images.githubusercontent.com/8266281/83292483-71affe80-a1ea-11ea-93ba-10366f258ab8.png)

Single click -> select only the top level (main) dataset, double-click -> select the top level and all underlying datasets.
This is behaviour similar to what is often used in various installers, which ask which features to install in the form of a tree-like list.

rozyczko:
Manually selecting/deselecting sub-datasets would automatically put the main dataset checkbox in the `indeterminate` state

butlerpd:
so the 90% use case will be in my estimation selecting only the top level data. If the selection box is checked does that mean only it is checked? or that all the hidden ones that most people don't even know about are checked? how do I tell the difference visually? or does the second click unroll the tree for me and shows the others checked?

I have to say that I think the manual unrolling and checking would be safer for users.  It will be very rare when what people want is really all the plots at once.  What will be desired way over 50% of the time is just the parent data and from time to time we will want a couple of the children, sometimes with, sometimes without, the parent plotted together.

So the more I think about it the more I think the tristate will be more confusing than helpful? @smk78 or others .. what do you think?

RichardHeenan:
[...]

PR #1748, opened 2021-01-16
Make new child plots unselected by default

butlerpd:
Any perspective creating child data and adding them to the datamanager now default to them being unselected. The user can unroll the tree and check whichever plots they would like to work with.

If a perspective needs to have the child plot it is creating and sending to the data manager selected instantly it will need to deliberately set the checkstate at Checked.  This should normally be avoided as it causes no end of confusion to the user.

This should resolve issue #1566.  I suspect this will reduce confusion on the Invariant as well and maybe other places.

Note to Reviewers. This also reverts the last commit in the slicer series PR #1721 that was merged a few days ago which did a manual force uncheck. That is no longer needed.  This was tested on win 10 only.

### DIFF
--- a/src/sas/qtgui/Utilities/GuiUtils.py
+++ b/src/sas/qtgui/Utilities/GuiUtils.py
@@ -411,7 +411,7 @@ def createModelItemWithPlot(update_data, name=""):
 
     checkbox_item = HashableStandardItem()
     checkbox_item.setCheckable(True)
-    checkbox_item.setCheckState(QtCore.Qt.Checked)
+    checkbox_item.setCheckState(QtCore.Qt.Unchecked)
     checkbox_item.setText(name)
 
     # Add "Info" item
--- a/src/sas/qtgui/Plotting/SlicerParameters.py
+++ b/src/sas/qtgui/Plotting/SlicerParameters.py
@@ -352,7 +352,6 @@ class SlicerParameters(QtWidgets.QDialog, Ui_SlicerParametersUI):
                 # Add plot to the DataExplorer tree
                 new_name, _ = os.path.splitext(os.path.basename(filepath))
                 new_item = GuiUtils.createModelItemWithPlot(data, name=new_name)
-                new_item.setCheckState(QtCore.Qt.Unchecked)
                 self.parent.manager.updateModelFromPerspective(new_item)
 
                 items_for_fit.append(new_item)


src/sas/qtgui/Utilities/GuiUtils.py
```diff
@@ -411,7 +411,7 @@ def createModelItemWithPlot(update_data, name=""):
 
     checkbox_item = HashableStandardItem()
     checkbox_item.setCheckable(True)
-    checkbox_item.setCheckState(QtCore.Qt.Checked)
+    checkbox_item.setCheckState(QtCore.Qt.Unchecked)
```
rozyczko, line 414: This solves the immediate issue.
However, maybe we can extend this method to allow for creation of checked items?
If we add an optional argument to the arg list and make the default `Unchecked`, then we can later call this method with this argument set if we ever need a checked item.
`def createModelItemWithPlot(update_data, name="", is_checked=False)`
...
`checkbox_item.setCheckState(QtCore.Qt.Checked if is_checked else QtCore.Qt.Unchecked)`

rozyczko (approved):
Looks good, potential improvement suggestion added.
Feel free to merge with or without changes.

src/sas/qtgui/Utilities/GuiUtils.py
butlerpd: True .. but can we think of a use case where `createModelItemWithPlot` should ever be used with `checked=True`?  I have thought about it but have yet to think of a case where that would be the right answer.  I figured if there is some very rare case that we cannot think of now, you can still do the same by adding the line
``` python
     new_item.setCheckState(QtCore.Qt.Checked)
```

My naive thinking is that it would be best not to add complexity to a function that will likely never be used? but am happy to have programming experts tell me otherwise.  I am a complete white belt in code design :-)

rozyczko: > but have yet to think of a case where that would be the right answer

Then it's moot. The only reason for my comment was if there would be a case for it. I agree that making the function more complex is not sensible, especially if making it work with a very small additional effort is possible.

krzywon (approved):
This seems reasonable.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win

wpotrzebowski:
ready for testing on OSX

Issue #1344, opened 2019-06-04
Slicer options are too limiting


butlerpd:
Slicers currently always require symmetric + and -  Q and then average the + and - sides.  The first is not helpful in any case (though perhaps makes the data "folding" easier?). However there are a number of circumstances where you would want to see the two sides separately (think Couette tangential measurements or grazing incidence etc.).

Ultimately it would be nice to define arbitrary vertical and horizontal slices (not through zero).

NOTE:  For the "folding" part of the problem, the code seems to already be set up to do this using a flag: self.fold.  Currently this is set to TRUE.  making this settable by the user is all that is needed.  Then for the Gui add a checkbox to go from True to False?

NOTE2: this all seems to reside in sasgui a the moment and should probably be moved to sascalc.

krzywon:
This might be fixed from a computational standpoint, where fold no longer must be True, but is this implemented on the GUI side?

butlerpd:
Oops -- This should have been closed with the slicer PR merge. everything requested and more have been addressed there.

PR #1764, opened 2021-01-25
Slicer extension 1344

butlerpd:
This is a start at addressing #1344.  It only addresses the "folding" issue, and only for the boxslicer the code for which, as stated in the issue, was already present but was hard coded to `fold = True` in the GUI and `fold = False` in sascalc.

The GUI is particularly ugly at present in that the slicer parameter editor assumes all parameters are floats.  This works since any number other than 0 is true and 0 is false but ultimately allowing for a parameter by parameter editor would be the right answer making the fold boolean a checkbox.  This also is probably not possible for 5.0.4-rc1 in which case I would recommend that we merge this but keep the branch active. It still allows the functionality to be changed and should be a benign change?

### DIFF
--- a/src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
@@ -233,6 +233,7 @@ class BoxInteractor(BaseInteractor, SlicerModel):
         params["x_max"] = numpy.fabs(self.vertical_lines.x)
         params["y_max"] = numpy.fabs(self.horizontal_lines.y)
         params["nbins"] = self.nbins
+        params["fold"]= self.fold
         return params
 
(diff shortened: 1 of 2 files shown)


butlerpd:
ready for testing on Win

butlerpd:
Note that the functionality may be a bit painful due to the outstanding Slicer Parameter Editor outlined in issues #1766.  The Editor is the only way to change the number of points or this flag.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski (approved):

wpotrzebowski:
Code looks good but I would agree that in the long term we need some kind of checkbox rather than number field. I've tested functionality but wasn't sure if I am doing right thing (as I have little experience with slicers functionality in general).

PR #1768, opened 2021-01-26
Define proper signatures for validate methods. Fixes #1766

### LINKED ISSUE #1766: Slicer Parameter Editor Locks (fails to edit)
The slicer parameter editor seems to be able to easily get into a state where it no longer responds to input changes.  The editor types the changes but then reverts immediately back to the original value when the return key or the apply button is hit.  Furthermore, once in that state it is very hard to get out.  While in that sate every attempt at editing a parameter throws the following error in the log:
```
19:19:58 - ERROR: Traceback (most recent call last):
  File "site-packages\sasview-5.0.3-py3.6.egg\sas\qtgui\Plotting\SlicerParameters.py", line 508, in setModelData
TypeError: validate() takes 1 positional argument but 3 were given
```
The error is not thrown of course when things work suggesting someplace calling it with a different number of parameters?  I suspect it is tied up in how the editor is tied to different plots since each plot brings up another instance of the editor but each editor's batch tab controls all the plots?  There does seem to be a variety of ways this can happen when working extensively with the slicer but the following seems to be fairly constantly reproducible:
1. Load some 2D data sets
2. Send one to `Create New` plot
3. Right click on the plot and chose the `Box Averaging in Qx`
4. right click again on the plot and chose `Edit Slicer Parameters`
5. try changing the number of points
**FAIL**

Now try the following
1. Load some 2D data sets
2. Send one to `Create New` plot
3. right click again on the plot but chose `Edit Slicer Parameters` first
4. From the drop down in the `Slicer Parameters` Panel choose the `Box Interactor X` 
5. Now try changing the number of points
**SUCCESS**

rozyczko:
`box x` and `box y` interactors didn't have correct signatures for the `validate` method. Actual validation is not required at this stage, since values will be checked at apply time in the slicers and if incorrect, a meaningful message is displayed.
However, we need to at least define those methods properly...

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter2D.py
+++ b/src/sas/qtgui/Plotting/Plotter2D.py
@@ -246,7 +246,7 @@ class Plotter2DWidget(PlotterBase):
 
         self.param_model = None
         validator = None
-        if self.slicer is not None:
+        if self.slicer is not None and not isinstance(self.slicer, BoxSumCalculator):
             self.param_model = self.slicer.model()
             validator = self.slicer.validate
         # Pass the model to the Slicer Parameters widget
--- a/src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
@@ -18,7 +18,7 @@ class BoxInteractor(BaseInteractor, SlicerModel):
         self.markers = []
         self.axes = axes
         self._item = item
-        #connecting artist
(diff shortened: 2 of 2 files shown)


wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win

krzywon (approved):
Code looks good and I tested running from source. The errors mentioned in #1766 are still thrown in main, but are fixed here.

PR #1771, opened 2021-01-27
Try to run tests everywhere rather than fast-fail

llimeht:
At the moment, pyopencl breakage on macos is preventing the tests being run on other platforms as the test matrix uses fast-fail by default.


wpotrzebowski:
ready for testing on OSX

Issue #1689, opened 2020-10-21
Project load of Theory items causes the Fit panel to go blank


rozyczko:
The whole FitPage with no data is bit weird in that you can only (I think) do it first thing in a fresh sasview session.

In 5.0.3 release on Windows, I set up no data, sphere, smeared 15%, then tried save project, closed sasview, started again, load project comes up with an completely blank FitPage. needs investigation!

_Originally posted by @RichardHeenan in https://github.com/SasView/sasview/issues/1686#issuecomment-709252682_

krzywon:
This might be a related to #1657 (though worded differently) which is already fixed in PR #1675.

krzywon:
@rozyczko and @RichardHeenan I just tested this using main and think it is fixed. Could one or both of you test when you get a chance and either comment or close?

RichardHeenan:
**In local Win10 build of Main, Calculated coreshellsphere with default Q range, (i.e. did not send any data to FitPage)
[theory_cssphere_smeared.json.txt](https://github.com/SasView/sasview/files/5879579/theory_cssphere_smeared.json.txt)**

Save project (file attached) gave 
10:35:14 - INFO: data cannot be serialized to json: <class 'PyQt5.QtWidgets.QLineEdit'>
10:35:14 - INFO: data cannot be serialized to json: <class 'method'>
10:35:14 - INFO: data cannot be serialized to json: <class 'PyQt5.QtWidgets.QLineEdit'>
10:35:14 - INFO: data cannot be serialized to json: <class 'method'>

**Start new session of sasview, load project gave only some title info in the M1 workspace and errors below. (The json seems to be missing the parameter values from the model? May be asking the impossible here to expect this to work???)**

10:36:05 - ERROR: Traceback (most recent call last):
  File "C:\sasview5\sasview\src\sas\qtgui\Perspectives\Inversion\InversionPerspective.py", line 459, in check_q_low
    q_min = min(self._calculator.x)
ValueError: min() arg is an empty sequence

10:36:05 - ERROR: Traceback (most recent call last):
  File "C:\sasview5\sasview\src\sas\qtgui\Perspectives\Inversion\InversionPerspective.py", line 475, in check_q_high
    q_max = max(self._calculator.x)
ValueError: max() arg is an empty sequence

10:36:05 - ERROR: Traceback (most recent call last):
  File "C:\sasview5\sasview\src\sas\qtgui\Perspectives\Inversion\InversionPerspective.py", line 459, in check_q_low
[...]

PR #1774, opened 2021-01-29
5.0.4 Project Load Fix

krzywon:
This fixes a project load issue noted by @RichardHeenan in a [comment in 1689](https://github.com/SasView/sasview/issues/1689#issuecomment-768239201). This fixes the issue with loading projects with 3 (or more) fits. I am working on the errors thrown in another branch.


rozyczko (approved):
Good code.
The only problematic thing I can see is if you had empty `self.tabs` list. This would throw a ValueError from `max`. However, self.tabs is never empty, as deleting the last tab always creates a new tab in the perspective with incremented `tab_id`

krzywon:
Good catch, @rozyczko. I'll add a `default` value to max for completeness. Would `0` be an appropriate value? An empty list implies no tabs exist and `self.maxIndex` would be set to `1`.

Issue #1772, opened 2021-01-27
Issues noted during testing of 5.0.4 RC1 & RC2


smk78:
1. The Installer is still labelled as 5.0.3
![image](https://user-images.githubusercontent.com/10938679/105989900-a6d26080-6099-11eb-932c-5e4bbbef5504.png)

2. The installer is not offering a choice of installation path, but worse, instead of creating a sensible folder name with 5.0.4 in it, it is overwriting 5.0.3!

smk78:
3. Documentation is still tagged as 5.0.3.
 Did the version number not get updated during the release process?

rozyczko:
I guess we have not done that... 

https://github.com/SasView/sasview/pull/1773

smk78:
4. Are we happy with the behaviour of Change Name in the Data Explorer? Data in plots is always being labelled with whatever SasView is storing internally as the 'Title', completely irrespective of what you ask Change Name to do. So you can have datasets in the Data Explorer called one thing, but if you plot the data it is called something else.

5. Also, if you load a dataset, then delete it. Then load it again, it gets labelled as [1], [2], etc,...

6. And when you delete a dataset this appears in the Log Explorer:
```
  File "site-packages\sasview-5.0.4-py3.6.egg\sas\qtgui\Perspectives\Inversion\InversionPerspective.py", line 475, in check_q_high
ValueError: max() arg is an empty sequence
```

smk78:
7. If you send data to Fitting and do Show Plot, the first time you grab a slider and drag it, it moves smoothly. Thereafter, or the other slider, does nothing until you unclick, at which point it jumps to the new position.

butlerpd:
to @smk78 point 7, it is already itemized in #1762 as left to do (along with quite a few things) with the sliders.  It actually updated the q range numbers in the box on the panel but will not move until you release the mouse. at which point a recalculation and replot is done which takes some time. There is an underlying behavior question that is larger than the slider that is at play there I think.

smk78:
These may already be known issues (in fact, I'm pretty certain I've seem at least one of these mentioned in an issue thread), but:

8. In the Inversion perspective, estimating the background doesn't seem to do anything, at least, the background level box never displays anything but 0.0, but when you click Calculate the radio button swaps to Manual Input.

9. There is some very weird stuff going on with editing the Total Q Range Min/Max boxes.

smk78:
[...]

PR #1775, opened 2021-01-29
5.0.4: Fix errors on data deletion and project load

krzywon:
This corrects issues resulting in errors thrown when data is deleted or new projects are loaded noted by @smk78 in #1772 points [6](https://github.com/SasView/sasview/issues/1772#issuecomment-768257509) and [13](https://github.com/SasView/sasview/issues/1772#issuecomment-768422908) and by @RichardHeenan in multiple comments of #1689.

### DIFF
--- a/src/sas/qtgui/Plotting/QRangeSlider.py
+++ b/src/sas/qtgui/Plotting/QRangeSlider.py
@@ -34,17 +34,6 @@ class QRangeSlider(BaseInteractor):
                                        input=self.data.slider_high_q_input, setter=self.data.slider_high_q_setter,
                                        getter=self.data.slider_high_q_getter)
         self.has_move = True
-        ######################################################
-        # Reset all slider values in data so save/load does not choke on them
-        # Remove once slider definition moved out of PlotterData
-        self.data.slider_low_q_setter = None
-        self.data.slider_high_q_setter = None
-        self.data.slider_low_q_input = None
-        self.data.slider_high_q_input = None
-        self.data.slider_update_on_move = False
-        self.data.slider_low_q_getter = None
-        self.data.slider_high_q_getter = None
-        ######################################################
         self.update()
 
     def validate(self, param_name, param_value):
--- a/src/sas/qtgui/Plotting/QRangeSlider.py
+++ b/src/sas/qtgui/Plotting/QRangeSlider.py
@@ -68,8 +68,6 @@ class QRangeSlider(BaseInteractor):
         Clear this slicer and its markers
         """
         self.clear_markers()
-        self.line_max.remove()
-        self.line_min.remove()
 
     def update(self, x=None, y=None):
         """
@@ -102,9 +100,10 @@ class QRangeSlider(BaseInteractor):
 
     def clear_markers(self):
         """
-        Should be no way to clear the markers
+        Clear each of the lines individually
         """
-        pass
+        self.line_min.clear()
+        self.line_max.clear()
 
     def draw(self):
         """
@@ -150,6 +149,7 @@ class LineInteractor(BaseInteractor):
         return True
 
     def clear(self):
+        self.clear_markers()
         self.remove()
 
     def remove(self):
@@ -241,6 +241,10 @@ class LineInteractor(BaseInteractor):
 
     def clear_markers(self):
         """
-        Should be no way to clear the markers
+        Disconnect the input and clear the callbacks
         """
-        pass
+        if self.input:
+            self.input.textChanged.disconnect(self.inputChanged)
+        self.setter = None
+        self.getter = None
+        self.input = None
--- a/src/sas/qtgui/Plotting/QRangeSlider.py
+++ b/src/sas/qtgui/Plotting/QRangeSlider.py
@@ -34,6 +34,17 @@ class QRangeSlider(BaseInteractor):
                                        input=self.data.slider_high_q_input, setter=self.data.slider_high_q_setter,
                                        getter=self.data.slider_high_q_getter)
         self.has_move = True
+        ######################################################
+        # Reset all slider values in data so save/load does not choke on them
+        # Remove once slider definition moved out of PlotterData
+        self.data.slider_low_q_setter = None
+        self.data.slider_high_q_setter = None
+        self.data.slider_low_q_input = None
+        self.data.slider_high_q_input = None
+        self.data.slider_update_on_move = False
+        self.data.slider_low_q_getter = None
+        self.data.slider_high_q_getter = None
+        ######################################################
         self.update()
 
     def validate(self, param_name, param_value):
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -396,6 +396,19 @@ class DataExplorerWindow(DroppableDataLoadWidget):
             item = model.item(i)
             data = GuiUtils.dataFromItem(item)
             if data is None: continue
+            ###############################################
+            # FIXME: Temporary patch to remove q-range slider params on project/analysis save
+            # Remove when q-range sliders are moved out of PlotterData
+            data.show_q_range_sliders = False
+            data.slider_update_on_move = True
+            data.slider_low_q_input = None
+            data.slider_high_q_input = None
+            data.slider_low_q_setter = None
+            data.slider_low_q_getter = None
+            data.slider_high_q_setter = None
+            data.slider_high_q_getter = None
+            # End temporary patch
+            ###############################################
             # Now, all plots under this item
             name = data.name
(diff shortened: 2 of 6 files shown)


butlerpd:
ready for testing on Win

wpotrzebowski:
ready for testing on OSX

src/sas/qtgui/Plotting/QRangeSlider.py
```diff
@@ -230,6 +241,10 @@ def onRelease(self, ev):
 
     def clear_markers(self):
         """
-        Should be no way to clear the markers
+        Disconnect the input and clear the callbacks
         """
-        pass
+        if self.input:
+            self.input.textChanged.disconnect(self.inputChanged)
```
rozyczko, line 247: This will get disconnected when the slider is closed/removed but explicit disconnect is always good. We should probably strive at having disconnects for each signal connect.

rozyczko (approved):
Good code, very important fixes which should be in the code since the beginning. Thanks for fixing!

wpotrzebowski:
Tested several scenarios with project saving/loading and deleting files and it seems to work fine. 
However I run into error when I try open project saving dialog and click on Cancel:
```
20:12:16 - ERROR: Traceback (most recent call last):
  File "site-packages/sasview-5.0.4-py3.6.egg/sas/qtgui/MainWindow/GuiManager.py", line 693, in actionSave_Project
TypeError: expected str, bytes or os.PathLike object, not NoneType 
```
I suppose this may be some other issue though and program seems to recover from it (doesn't crash).

krzywon:
I made a small change to address the cancel issue as well. Should I create new win and OSX builds to test?

wpotrzebowski:
ready for testing on OSX

butlerpd:
I am finding some serious functionality issues. I have so far only tested the q sliders (using windows installer) as they were a large part of the code touched it seems. The main problem so far seems to be with the P(r) perspective.  In this case I find:
* the sliders may initially be dragable graphically (the point of having them) or not. I have yet to track down precisely what triggers them to become dragable. However once dragged (see next point) and calculate pressed they are no longer dragable. 
* Once they are dragable (the line highlights in red) and one starts to drag, it never lets go. i.e. it keeps moving with the mouse no matter how much one clicks or moves off the graph and back or clicks on another graph and comes back. I have found that clicking on compute again stops it cold ... but they are no longer dragable (see above).
* Finally, and most problematic is editing the min or max box directly.  Here the problem seems to be linked to number that evaluates to zero.  So if qmin is 0.00154 and I backspace one a time I get 0.0015, 0.001, **BLANK** typing 0 remains blank. typing 0. yields . and 0.0 goes back to blank. typing 0.01 will often yield 0.01xx. I do not know where xx came from. The error thrown in the console every time is:
 ```
14:01:48 - ERROR: Traceback (most recent call last):
  File "site-packages\sasview-5.0.4-py3.6.egg\sas\qtgui\Plotting\QRangeSlider.py", line 184, in inputChanged
  File "site-packages\sasview-5.0.4-py3.6.egg\sas\qtgui\Plotting\QRangeSlider.py", line 177, in _get_q
ValueError: could not convert string to float:
```
* However changing the editable box (when it works correctly)  **does** change the slider positions on the graph.

In the case of the Invariant for which the sliders are not active, the qmin and qmax are not editable (is this a known problem or new?)

Finally, and strangely to me, the sliders and max min editable boxes work perfectly as before in the case of both Corfunc and Fitting.

wpotrzebowski:
I can confirm that error on "Cancel" has been now fixed. (Tested on OSX).

butlerpd:
Regarding @RichardHeenan in #1689, the QLineEdit errors **continue** to be thrown when saving the project with a single model and no data.  The same error mentioned by @smk78 in point [13](https://github.com/SasView/sasview/issues/1772#issuecomment-768422908). However the errors are no longer thrown when loading those projects back into a fresh instance of SasView.  However However in the case of the model only project of @RichardHeenan the fitpage is empty (no model is chosen) and the theory box in dataexplorer has one element titled M1 instead of M1 [data].

butlerpd:
deleted data do not cause errors any more

butlerpd:
Further research suggests that the edit box problem is related to the issue of the editor validating/calculating after each key stroke.  In fact it seems to maybe be slightly better than originally. One problem is that the value **must** be between qmin and qmax of the data. So before this PR, of the data qmin is 0.0055 and that is the default in the min q box, as I delete the first 5 (to replace say with 0.008) the value would effectively be 0.005 momentarily which is less than qmin and thus not allowed so nothing would happen.  Now it will delete but deleting the second makes it go blank. Further inserting a 1, 2, or 3 in front of 0.0055 will make the number jump back to 0.0055 (this may be OK even if a bit enerving for those not in the know).  On the other hand inserting an 8 in front of 0.0055 does so but jumps the cursor back to the end of 0.00855.  This is just very enerving.  I will need to see how far back this goes or if it is new with the sliders.

wpotrzebowski:
ready for testing on OSX

krzywon:
I fixed this by enabling the interactive features of the plots after clearing data for replacement.
> - the sliders may initially be dragable graphically (the point of having them) or not. I have yet to track down precisely what triggers them to become dragable. However once dragged (see next point) and calculate pressed they are no longer dragable.

This was related to reinititializing the q-slider values in PlotterData to fix the save/load errors. I moved this reset into the save project routines and things seem better now.
> - Once they are dragable (the line highlights in red) and one starts to drag, it never lets go. i.e. it keeps moving with the mouse no matter how much one clicks or moves off the graph and back or clicks on another graph and comes back. I have found that clicking on compute again stops it cold ... but they are no longer dragable (see above).

I changed the box to update when editing is finished, not when text is changed. This should be more stable now.
> - Finally, and most problematic is editing the min or max box directly. Here the problem seems to be linked to number that evaluates to zero. So if qmin is 0.00154 and I backspace one a time I get 0.0015, 0.001, BLANK typing 0 remains blank. typing 0. yields . and 0.0 goes back to blank. typing 0.01 will often yield 0.01xx. I do not know where xx came from. The error thrown in the console every time is:

This was not present running the code locally, but I tested the last Windows build and also saw this error suggesting the merge might revert something. I am hoping my recent changes fixed this, and I will generate a new windows installer to check. Otherwise, I will need to merge the release_5.0.4 branch and fix what the merge is breaking.
> Regarding @RichardHeenan in #1689, the QLineEdit errors continue to be thrown when saving the project with a single model and no data. The same error mentioned by @smk78 in point 13.

krzywon:
ready for testing on Win

krzywon:
ready for testing on OSX

butlerpd (approved):
New code was not reviewed but functionality was tested with the installer build on windows and seems ready for merging. Anything else should be moved to other issues/PR.  In particular:
1. fix to error being thrown when using the Edit Mask menu item is fixed.  I still think the final/proper fix should disable the menu item just like others are disabled when not usable and as the button that has the same function is disabled.  However, not only is the error not thrown a more useful message is added to the logger and a popup explains to the user why it does not work and what they need to do. **nice**
2. Inversion sliders work very nicely and smoothly now - seems even better than originally. Of course all sliders have a bit of sluggish feel to them but that detail would be a future major project I think.  Very nice
3. Inversion q limits editor also work very smoothly now. The only minor niggle is that hitting return (or clicking away) does not update the slider position as it does in the corfunc perspective for example. However as soon as you hit calculate the sliders move appropriately.  Plenty good enough for this release. Eventually it would be nice for all perspectives to behave similarly.
4. Fitting perspective.  the sliders seem to be working much more smoothly now as well -- in fact look just like they do with the P(r) - **I like it!**
5. The QLineEdit errors on saving a project seems to be fixed. At least I can no longer make it happen.
6.  The problem described by @RichardHeenan on loading a project with only a theory model remains.  That should however I think be moved to another issue/ticket as I don't think it is related to this PR
7. As discussed by @smk78 the plot state of child data is not respected on loading a project but they are loaded in the data manager.  This is probably not related to this PR and IMO should be submitted as a separate ticket.
8. A new observation but again probably should not be part of this PR. Load a data file and send to inversion. set the appropriate parameters and press `calculate` then save as project. Load a fresh sasview and load the project. Everything loads correctly, including the slider positions ... almost.  The number of terms does not seem to be respected. Moreover hitting calculate to display the plots leads to a P(r) plot that is a damped ringing centered at zero no matter what number of terms is chosen. Only by changing the Dmax ever so slightly and calculation will the computation return something sensible. It is really baffling what calculation is being reported instead of the P(r) and why. strange what computation it may be doing instead of P(r). One other observation is that the I_{fit} and P_{fit} subplots are loaded correctly as evidenced by sending them to Plot before recalculating anything.

Bottom line I strongly recommend merging with the caveat that the remaining items above be properly documented in new issues.

Issue #1177, opened 2019-03-30
QT5 GUI - rename FitPage0 to Fitpage1 or rename M1 to M0 (Trac #1136)


RichardHeenan:
The first !FitPage in the QT gui is FitPage0 but has models M1 and parameters M1.radius etc. This will make constraints very confusing!

Suggest rename one or the other, preferably to my mind start at FitPage1 for backwards compatibility with earlier versions.

(putting this in beta approx project for now, though is not really part of this)

Migrated from http://trac.sasview.org/ticket/1136
```json
{
    "status": "closed",
    "changetime": "2018-08-21T11:15:08",
    "_ts": "2018-08-21 11:15:08.987585+00:00",
    "description": "The first !FitPage in the QT gui is FitPage0 but has models M1 and parameters M1.radius etc. This will make constraints very confusing!\n\nSuggest rename one or the other, preferably to my mind start at FitPage1 for backwards compatibility with earlier versions.\n\n(putting this in beta approx project for now, though is not really part of this)",
    "reporter": "richardh",
    "cc": "",
    "resolution": "fixed",
    "workpackage": "Beta Approximation Project",
    "time": "2018-07-18T16:37:42",
    "component": "SasView",
    "summary": "QT5 GUI - rename FitPage0 to Fitpage1 or rename M1 to M0",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView 5.0.0",
    "owner": "richardh",
    "type": "defect"
}
```

rozyczko:
Trac update at `2018/07/20 08:13:46`: 
* **piotr** changed _comment0 from "FittingPerspectiveTest.py will also need to modified" to "1532074491883495"
* **piotr** commented:


`FittingPerspectiveTest.py`  will also need to modified

RichardHeenan:
Trac update at `2018/08/21 11:15:08`: 
* **richardh** commented:

Has been done, so closing this ticket.


* **richardh** changed resolution from "" to "fixed"
* **richardh** changed status from "new" to "closed"

PR #1777, opened 2021-02-02
Exclude masked data points from slicer calculations

### LINKED ISSUE #1776: Slicers Using Masked Data Points in Calculation
User @murphyryanp notes masked 2D data points appear to be added in when calculating slicer sums. The masked points should be excluded in this case, or the user should have an option of what to do with masked points.

krzywon:
The _Slab, Boxsum, Ring, and _Sector slicer calculators now exclude masked points from their calculations. The CircularAverage calculator already excluded such points.

NCNR VSANS data - annular average with masked data points included in the calculation (using release_5.0.4 branch):
slicer_mask_not_applied

NCNR VSANS data - annular average with masked data points excluded from the calculation (using this branch):
slicer_mask_applied

Note: No new unit tests were added. None of the current tests use masked data.

Fixes #1776.

### DIFF
--- a/src/sas/sascalc/dataloader/manipulations.py
+++ b/src/sas/sascalc/dataloader/manipulations.py
@@ -347,6 +347,7 @@ class _Slab(object):
         err_data = data2D.err_data[np.isfinite(data2D.data)]
         qx_data = data2D.qx_data[np.isfinite(data2D.data)]
         qy_data = data2D.qy_data[np.isfinite(data2D.data)]
+        mask_data = data2D.mask[np.isfinite(data2D.data)]
 
         # Build array of Q intervals
         if maj == 'x':
@@ -371,6 +372,9 @@ class _Slab(object):
 
         # Average pixelsize in q space
         for npts in range(len(data)):
+            if not mask_data[npts]:
+                # ignore points that are masked
+                continue
             # default frac
             frac_x = 0
             frac_y = 0
@@ -506,6 +510,7 @@ class Boxsum(object):
         err_data = data2D.err_data[np.isfinite(data2D.data)]
         qx_data = data2D.qx_data[np.isfinite(data2D.data)]
         qy_data = data2D.qy_data[np.isfinite(data2D.data)]
+        mask_data = data2D.mask[np.isfinite(data2D.data)]
 
         y = 0.0
         err_y = 0.0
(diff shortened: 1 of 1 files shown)


butlerpd (approved):
The code is clean enough that I think we can merge this, unless wojciech wants to test with OSX but we would need to add an example masked 2D data to test.  The code is only in the slicer calculator file which touches nothing else and duplicates the existing functionality in the Circular Average Slicer which has been there for some time without obvious deleterious consequences elsewhere.
Given the release of RC2 I vote we merge and put a ticket in to add unit tests later (including a test data with mask)

rozyczko (approved):
code looks good.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
I took a quick look on this PR. I don't have much experience with this type of analysis, so it is bit difficult to test functionality but what I tested worked fine. I've found small bug though. When no file is selected in Data Manager and try to do "Fitting -> Edit Mask" the error occurs: 
```
  File "site-packages/sasview-5.0.4-py3.6.egg/sas/qtgui/MainWindow/DataExplorer.py", line 1653, in extShowEditDataMask
  File "site-packages/sasview-5.0.4-py3.6.egg/sas/qtgui/MainWindow/DataExplorer.py", line 1681, in showEditDataMask
TypeError: must be str, not IndexError
```
And I think we used to have proper error handling before.

butlerpd:
Nice find @wpotrzebowski.  I did not even know that menu item existed.  From the code I cannot see however that this is related to this PR.  Doing a quick check on my build of the #1775 PR (which does NOT include any of the work in #1177 and indeed the error is there as well.  I suspect it goes back quite a ways?

NOTE: this is a **new** feature in 5.x (by request as I recall) the mask editor was only available from the fit panel in 4.x. It still is in 5.x under the `Fitting Options` tab.

The problem is that the mask editor button on the fitpage (including 5.x) will be greyed out if there is no 2D data however the menu item is **not** allowing the user to do something impossible.  This menu item should be greyed out as well as the button when the function is not available.  I suspect this has been so since its introduction. 

Thus I think we can merge this PR. I will copy the notes above into a new issue

butlerpd:
I have now confirmed this is in the PR 1775 build also as I suspected.  In
fact I did not know about the Mask--> edit menu item.. I tink that was
added in 5.x.  The place it used to be still exists in 5.x: a button on the
fit panel (in the case of 5.x  the "Fit Options" tag.  If there is no data
in the fitpage that button is greyed out properly ... but the menu item is
NOT greyed out.  As I say that has probably always been the case.

I will add that comment and merge the masking PR

PS: yes .. I can now verify that this is new to 4.x

Stay safe
Paul

PR #1779, opened 2021-02-09
Retain graph range on redraws

### LINKED ISSUE #1750: Graph range reset to default values when replot
I used 'Change Scale' to view the data as a Guinier plot. The plot automatically adjusts the scale, which looks as shown on the image for data with large error bars:
image

I can re-adjust the scale, but I can only do that if I close the Linear Fit window, which forces me to start the fit again. When I restart the fit, the scale is again reset to inconvenient values. 
The Linear Fit would be much easier to use if the option to change the graph range (or use the zoom tool) remains active while the fit window is open.

krzywon:
This PR retains the graph range when data are replotted, e.g. after a second fit, and ensures the minimum value of a linear fit is always non-negative.

I built this off of `release_5.0.4` and am currently pointing this PR there, but this could also go into `main`.

Fixes #1750

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -456,6 +456,7 @@ class PlotterWidget(PlotterBase):
         if self.setRange.exec_() == QtWidgets.QDialog.Accepted:
             x_range = self.setRange.xrange()
             y_range = self.setRange.yrange()
+            self.setRange.rangeModified = True
             if x_range is not None and y_range is not None:
                 self.ax.set_xlim(x_range)
                 self.ax.set_ylim(y_range)
@@ -492,7 +493,7 @@ class PlotterWidget(PlotterBase):
         if fit_dialog.exec_() == QtWidgets.QDialog.Accepted:
             return
 
-    def replacePlot(self, id, new_plot):
+    def replacePlot(self, id, new_plot, retain_dimensions=False):
         """
         Remove plot 'id' and add 'new_plot' to the chart.
         This effectlvely refreshes the chart with changes to one of its plots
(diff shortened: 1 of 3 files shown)


butlerpd:
ready for testing on Win

butlerpd (approved):
Functionality tests out very nicely indeed.  a big improvement and nothing else seems to have fallen over though more problems have been identified in the testing. 

The code changes are simple enough and look pretty good to me. Further it looks contained enough that it should not affect anything outside of the linearized fits except for saving the view bounds?

This is a big problem for the bio folks at the neutron school (and presumably bio in general) and I recommend merging into 5.0.4, preferably for an RC3 prior to the second group at the neutron school if possible.

butlerpd:
ready for testing on OSX

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -663,8 +670,12 @@ def onFitDisplay(self, fit_data):
         self.fit_result.title = 'Fit'
         self.fit_result.name = 'Fit'
 
-        # Plot the line
-        self.plot(data=self.fit_result, marker='-', hide_error=True)
+        if self.fit_result.name in self.plot_dict.keys():
```
rozyczko, line 673: This is fine but it just made me think if the data explorer name change fix also updates the plot dictionary.
If not, then datasets with manually changed names won't get replotted properly with retained dimension.

However, this is related to the name change functionality, not this issue so let's merge this commit and potentially address the plot_dict update later as a separate issue.

rozyczko (approved):
Code looks good.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Linux

butlerpd:
This was tested by @skrueger111 on a Mac and works well.

> The Mac version works.  I was able to do Guinier fits and Kratky plots without the plot rescaling every time I did a new fit.  In fact, the plot didn't rescale at all unless I adjusted it manually.  So, if you choose a wider range than you need initially (which might take a couple of tries once you have an idea of what the Guinier range is on a lny vs x2 scale), you are all set.  The fitted curve adjusts when you change the fit, but that is all. 

> Susana, I can upload the sasview5.dmg file to our shared drive in case any of our students want to try it either during class or after.  As long as the Mac folks aren't running Catalina, the install process should be quick.  I was able to get it to run on Catalina after giving it permission to run.
> 
> I like the new options for P(r) as well.   The horizontal line at y=0 is very helpful and the vertical bars make it easy to change the q-range.
> 
> Can vertical bars also be added to the Guinier (and other) fitting options that are accessible via the Change Scale option?  Since we are dealing with q*2 or other options when we rescale, it would be easier to just move the bars to change the range rather than to guess what the equivalent q value is (which I know is included in the Fit box, but it is just by trial and error that you hit on the values you want).  I don't recall if this option was available in SasView4.

Note the answer to the last question is yes it did work in 4.x but was added very late in the game by an ISIS sandwich student if I recall? Turns out that because the linearized fit code is so clunky and in need of serious refactoring it was not possible to get these working for 5.0.4.

PR #1784, opened 2021-02-10
Linear Fit Rg*Qmin & Rg*Qmax

### LINKED ISSUE #1782: Rg*Qmax and Rg*Qmin are inverted in the Gunier linear fit
Rg*qmax is an important quantity for verifying the validity of the linearized Rg calculation.  Roughly it should be less than 1 though for a sphere it can be much as high as 2.

The linear fit for the Rg reports Rg*qmin in the box labelled Rg*qmax and the Rg*qmax value is reported in the Rg*qmin box. 

This should be very simple fix but is quite confusing at first.

krzywon:
The Rg*Qmin and Rg*Qmax are now displaying in the correct text boxes in the Linear Fit window. I also corrected the spelling of Guinier (from Guiner) in the window.

I have this pointing to main for the next release after 5.0.4.

Fixes #1782

### DIFF
--- a/src/sas/qtgui/Plotting/LinearFit.py
+++ b/src/sas/qtgui/Plotting/LinearFit.py
@@ -48,14 +48,14 @@ class LinearFit(QtWidgets.QDialog, Ui_LinearFitUI):
         self.bg_on = False
 
         # Scale dependent content
-        self.guiner_box.setVisible(False)
+        self.guinier_box.setVisible(False)
         if (self.yLabel == "ln(y)" or self.yLabel == "ln(y*x)") and \
                 (self.xLabel == "x^(2)"):
             if self.yLabel == "ln(y*x)":
                 self.label_12.setText('<html><head/><body><p>Rod diameter [Å]</p></body></html>')
                 self.rg_yx = True
             self.rg_on = True
-            self.guiner_box.setVisible(True)
+            self.guinier_box.setVisible(True)
 
         if (self.xLabel == "x^(4)") and (self.yLabel == "y*x^(4)"):
             self.bg_on = True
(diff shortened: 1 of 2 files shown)


butlerpd:
I would like to lobby for putting this into an RC3.  My reasoning is as follows:
1. The swap is completely trivial so might as well be added now (L231 and L233)
2. The many changes to the spelling are more dangerous as missing one or mistyping one will cause a failure of the code. however that is only used in the linear fit section of the code and nowhere else (nothing else even really knows about the linear fits but that is another discussion) - Thus it should be fairly easy to test thoroughly being so localized.
3. This section of code is not used much so far but one of the neutron school groups is using it. If we can make an RC3 that includes this by the start of group 2 it may be our best chance to get some real testing of the changes anyway.

butlerpd:
ready for testing on Win

src/sas/qtgui/Plotting/LinearFit.py
```diff
@@ -224,13 +224,13 @@ def fit(self, event):
...
-        self.txtGuiner2_Err.setText(err)
+        self.txtGuinier_2.setText(value)
+        self.txtGuinier2_Err.setText(err)
 
         value = formatNumber(rg * self.floatInvTransform(self.xminFit))
-        self.txtGuiner_3.setText(value)
+        self.txtGuinier_4.setText(value)
         value = formatNumber(rg * self.floatInvTransform(self.xmaxFit))
-        self.txtGuiner_4.setText(value)
+        self.txtGuinier_3.setText(value)
```
rozyczko, line 233: so this is the actual fix, all the rest is typos being corrected in source, which might not be that important.
Assuming unit tests pass after this (I would expect some elements are being tested by property name), this is ready for merge.

rozyczko (approved):
ready for merge assuming unit tests pass after GUI element name change.

src/sas/qtgui/Plotting/LinearFit.py
butlerpd: so are you OK with merging it into 5.0.4 RC3 (assuming tests pass)? or are you concerned that it should remain only in main at this point?

krzywon: Running the tests (locally) I don't see any failures related to the Linear Fit. For some reason, no unit tests touch the Guinier text boxes.

rozyczko: Then it's moot. I'm OK with it being merged.

krzywon:
I created a parallel PR (#1786) for this that is based off `release_5.0.4` branch. If this is going into 5.0.4, this PR should be closed, or vice-versa.

PR #1786, opened 2021-02-10
Linear Fit Rg*Qmin & Rg*Qmax

### LINKED ISSUE #1782: Rg*Qmax and Rg*Qmin are inverted in the Gunier linear fit

krzywon:
The `Rg*Qmin` and `Rg*Qmax` are now displaying in the correct text boxes in the Linear Fit window. I also corrected the spelling of Guinier (from Guiner) in the window.

I built this branch off of release_5.0.4 instead of main. This PR should replace #1784 if we plan to add this to 5.0.4 RC3/final.

Fixes #1782

### DIFF
--- a/src/sas/qtgui/Plotting/LinearFit.py
+++ b/src/sas/qtgui/Plotting/LinearFit.py
@@ -48,14 +48,14 @@ class LinearFit(QtWidgets.QDialog, Ui_LinearFitUI):
         self.bg_on = False
 
         # Scale dependent content
-        self.guiner_box.setVisible(False)
+        self.guinier_box.setVisible(False)
         if (self.yLabel == "ln(y)" or self.yLabel == "ln(y*x)") and \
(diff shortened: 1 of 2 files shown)


butlerpd (approved):
See PR #1784 for the code review comments and the argument for moving these commits to be included in 5.0.4.  Testing of a jenkins build is not currently possible due to servers being down. However, as noted in #1784, the code change is trivial enough and runs locally so should be safe. Further this is for an RC3 so any issues will be caught anyway.  Thus I approve this PR

Issue #1754, opened 2021-01-18
Delete Data does not remove data or plots from Fitting, P(r) or Inversion


smk78:
Pulled and built 5x main on W10 to test #1742 (verifying @butler 's observations) on the P(r) perspective. But then, with the data still checked in the data explorer I selected Delate Data. The confirmation box appeared warning that all data and dependent datasets would be deleted, but on confirming 'yes' nothing happened! The data and plots persist!

The console reports:
```
17:42:39 - ERROR: Traceback (most recent call last):
  File "D:\SasViewCodeDev\sasview-qt5\src\sas\qtgui\MainWindow\DataExplorer.py", line 695, in deleteFile
    self.closePlotsForItem(item)
  File "D:\SasViewCodeDev\sasview-qt5\src\sas\qtgui\MainWindow\DataExplorer.py", line 1821, in closePlotsForItem
    current_plot_items[plot_name] = PlotHelper.plotById(plot_name).item
AttributeError: 'PlotterWidget' object has no attribute 'item'
```
Checking the Inversion and Fitting perspectives I find the same thing.

Moreover, having first created plots with P(r) and then Inversion, when I then sent a third dataset to Fitting and did Show Plot, only the chosen model function appeared, the experimental data did not display. Only when I manually closed all the open plots and then did Show Plot again could I get the normal data + model plot.

butlerpd:
Re the last point, I have also recently noticed that if you send a data set to "create new plot" and THEN send it to fitting and "show plot" you also only get the theory curve. It seems that between the fitting and plotting code, SasView thinks the data is already plotted and just needs to update the model curve (true) but does so in a new plot.

butlerpd:
On the other hand I am not seeing the first problem (i.e. "works for me" 😄 ).  I tested with the PR #1719 build I was working on.  It was branched just before the new year it seems so something may have happened in the meantime. I need to check the latest build from main I guess.

krzywon:
@smk78 are you still seeing this issue?

smk78:
Having locally built release_5.0.4 branch on x64/W10.

The reported problem now seems fixed. Data/plots in the Fitting, Inversion and Invariant (no plots there anyhow) perspectives are all removed if the data source in the Data Explorer is deleted.

The only foible is that in the Corfunc perspective the data and plots in the 'Scattering data' window are removed, but any transforms in the 'Real Space Correlations' window are not! This suggests these generated data are not affiliated to the input data?

PR #1787, opened 2021-02-12
Corfunc Perspective: Reset Plots on Data Deletion

krzywon:
This fixes an issue noted in #1754 where the Corfunc perspective 'Real Space Correlation' plot is not reset on data deletion.


butlerpd (approved):
Tested using a local build of the branch since the Jenkins windows build servers remain down. This fixes the problem noted with Corfunc. I could not duplicated the PQ and SQ problem noted by @smk78 so not sure what to test for there. Otherwise corfunc seems to behave as usual
Given this fix only affect the corfunc perspective  (`CorfuncPerspective.py`) I recommend it can be merged.

wpotrzebowski:
ready for testing on OSX

PR #1788, opened 2021-02-12
Always show path selection on installation

krzywon:
This changes the INNO setup to always allow the user to select an installation path. This fixes [point 2](https://github.com/SasView/sasview/issues/1772#issue-795052995) of #1772.

Documentation on the flag I am adding: [DisableDirPage](https://jrsoftware.org/ishelp/index.php?topic=setup_disabledirpage)

The windows build server will need to be back online before this can be verified.

### DIFF
--- a/installers/installer.iss
+++ b/installers/installer.iss
@@ -21,6 +21,7 @@ AppUpdatesURL={#MyAppURL}
 DefaultDirName=c:\{#MyAppName}-{#MyAppVersion}
 DefaultGroupName={#MyAppName}-{#MyAppVersion}
 DisableProgramGroupPage=yes
+DisableDirPage=no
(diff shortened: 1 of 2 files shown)


rozyczko (approved):
Yes! This `disable` is what triggers asking for the directory location.
It did seem to be set to `yes` before (default?), since it worked as expected at some point.

I rebuilt the installer locally and it works as expected.

butlerpd:
If the jenkins link cannot be restored I would recommend we merge this anyway. Since it only touches the installer files (and only adds the one flag) it is hard to see how this could affect the functionality in anyway.  If we do a manual RC3 release that will at least provide a level of testing?

wpotrzebowski:
ready for testing on OSX

PR #1790, opened 2021-02-16
Limit adding [n] to name on data load and name change

### LINKED ISSUE #1757: Data name dictionary should be purged on name change and deletion
Outlined in PR #1713, when changing the name of a loaded data set, then changing back to any name previously used, the name is appended with a [1] (or [2], etc.). The name is sent through the unique name checker causing this, but the fix was too large and the issue is not a large enough issue to fix in the PR.

krzywon:
The display name dictionary `DataManager.data_name_dict` was changed from `{str: int}` to `{str: [int]}` to better track display names. The dictionary is now purged whenever data is deleted or a display name is changed. Unit tests were updated to account for this change.

Fixes #1757 and [point 5](https://github.com/SasView/sasview/issues/1772#issuecomment-768257509) of #1772.

This could wait until post 5.0.4 final, but this branch is based of `release_5.0.4`.

### DIFF
--- a/src/sas/qtgui/MainWindow/DataManager.py
+++ b/src/sas/qtgui/MainWindow/DataManager.py
@@ -149,10 +149,11 @@ class DataManager(object):
         name = name[0:max_char]
 
         if name not in self.data_name_dict:
-            self.data_name_dict[name] = 0
+            self.data_name_dict[name] = [0]
         else:
-            self.data_name_dict[name] += 1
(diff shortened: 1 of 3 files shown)


rozyczko (changes requested):

butlerpd:
DO NOT MERGE .. wait till 5.0.4 branch is merged back into main then point this to main This will be for a future release.

butlerpd:
Now that the release is out, this needs to be rebased against main so we can merge

PR #1792, opened 2021-02-16
Force redraw of q-range sliders on input change

krzywon:
Q-range sliders are redrawn when their linked inputs are changed. Previously, this only happened when the plot regained focus.

This fixes [point 9](https://github.com/SasView/sasview/issues/1772#issuecomment-776922573) of #1772.

### DIFF
--- a/src/sas/qtgui/Plotting/QRangeSlider.py
+++ b/src/sas/qtgui/Plotting/QRangeSlider.py
@@ -172,9 +172,9 @@ class LineInteractor(BaseInteractor):
         """ Track the input linked to the x value for this slider and update as needed """
         self._get_q()
         self.y_marker = self.base.data.y[(np.abs(self.base.data.x - self.x)).argmin()]
-        self.update()
+        self.update(draw=True)
 
-    def update(self, x=None, y=None):
+    def update(self, x=None, y=None, draw=False):
         """
         Update the line position on the graph.
         """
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Plotting/QRangeSlider.py
```diff
@@ -172,9 +172,9 @@ def inputChanged(self):
         """ Track the input linked to the x value for this slider and update as needed """
         self._get_q()
         self.y_marker = self.base.data.y[(np.abs(self.base.data.x - self.x)).argmin()]
-        self.update()
+        self.update(draw=True)
 
-    def update(self, x=None, y=None):
+    def update(self, x=None, y=None, draw=False):
```
rozyczko, line 177: Instead of extending the scope of `update` you could just add `self.base.draw()` after each `update` call with `draw=True`.
This will make `update` method more consistent in terms of functionality but will increase code duplication slightly.
I prefer to pass as few (unrelated) arguments as possible to a method to have it as clear/testable as possible.
But it's more of a stylistic issue than functionality - your call :)

rozyczko (approved):
good code.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win

wpotrzebowski:
ready for testing on Linux

butlerpd (approved):
installer build seems to work well and smoothly in all perspectives that have sliders.  We should be able to merge this

PR #1793, opened 2021-02-17
Add CSV Writer and Allow freeform file extensions

### LINKED ISSUE #1780: "Save points as a file" does not accept freeform file extensions
When I want to save a fit, I click on the graph, go to the model, select "save points as a file". When I then select "text files", but want to save my file with a different extension, e.g. "my_beautiful_fit.csv", it seems to work, the file selection window is closed, but no file is written. 

This can be very annoying to find out that there are no files after exporting the fits of several optimizations. 

OS: Mac OS X, SasView 5.0.3

krzywon:
Data1D can now be output in CSV format, and freeform file extensions can be provided by the user regardless of the file type selected in the save menu. Unit tests and documentation were modified and expanded to account for the changes I made.

Fixes #1780

### DIFF
--- a/src/sas/qtgui/MainWindow/media/data_formats_help.rst
+++ b/src/sas/qtgui/MainWindow/media/data_formats_help.rst
@@ -18,23 +18,22 @@ SasView reads several different 1D SAS (*I(Q) vs Q*), 2D SAS(*I(Qx,Qy) vs (Qx,Qy
 1D SAS Formats
 --------------
 
-SasView will read ASCII ('text') files with 2 to 4 columns of numbers in the following order: 
-
-    *Q, I(Q), ( dI(Q), dQ(Q) )*
-    
-where *dQ(Q)* is the instrumental resolution in *Q* and assumed to have originated 
-from pinhole geometry.
-
-Numbers can be separated by spaces or commas.
-
-SasView recognises the following file extensions which are not case-sensitive:
+SasView recognizes 1D data in ASCII, CSV, CanSAS XML, NXCanSAS, and AntonPaar SAXSESS formats with the following extensions:
 
 *  .TXT
 *  .ASC
 *  .DAT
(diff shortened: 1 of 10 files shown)


src/sas/qtgui/MainWindow/media/data_formats_help.rst
```diff
@@ -18,23 +18,22 @@ SasView reads several different 1D SAS (*I(Q) vs Q*), 2D SAS(*I(Qx,Qy) vs (Qx,Qy
...
-SasView recognises the following file extensions which are not case-sensitive:
+SasView recognizes 1D data in ASCII, CSV, CanSAS XML, NXCanSAS, and AntonPaar SAXSESS formats with the following extensions:
 
 *  .TXT
 *  .ASC
 *  .DAT
-*  .XML (in canSAS format v1.0 and 1.1)
+*  .XML, or .COR (in canSAS XML v1.0 and v1.1 formats)
+*  .CSV
+*  .H5, .NXS, or .HDF (in NXcanSAS v1.0 and v1.1 formats)
```
toqduj, line 28: Probably should support .hdf5 extension too...

toqduj (approved):
Tested in practice too. Seems to work fine.

butlerpd:
I note that @toqduj added hdf5 to `associations.py` but forgot to add that in the `data_formats_help.rst` help file.  Once that is added, as was agreed at the March 30 2021 fortnightly meeting this can be merged

butlerpd:
Realizing that both @toqduj and @krzywon are out at the moment I thought I'd make the simple edit myself but now realize that it probably best to have one of you check that I understood properly what was being done when I edited the help file.  Thus I will leave for @krzywon to merge unless he thinks I missed the point.

krzywon:
I updated the help file for the 2D data formats as well. @butlerpd you change looks good. Have a look at my change and merge when ready.

PR #1797, opened 2021-02-23
Batch Fitting Fixes for Options and Plotting Issues

### LINKED ISSUE #1446: 5.0 dI uncertainty unavailable in batch mode
The `dI` option for uncertainty (that is, uncertainty recorded in the data file) is marked as unavailable in the 5.0 GUI.  

To reproduce, load `test/1d_data/latex_smeared.xml` and *Send data to* *Fitting* with *Batch mode* checked.  Choose the *Fit Options* tab.  In the *Weighting* panel, *Use dI Data* is displayed as gray and cannot be checked. 

This is failing when running directly from the ESS_GUI branch as of 2020-01-23.

*Use dI Data* is available in the 4.2.2 release.

### LINKED ISSUE #1794: Batch fitting in 5.x returns different parameters to single fits in 5.x
User Y-CC reports:

_We recently encountered confusing fitting outcomes when we are using the batch mode to fit our data in SasView. Our data came from an operando measurement, so there are several SAXS datasets along the measuring time. We first tried to fit some of them individually (single fit) and we decided that only a certain Q-range (0.015 – 0.3 Å-1) is of our interest and got a reasonable fit within that range. However, when we switched to the batch mode and tried to fit several datasets with the same model and initial parameters copied from the single fit, the fitting results are quite different from the single fit for the same dataset. It seems like SasView fitted a much larger Q-range than the range that we assigned but it is unclear which Q-range that was. We wonder if you have come across similar issues or we are missing some critical steps here. Please find the log file as attached. I performed again the single and batch fitting right before writing this email, so hopefully you can find the latest activities at the bottom of the log._ 

[sasview.log](https://github.com/SasView/sasview/files/6023127/sasview.log)

### LINKED ISSUE #1795: Display of batch fitting results is broken in 5.0.4
Testing 5.0.4 RC3 whilst trying to assist User Y-CC (also see #1794).

When batch fitting concludes the Batch Fitting Results table is generated. According to the Help, you should be able to select a row and click Plot, or right-click on a dataset name and select Plot Selected Fits, to plot a graph of that model-fit on the selected dataset. This works in 5.0.3 but not in 5.0.4. 

As this renders Batch Fitting useless, because there is no other way to display or even obtain the batch fits, I'm making this issue a blocker on the release of 5.0.4.

### LINKED ISSUE #1796: Batch Fitting does not respect Q-range for fitting
Testing 5.0.4 RC3 whilst trying to assist User Y-CC (also see #1794 & #1795). But affects all versions of 5.0.x.

When Batch Fitting you are able to set the min and max Q values for the fit just as when using normal Single Fitting.

In 4.2.2 these Q-limits are respected by the fit. In all versions of 5.0.x they are ignored and the fit is performed over the full Q-range.

As this renders Batch Fitting pointless (you are not fitting what you think), I'm making this issue a blocker on the release of 5.0.4.

krzywon:
This PR ensures the batch fit page logic element isn't orphaned when a batch fit page is created which rendered all Q range and dI options useless. This work also ensures the q-range isn't improperly reset after a batch fit. Plots can also now be found when clicking the Plot button from the batch fit window.

Fixes #1796
Fixes #1795
Fixes #1794
Fixes #1446

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -210,6 +210,8 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             for data_item in value:
                 logic = FittingLogic(data=GuiUtils.dataFromItem(data_item))
                 self._logic.append(logic)
+            # Option widget logic was destroyed - reestablish
+            self.options_widget.logic = self._logic[0]
             # update the ordering tab
             self.order_widget.updateData(self.all_data)
 
@@ -1758,9 +1760,6 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
 
             # Switch indexes
             self.data_index = res_index
-            # Recompute Q ranges
-            if self.data_is_loaded:
-                self.q_range_min, self.q_range_max, self.npts = self.logic.computeDataRange()
 
             # Recalculate theories
             method = self.complete1D if isinstance(self.data, Data1D) else self.complete2D
--- a/src/sas/qtgui/Perspectives/Fitting/OrderWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/OrderWidget.py
@@ -35,7 +35,7 @@ class OrderWidget(QtWidgets.QWidget, Ui_OrderWidgetUI):
             if not hasattr(item, 'data'): continue
             dataset = GuiUtils.dataFromItem(item)
             if dataset is None: continue
-            dataset_name = dataset.filename
+            dataset_name = dataset.name
             self.order[dataset_name] = item
             self.lstOrder.addItem(dataset_name)
 
--- a/src/sas/qtgui/Utilities/GridPanel.py
+++ b/src/sas/qtgui/Utilities/GridPanel.py
@@ -356,11 +356,11 @@ class BatchOutputPanel(QtWidgets.QMainWindow, Ui_GridPanelUI):
         for i_row, row in enumerate(data):
             # each row corresponds to a single fit
             chi2 = row[0].fitness
-            filename = ""
+            name = ""
             if hasattr(row[0].data, "sas_data"):
-                filename = row[0].data.sas_data.filename
+                name = row[0].data.sas_data.name
             widget.setItem(i_row, 0, QtWidgets.QTableWidgetItem(GuiUtils.formatNumber(chi2, high=True)))
-            widget.setItem(i_row, 1, QtWidgets.QTableWidgetItem(str(filename)))
+            widget.setItem(i_row, 1, QtWidgets.QTableWidgetItem(str(name)))
             # Now, all the parameters
             for i_col, param in enumerate(param_list[2:]):
                 if param in row[0].param_list:


rozyczko (commented):
Changes look sensible.
Attaching logic to the options widget is indeed correct.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Win

butlerpd (approved):
testing on Windows:
1.  dI is not only selectable now it is the default if dI exists just as it is for normal fits.  I have not tested what happens if some of the data has dI and others do not
2. Selecting a row and choosing plot does (eventually) bring up the plots (all that are relevant:residuals, fit, distributions etc)
3. Testing RC3 the q range was always respected so maybe I don't know the sequence that caused "orpaning" of the batch fit page?  It always worked for me so perhaps someone else should test.

However, the plotting still has issues. For example clicking on a data set that was plotted already does nothing, not even bring that plot into focus.  Further, while rerunning the batch fit with different parameters (dI or no dI, differ q ranges etc) creates a new tabbed table of results each time (with different results) all tabs seem to be associated to the same plots.

This plot behavior however is consistent, and in fact already a slight improvement over, 5.0.3 where the underlying fit plot goes blank when trying to reselect it. I recommend therefore that this PR be merged and that a separate ticket be created regarding the batch plotting.

smk78:
@butlerpd  said:
> 3. Testing RC3 the q range was always respected so maybe I don't know the sequence that caused "orpaning" of the batch fit page?  It always worked for me so perhaps someone else should test.

Having carefully re-examined this, testing 4.2.2, 5.0.1, 5.0.3 and 5.0.4RC3 it appears @butlerpd is correct: the q-range was always respected!

[PR1797.pdf](https://github.com/SasView/sasview/files/6152519/PR1797.pdf)

I can only assume that previous testing didn't spot/forgot that batch fitting in 5.0.x was not honouring the use of dI if present in the data. But as noted above in 1) this is now cured.

> However, the plotting still has issues. For example clicking on a data set that was plotted already does nothing, not even bring that plot into focus. Further, while rerunning the batch fit with different parameters (dI or no dI, differ q ranges etc) creates a new tabbed table of results each time (with different results) all tabs seem to be associated to the same plots.

I note that if you display plots by selecting data from the batch results window, you cannot then plot any other set of data from any results tab unless you delete the existing plots first. This is inconvenient, but should be ticketed separately.

In summary, this PR is an improvement over 5.0.3 and looks good to merge.

krzywon:
Color me surprised (and I stand corrected) about the Q-range! I'm happy to hear the `dI` issue is resolved.

One final question - are the issues I note as fixed actually fixed by this PR?

butlerpd:
issues #1446  and #1794 are the same issue as noted in the comments to #1794 and both are thus fixed.  #1796 either is the same issue again or was not a problem apparently so also fixed.  #1795 is debatable I suppose from the way it is worded. However the spirit of it was that a feature that was working in 5.0.3 had been broken.  That **is** fixed by this PR so for the purposes of 5.0.4 I believe we are done. However we should ticket some of the annoyances with the batch fitting plots before we forget them.

Given Steve's review and the discussion at the fortnightly meeting I think we can merge this branch into 5.0.4 and close all 4 issues.

smk78:
Yes, I would concur. The issues as noted for this PR are fixed. Many thanks @krzywon !

Issue #1767, opened 2021-01-26
Too frequent fit update on parameter changes


butlerpd:
In many of the perspectives, editing of parameters triggers a recalculation. The problem is that in many cases this appears to happen for every key stroke.  This is particularly painful for example in the Fitting perspective with a complex model (computationally).  Changing the Q range triggers a recomputation of the model on every backspace or number typed (unless one types fast enough that some recalculations get skipped).

In most cases the computation should wait for either an apply or calculate button to be clicked or, at the very least wait for a return to signal the need for a recalc.  As it is changing the Q range for a fit can be unusable in some cases.

rozyczko:
Hmm are you sure the recalculation happens after each keystroke and not on leaving the edit field?
I can reproduce the recalculation request on hitting TAB, Enter or clicking outside the control but not on a simple keystroke.
Can you specify in which other fields (apart of Q min, Qmax) do you see this behaviour?

rozyczko:
Ah, this does happen but only in Jeff's Q range bars branch

butlerpd:
Ooops! Sorry @rozyczko and great sleuthing to figure it out anyway!. I was pulling different branches and testing and failed to note which one I was testing to get this (not sure I realized - my bad)

butlerpd:
Also Re @rozyczko previous question, one other place I know for sure I see this behavior (and I don't think it is restricted to just the slider branch) is when editing the Dmax in the P(r) perspective but there it is launching the background calculations which would be ok except that it throws errors onto the console until the final (reasonable) Dmax is entered.  Will look for other places and document here if/when I find them.

rozyczko:
I'm confused by this change, to be honest.
The original complaint was about model updates on each keystroke, and this was already fixed by you in 5.0.3.
Subsequent issue described by Paul had to do only with the Dmax line edit in P(r).

The current behaviour of the model being recalculated at each parameter change automatically has been with 5.0 since the beginning and I believe was widely accepted as it gives an easy way to see the effect of parameter modification on results.
Why would we want to change how this works all of a sudden? Are there any complaints about the current behaviour?

krzywon:
My PR was meant to draw out some discussion, honestly. The computation behavior of the fitting perspective is different between 4.2.2 and 5.0. In 4.2.2, the model is **not** recalculated after every parameter change, only when the compute or fit button is pressed. #1798 sets the behavior of 5.0.x to match that of 4.2.2.

[...]

PR #1798, opened 2021-02-24
Add Compute Button and Limit Fitting Computations

krzywon:
This PR adds a 'Compute' button to fit pages and limits recalculation to when the user presses this button, or hits the Enter or Return keys. This matches the behavior in 4.2.2.

Refs #1767

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -109,6 +109,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
     batchFittingFinishedSignal = QtCore.pyqtSignal(tuple)
     Calc1DFinishedSignal = QtCore.pyqtSignal(dict)
     Calc2DFinishedSignal = QtCore.pyqtSignal(dict)
+    keyPressedSignal = QtCore.pyqtSignal(QtCore.QEvent)
 
     MAGNETIC_MODELS = ['sphere', 'core_shell_sphere', 'core_multi_shell', 'cylinder', 'parallelepiped']
 
@@ -555,6 +556,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         self.cbFileNames.setVisible(False)
         self.cmdFit.setEnabled(False)
         self.cmdPlot.setEnabled(False)
+        self.cmdCompute.setEnabled(False)
         self.chkPolydispersity.setEnabled(True)
         self.chkPolydispersity.setCheckState(False)
         self.chk2DView.setEnabled(True)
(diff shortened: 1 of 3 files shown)


butlerpd:
Plot does the same thing as compute currently  we should consolidate this and use a better button name
call it plot/update and add a tooltip

butlerpd:
ready for testing on Win

butlerpd:
ready for testing on Win

butlerpd (changes requested):
I can confirm that compute/show now exists and works. However the `CR` key (or enter) does nothing despite the advertisement. Wondering if the problem is that the code only "recalculates plot data" but doesn't force a replot of that data?  At any rate the enter key not working needs to be fixed I guess before we can merge.

butlerpd:
another question: If the model is already calculated (as it is to start with) does the button force a recalculation anyway?

krzywon:
> If the model is already calculated (as it is to start with) does the button force a recalculation anyway?

Yes.

> However the `CR` key (or enter) does nothing despite the advertisement.

Testing the PR build, I can trigger a recalculation, but only when the data is already plotted and when the Fit panel has focus. I agree this should generate a new plot if one isn't present already but the fitting perspective should not capture the Enter key anytime it wants.

butlerpd:
Humm... I think we are saying the same thing?  To be clear, from my dumb user perspective, I enter a new value for one of the parameter and hit enter and absolutely nothing happen!! I think what you are saying is that it does "recalculate" which is what the code looks like it is doing.  But  I was unable to figure anyway to get the plot to update (the only recalculate that I ever see and thus that matters to me 😄 ) after entering a new value other than hitting the computer button.  I think we agree that hitting the enter key from random places should not trigger a "recalculate" ... and by "recalculate," as a user, I mean recalculate and re-plotted.

Is this the  behavior you get? Or are you saying you actually get it to replot properly when entering a value followed by a return?

krzywon:
> Is this the behavior you get? Or are you saying you actually get it to replot properly when entering a value followed by a return?

One other thing I see now - if you are editing a parameter (cursor in the param value box), the enter key exits the input box without an update. Another enter keypress is needed. The 1st enter should do both.

Things I need to fix:
- Enter/CR keys should generate a new plot if one isn't already plotted.
- Enter/CR keys should generate a calculation while editing a parameter.

butlerpd:
Ahhh.  You have to press CR twice to get it to plot the new curve.  That is one thing I did not try :-)  Yes one time should do it (exit the box and update the data/plot as you exit).

krzywon:
My latest changes should fix both of these now.

> Things I need to fix:
> - Enter/CR keys should generate a new plot if one isn't already plotted.
> - Enter/CR keys should generate a calculation while editing a parameter.

krzywon:
ready for testing on Win

butlerpd (approved):
The PR now behaves as advertized.  The code seems reasonable to the untrained eye given the rather convoluted behavior that is given to the `compute/plot` button.  I think that button behavior could be improved from the point of view of a researcher wanting to fit data and if that change is made there may be some code simplification possible.

That however is another ticket.  In the meantime I believe this can be merged

Issue #1732, opened 2021-01-13
Unit testing of qtgui


llimeht:
Python particularly rewards unit testing as there's only so much that static checkers such as pylint or mypy will pick up.

`src/sas/qtgui/GUITests.py` exists and has a number of tests that exercise parts of the `qtgui` tree, but those tests do not currently pass and so they are not enabled on PRs.

Some of the errors/failures appear to be errors in the tests, while some are errors in the code being tested.

A summary of the current situation is:

```
Running 56 test cases for calculatorsSuite
......F..F......F..........................FE...E......E

Running 56 test cases for mainSuite
..........................................F............F

Running 114 test cases for fittingSuite
..................................................................F...........E...................................

Running 85 test cases for plottingSuite
FF.EEEEE........Segmentation fault (core dumped)
```
(the segfault stopped the test prior to it getting to `utilitiesSuite` and `perspectivesSuite`)

A bit of work to do.

rozyczko:
I'm working on the `plottingSuite` tests as part of #1721

RichardHeenan:
Are any of these errors suitable for "scientists" (like myself) to investigate, or will they require a software guru?

krzywon:
@RichardHeenan you can check which tests are failing by running `python GUItests.py` from within the `qtgui` directory (make sure you've run setup.py first). Exception tracebacks and failed tests will be listed.

You can run subsections of the tests by adding one of the following arguments:
-  `calculatorSuite` runs the gui tests for the calculator apps like the data operation panel, SLD calculator, etc.
- `mainSuite` runs the tests for the main window and data browser
- `fittingSuite` runs all fitting perspective gui-related tests
- `plottingSuite` test the plotting routines
- `utilitiesSuite` tests things like the python window, logger, results window, etc.
- `perspectivesSuite` tests the corfunc, invariant, and P(r) perspectives

krzywon:
One note - PR #1715 separates the `perspectivesSuite` into `corfuncSuite`, `invariantSuite`, and `inversionSuite` and fixes all tests for the invariant perspective.

rozyczko:
I think this is critical enough to warrant adding to the hackathon list.

RichardHeenan:
I had a quick look at the errors list from GUITEst as was suggested,  a few of them may be due to test data issues, so I may be able to help there.

RichardHeenan:
I have had a look at some of this, after todays big PR's to Main for the Invariant etc.

If I do 
[...]

PR #1799, opened 2021-02-25
Various P(r) GUI Fixes

### LINKED ISSUE #1742: PR parameter buttons greyed out at wrong times
The P(r) perspective uses three parameters with which to do the computation, two of which, number of terms and the regularization constant, have suggested valued provided. These suggestions are based on background computations initiated as soon as any of the 3 parameter values changes. They show up on a button in the GUI and, if different from the current value, clicking that button will make the suggested value become the chosen value the actual value. provides two buttons to help the user choose the best parameters for the calculation buttons seem to grey out (inactive) when they should be active.P(r) perspective.  Normally the buttons are grey while the computation goes on and are ungreyed as soon as the background computation is finished and a suggested value available.

It appears as if the number of terms will grey out as soon as it is clicked and will mostly stay that way until a new value is computed that is different from what it was before, though it has been caught in a cycle when it would not ungrey even then.  The regularization parameter is much more robust but does also at times grey out.

This suggests the link between those computations and the GUI is not that robutst?

### LINKED ISSUE #1765: P(r) does not respect setting of manual background
The P(r) defaults to automatically estimating the background subtraction needed. It is critical to have the background properly subtracted from the data in order for the inversion to work.  For many users and cases that subtraction is not done and thus the automatic estimation is appropriate.  But sometimes the estimate is wrong and one would like to set it manually.

Currently there is a radio button feature to choose either "automatic" or "manual" subtraction. However, if one chooses manual and inputs a number, that number is ignored and reset to the estimated value. Note that the radio button continues to indicate "manual"

krzywon:
This PR fixes a handful of bugs associated with the P(r) Inversion perspective.

- The logic for activating the parameter buttons (No. of Terms and Reg. Constant) has been simplified. They are active whenever data is loaded and no calculation is occurring. Fixes #1742 
- The manual background input is now accepted and the toggle is not switched to manual input after every fit. Fixes #1765
- A few unticketed save/load issues, namely values not being updated properly, were also fixed.
- A few of the unit tests were fixed or updated but a couple of tests in the `inversionPerspectiveSuite` are throwing errors that I was unable to fix here. Refs #1732

This branch was built off of `release_5.0.4`, but this PR could wait until after 5.0.4 final.

### DIFF
--- a/src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
+++ b/src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
@@ -452,10 +452,7 @@ class InversionWindow(QtWidgets.QDialog, Ui_PrInversion):
     def check_q_low(self, q_value=None):
         """ Validate the low q value """
         if not q_value:
-            q_value = float(self.minQInput.text()) if self.minQInput.text() else ''
-        if q_value == '':
-            self.model.setItem(WIDGETS.W_QMIN, QtGui.QStandardItem(q_value))
-            return
+            q_value = float(self.minQInput.text()) if self.minQInput.text() else '0.0'
         q_min = min(self._calculator.x) if any(self._calculator.x) else -1 * np.inf
         q_max = self._calculator.get_qmax() if self._calculator.get_qmax() is not None else np.inf
         if q_value > q_max:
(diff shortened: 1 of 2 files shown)


butlerpd:
DO NOT MERGE .. wait till 5.0.4 branch is merged back into main then point this to main This will be for a future release.

butlerpd:
5.0.4 is now out so this can now be pointed to main?

butlerpd:
ready for testing on Win

butlerpd:
This works **much** better than before.  Certainly both #1742 and #1765 are now fixed.  I cannot really speak to the unit tests and of course much less the undocumented save/load issues that were presumably found while working on this?  Code sees OK to me but I am probably not the best qualified to judge?

Whatever the case, functionally this is ready to merge. do we need a separate code review?

butlerpd:
This has been approved at fortnightly meeting

PR #1559, opened 2020-05-28
Major Update to the Data Loader System (ESS_GUI)

krzywon:
This is a major refactor of the loader system and touches on multiple tickets and many files. ~~Due to the large number of changes and substantial changes done within qtgui, I do not plan on applying these changes to master unless there is a strong demand for it. Splitting this into multiple PRs might be possible, but there is a decent overlap in work done and files touched for each item fixed, so the separation may not be very clean.~~

**UPDATE:** This has been separated and finalized.

Smaller PRs created from this:
1. #1586 (ESS_GUI): Data loader fixes and enhancements for v5.0.3
2. #1602 (ESS_GUI) and #1619 (wxPython): Simplify Data Loader System
3. #1620 (ESS_GUI) and #1634 (wxPython): Organize unit test directory
4. #1625 (ESS_GUI) and #1635 (wxPython): Allow .dat reader to recognize rectangular data sets
5. #1652 (ESS_GUI): Display sample title instead of filename in data explorer
6. #1679 (ESS_GUI) and #1680 (wxPython): Dataloader fixes including CanSAS XML Refactor
7. #1802 (ESS_GUI): Retain native units on data load

Changes made in this PR and tickets associated with each change (NT == No Ticket):
- ~~#1465: The as-loaded units for each data object are now retained and kept as part of the data object. When data is sent to a perspective, the data is converted to sasview native units (if possible), otherwise, an warning message is given. A converter was also added to plots to allow for scaling data. Multiple unit tests were added or modified to be sure data is loading as expected.~~ (7)
- ~~#1544: The .DAT reader can now recognize rectangular data sets. Unit tests for rectangular and square data sets were added.~~ (4)
- ~~#1243: The data name displayed in the Data Browser now defaults to the sample title rather than the file name. Saving and loading projects also now follow this paradigm.~~ (5)
- ~~#1303: The canSAS XML reader has been completely overhauled using a psuedo-switch statement rather than nested if-elses for maintainability (and a small load speed increase).~~ (6)
- ~~#1002: Transmission spectra are now saved when saving in the canSAS XML format. Unit tests were added to verify this is true.~~(1)
- ~~#726: The canSAS XML reader now defaults to cansas v1.1.~~(1)
- ~~#506: Partly touches on this ticket by removing 2 exec calls in associations.py.~~ (2)
- ~~NT: Changed how the loader system falls to the 'default' readers, simplifying the error handling, allowing for better error capturing. Unt tests were updated to conform to these changes.~~ (2)
[...]

PR #1802, opened 2021-03-10
Retain native units on data load

### LINKED ISSUE #1465: Units selection for input data, simulated data, plotted data
Hi all, 

We're exclusively using 1/(m sr) for scattering cross-section and 1/nm for scattering vector in our lab (although this would be 1/m if I had my way, but never mind that). However, SasView is exclusively working in 1/cm and 1/Å for its units. This makes working with SasView difficult for us in three cases:
  1. When loading data without units (three-column data), the wrong units are assumed. A preference menu where the default units can be specified would be helpful. 
  2. When simulating data without an input dataset, it assumes 1/cm and 1/Å, which is hard for us to use. Likewise, dimensions cannot be entered in nm for the models, have to be input in Å. A preference menu where the output units for plotting (and maybe model entry) can be specified would be helpful. 
  3. When loading a dataset with units, the plotted data units should follow the units of loaded data.

### LINKED ISSUE #1575: The celsius/kelvin conversion in nxsunit is not correct.
Celsius and kelvin temperature units are currently considered equivalent, with no offset between them.

A related implementation from another project is attached, with support for celsius, kelvin and Fahrenheit.

[nxsunit.txt](https://github.com/SasView/sasview/files/4714202/nxsunit.txt)

krzywon:
This was supposed to be the 7th, and final, PR separating SasView/sasview#1559. Due to the size, I am separating it into at least 7 more pull requests. I am using this to track the work as I go.

Anticipated pull requests:
 - sasdata package
   - refactor nxsunit - sasview/sasdata#13
   - separate current data units from units that were loaded
   - helper methods for scaling data and no longer scale to sasview native units on loading
 - sasview package
   - preferences window - #2167
   - ability to scale data to any compatible units and only scale to sasview native units on fitting
   - data loading preferences
   - plotting preferences


toqduj (commented):
Looking very good so far. However, there is one usability issue remaining (also because Olivier Tache brought it up): we need to be able to set the input/output units. for Q (+Q Uncertainties) and I (+I Uncertainties). 
I'm not sure, but an easy way of setting these might be by adding two sub-menus in the "file" menu, called "Set in/out Q units", and "Set in/out I units". These could contain a list of options, one of which has a checkmark next to it, similar to the example in the attached screenshot. Would that be possible?

Screen Shot 2021-03-17 at 10 06 48

krzywon:
Thanks for reviewing and the suggestion, @toqduj. My original plan was to add the feature you are suggesting, but didn't think it would be useful.

Would extra inputs on the data explorer be helpful or just add clutter? I'm thinking an extra section, just below the load data button, with two text boxes for default I and Q units (if none are found in the data file). I would sanitize the inputs and give a warning if the units aren't recognized. This would only effect the data as-loaded.

Exported data take the current state of the data including units (x_unit and y_unit).

krzywon:
Here is a quick mock-up of what I'm proposing. Input welcome.

unit_text_boxes

gonzalezma:
It seems good to me, but if you are planning to allow a completely free input from the user, I guess the parsing and the handling of the units may become difficult. What about just a choice menu with the most usual choices (A-1, nm-1, cm-1, what else?).

smk78:
How would what is proposed work if the data you want to use, say, for simultaneous fitting, had a mix of Q & I units?

butlerpd:
It does look visually appealing but I am getting very concerned (I believe I have made that comment before 😄  of the amount of stuff permanently on the GUI.  It is a delicate balance between making it easy for expert users to find all the bells and whistles and making it easy for new users to access. I think we should work to minimize the clutter that puts a huge barrier to entry whenever possible.

Being on a zoom call with Adrian Rennie and @ajj they both agreed that this should not be part of the permanent GUI as envisioned.  The suggestion was that if the file being read does not have units it should pop up asking for units at that time.  Also instructions on how to format the units so they will be properly recognized would be helpful in said popup I guess.

Not sure this is the right answer (one has to think through the various consequences), but I think we need to start asking the question regularly when adding to the main page of the GUI: is this absolutely necessary to add to the main page?

My 2 cents anyway. Should we have a discussion at our fortnightly meeting about GUI structure?

toqduj:
In light of the discussion, I would still have indicated a "default units" in the menus. It's out of the way but can be adjusted if necessary by expert users. It also then defines the output units in graphs..

src/sas/qtgui/Plotting/Plotter2D.py
```diff
@@ -222,6 +225,18 @@ def onClearSlicer(self):
         if self.slicer_widget:
             self.slicer_widget.setModel(None)
 
+    def onUnitsChange(self):
```
rozyczko, line 228: This method is identical to that in `Plotter.py`. Can it be refactored to `PlotterBase`?

rozyczko (commented):
Probably doing something wrong but I got it to fail with a simple conversion:
`cyl_400_40.txt` dataset, converting Q to `nm^{-1}` and I to `cm^{-1}` gave me:

```
12:14:48 - ERROR: Traceback (most recent call last):
  File "D:\projects\sasview\src\sas\qtgui\Plotting\Plotter.py", line 404, in onUnitsChange
    plot.convert_q_units(self.units.cbX.currentText())
  File "D:\projects\sasview\src\sas\sascalc\dataloader\data_info.py", line 185, in convert_q_units
    self.x_unit = unit
  File "D:\projects\sasview\src\sas\qtgui\Plotting\Plottables.py", line 456, in __setattr__
    object.__setattr__(self, name, value)
  File "D:\projects\sasview\src\sas\sascalc\dataloader\data_info.py", line 117, in x_unit
    self.x = self.x_converter(self.x, x_unit)
  File "D:\projects\sasview\src\sas\sascalc\data_util\nxsunit.py", line 478, in __call__
    raise exc
KeyError: 'n m^{-1} is not a Q'
```

butlerpd:
I would agree with @toqduj. Having it as a menu item is probably best.  considering most files do not have units, or at least not units that are recognized by SasView, Popping up a dialog every time will become really old -- particularly when loading a folder full of them. In light of the discussion I would alter my suggestion then to: read in units from the file. If none assume the units as SasView always has (no behavior change = minimal surprises/confusion for users). Provide menu item for expert users to be able to change the default units etc.

butlerpd:
Does this have implications for data converter?

pkienzle:
Wouldn't it be easiest to modify the reduction program to store the units in the file? This could be as simple as "# key: value" pairs before the column data, which makes it pretty easy to write and parse. With value encoded as JSON it is very flexible. The ascii reader would need to define the expected keys and the layout of the corresponding values.

If you can't modify the reduction code, maybe a file reader that sets the units based on file extension?

Shouldn't display units be a user preference independent of storage units? If you are used to looking at curves with Q as 1/nm or 1/m wouldn't you still want to see those units even if you are using data from a different instrument?

Similarly for model parameters, would a user want to see SLDs in units of 1/m²?

Internal units are fixed by the calculation engine(s). We could carry units with us on every input to the calculation, but that amounts to the same thing just with more bookkeeping.

Note that there are consequences to the choice of internal unit. Some fitting engines use step size in parameter space for derivative estimation and for stopping conditions. If SLDs are expressed as 1/m² then values are going to be in the range of 1e-26, and all reasonable steps will be infinitesimal. Parameter values will ideally be in [0.01, 1000], which gives 6-9 digits of precision on the parameter with the usual 1e-8 step size for numerical derivatives.

krzywon:
Here is another mock-up for a default units selection tool. This one uses a cascading menu, not only for loading data but also when plotting. These menus would be dynamically populated using the same methods used to populate the combo boxes in the unit change window. I plan to optimize those methods by removing duplicates (e.g. we only need one of 1/c_m^2 and 1/cm^2). The SasView defaults of 1/cm for intensity and 1/A for Q would be auto-selected on load unless changed, and then the user-selected option would persist between SasView instances.

sasview-default-units-mock-up

butlerpd:
I assume this is to set the default behavior for the user? but does not allow changes after loading for specific data which you realize you got from somebody who uses 1/m instead of your usual 1/cm?... or even plots where there may be a particular data set that makes more sense to plot in different units from your default?

If I interpret correctly, we would still then probably need the ability to do one-off fixes (perhaps as an option in the data context menu?).  Having global defaults is important, particularly for plotting, but also for reading data if your usual data comes in 1/m and is not in canSAS format so that it can so render.  Ultimately this should probably go into the local_config file.  We would then need the preferences menu that allows the user to change those values as I think is documented in another ticket.  But for an immediate implementation the suggested mockup would do it I think?

krzywon:
@butlerpd, your assumptions are correct. The 'Default I|Q Units on Loading' options would only be applicable when loading data. The preferences would be stored in the local config which is how I would know what values to select when opening SasView.

This PR already covers the case of units for individual plots. The selected menu option would still be the units used when the plot is initially loaded, but the user could always change the units scale for one plot and not another.

I'll have to think more about how to handle datasets where the incorrect units were selected for loading. A context menu option seems like the way to go if we allow that. Two unit menu options, one that scales the data to a particular unit and another that changes the units of the existing data but does not scale the data may create confusion, so if we decide to go that route, I'll have to document the options.

butlerpd:
The issue with only providing a default units for data that came in via the generalized reader for example is that not all data are ever going to be in the same units.  In particular a lot of us still use 1/A but an increasing number of places are moving to SI units (which **coud be** 1/m but are usually in 1/nm)  So fixing it to always being one thing is not a good solution I think.

I think any confusion goes away in the longer term as the addition of those options as separate options in the file menu is not the proper way to deal with configuration items. We have long needed a proper "preferences" menu item which would hide all of that. Unfortunately, that is probably not going to happen in the short term so some thought on avoiding confusion between setting a **permanent preference** vs a true context option is going to be needed in the short term I guess?

butlerpd:
This is getting too big and complicated. @krzywon says we should rethink. Closing this PR but keeping the branch for further work

krzywon:
I've started the separation of this PR in sasview and sasdata. I'm reopening this PR as a draft so I can track my separation progress.

lucas-wilkins:
This gets a star too :)

Issue #430, opened 2019-03-30
Need to completely separate GUI from computational code (Trac #295)


butlerpd:
Migrated from http://trac.sasview.org/ticket/295
```json
{
    "status": "assigned",
    "changetime": "2018-09-05T13:18:10",
    "_ts": "2018-09-05 13:18:10.853057+00:00",
    "description": "This is a mojor Refactoring need and critical to the future of the project.\n\nSuggestion during discussion of this issue at fortnihgtly meeting of Oct 21, 2014 is to bundle all of calculation (data loading, SANS models, P(r) etc.) into one package and UI into another.",
    "reporter": "butler",
    "cc": "",
    "resolution": "",
    "workpackage": "SasView Framework Enhancements",
    "time": "2015-01-31T17:02:54",
    "component": "SasView",
    "summary": "Need to completely separate GUI from computational code",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView 5.0.0",
    "owner": "richardh",
    "type": "enhancement"
}
```

RichardHeenan:
Trac update at `2016/03/24 18:39:03`: **richardh** commented:

If we are going to rewrite the sasview gui in py tk, then what are the pros & cons of putting sasview inside Mantid and using its workspaces and plotting etc ?

Richard

butlerpd:
Trac update at `2017/10/23 15:05:17`: **butler** changed milestone from "SasView WishList" to "SasView 5.0.0"

rozyczko:
Will be addressed as part of the headless SasView project for live analysis (for 5.1)

PR #1805, opened 2021-03-20
Release notes updated

wpotrzebowski:
Release notes for 5.0.4 release update

### DIFF
--- a/docs/sphinx-docs/source/user/RELEASE.rst
+++ b/docs/sphinx-docs/source/user/RELEASE.rst
@@ -32,13 +32,12 @@ of 5.0.x:
 
 There are also some new features in this version:
 
-* A new generalized 3D description of magnetic SANS with an arbitrary field/polarisation
-  axis is introduced. In particular, this facilitates the description of crystals with
-  strong magnetic anisotropy, or data from experiments where only a small magnetic field
-  is used to set the polarisation, or where an AC field configuration is used for time-resolved
-  SANS (TISANE).
-* The slider bars for selecting the q-range for fitting that featured in earlier
-  versions of SasView have been re-introduced by popular request.
+* Though not strictly a new feature, the functionality and operation of parameter
+  constraints has been significantly overhauled for this version.
+* The slicer functionality has been significantly overhauled and made to work properly.
+* The slider bars on plots for selecting the q-range for fitting that featured in earlier
+  versions of SasView have been re-introduced by popular request, although they do not
+  yet work on linearized plots.
 * It is now possible to swap the dataset used to create an existing FitPage for a different
   dataset. This removes the need to re-generate a complex model (eg, featuring many parameters
   and/or constraints) many times over to use it to fit several datasets.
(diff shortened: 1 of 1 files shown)


wpotrzebowski (commented):
Looks good to me!

butlerpd:
First note: I agree that we should not let perfection be the enemy of the good and in the absence of a good clean way of sorting the release notes, I agree we should generally accept this. That said, to be clear, the way we use issues and close them I think we are over stating what was done. Often issues are closed because PART of the issue was dealt with and the rest is put into a new issue. Also there are issues that get closed sometimes well after they were fixed in a previous version.  The change is often documented in that previous version if the commits list was used in generating those notes which points to the fact that we also may be understating by not looking at the list of commits since 5.0.3.

The key for me however is that we **must** be absolutely resolute that once a tag is created no changes of any kind (not even the deletion or addition of a single character) should be made to that code anymore.  The point of the tag is to be a permanent marker of the exact state at a given time.  So if we change our mind we will need to make a new tag.  Changes can of course be made on the web page where the release is stored IMO.

butlerpd (changes requested):
subject to caveat left in comments there are two issues
1. line 116 and 119 are duplicates - one of these should be deleted
2. the notes suggest that the magnetic PR recently merged into main of sasmodels is part of 5.0.4.  I thought we had specifically agreed that was **not** the case.  the unintended knock on effects of such a major change on sasview as a whole has not been tested.  This speaks to both the notes and at what point sasmodels is being tagged.  This must be addressed before we can release.

smk78:
@butlerpd said:
> subject to caveat left in comments there are two issues
> 
> 1. line 116 and 119 are duplicates - one of these should be deleted

Have removed line 116

> 3. the notes suggest that the magnetic PR recently merged into main of sasmodels is part of 5.0.4.  I thought we had specifically agreed that was **not** the case.  the unintended knock on effects of such a major change on sasview as a whole has not been tested.  This speaks to both the notes and at what point sasmodels is being tagged.  This must be addressed before we can release.

On checking, sasmodels PR #430 was merged into sasmodels _master_, not the _release_5.0.4_ branch, so it is **not** included in this release.

> 2. sasmodels #430 is a new feature (and probably not part of this release? see next) and not a bug fix

Pursuant to Point 3 above, have deleted line 127 as was (Fixes sasmodels #430...)

Have also removed the corresponding bullet point under 'new features in this version'.

> 4. The slider bars have not been re-instated completely as they do not yet work on linearized plots

Have revised the bullet point to read: _The slider bars on plots for selecting the q-range for fitting that featured in earlier versions of SasView have been re-introduced by popular request, although they do not yet work on linearized plots._

In addition, have added the following text under Known Issues in this version: _At this time, the reinstated slider bars on plots for selecting the q-range for fitting do not work on linearized plots._ 

**However**, having played around with a local build of the _release_5.0.4_ branch I have to say I am not entirely sure what we mean by a 'linearized plot'...!

> 5. I believe the slicer functionality (a parity requirement) is new to 5.0.4. There was a very simple and broken implementation in 5.0.3 that was mostly useless as I recall

Have added the following bullet under 'new features in this version': _The slicer functionality has been significantly overhauled and made to work properly._

> 6. There were also quite a few additions to the constraints functionality by Nico well into September. 5.0.3 was released on July 7. Thus 5.0.4 contains quite a few improvements to the constrained fitting feature as well. There are of course several yet to be finished but they will have to go to a future release.

Have added the following bullet under 'new features in this version': _Though not strictly a new feature, the functionality and operation of parameter constraints has been significantly overhauled for this version._

PR #1807, opened 2021-03-24
1804: Remove redundant checks so dx==0 is allowed

### LINKED ISSUE #1804: v5 zero value for dx causes resolution smearing to not happen
Having just had a data set where one single dX value was zero, and then spent ages trying to work out why I could not ask for Q resolution smearing (which I would have expected to be on by default), could the loader please flag this as an issue.

This may perhaps be the reason behind some other mysterious complaints ?

krzywon:
This changes how `SmearingWidget.py` determines which resolution should be used. The data checks that have been removed from the widget in this PR are already being performed in `src/sas/sascalc/fit/qsmearing.py`. On top of that, the checks were inconsistent in the two places.

fixes #1804

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
@@ -7,10 +7,13 @@ from PyQt5 import QtCore
 from PyQt5 import QtGui
 from PyQt5 import QtWidgets
 
-from sas.sascalc.fit.qsmearing import smear_selection
+from sas.sascalc.fit.qsmearing import smear_selection, PySmear, PySmear2D
 from sas.qtgui.Plotting.PlotterData import Data1D
 from sas.qtgui.Plotting.PlotterData import Data2D
 import sas.qtgui.Utilities.GuiUtils as GuiUtils
+
+from sasmodels.resolution import Slit1D, Pinhole1D
+from sasmodels.sesans import SesansTransform
 
 # Local UI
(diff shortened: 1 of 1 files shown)


wpotrzebowski:
ready for testing on OSX

src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
```diff
@@ -146,7 +149,8 @@ def updateKernelModel(self, kernel_model=None, keep_order=False, old_index=None)
             self.setElementsVisibility(False)
             return
         # Find out if data has dQ
-        (self.smear_type, self.dq_l, self.dq_r) = self.getSmearInfo()
+        self.current_smearer = smear_selection(self.data, self.kernel_model)
+        self.getSmearInfo()
```
rozyczko, line 153: Is there any reason to have `smear_type`, `dq_l` and `dq_r` as state? The only place we are using them is in populating the boxes, and the getter is only called once. I think keeping these three as local variables is slightly cleaner.

rozyczko (approved):
Minor stylistic issues

src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
krzywon: `smear_type`, `dq_l `and `dq_r` are used when the Q-resolution combo box changes to dQ, which could happen when a user wants to check a fit without any Q resolution effects and then switches back to the original fit.

PR #1809, opened 2021-03-29
Pin versions of qtconsole and PyZMQ to enable QtConsole.

rozyczko:
qtconsole=4.7.5
pyzmq=19.0.1

probably a bit too conservative, but this is what ran fine on 5.0.3


wpotrzebowski (approved):

wpotrzebowski:
Since it is touching commercial yaml file we can only test it by merging to release branch, which I suggest  doing

PR #1829, opened 2021-04-25
Release 5.0.4

wpotrzebowski:
Merging release_5.0.4 after it's been released


butlerpd (approved):
This PR is to merge the final changes rolled out in Release 5.0.4 back into the main branch.  The only failing test is the travis CI which seems to be failing because it was set up wrong (no python 3.7 installed).  All the github actions are passing as they should for released code. Thus I see no reason not to merge this.

PR #1849, opened 2021-05-13
Update h5py version for all operating systems

### LINKED ISSUE #1848: Increment h5py package
NXcanSAS files produced using Igor Pro 8 are loading into SasView using the h5py package released with SasView (v2.9.0 win, v2.8.1 osx), but the Igor Pro 9 beta added stricter HDF5 file writing capabilities. At least one issue related to this was fixed in h5py v2.10.0. Locally, I was able to locally upgrade to v3.1 without upgrading any other packages.

This likely affects other NXcanSAS files.

krzywon:
This pins the h5py version to 3.1 for all operating systems.

Fixes #1848


krzywon:
ready for testing on OSX

butlerpd (approved):
Looks good but the conflict needs to be fixed first

rozyczko (commented):
My standard comment about being wary of using `>` in versioning - future versions of `h5py` might start pulling in updated dependencies which will break the installer.
Also, we are locking `hdf5` versioning, maybe this also should be released if `h5py` is using current?

PR #1853, opened 2021-05-18
Adding pip dependency to yml file so that conda doesn't complain

wpotrzebowski:
When trying to create env from yml file with recent versions of conda it throws thousands of warnings:
`Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.`

I observed it on OSX and added pip to developers build (min file). It needs to be explored for other platforms.


wpotrzebowski:
ready for testing on Linux

rozyczko (approved):
tested on windows, behaves as expected

Issue #1855, opened 2021-05-18
Update python and dependent modules - the pyinstaller step.


rozyczko:
Python version and external modules we rely on needs updating.
Currently we freeze the versioning of most modules for stability but it has been a while since we upgraded the major components like MPL, numpy, pyopencl etc.

I tested SasView with the most up-to-date versions and the dev build works perfectly well.
However, packaging it with pyinstaller fails, so fixing this is the main task here.

butlerpd:
@rozyczko can you update? with github actions and other things going on is this still a thing even? or should we close the ticket?

rozyczko:
Indeed, GH actions solved this problem. Closing.

PR #1856, opened 2021-05-18
Add py3.9 remove py3.6 from github actions

llimeht:
First step towards #1855 is to see what we can already do in the GitHub Actions CI for tests, docs and installer, which this PR does.

 * the pytest tests pass (with the usual failures not actually causing the tests to fail!)
 * the GUI tests still fail in the normal ways and that failure is suppressed
 * building the docs succeeds
 * building the installer succeeds (but that is not the same as saying that the installer works)

The installer is at https://github.com/llimeht/sasview/actions/runs/852749020 if windows and mac users would like to test it out.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8]
+        python-version: [3.9]
 
     steps:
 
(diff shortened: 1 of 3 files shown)


wpotrzebowski:
ready for testing on Windows

wpotrzebowski:
ready for testing on Win

wpotrzebowski (approved):

rozyczko (approved):
Adding 3.9 is good. Removing 3.6 is even better.

PR #1859, opened 2021-05-20
Fixes to invariant plot high q extrapolation

### LINKED ISSUE #1749: Invariant not extrapolating over listed high-q
The invariant perspective does not extrapolate over the entire high-q range listed in the GUI (default 10 Angstrom<sup>-1</sup>). The total number of points in the extrapolated data is always 500. The low-q extrapolation goes to the expected low-q value, but often using fewer points.

Example: load test\1d_data\AOT_Microemulsion-Drop_Contrast.xml into the the invariant perspective, with the extrapolation limit at the default and the extrapolation ends at 0.843 Angstrom<sup>-1</sup> over 500 points.

This issue has been present since at least release v4.2.2.

More comments and examples can be found in PR #1718.

phgilbert:
Removes 3x upper limit on invariant high Q extrapolation plot, and replaces with user defined upper q limit.

Fixes #1749 (High Q Part)

### DIFF
--- a/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
+++ b/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
@@ -23,8 +23,6 @@ from .InvariantUtils import WIDGETS
 Q_MINIMUM = 1e-5
 # The maximum q-value to be used when extrapolating
 Q_MAXIMUM = 10
-# the ratio of maximum q value/(qmax of data) to plot the theory data
-Q_MAXIMUM_PLOT = 3
 # Default number of points of interpolation: high and low range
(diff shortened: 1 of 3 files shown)


butlerpd:
ready for testing on Win

rozyczko (commented):
Code looks good, would be useful to add a unit test for this change.
Do the current unit tests pass?

krzywon (approved):
I don't see a way to make the test I suggested work without completely reworking `InvariantPerspectiveTest.py`.

I suggest we merge this as-is and I'll ticket the issue for the invariant tests.

wpotrzebowski:
ready for testing on OSX

PR #1862, opened 2021-05-23
Increment version number

butlerpd:
This just increments the version number and adds the alpha.1 tag post release 5.0.4 for ease of verifying which version one is working with/testing as suggested in release notes (and was originally meant to be done automatically by adding the commit hash to the about box which has never worked on the build systems).


rozyczko (approved):
These changes look good and straightforward.

It looks like the version check (comparing the two version strings) also works properly, so as long as we stick to proper naming (`-alpha.n+1` after `-alpha.n` and `-beta.x` after `-alpha.x`) we should be fine

Issue #1870, opened 2021-06-18
Issues with Image Viewer generated data


rozyczko:
2 issues:

1. Import an image to the Image Viewer, e.g. image_data/ISIS_98940_greyscale_png.png
2. Save data as dataset
3. Send the dataset to fitting
4. Select a model and click on Show Plot
5. Nothing gets shown

Additionally, attempting to update a 2D plot one gets an error about `plot` not having `plot_dict` attribute.

PR #1871, opened 2021-06-18
Image Viewer bug fix

rozyczko:
Set the correct 2D dataset name.
Avoid querying non-existing attributes.
Addresses #1870


krzywon (approved):
Looks good. Tested locally and the issue seems to be resolved.

Issue #1872, opened 2021-06-18
New version string confuses StrictVersion check


rozyczko:
StrictVersion() call on the new format version string `5.0.5-alpha.1` seems to confuse distutils.version.StrictVersion() check with default versioning of distutils (3.6.13)

Attempting to start SasView on either Linux or Windows fails with
```
ValueError: invalid version number '5.0.5-alpha.1'
```

butlerpd:
Interesting --- I had forgotten we do use the version number for one thing besides printing the string in various places: "checking whether I am the latest version."  I guess that is where we are using StrictVersion rather than LooseVersion?

As I understand it, PEP 440 covers StrictVersion? If so I note that besides a, b and rc (instead of alpha, beta etc,) a dev and post tags are also allowed?  Given the rationale for adding the tag after a release, the dev tag may be more appropriate than a(lpha). From what I can tell we should avoid using the post tag.  So, my suggestions is that we replace the original suggestion of using alpha for nightly builds with dev.  Thus immediately upon release of 5.0.4 for example, the version should be incremented to: 5.0.5dev1

It may also be appropriate in the init.py to add a comment indicating strict versioning only.

butlerpd:
Of course the above proposal assumes that dev resolves as preceding a which precedes b which precedes rc.  It should based on the description but {a,b,rc} follow the natural lexical order but dev would not so we should probably check this before adopting it.

butlerpd:
This should have been closed with the merging of #1874 but apparently we failed to link it to that PR. Closing it now

PR #1874, opened 2021-06-21
Fixing version number

wpotrzebowski:
The previous proposal with 5.x.x-alpha1 didn't work with StrictVersion from distutils. The alternative is to use a1 label e.g. 5.0.5a1


wpotrzebowski:
ready for testing on OSX

butlerpd:
Does this mean that the `dev` nomenclature suggested in #1872 from PEP 440 not resolve properly?

wpotrzebowski:
ready for testing on Windows

wpotrzebowski:
ready for testing on Win

docs/sphinx-docs/source/conf.py
```diff
@@ -79,7 +79,7 @@
 # The short X.Y version.
 version = '5.0'
 # The full version, including alpha/beta/rc tags.
-release = '5.0.5-alpha.1'
+release = '5.0.5a1'
```
krzywon, line 82: The existing docstring is no longer valid and should be changed. Adding a link to the StrictVersion rules should suffice.

http://epydoc.sourceforge.net/stdlib/distutils.version.StrictVersion-class.html

src/sas/sasview/__init__.py
```diff
@@ -1,5 +1,5 @@
 from distutils.version import StrictVersion
-__version__ = "5.0.5-alpha.1"
+__version__ = "5.0.5a1"
 StrictVersion(__version__)
```
krzywon, line 3: Same as my other comment.

krzywon (changes requested):
This fixes the loading issue, but the docstrings need to match the functionality or someone might repeat this mistake in the future.

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
ready for testing on Linux

krzywon (approved):

krzywon:
I havent looked at why the builds are failing in this case, but everything here looks good now.

wpotrzebowski:
> I havent looked at why the builds are failing in this case, but everything here looks good now.

Yes, builds are still failing. I am trying to look into this.

Issue #1800, opened 2021-03-03
No Check Model functionality in 5.x


smk78:
In 4.x, the Advanced Plugin Model Editor incorporates functionality to check the validity of the code in the selected plugin (Run > Check Model):

![image](https://user-images.githubusercontent.com/10938679/109783959-75056a00-7c02-11eb-9c7e-e03806dc7a8f.png)

The output is extremely valuable for debugging plugin models:

![image](https://user-images.githubusercontent.com/10938679/109783895-5f904000-7c02-11eb-9240-698911dde17c.png)

In this instance, for example, highlighting that the User had used _define random():_ instead of _def random():_

In 4.x this output **is not** duplicated in either the Console or _sasview.txt_.

Unfortunately there is no equivalent Check Model functionality in 5.x (the following screenshots are from 5.04 RC3):

![image](https://user-images.githubusercontent.com/10938679/109785331-ee518c80-7c03-11eb-9068-ed8c019d50e1.png)

![image](https://user-images.githubusercontent.com/10938679/109785526-1ccf6780-7c04-11eb-8410-ccdde4f602db.png)

The issue here is that if 5.x cannot compile a plugin then it simply passes over it. Nothing to say it could not compile a plugin is output to the Log Explorer or _sasview.txt_. This actually makes 5.x pretty much useless for developing or tweaking a plugin by editing the code.

Can the Check Model functionality be reinstated in 5.x?

rozyczko:
Yes, this is indeed missing. `Wx` had a nice python editor/syntax checker built in, but `Qt` has nothing like that so far.
Reimplementing this functionality in Qt will be a pretty hefty effort, unless there is a ready component out there we can reuse.

rozyczko:
QScintilla2 might be a solution to that... investigating

rozyczko:
using `ast` along with showing the results with `traceback` seems to be doing the trick.

butlerpd:
Just did a quick search on ast -- that does seem promising!!

rozyczko:
Added a syntax check for the edit model window.
Also, added support for line numbers and current row highlighting

branch: `improved_model_check`

krzywon:
This was fixed with #1875

PR #1875, opened 2021-06-21
Syntax checker in the model editor.

rozyczko:
Also, replaced QTextEdit with a nicer editor, with line number printout and current row highlight

Addresses issue #1800

### DIFF
--- a/src/sas/qtgui/Utilities/CodeEditor.py
+++ b/src/sas/qtgui/Utilities/CodeEditor.py
@@ -1,0 +1,96 @@
+
+from PyQt5.QtCore import Qt, QRect, QSize
+from PyQt5.QtWidgets import QWidget, QPlainTextEdit, QTextEdit
+from PyQt5.QtGui import QColor, QPainter, QTextFormat
+
+
+class QLineNumberArea(QWidget):
+    def __init__(self, editor):
+        super().__init__(editor)
+        self.codeEditor = editor
+
+    def sizeHint(self):
+        return QSize(self.editor.lineNumberAreaWidth(), 0)
+
+    def paintEvent(self, event):
+        self.codeEditor.lineNumberAreaPaintEvent(event)
+
+
+class QCodeEditor(QPlainTextEdit):
+    def __init__(self, parent=None):
+        super().__init__(parent)
+        self.lineNumberArea = QLineNumberArea(self)
+        self.blockCountChanged.connect(self.updateLineNumberAreaWidth)
+        self.updateRequest.connect(self.updateLineNumberArea)
+        self.cursorPositionChanged.connect(self.highlightCurrentLine)
+        self.updateLineNumberAreaWidth(0)
+
+    def lineNumberAreaWidth(self):
+        digits = 1
+        max_value = max(1, self.blockCount())
+        while max_value >= 10:
+            max_value /= 10
+            digits += 1
+        space = 3 + self.fontMetrics().width('9') * digits
+        return space
+
+    def updateLineNumberAreaWidth(self, _):
+        self.setViewportMargins(self.lineNumberAreaWidth(), 0, 0, 0)
+
+    def updateLineNumberArea(self, rect, dy):
+        if dy:
+            self.lineNumberArea.scroll(0, dy)
+        else:
+            self.lineNumberArea.update(0, rect.y(), self.lineNumberArea.width(), rect.height())
+        if rect.contains(self.viewport().rect()):
+            self.updateLineNumberAreaWidth(0)
+
+    def resizeEvent(self, event):
+        super().resizeEvent(event)
+        cr = self.contentsRect()
+        self.lineNumberArea.setGeometry(QRect(cr.left(), cr.top(), self.lineNumberAreaWidth(), cr.height()))
+
+    def highlightCurrentLine(self):
+        extraSelections = []
+        if not self.isReadOnly():
+            selection = QTextEdit.ExtraSelection()
+            lineColor = QColor(Qt.yellow).lighter(160)
+            selection.format.setBackground(lineColor)
+            selection.format.setProperty(QTextFormat.FullWidthSelection, True)
+            selection.cursor = self.textCursor()
+            selection.cursor.clearSelection()
+            extraSelections.append(selection)
+        self.setExtraSelections(extraSelections)
+
+    def lineNumberAreaPaintEvent(self, event):
+        painter = QPainter(self.lineNumberArea)
+
+        painter.fillRect(event.rect(), Qt.lightGray)
+
+        block = self.firstVisibleBlock()
+        blockNumber = block.blockNumber()
+        top = self.blockBoundingGeometry(block).translated(self.contentOffset()).top()
+        bottom = top + self.blockBoundingRect(block).height()
+
+        # Just to make sure I use the right font
+        height = self.fontMetrics().height()
+        while block.isValid() and (top <= event.rect().bottom()):
+            if block.isVisible() and (bottom >= event.rect().top()):
+                number = str(blockNumber + 1)
+                painter.setPen(Qt.black)
+                painter.drawText(0, top, self.lineNumberArea.width(), height, Qt.AlignRight, number)
+
+            block = block.next()
+            top = bottom
+            bottom = top + self.blockBoundingRect(block).height()
+            blockNumber += 1
+
+
+if __name__ == '__main__':
+    import sys
+    from PyQt5.QtWidgets import QApplication
+
+    app = QApplication(sys.argv)
+    codeEditor = QCodeEditor()
+    codeEditor.show()
+    sys.exit(app.exec_())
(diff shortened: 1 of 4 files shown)


src/sas/sasview/__init__.py
```diff
@@ -1,5 +1,5 @@
 from distutils.version import StrictVersion
-__version__ = "5.0.5-alpha.1"
+__version__ = "5.0.5a1"
```
krzywon, line 2: This fix is already being addressed in #1874 and should probably be removed.

src/sas/qtgui/Utilities/CodeEditor.py
```diff
@@ -0,0 +1,96 @@
...
+        digits = 1
+        max_value = max(1, self.blockCount())
+        while max_value >= 10:
+            max_value /= 10
+            digits += 1
+        space = 3 + self.fontMetrics().width('9') * digits
+        return space
+
+    def updateLineNumberAreaWidth(self, _):
+        self.setViewportMargins(self.lineNumberAreaWidth(), 0, 0, 0)
```
krzywon, line 38: I'm wondering why you are passing a value that you immediately throw away?

src/sas/qtgui/Utilities/CodeEditor.py
```diff
@@ -0,0 +1,96 @@
...
+        self.cursorPositionChanged.connect(self.highlightCurrentLine)
+        self.updateLineNumberAreaWidth(0)
+
+    def lineNumberAreaWidth(self):
+        digits = 1
+        max_value = max(1, self.blockCount())
+        while max_value >= 10:
+            max_value /= 10
+            digits += 1
+        space = 3 + self.fontMetrics().width('9') * digits
```
krzywon, line 34: Where do the 3 and 9 come from?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -299,6 +299,36 @@ def updateFromPlugin(self):
...
+        successfulCheck = True
+        try:
+            ast.parse(model_str)
+
+        except SyntaxError as ex:
+            msg = "Error building model: " + str(ex)
+            logging.error(msg)
+            # print three last lines of the stack trace
+            # this will point out the exact line failing
+            last_lines = traceback.format_exc().split('\n')[-4:]
```
krzywon, line 317: Is there more in the traceback related to the type of error? The line number and line of code can be helpful, but not in all cases, e.g.
```
y = ((m*x + b)

return y
```
Gives the error:
```
ERROR: Error building model: invalid syntax (<unknown>, line 36)
ERROR:     return y
```

krzywon (changes requested):
This is a good step forward and nearly there, but the code could use more documentation and I've made a few inline comments that could use addressing.

smk78 (commented):
Hmm. So I built and ran the branch. But I don't actually see any functionality to check a model! What am I missing?

rozyczko:
> Hmm. So I built and ran the branch. But I don't actually see any functionality to check a model! What am I missing?

Fitting -> Edit Custom Model -> Load plugin...

Once a plugin is loaded, modify it with wrong syntax and try saving. Saving triggers the initial syntax check and if not successful, shows similar visual cues as with the simple Plugin Definition check.

Untitled

src/sas/qtgui/Utilities/CodeEditor.py
rozyczko: I'm wondering about that too :)
Fixed

src/sas/qtgui/Utilities/CodeEditor.py
rozyczko: explanation added

src/sas/qtgui/Utilities/TabbedModelEditor.py
rozyczko: I looked at various exceptions and it does not seem anything over the last 3 lines is all that useful.
E.g. the above gives

```
4:38:26 - ERROR: Error building model: invalid syntax (<unknown>, line 35)
14:38:26 - ERROR: Traceback (most recent call last):
  File "C:\Users\piotr\projects\sasview\src\sas\qtgui\Utilities\TabbedModelEditor.py", line 310, in checkModel
    ast.parse(model_str)
  File "C:\Anaconda3\envs\qt5_win\lib\ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 35
    return y
```

The extra information is about the AST module, so I think it would be confusing to the user.

src/sas/sasview/__init__.py
rozyczko: Of course. I just hijacked it to be able to run this branch and test it. Otherwise, it requires manual modification

rozyczko:
> if you load a plugin with errors already in it ... then the Save button is disabled, so you sit there asking yourself 'ok, now what?'!

Ah! Great use case!
I added the model check on plugin load. Also, enabled the `Save` button for loaded models, so clicking on it is immediately possible, if needed. This, of course, will apply another check and re-save the plugin only when the syntax is correct.

krzywon (approved):

smk78:
I just pulled and built the branch again, but I can't run it because of the version string problem (#1874)!
```
(qt5) D:\SasViewCodeDev\pr1875>python run.py
  File "run.py", line 138, in <module>
    run_sasview()
  File "D:\SasViewCodeDev\pr1875\src\sas\qtgui\MainWindow\MainWindow.py", line 106, in run_sasview
    mainwindow = MainSasViewWindow(screen_resolution)
  File "D:\SasViewCodeDev\pr1875\src\sas\qtgui\MainWindow\MainWindow.py", line 38, in __init__
    from .GuiManager import GuiManager
  File "D:\SasViewCodeDev\pr1875\src\sas\qtgui\MainWindow\GuiManager.py", line 22, in <module>
    import sas.qtgui.Utilities.LocalConfig as LocalConfig
  File "D:\SasViewCodeDev\pr1875\src\sas\qtgui\Utilities\LocalConfig.py", line 8, in <module>
    import sas.sasview
  File "D:\SasViewCodeDev\pr1875\src\sas\sasview\__init__.py", line 3, in <module>
    StrictVersion(__version__)
  File "D:\Anaconda3\envs\qt5\lib\distutils\version.py", line 40, in __init__
    self.parse(vstring)
  File "D:\Anaconda3\envs\qt5\lib\distutils\version.py", line 137, in parse
    raise ValueError("invalid version number '%s'" % vstring)
ValueError: invalid version number '5.0.5-alpha.1'
```

rozyczko:
@smk78 that's because I removed the version number fix/workaround I had in this branch code originally, after @krzywon (correctly) complained about including it here.
To test the model check, just comment out 
```
StrictVersion(__version__)
```
line in `src\sas\sasview\__init__.py` as a temporary fix

smk78:
Thanks @rozyczko . That did the trick.

Ok, so now when I run the Model Editor and load a plugin with (known) errors this happens:
image
image

The highlighting line 1 and not line 70 is a bit confusing:
image

and clicking Save at this point doesn't change the view, but the red border is good and the log explorer is clear about the problem. It seems the highlighting is just where the input focus is (top of the file initially)?

I notice that if I hover the mouse pointer in the window a popup appears:
image

but it takes a while to appear (too long I think) and the absence of the line number in the popup is a pity. **Can anything be done about both of these?**

If I then change _define_ in line 70 to _def_ and try Save it correctly picks up that there is a second error:
image
image

So that's good.

If I add in the missing ] and Save again the red border vanishes and the Log Explorer confirms the fix:
image
image

rozyczko:
Setting the cursor on the 1st error line has been implemented.

Changing the tooltip activation time seems really difficult, requiring custom widget signalling and non-trivial code changes.
I would like to avoid it, if possible.

smk78:
The cursor does not move to the failing location on each Save in what I am testing...

smk78:
Pulled and built again.
The code highlighting seems to jump to the line of code before the one with the error (So 67 in this case and not 70):
image

When I fix the error at 70, it then jumps to 75 and not 81:
image

rozyczko:
@smk78 : what does the error message say? Does it point to the correct line? The parser is just taking the line mentioned in the error and putting the cursor on it.
Can you please include the plugin file?

smk78:
Yes, the Log Explorer (and cmd window since I'm running a build) show the correct line number as shown in the screen shots above (you may have skipped over them: they're under the editor windows). Here's the model I've been using:
[torus.zip](https://github.com/SasView/sasview/files/6714627/torus.zip)

rozyczko:
oh.. OHH... I see now. Some of the lines are too wide and they just wrap.
The pointer set doesn't know wrapped lines from regular lines and just places the cursor on the n-th line in the editor.
Try resizing the editor so that no line wraps and the position of the cursor is correct.

I need to find a fix for that.

rozyczko:
Fixed.
Also fixed the case when one loads a good model after having an incorrect model opened.

smk78:
Yes! That's working now! Good to merge as far as I'm concerned!

wpotrzebowski:
Build system is fixed now (we were waiting for it). It should be ok to merge now.

Issue #1866, opened 2021-06-03
Issue loading data with 5.0.4 on Windows


wpotrzebowski:
I've been teaching SasView 5.0.4 as SANS course and one of the students had a problem with loading data.
When she tried to open "Load Data" dialog it took ~30 for Windows to come. Drag and dropping worked fine. 
She used latest version of Windows pushed by University update.

smk78:
Yes, I've experienced this now and again too. Including with previous versions. See https://github.com/SasView/sasview/issues/1445 Unfortunately I never did nail down the cause.

butlerpd:
@wpotrzebowski and/or @smk78, should we try to merge these two tickets? Or do we think they are clearly distinct issues that should have separate tickets?

smk78:
I think the reference here will suffice.

rozyczko:
Is this happening on the `File Open` dialog show or when a file (files) are selected in the dialog and the actual file load starts?

rozyczko:
This might be a manifestation of an old (ancient!) bug in Qt, which so far, hasn't been properly fixed
https://bugreports.qt.io/browse/QTBUG-13182

There are some workarounds listed, which I will try to implement and ask you all to test, if possible.

smk78:
> Is this happening on the `File Open` dialog show or when a file (files) are selected in the dialog and the actual file load starts?

As I recall, it was pretty much when the file chooser window appeared; the folder contents part would remain blank for a long time.

rozyczko:
The corresponding PR has been merged to `main` meaning we can close this issue?

butlerpd:
Not sure why nobody answered you @rozyczko but it seems like this should be closed so will do so now

PR #1876, opened 2021-06-21
Attempt to address file load widget sluggishness issue. #1866

rozyczko:
a simple extra option passed to the file open widget apparently fixes it for some.
https://bugreports.qt.io/browse/QTBUG-13182?focusedCommentId=201836&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-201836


smk78 (approved):
This looks like a nice and simple fix. Proving it's a fix may be trickier however (though at least we may have a clue to the cause now)! I suggest we merge this and see how we go.

krzywon (approved):
Ditto to what @smk78 said.

PR #1877, opened 2021-06-22
Data import/export syntax fixes

krzywon:
This fixes two syntax issues in data import/export files, as noted by @pkienzle. There is no associated ticket for this.


src/sas/sascalc/file_converter/nxcansas_writer.py
```diff
@@ -276,7 +276,7 @@ def _write_h5_vector(entry, vector, names=['x_position', 'y_position'],
             trans_entry.attrs['signal'] = 'T'
             trans_entry.attrs['T_axes'] = 'T'
             trans_entry.attrs['name'] = trans.name
-            if trans.timestamp is not '':
+            if trans.timestamp != '':
```
rozyczko, line 279: if `timestamp` is a string, then why not compare it directly?
```
if not trans.timestamp:
```

butlerpd (approved):
At today's fortnightly meeting Piotr says he approves the changes.

PR #1884, opened 2021-07-08
1814: Q-range reset

### LINKED ISSUE #1814: In 5.0.4 release: Reset Q limits button has no effect
Reported by user GregS at ISIS:

_I was using the newly released 5.0.4 today, and I noticed a bug when adjusting the Q limits on some fitting._
 
_If you adjust the limits, you cannot reset to the data Q_min and Q_max using the “reset” button. Pressing the button doesn’t do anything. You can revert to a Q limit by choosing a value that is much bigger or smaller than the one from the data._
 
_This was not the behaviour in previous versions of SasView 4 or 5. Pressing the reset button restored the data Q limits._

krzywon:
This fixes the issue related to resetting the Q-range that was introduced in v5.0.4. Unit tests were added to the `fittingSuite` of `GUITests.py` to ensure the values are reset properly.

Fixes #1814

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -476,6 +476,54 @@ class PlotterWidget(PlotterBase):
         self.setRange.txtYmax.setText(str(y_range[1]))
         self._setGraphRange()
 
+    def _zoom_handler(self, event):
+        """
+        Explicitly call toolbar method to ensure range is changed. In MPL 2.2, local events take precedence over the
+        toolbar events, so the range isn't zoomed until after _zoom_pan_handler is run.
+        """
+        self.toolbar.release_zoom(event)
+        self._zoom_pan_handler(event)
+
+    def _pan_handler(self, event):
+        """
+        Explicitly call toolbar method to ensure range is changed. In MPL 2.2, local events take precedence over the
+        toolbar events, so the range isn't panned until after _zoom_pan_handler is run.
+        """
+        self.toolbar.release_pan(event)
+        self._zoom_pan_handler(event)
+
+    def _home(self, event):
+        """
+        Catch home button click events
+        """
+        self.onResetGraphRange()
+
+    def _back(self, event):
+        """
+        Catch back button click events
+        """
+        self.toolbar.back()
+        self._zoom_pan_handler(event)
+
+    def _forward(self, event):
+        """
+        Catch forward button click events
+        """
+        self.toolbar.forward()
+        self._zoom_pan_handler(event)
+
+    def _pan(self, event):
+        """
+        Catch pan button click events
+        """
+        self.canvas.mpl_connect('button_release_event', self._pan_handler)
+
+    def _zoom(self, event):
+        """
+        Catch zoom button click events
+        """
+        self.canvas.mpl_connect('button_release_event', self._zoom_handler)
+
     def onSetGraphRange(self):
         """
         Show a dialog allowing setting the chart ranges
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -181,20 +181,6 @@ class PlotterWidget(PlotterBase):
         ax.set_xscale(self.xscale, nonposx='clip')
         ax.set_yscale(self.yscale, nonposy='clip')
 
-        # define the ranges
-        if hasattr(self, 'setRange') and self.setRange.rangeModified:
-            # Assume the range has changed and retain the current and default ranges for future use
-            modified = self.setRange.rangeModified
-            x_range = self.setRange.xrange()
-            y_range = self.setRange.yrange()
-        else:
-            # Use default ranges given by matplotlib
-            x_range = self.ax.get_xlim()
-            y_range = self.ax.get_ylim()
-            modified = False
-        self.setRange = SetGraphRange(parent=self, x_range=x_range, y_range=y_range)
-        self.setRange.rangeModified = modified
-
         # Draw non-standard markers
         l_width = markersize * 0.4
         if marker == '-' or marker == '--':
@@ -260,6 +246,23 @@ class PlotterWidget(PlotterBase):
         if self.xLabel and not is_fit:
             ax.set_xlabel(self.xLabel)
 
+        # define the ranges
+        if hasattr(self, 'setRange') and self.setRange.rangeModified:
+            # Assume the range has changed and retain the current and default ranges for future use
+            modified = self.setRange.rangeModified
+            default_x_range = self.setRange.defaultXRange
+            default_y_range = self.setRange.defaultYRange
+            x_range = self.setRange.xrange()
+            y_range = self.setRange.yrange()
+        else:
+            # Use default ranges given by matplotlib
+            x_range = default_x_range = self.ax.get_xlim()
+            y_range = default_y_range = self.ax.get_ylim()
+            modified = False
+        self.setRange = SetGraphRange(parent=self, x_range=x_range, y_range=y_range)
+        self.setRange.rangeModified = modified
+        self.setRange.defaultXRange = default_x_range
+        self.setRange.defaultYRange = default_y_range
         # Go to expected range
         self.ax.set_xbound(x_range[0], x_range[1])
         self.ax.set_ybound(y_range[0], y_range[1])
@@ -470,8 +473,9 @@ class PlotterWidget(PlotterBase):
         if self.setRange.exec_() == QtWidgets.QDialog.Accepted:
             x_range = self.setRange.xrange()
             y_range = self.setRange.yrange()
-            self.setRange.rangeModified = True
             if x_range is not None and y_range is not None:
+                self.setRange.rangeModified = (self.setRange.defaultXRange != x_range
+                                               and self.setRange.defaultYRange != y_range)
                 self.ax.set_xlim(x_range)
                 self.ax.set_ylim(y_range)
                 self.canvas.draw_idle()
@@ -480,11 +484,12 @@ class PlotterWidget(PlotterBase):
         """
         Resets the chart X and Y ranges to their original values
         """
-        # Clear graph and plot everything again
-        mpl.pyplot.cla()
-        self.ax.cla()
-        for ids in self.plot_dict:
-            self.plot(data=self.plot_dict[ids], hide_error=self.plot_dict[ids].hide_error)
+        x_range = self.setRange.defaultXRange
+        y_range = self.setRange.defaultYRange
+        self.setRange = SetGraphRange(parent=self, x_range=x_range, y_range=y_range)
+        # Go to expected range
+        self.ax.set_xbound(x_range[0], x_range[1])
+        self.ax.set_ybound(y_range[0], y_range[1])
 
         # Redraw
         self.canvas.draw_idle()
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -465,20 +465,34 @@ class PlotterWidget(PlotterBase):
 
         self.canvas.draw_idle()
 
+    def _zoom_pan_handler(self, event):
+        if not self.setRange:
+            self.setRange = SetGraphRange(parent=self)
+        x_range = self.ax.get_xlim()
+        y_range = self.ax.get_ylim()
+        self.setRange.txtXmin.setText(str(x_range[0]))
+        self.setRange.txtXmax.setText(str(x_range[1]))
+        self.setRange.txtYmin.setText(str(y_range[0]))
+        self.setRange.txtYmax.setText(str(y_range[1]))
+        self._setGraphRange()
+
     def onSetGraphRange(self):
         """
         Show a dialog allowing setting the chart ranges
         """
         # min and max of data
         if self.setRange.exec_() == QtWidgets.QDialog.Accepted:
-            x_range = self.setRange.xrange()
-            y_range = self.setRange.yrange()
-            if x_range is not None and y_range is not None:
-                self.setRange.rangeModified = (self.setRange.defaultXRange != x_range
-                                               or self.setRange.defaultYRange != y_range)
-                self.ax.set_xlim(x_range)
-                self.ax.set_ylim(y_range)
-                self.canvas.draw_idle()
+            self._setGraphRange()
+
+    def _setGraphRange(self):
+        x_range = self.setRange.xrange()
+        y_range = self.setRange.yrange()
+        if x_range is not None and y_range is not None:
+            self.setRange.rangeModified = (self.setRange.defaultXRange != x_range
+                                           or self.setRange.defaultYRange != y_range)
+            self.ax.set_xlim(x_range)
+            self.ax.set_ylim(y_range)
+            self.canvas.draw_idle()
 
     def onResetGraphRange(self):
         """
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -484,12 +484,11 @@ class PlotterWidget(PlotterBase):
         """
         Resets the chart X and Y ranges to their original values
         """
-        x_range = self.setRange.defaultXRange
-        y_range = self.setRange.defaultYRange
-        self.setRange = SetGraphRange(parent=self, x_range=x_range, y_range=y_range)
-        # Go to expected range
-        self.ax.set_xbound(x_range[0], x_range[1])
-        self.ax.set_ybound(y_range[0], y_range[1])
+        # Clear graph and plot everything again
+        mpl.pyplot.cla()
+        self.ax.cla()
+        for ids in self.plot_dict:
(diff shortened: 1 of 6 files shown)


butlerpd:
ready for testing on Win

butlerpd:
Trying to test this I think has uncovered a rat's nest of strange behavior. I can get different behaviors every time I start the program again. Know I have two different behaviors between fitpage1 and fitpage2 indicating there is some strong path dependent issues.   We have three "reset" that all interact with the same thing but in different ways and each even seems to react differently with a theory curve only vs data and model.  documenting this carefully is going to take some time. However I can report two things that relate directly to what this PR thought it was changing:
1. in 5.0.4 it seems that the `Reset` button in the Fit Options tab did not reset the fitting q range as the box it is in suggests.  That STILL happens ... sometimes.
2. However so far it seems to do something more often than before? But rather than just resetting the fitting range, it also resets the plot viewport **and** resets the number of points (assuming we are only working with a theory curve) to the default 50 points.

Also the "home" icon which should reset the plot viewport if the navigation toolbar is available only seems to reset to the original x, y from when the plot was first started. If the parameters change, that is not recognized.  On the other hand, the `Reset Graph Range` from the drop down will resize to **exactly** the current theory curve's min max ... with no extra buffer.  Sometimes this gets fixed when changing a parameter value and sometimes not.

In short - this will need a lot more investigating and probably more changes before it is ready to merge from what I can tell

krzywon:
> 1. in 5.0.4 it seems that the `Reset` button in the Fit Options tab did not reset the fitting q range as the box it is in suggests.  That STILL happens ... sometimes.

Can you reliably repeat the situation where it does not reset? I'll try the installer and other scenarios, but when running from source, I see the range resetting properly when I have multiple fit tabs open.

> 2. However so far it seems to do something more often than before? But rather than just resetting the fitting range, it also resets the plot viewport **and** resets the number of points (assuming we are only working with a theory curve) to the default 50 points.

> On the other hand, the `Reset Graph Range` from the drop down will resize to **exactly** the current theory curve's min max ... with no extra buffer. Sometimes this gets fixed when changing a parameter value and sometimes not.

These both mimic the behavior in v5.0.0 - v5.0.3. I could change the number of points resetting behavior in this PR, but a new ticket would be more appropriate for the reset graph range issue.

> Also the "home" icon which should reset the plot viewport if the navigation toolbar is available only seems to reset to the original x, y from when the plot was first started. If the parameters change, that is not recognized.

Another thing that was present in v5.0.3 that I think is beyond the scope of this PR.

butlerpd:
Ha ... If I could reproduce anything reliably the comment would have been shorter. And I'd have a better idea of where things are going wrong 😄. I tried exiting and restarting, re-installing and trying different sequences  of things. Sometimes things worked, sometimes they didn't, but I could not, during the course of over an hour, figure out any sequence that would reliably reproduce when things worked right vs when they would not. Hence my point of needing to do a lot more testing. I find this particularly worrisome.

Regarding whether the other comments are out of scope, In principle yes. However my testing so far suggests to me that they are all intimately linked.  Actually decoupling the `Reset` from the plot window q range might in fact make this PR more robust as I suspect it is the plot window q range that is unstable. Possibly because these are three different ways (currently this includes **this PR** also) to achieve the same thing without using 100% the same code?

krzywon:
> Actually decoupling the `Reset` from the plot window q range might in fact make this PR more robust as I suspect it is the plot window q range that is unstable. Possibly because these are three different ways (currently this includes **this PR** also) to achieve the same thing without using 100% the same code?

These are 3 distinctly different features that are supposed to have different behaviors.
- Home Button
  - Functionality: This is a built-in feature of Matplotlib. This is supposed to reset the chart to its default view resetting the graph range, toggling title and legend visibility, resetting the title and axis labels, but should not affect the plotted data.
  - Code Location: https://matplotlib.org/3.2.2/users/navigation_toolbar.html
- Reset Graph Range
  - Functionality: Resets the display range of the chart to min and max of the I and Q values in the displayed data without touching any other plot changes. This should probably add a small buffer to the plotted range, so the range is the same as when the plot is first generated.
  - Code Location: sas.qtgui.Plotting.Plotter.py -> PlotterWidget.actionResetGraphRange triggers PlotterWidget.onResetGraphRange()
- Reset Fit Range
  - Functionality: Resets the upper and lower bounds and number of points of the fitted Q data. This is not supposed to change the plotted range.
  - Code Location: sas.qtgui.Perspectives.Fitting.OptionsWidget.py -> OptionsWidget.cmdReset triggers OptionsWidget.onRangeReset()

What I can do in this PR:
- Fix the graph range reset to add a buffer to the outer edges of the graph.
- Ensure fitted q range changes do not trigger any graph range changes, though I don't yet see how that could happen.
- Update the number of **fitted** points in the GUI - this isn't changing when the fitting range is changed.

butlerpd:
> These are 3 distinctly different features that are supposed to have different behaviors.

As I suspected then! Because for the user they do **exactly** the same thing regarding the plot x and y view range.  The fact that they are done from 3 completely distinct sources seems problematic to me. Indeed, I suspect that the problem comes from trying for example to set the range from the drop down but using the plot navigation to reset it, or vice versa .. or use the `Reset` on the fitting q range in the FitPanel? At some point I suspect the system gets into a state where it does not know that the plot range has changed (because it was done behind my back and maybe I reset a flag so that the original code now also thinks it has been reset?) so "resetting" does nothing.

>* Reset Fit Range
>   * Functionality: Resets the upper and lower bounds and number of points of the fitted Q data. This is not supposed to change the plotted range.

Is this the behavior you get? This is one thing that **was** consistent in all my tests as best I can remember: `Reset` either did nothing at all OR it did both: reset the Q limits for fitting and the plot range to show all the data.

That said I could test again as I tested it with theory only more than with data (the fit q range boxes are editable at this point even though now that I think about it that doesn't make a whole lot of sense) so it is possible that I didn't notice that with data 

And I now realize WHY it is changing the number of points in the theory curve 😄 ... For a fit, changing the fit q range changes the number of points the model should calculate.  Aside from making sure that `Reset` does not change the plot range there are two options that would work from the usability perspective.  Leave as is but make sure the number of points to use in the model doest not change when in theory only mode, or disable (grey out) all the fitting range boxes and buttons when there is no actual data.  I'm not sure which is better from a code logic or UI best practices perspective.

krzywon:
I'll see what I can do about all of this. The 3 are supposed to be unique, and that should be obvious to users.

butlerpd:
That said you may be right that fixing the problem with `Reset` also resetting the plot view (and dealing in some way with the changing number of theory points if possible) is enough for this PR and the rest addressed in a different PR?  unless it is easy enough to do in one?

krzywon:
ready for testing on Win

krzywon:
I've now fixed the following in this PR. Note, I have **not** verified that all unit tests that were passing still are and I will do that before this PR is merged.

> - Fix the graph range reset to add a buffer to the outer edges of the graph

The graph is completely redrawn now and takes the `matplotlib` default ranges.

> - Ensure fitted q range changes do not trigger any graph range changes

Q-range changes and fits trigger a plot redraw which was causing this. I added a way to store existing ranges between redraws.

> - Update the number of fitted points in the GUI

This is now fixed.

krzywon:
My latest commit has cleaned up the graph range resetting issues and ensures all unit tests that were previously passing are now passing again. This should be ready for review again.

krzywon:
ready for testing on Win

rozyczko:
The code looks good.
Behaviour for the range reset functionality is as expected.

> in 5.0.4 it seems that the Reset button in the Fit Options tab did not reset the fitting q range as the box it is in suggests. That STILL happens ... sometimes.

I think the issue is with signaling here. Fitting range textboxes in `Fit Options` have slots for both `textChanged` and `editingFinished`. You can see that when you modify the values - each modification causes the min/max bars to be moved, but the corresponding chart update does not happen until the `editingFinished` signal is fired by exiting the textbox.
This can potentially be fixed by making the range bars - range text box connection fire only on `onMouseRelease` rather than `onMouseMove` ?

rozyczko (approved):
looks good to me!

butlerpd:
As part of functionality testing I find that the problem of the graph window resetting when the fit button is pressed is still present. However it is not mentioned in the statement of the problem of this PR. Is this the expected behavior for this PR? In other words is this being, or has this been, addressed by another issue/PR?

To reproduce, using this installer build, start sasview. Load the 100nm pinhole sphere data and send to fitting. Select Sphere/Sphere for the model. hit show/plot.  On the plot zoom into a small region of the data.  set the scale, background an radius as fit parameters and hit fit. Once the fit is completed the graph is reset as if the `Reset Graph Range` menu item in the drop down had been selected. rezooming in and changing a parameter value and hit return again rests the graph range.

As long as this was a) not fixed in the main when this branch was created and b) is meant to be addressed elsewhere and not here this comment can be ignored... but more comments to come relating to what has been discussed here

butlerpd:
This seems to be behaving a lot better than before.  A few niggles I have found so far which could be punted to another ticket in the interest of perfection not being the enemy of the good.

* The zoom and reset range (home) on the optional graph toolbar and the drop down menu not behave essentially the same.  While @krzywon suggests they are meant for different purposes, from the user point of view they are just two ways to achieve the same thing with the graph toolbar being much easier than the dropdown (particularly for zooming in -- except for the rare case where the user wants fine control of exactly the xmin, xmax, ymin, ymax)
* One small issue that can probably be ignored unless brought up by users, if the range is set by the drop down only the dropdown can reset the range.  As long as the zoom is done graphically with the graph toolbar both the home button on the toolbar and the `Reset Graph Range` dropdown will behave the same.
* The `reset` fitting Q range and the other zooms now seem to be well separated!!
  * An interesting factoid, in theory mode, where the user gets to select the number of points to plot (associated with the full data range as opposed to the fitting data range) that number gets reset to the default (currently 50) gets reset with the fitting range reset but the log/linear checkbox next to that number is not reset.  I don't think this merits any fixing as resetting fitting range in a theory plot has no meaning in principle. and keeping the log/linear choice during resetting of an actual fit is the correct answer. Just FYI
  * One real annoyance which is for another issue I think, if the graph toolbar zoom is on (as it will often be unless one remembers to unclick it after a zoom) when one tries to move the slider bar, both the slider bar move AND the zoom are simultaneously activated so that after moving the slider there is almost no data visible. This can be fixed by clicking home or reset.
* Adding data to a plot that requires a larger q or intensity range than exists on the plot causes the graph to resize when sending to the graph.  It also appears to reset the default graph range for both the drop down reset range and the home button. Good!! However there do seem to be a few issues that can crop up when the toolbar is activated after the range is set by the dropdown:
  * Set the range via the dropdown. Then activate the toolbar from the drop down. Click home and nothing happens. This is a minor issue as long as ** the wrong** sequence is not entered (see later).  It appears that zooming in from the toolbar, at least with the full range displayed will set the range for the home button.
  * However, if immediately after clicking home the first time (with the restricted range) one uses the toolbar zoom (restricting further) then hit home, the home button now seems to be "permanently" stamped with that original restricted range while `Reset Graph Range` from the drop down works as expected.
  * Next, with the plot range set to the restricted range given by hitting the home button, one adds another data set to the graph (in my case one that was well out of the range of the current graph). The graph window does NOT reset but behaves as if the zoom had been set.  However now, **BOTH** the `Reste Graph Range` and the home button have the same default plot range which is the restricted range that the home button was stuck on (i.e. now both ways are stuck - btw, I believe if this is a fit plot, activating a fit can get the dropdown limits to work again but not the home button limits .. needs more testing).
  * Create a plot. activate the toolbar. set a restricted range using the toolbar zoom. Add another data set that is outside the range of the original data set loaded.  The dropdown `Reset Graph Range` is now stuck on the restricted viewport.  However using the home button will rezoom to the original data set but the new data set will not be visible (note even this is **not** the behavior if one adds the second data set while the viewport is in the "home" position)
  * Create a plot. activate the toolbar. set a restricted range using the dropdown menu. Add another data set that is outside the range of the original data set loaded.  Now **both** the dropdown `Reset Graph Range` and the home button are stuck on the restricted viewport.  zooming in and back out does nothing to fix it.

There may be other permutations but this seems like a good start 😄 I'm not sure if these should stop us from merging the distinct improvements.  Again I think this comes from making the dropdown not call the toolbar functions (or co-opting the toolbar as we did in 4.x) so that trying to make it always work as the user would expect may be quite difficult?  To be fair the dropdown was created long before we thought to add the toolbar for the much simpler zoom feature that we had lost in going to 5.x so trying to use the toolbar methods would not have been a thing I guess.

Thoughts?

krzywon:
> As part of functionality testing I find that the problem of the graph window resetting when the fit button is pressed is still present. However it is not mentioned in the statement of the problem of this PR. Is this the expected behavior for this PR? In other words is this being, or has this been, addressed by another issue/PR?

@butlerpd This is fixed in this branch now. Please retest when you get a chance. Your other, longer, message is well beyond the scope of this PR and will likely require a series of separate tickets to discuss and address them.

wpotrzebowski:
@butlerpd can you please check this PR?

krzywon:
After a phone call with @butlerpd, I think we found the disconnect. Locally, I am running the same environment as the GH actions build, but @butlerpd is running the same environment as Jenkins. Testing the different builds, the behavior in the GH actions matches what I see locally, but the Jenkins build matches the behavior Paul sees. This seems to be a matplotlib version discrepancy.

We will need to finalize which build we use for releases before I can finalize this.

krzywon:
The latest commit should now work on both `matplotlib` version 2.2.3 **and** 3.4.3, without the need to check the installed mpl version. In mpl2, the events in `Plotter.py` always happen before the events that happen in the mpl toolbar class. The opposite behavior is seen in mpl3.

Without calling the toolbar methods, the graph range before the pan/zoom/home happens was being stored. I am now explicitly calling the built-in toolbar methods before calling the graph range methods in Plotter.py and PlotterBase.py to be sure they occur in the proper order.

butlerpd:
ready for testing on Win

butlerpd (dismissed):
### FUNCTIONALITY REVIEW ONLY
**Jenkins PR build (including matplotlib 2):**
BRAVISIMO!! 👍 
Finally-the zoom, set range, reset range and all of its permutations seems to be behaving rock solidly as expected including on fits (what prompted the initial issue) and a number of the inconsistent behaviors noted in prior testing -- no doubt due to the tying of the toolbar actions with the drop down menu actions.  zoom now sees to be working beautifully.

**Github Actions PR build (including matplotlib 3):**
The behavior seems to be the same as for the Jenkins build but I cannot give it the "rock solid" stamp due to the impossibility of properly testing fitting plots in this build. However, a cursory testing of the main branch installer from github actions (which does not yet have this PR merged in), exhibits similar plotting problems when fitting. While this PR may or may not still be exacerbating the issue it is clearly not the initial culprit so am approving this PR

NOTE: I believe that @krzywon also requested a re-review of the code from @rozyczko prior to merging due to the amount of code changes since @rozyczko last review?

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -448,32 +465,97 @@ def onRemoveText(self):
...
+        """
+        Resets the chart X and Y ranges to their original values
+        """
+        # Clear graph and plot everything again
+        mpl.pyplot.cla()
+        self.ax.cla()
+        self.setRange = None
+        for ids in self.plot_dict:
+            self.plot(data=self.plot_dict[ids], hide_error=self.plot_dict[ids].hide_error)
+
```
rozyczko, line 555: `Plot` accepts more arguments:
```
def plot(self, data=None, color=None, marker=None, hide_error=False, transform=True):
```
If your plot has non-default marker or color this probably will reset those properties.

src/sas/qtgui/Plotting/PlotterBase.py
```diff
@@ -255,6 +260,42 @@ def onResize(self, event):
...
+        """
+        Catch home button click events
+        """
+        pass
+
+    def _back(self, event):
+        """
+        Catch back button click events
+        """
+        pass
```
rozyczko, line 279: Shouldn't the base method just fire the corresponding MPL toolbar event instead of passing?
e.g. `self.toolbar.back()`
Or alternatively, should it require implementation in derived class by adding a `NonImplemented` exception?

rozyczko (dismissed):
Code looks good, a few questions/comments added.
Functionality is OK.

src/sas/qtgui/Plotting/Plotter.py
krzywon: The markers and colors are stored in each data set held within `self.plot_dict`. If either `color | marker` is not passed to `self.plot`, the respective value is taken from the passed data set. I'd have to do the same here, so I think it would be redundant.

krzywon:
I accidentally dismissed reviews by @butlerpd and @rozyczko. Could you both review one last time?

butlerpd:
ready for testing on Win

butlerpd (approved):
Not as thorough a test as last time but all still seems to be working after the latest edits in response to the code review. Note: I did not bother trying to test the github installer version due to its problems but don't think that should stop merging.

rozyczko (approved):

PR #1887, opened 2021-07-13
minor change to sas_gen.py which should fix #1886

### LINKED ISSUE #1886: The 'Save SLD Data' button in the generic scattering calculator saves sld data with rows and columns flipped
When saving loaded data in the generic scattering calculator as a .sld file the rows and columns of the data are saved the wrong way around. e.g. the initial file:
```
X  Y  Z  SLDN  SLDMx  SLDMy  SLDMz
11 12 13 14 15 16 17 18
21 22 23 24 25 26 27 28
```
would be resaved by the calculator as:
```
X  Y  Z  SLDN SLDMx  SLDMy  SLDMz VOLUMEpix
1.100000000000000000e+01 2.100000000000000000e+01
1.200000000000000000e+01 2.200000000000000000e+01
1.300000000000000000e+01 2.300000000000000000e+01
1.400000000000000000e+01 2.400000000000000000e+01
1.500000000000000000e+01 2.500000000000000000e+01
1.600000000000000000e+01 2.600000000000000000e+01
1.700000000000000000e+01 2.700000000000000000e+01
1.800000000000000000e+01 2.800000000000000000e+01
```
As a result the new .sld files cannot be reloaded by the calculator.

rjbourne:
Use np.column_stack() to correctly arrange the outputted data for the sld file, so that it is in the correct order.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ default_categories.json
 /test/sasdataloader/data/isis_1_0_write_test.xml
 /test/sasdataloader/data/isis_1_1_write_test.xml
 /test/sasdataloader/data/write_test.xml
+/test/sascalculator/data/write_test.sld
 /test/fileconverter/data/export2d.h5
 **/logs
 tests.log
--- a/src/sas/sascalc/calculator/sas_gen.py
+++ b/src/sas/sascalc/calculator/sas_gen.py
@@ -673,7 +673,7 @@ class SLDReader(object):
         else:
             mx, my, mz = data.sld_mx, data.sld_my, data.sld_mz
         vol = data.vol_pix if data.vol_pix is not None else np.ones_like(x)
-        columns = np.vstack((x, y, z, sld_n, mx, my, mz, vol))
+        columns = np.column_stack((x, y, z, sld_n, mx, my, mz, vol))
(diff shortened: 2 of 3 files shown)


wpotrzebowski:
ready for testing on Linux

wpotrzebowski:
ready for testing on OSX

wpotrzebowski (commented):
`column_stack` looks like more logical choice here. Does it mean it was giving incorrect output before? It would be great if you can add unit test for this fix. You can take a look on what's been available here https://github.com/SasView/sasview/tree/main/test/sascalculator

P.S. Ignore failing OSX PR builder. There is unrelated problem with this build

rjbourne:
Yes - the previous output had its rows and columns swapped, which meant that Sasview recognized it an an invalid file type and rejected it. Interestingly when the file is rejected Sasview throws a runtime exception which means that to the user it simply appears as if the file is still loading. It would probably be better for the error to be shown in the log and the loading visibly stopped so the user can see there is a failure rather than the file appearing to never load.

I can definitely add in a unit test, it should be fairly easy to save a small datafile and confirm that everything is stored as expected

wpotrzebowski (approved):

PR #1890, opened 2021-08-03
canSAS XML Reader: Load multiple SASdata in a single SASentry

### LINKED ISSUE #1889: CanSAS XML reader cannot read multiple SASdata elements within a single SASentry
From @gonzalezma:

> Guillaume, one of our Mantid developers (in cc), wants to export several
> measurements into a single xml file using more than one <SASdata> block.
> The problem is that when reading this file in SasView only the last
> block is kept in the Data Explorer.
>
> I found the same behaviour when trying to read one of the official
> canSAS examples:
>
> https://github.com/canSAS-org/1dwg/blob/master/examples/cs_af1410.xml
>
> This one is more complex, because it contains 10 different SASentries
> which are correctly read, but each SASentry has 2 SASdata and only the
> last one is read by SasView.
>
> So this seems like an issue, but before opening one I wanted to check
> with those of you having more experience with the cansas format that I'm
> not overlooking or misunderstanding something.

krzywon:
The CanSAS XML reader now properly loads in multiple SASdata elements held within a single SASentry element. Unit tests were added to ensure this remains true.

fixes #1889

### DIFF
--- a/src/sas/sascalc/dataloader/readers/cansas_reader.py
+++ b/src/sas/sascalc/dataloader/readers/cansas_reader.py
@@ -110,7 +110,13 @@ class Reader(XMLreader):
                         bad_xml = INVALID_XML.format(basename + self.extension)
                         bad_xml += invalid_xml
                         self.current_datainfo.errors.append(bad_xml)
-                self.data_cleanup()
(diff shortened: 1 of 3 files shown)


gonzalezma (approved):
Tested and works nicely.

butlerpd:
Code looks straight forward?  Suggest we can merge.

PR #1892, opened 2021-08-11
Ticket 1406 cancel calculation

### LINKED ISSUE #1406: Add cancel button to generic scattering panel
Generic scattering calculator is breaking the computation across Q values in order to give a progress update.  Add a test for calcthread.isquit() in the loop and fill remaining Iq values with a constant if it is encountered.

rjbourne:
This PR adds a cancel button to the generic scattering calculator to cancel computations between each chunk of Q values. If the computation is cancelled the remainder of the Iq values are filled in with `NaN`. The ticket for this request suggests using `calcthread.isquit()`, however the current code uses twisted.internet threads as opposed to the calcthread class in sasview. As far as I can tell there is no inbuilt way to detect an interruption in these threads, so a boolean flag is used. Upon cancelling the calculation the compute/cancel button becomes disabled until the current chunk of Q values have been computed.

Currently the areas of the graph with `NaN` values are simply not updated, which can cause odd effects on the graph when cancelling a calculation. This is because the `numpy.nan_to_num` function in the current code is not correctly implemented, an issue fixed in https://github.com/SasView/sasview/pull/1888.

fixes #1406

### DIFF
--- a/src/sas/qtgui/Calculators/GenericScatteringCalculator.py
+++ b/src/sas/qtgui/Calculators/GenericScatteringCalculator.py
@@ -609,8 +609,11 @@ class GenericScatteringCalculator(QtWidgets.QDialog, Ui_GenericScatteringCalcula
                 self._create_default_2d_data()
                 inputs = [self.data.qx_data, self.data.qy_data]
             logging.info("Computation is in progress...")
-            self.cmdCompute.setText('Wait...')
-            self.cmdCompute.setEnabled(False)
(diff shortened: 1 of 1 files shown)


rozyczko (approved):
This looks good and performs as expected.
Using QThreads should probably be preferred in this Qt app, but back when I wrote it, it didn't really behave properly.
It could probably be good to remove twisted and use QThreads throughout.

Issue #1861, opened 2021-05-23
Add the SasView version number to the logfile and top info bar


butlerpd:
When running SasView the top bar contains the icon and the word "SasView" it would be useful to have it contain the version number (i.e.: [icon.png]SasView 5.0.4.  Furthermore it would be really useful for debugging purposes when users send their log file to us to have the first line sent to the log include the version number. In other words, in place of
`hh:mm:ss - INFO: ---SasView session started---`
we have:
`hh:mm:ss - INFO: ---SasView 5.0.4 session started---`

And as long as we are at it, I notice that when we moved to default installing to `C:\sasviewV.v.v\` we seem to have dropped adding the version number to the name of the exe run file. It would be good to add that back please?

butlerpd:
![SasViewTopLeft](https://user-images.githubusercontent.com/10955478/119266733-ee857680-bbb9-11eb-9189-65f4aa4b7421.PNG)
Just to be clear about what is meant by the first part

and for the second part (though this is the console log but should be the same as what is in `c:\user\username\.sasview\sasview.log`

![SasViewBottomLeft](https://user-images.githubusercontent.com/10955478/119266802-33a9a880-bbba-11eb-9d9a-e8f47debcd7a.PNG)

PR #1893, opened 2021-08-18
#1861, adding the SasView version number

Caddy-Jones:
Attempting to solve issue #1861

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -13,6 +13,9 @@ from PyQt5.QtCore import Qt, QLocale, QUrl
 import matplotlib as mpl
 mpl.use("Qt5Agg")
 
+from sas.sasview import __version__ as SASVIEW_VERSION
+from sas.sasview import __release_date__ as SASVIEW_RELEASE_DATE
+
 from twisted.internet import reactor
 # General SAS imports
 from sas import get_local_config, get_custom_config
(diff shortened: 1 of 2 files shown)


src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -100,9 +103,12 @@ def __init__(self, parent=None):
         handler.messageWritten.connect(self.appendLog)
 
         # Log the start of the session
-        logging.info(" --- SasView session started ---")
+        version = f"{SASVIEW_VERSION.split('.')[0]}.{SASVIEW_VERSION.split('.')[1]}"
```
krzywon, line 106: This only gives the major and minor version information, but not which point release is being used (5.0 vs 5.0.5). The full version is given in the next line, but this seems unnecessary.

Perhaps combine the two lines:
`logging.info(f" --- SasView session started: v{SASVIEW_VERSION}, {SASVIEW_RELEASE_DATE} ---")`

krzywon (commented):
This only does half of what is asked for in #1861 and does not touch on the window title. I took a quick look at changing the window title and that is not as straight forward as I would have expected, so I can't fault you there. I'll update the ticket with what I know.

src/sas/qtgui/MainWindow/GuiManager.py
rozyczko: Adding something like this to MainWindow.py somewhere in `__init__.py` should do the trick
```
        from sas.sasview import __version__ as SASVIEW_VERSION
        self.setWindowTitle("SasView " + str(SASVIEW_VERSION))
```

rozyczko: it does.

Untitled

rozyczko:
Adding something like this to MainWindow.py somewhere in `__init__.py` should do the trick
```
        from sas.sasview import __version__ as SASVIEW_VERSION
        self.setWindowTitle("SasView " + str(SASVIEW_VERSION))
```

Untitled

krzywon:
@rozyczko - I had tried that but I added `self.setWindowTitle("SasView " + str(SASVIEW_VERSION))` before `self.setupUi(self)` the latter, I now see, resets the window title. Yes, it should be that easy!

smk78:
Ooo, it would be nice to get Mantid to do that too! :-)

rozyczko (commented):
This looks good to me!

butlerpd (approved):
Looks good.  as discussed at the SasView fortnightly meeting this should now be good to go, so approving it now.

Issue #1821, opened 2021-04-21
skriptable scattering calculator


dehoni:
The aim is to read sld files and generate a SAS pattern with the general scattering calculator out of a call within a skript. Finally it should  be possible to compare experimental and simulated data, calculate a cost function and find the minimum for a range of simulations where a parameter has been scanned.

dehoni:
There is an example how to use the scattering calculator in a script in the documentation: https://github.com/SasView/sasview/pull/1974/files#diff-6825133a3d2ef3f786d4f9f5f62aa70071a146400edebe3926b1396609de7bcb

PR #1899, opened 2021-08-24
Ticket 1882 coordinate systems

### LINKED ISSUE #1882: Visualisation and Calculation of Rocking Curve in Scattering Calculator
The Scattering Calculator should allow to rotate the sample coordinate system, against, field geometry and laboratory system of instrument. This would allow to calculate the scattering observed for doing rocking curves for ordered systems producing small-angle diffraction peaks.
A visual aid/plot showing the relation between the different coordinate system would be helpful to guide the user.

rjbourne:
This pull request introduces multiple different coordinate systems into the generic SANS calculator - allowing the sample, environment and polarisation vector to be independently oriented. This closes #1882. The functionality is currently ready for testing - however the code still depends on pull request #1888, which is not expected to significantly change before merging - after which main will need to be re-merged into this branch to resolve any conflicts. Note that this significantly inflates the number of changes github thinks have been made.

A work in progress account of the new GUI including coordinate systems is given on the branch https://github.com/SasView/sasview/tree/scattering-calculator-docs in the docs

The main approach is as follows:

 - GenericScatteringCalculator.py specifies the coordinate systems as seen and altered by the user - it provides a suitable way for the user to visualise the different coordinates and rotate them relative to each other.
 - sas_gen.py handles the actual conversions between coordinate systems - specified by just two rotations. This is to enable easy rotations if the calculation is to be scripted  as may be desired to create rocking curves or as per #1821.
 - geni.py expects all information in what are now called beamline coordinates and has not been altered.

The key changes in this PR are:

 - The addition of a new `set_rotations()` method in the GenSAS class - specifying rotations for the environment and sample coordinates.
 - The reconfiguration of the GUI to add 6 textboxes specifying yaw, pitch and roll of the coordinate systems relative to each other, as well as a 3D visualisation of the coordinate systems relative to beamline coordinates
 - A unit test for the rotations in the GenSAS class


rozyczko:
This looks impressive! Thanks for putting so much effort.

Before starting a proper code review let me just say my $0.02 about the framework used.
We don't really use QML/QtQuick in SasView for a variety of reasons. It looks good but connecting it with the Python logic is not as transparent as QtWidgets. It requires a completely different paradigm of thinking when creating an interface (declarative!) which significantly increases barrier of entry to new coders. It requires a separate UI editor (Composer). It also adds a non-negligible (in size and maintenance) dependency on additional modules (QtQuick).
This particular feature is probably OK, since it is self-contained and the interaction with the rest of SasView is in one direction only, but it does creates a precedent. 

I'd be happy to hear others' opinions on adding QML to SasView.

rjbourne:
That makes a lot of sense. I had hoped that the dependencies were already present in Sasview - my installation of Sasview 5.0.4 already contained `Qt5QML.dll` and  `Qt5Quick.dll`, although looking at it again now I see that the corresponding python extension modules are not present.

An alternative solution to the problem which only uses already existing components would be to use a matplotlib 3D graph and the already present `sas.qtgui.Plotting.Arrow3D` module. The upsides to this are all that you mentioned above, it adds no new dependencies or barriers to entry. Given matplotlib's speed with 3D plots I suspect the downside is that it would no longer be able to be a live view of the coordinate systems (although I can have a play with this to see how responsive it is). Perhaps it could be called up on command via a button. The user sets all the required setting and can then `view coordinates`.

Given that there are 6 different variables required to specify two 3D rotations I think that having some sort of visual representation is quite important. If you have any ideas about other ways to create them I'd be glad to hear them, I'm not overly familiar with all the functionality in Qt, or whether another of Sasview's dependencies could be used to address the issue, I didn't see anything in those I recognised, but I'm not familiar with many of them.

dehoni:
ready for testing on Win

dehoni:
ready for testing on Win

rjbourne:
The current version of Qt used in the developer environment does not support QML quick3D, meaning that the jenkins build of this ticket does not work. As discussed in one of the Tuesday meetings, the investigation into a dependency to enable faster and more user friendly 3D graphics is a longer term issue. The following commits:

 - fully separate the visualisation of the coordinate systems from the rest of the code. The visualisation is now located in just four methods of the calculator: `setup_display()`, `update_coords()`, `update_polarisation_coords()`, `set_polarisation_visible()`
 - implement a matplotlib version of the visualisation. This appears to be able to run smoothly on my computer, however further testing of performance is required. Additionally matplotlib's 3D support is still evolving, meaning that there is some discrepency in the functionality of version 2.2.3 (used by jenkins) and 3.4.3 (most recent version used by github actions). I have attempted to limit the amount of variation in the display due to this.

wpotrzebowski:
@dehoni to contact Naemi Leo about the review

wpotrzebowski:
Will not work with Jenkins mpl version

rjbourne:
I have set up Matplotlib 2.2.3 and Python 3.6 as per the developers environment in conda_qt5_win.yml and it appears to be working correctly, with the exception of stretched 3D aspect ratios, which seems to be a well known issue in this version of mpl. I'll trigger a Jenkins build and hopefully I can use that to work out where the problem is, or at least the difference in the environment. Are you getting errors/crashes or is it just not displaying?

rjbourne:
ready for testing on Win

rozyczko:
> @rozyczko and @wpotrzebowski, given this is a long standing bug from somewhere else and assuming the experts are happy that the functionality does what it says (even if there is always a bigger wish list) can we then move this to ready to merge? or do you still have concerns form your above comments?

I think the branch should be merged, considering importance of this feature.
The double legend bug and the additional requests/observations brought up by Naemi should be made as separate issues.

butlerpd:
@wpotrzebowski ... is this ready then to move to "ready to merge" or are you still working on the review?

wpotrzebowski:
ready for testing on OSX

wpotrzebowski (approved):

PR #1901, opened 2021-08-27
Improved add/multiply editor

rozyczko:
Added `@` to the list of operators and pretty-fied the label on the widget.


smk78 (approved):
Have pulled and built this branch locally on x64/W10. It does what it says! I also checked that plugins created loaded in the current release too, and they do.

However, I did spot a bug which is also present in the current release: if you use add/multiply and specify an _existing_ plugin model plugin name, and then change the model selection, clicking Apply overwrites the existing plugin model even if the 'Overwrite' box is unchecked! I will create a separate ticket for this.

Issue #1869, opened 2021-06-12
Make version of main dependencies easily accessible in distributables


butlerpd:
For debugging purposes it would be helpful to be able to easily see the version number of all dependencies uses.  This may be a lot but we could start with at least sasmodels and bumps perhaps?  It could probably be added to the about box or some config file during the build process.  Would need to verify that the distributable build (exe and dmg builds) carry this info as well as when doing local builds

Caddy-Jones:
Hi @butlerpd, I've had a look at the issue, and was wondering if you could clarify something for me. 

Are you thinking of having SasView document the current version number of all the dependencies currently installed within the virtual environment upon running SasView? Similar to running pip list/freeze, but outputting it to the config file or something.

Or is it better to have a list of dependencies created during installation by scanning through the SasView project scripts? In theory that list can then be used to set up the anaconda virtual environment using dynamic requirements for SasView, but this is probably overcomplicating things

Caddy-Jones:
Or maybe both?

butlerpd:
humm... you are way ahead of me @Caddy-Jones 😃 The issue was raised because, despite the yml files, local developer machines tend to have different versions of packages and, as was discussed at the last call, even github actions and the jenkins build machines use different versions of the dependency. Further, because of the hardcoding of the dependency versions on the Jenkins builds, it happens not infrequently that new features end up breaking the installer builds.  Then, for those of us on the help lists trying to respond to problems in different release versions of SasView it is sometimes useful to know what version the build machine had at the time the release was made.  Finally, for FAIR and Opent data etc initiatives, being able to identify which versions of dependent packages were used would probably be a good thing for transparency (e.g. the version of SasView used for the analysis was based on Scipy 1.5.2 which is known had a bug in the Legendre calculation - I'm making that up of course but hopefully makes the point 😃 

I will admit however that I do not understand enough myself to figure out which of your scenarios above would be the best approach without digging a bit.  Hopefully, the explanation of what I was suggesting we need to achieve helps? Otherwise maybe @rozyczko, @wpotrzebowski or @krzywon might be able to answer better?

andyfaff:
> even github actions and the jenkins build machines use different versions of the dependency

If both of these build systems make a new environment (venv or conda), then using a requirements.txt or an environment.yml should mean that the same dependency versions are used.

lucas-wilkins:
I think this will be resolved by #2177

lucas-wilkins:
[...]

PR #1903, opened 2021-08-31
1869 easily accessible dependencies

Caddy-Jones:
Hi guys

I’ve done some work on issue #1869, but need advice on the next steps.

This pull request does 2 things, 
On start up pip is used to generate a list of locally installed modules, and their version number. This only works for modules hosted on pip, so Matplotlib and bumps, but no sasmodules. 
On exit, python saves a dictionary of modules used while running this session of SasView as the key, with the accompanying version number as the value. This works for all modules*, including sasmodules.

There are a few areas I need clarification on.
1.	I am currently saving the outputs to 2 .txt files in a new folder called temp. This is a temporary measure, as I am unsure where best to save the outputs. The original issue mentioned saving the versions to the about box or some config file.
2.	Currently pip gathers the installed modules each time SasView is run. Since we cannot guarantee the user has internet access when running SasView, and the installed modules versions are unlikely to change, I think this should be moved to be part of the installation/ update.
3.	There are modules with version numbers that cannot be accessed, such as PyQt5. There is a commented out section of code which uses pip to try and get the module version, however this is a long process, so I decided to not include it. It could also be worth saving a list of modules where the version number could not be ascertained.


src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -104,6 +105,18 @@ def __init__(self, parent=None):
         # Log the python version
         logging.info("Python: %s" % sys.version)
 
+        import subprocess, json
+        installed_modules = {'python': sys.version}
+
+        # Get python modules installed locally
+        installed_modules_json = json.loads(subprocess.check_output("pip list -l --format=json"))
```
andyfaff, line 112: The following gets a string of pip installed packages in the current Python environment, without using a subprocess.

```
         try:
            from pip._internal.operations import freeze
        except ImportError:
            return ""

        return "\n".join(list(freeze.freeze()))
```
For conda installed packages you'd end up with something like:
`numpy @ file:///Users/runner/miniforge3/conda-bld/numpy_1629092076711/work`. To deal with those you'd have to use something like `subprocess.check_output("conda list")`, but conda may not be being used.

The pip approach works very well in a pyinstaller built standalone executable. When I build standalone executables of this type (perhaps sasview does the same), I always do it in a new virtual environment with all packages installed via a pip requirements file. Only using pip means the list is comprehensive. For the packages that need to be installed from source I do that with pip as well:
```
# in source dir
pip install .
# or
pip install git+https://.....
```

It seems that this short snippet would be more robust than the code in the `ModuleGatherer.py` file.

Caddy-Jones: Hi @andyfaff, thanks for the review

As far as I can tell, pip freeze / list etc get a list of modules installed on the system, either in a local virtual environment or globally. That’s is useful information to have and I aimed to make it accessible to the user. I see what you’re saying about the pip approach, and will also add the conda list function.

However I wanted to do something slightly different with the get_imported_modules function. My aim was to list every module that python imports, for example using an “import x” call, or a “from x import y as z” call; and get the exact version of that module. This is for 2 reasons. Firstly there is a chance that two versions of a module are present, one in the virtualenv and one on the system for example. I want to double check which one is being accessed. Secondly there may be modules in the virtual environment listed in pip that where not used in a session of SasView. Listing the imported modules will only show modules used in a session, so can help narrow down the search when attempting to bugfix. 

I couldn’t find any pip or built in function that does this accurately, hence the code in get_imported_modules.

I hope this made sense.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -104,6 +105,18 @@ def __init__(self, parent=None):
...
+        for mod in installed_modules_json:
+            installed_modules[mod['name']] = mod['version']
+
+        f = open("temp/installed_modules.txt", 'w')
```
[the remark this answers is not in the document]
Caddy-Jones: Thanks, I’ll change it. Although this is only a temporary solution as the modules information will probably best be saved to the config file or something.

src/sas/qtgui/MainWindow/GuiManager.py
butlerpd: That's a great point @Caddy-Jones. That could be a great help in tracking down failures that only occur for certain people I would guess?

Caddy-Jones: I hope so, I'm aiming to give the debuggers as much information as possible

wpotrzebowski:
@rozyczko will provide some hints to @Caddy-Jones

butlerpd:
ready for testing on Win

butlerpd (commented):
The code looks safe enough but am not comfortable judging beyond that.  Functionality does seem to work -- though it is a **lot** of modules which explains why @Caddy-Jones was worried about dumping the output too often.  My question now is the json output which does not seem to be compatible with the json viewers I had due to the use of 'xx' rather than "xx" for strings?
It would be helpful if there were an easy way to cut and paste the output into a tool to examine compared to, for example the required files or yml files?

That said, if this passes a better code review the concern above can be figured out later and turned into another issue.  This is already a great step forward.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,228 @@
...
+        logging.info(f"Imported modules\n"
+                     f"{imported_modules}")
+
+
+    def get_imported_modules(self):
+        """ Get a dictionary of imported module version numbers
+
+        Use a variaty of method, for example a module.version call, to attempt to get the module version of each
+        module that has been imported in this instance of running SasView. The sys.modules command lists the
+        imported modules. A list of modules who's version number cannot be found is also included.
```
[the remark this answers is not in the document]
Caddy-Jones: Done :)

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,228 @@
...
+            A dictionary with the module names as the key, with their respective version numbers as the value.
+        """
+
+        module_versions_dict = {'python': sys.version}
+        unattainable_modules = []
+
+        for mod in sys.modules.keys():
+
+            # Not a built in python module, which has no version, only the python version
+            if mod not in sys.builtin_module_names:
```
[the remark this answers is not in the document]
Caddy-Jones: Hi paul, I see how deep nested try except blocks are ahard to read, so i've reformatted the code to avoid this. 
I've also used the python library to generate a list of standard modules as you suggested, which works well.

Caddy-Jones:
Hi @butlerpd 

The pip command does list command does output a json file, however the log_installed_modules function converts the json to a python dictionary. As a result the output will not be compatible with a json viewer.

There isn't a critical reason for this change, it was just a stylistic choice to keep the outputs of both log_installed_modules and log_imported_modules the same, however both functions could easaliy be modified to output jsons if that would make life easier

It could also be an idea to compare the outputs to the yml file, and to make any differences easily visible in the functions outputs. Something like
{
    Modules different from yml file: {module: version, module_2: version_2,  etc...}

    Modules consistent with yml file: {module: version, module_2: version_2,  etc...}

    Modules whose version number could not be found: {module_1, module_2, etc...}
}
But I think this may be best answered in a separate pull request

butlerpd:
Thanks for the clarification.  The comparison might be nice but not sure that works once you make a build as I don't think the build slurps in the yml files?  So this could cause a problem when trying to build .. and to some extent duplicates the work of "check_packages.py" so not really necessary when not working with a built version?

Json output *might* help but then maybe not. I think that requires more thinking so I agree that all this can be put into a future ticket as desired.

I will try to test the new changes (responding to @pkienzle ) later today to make sure it doesn't break anything obvious to me at which point I am ready to approve.

butlerpd (approved):
Works as before for me (without building an installer) so this can now be merged.  I note that the following online json formatter/beautifier can deal with most of the issues and "fix" the "json" and render it in an easy to read format.
https://jsonformatter.curiousconcept.com/

Issue #1906, opened 2021-09-01
Easy Add/Multiply Editor overwrites an existing plugin model even if Overwrite box is unchecked


smk78:
While testing https://github.com/SasView/sasview/pull/1901 I noticed a bug **which is also present in the current release**: if you use add/multiply and specify an existing plugin model plugin name, and then change the model selection, clicking Apply overwrites the existing plugin model even if the 'Overwrite' box is unchecked!

rozyczko:
This indeed is a serious issue.
Fixed.

smk78:
Testing 5.0.5rc1 on W10/x64 this seems to be fixed:
![image](https://user-images.githubusercontent.com/10938679/150002121-81a9a790-50f5-4d88-a114-773b1ac467f1.png)

So am closing this issue.

PR #1907, opened 2021-09-01
Distinguish standard model list from full list with plugin models. #1906

rozyczko:
Addresses an issue found by @smk78 while testing #1901

### DIFF
--- a/src/sas/qtgui/Utilities/AddMultEditor.py
+++ b/src/sas/qtgui/Utilities/AddMultEditor.py
@@ -72,6 +72,7 @@ class AddMultEditor(QtWidgets.QDialog, Ui_AddMultEditorUI):
         self.setupSignals()
 
         self.list_models = self.readModels()
+        self.list_standard_models = self.readStandardModels()
 
         # Fill models' comboboxes
         self.setupModels()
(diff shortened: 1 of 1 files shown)


smk78 (approved):
Tested functionality on W10/x64.

This PR rectifies the unwanted behaviour I identified in https://github.com/SasView/sasview/pull/1901. A warning popup is now deployed if an attempt is made to create a new plugin (or recreate one) with the same name as an existing model in the plugins folder. The existing model can be overwritten only if the overwrite box is checked (the default is unchecked).

Thumbs up from me!

wpotrzebowski:
ready for testing on Win

src/sas/qtgui/Utilities/AddMultEditor.py
```diff
@@ -108,17 +109,28 @@ def setupModels(self):
...
+        """ Generate list of standard models """
+        s_models = load_standard_models()
         models_dict = {}
-        for model in models:
+        for model in s_models:
             if model.category is None or 'custom' in model.category:
                 continue
             models_dict[model.name] = model
 
         return sorted([model_name for model_name in models_dict])
```
wpotrzebowski, line 122: Should _readModels_ load standard models or all models?  Rather stylistic comment but _readModels_  and _readStandardModels_ seem to be quire redundant maybe we can remove some code duplication by passing arguments?

PR #1909, opened 2021-09-01
CI: use pip --cache to find dir

andyfaff:


andyfaff:
With the `python GUITests.py` it seems that there are a lot of tests failing. Are these known failures? Or do the test suites need updating? I don't think they're failures with this PR.

wpotrzebowski:
> With the `python GUITests.py` it seems that there are a lot of tests failing. Are these known failures? Or do the test suites need updating? I don't think they're failures with this PR.

I am afraid these tests having received much attention recently (particularly on Linux) and you are right they may be failing for different reason. I would suggest merging non-breaking changes from this PR and address tests issue in the separate PR/ticket.

.github/workflows/test.yml
```diff
@@ -131,5 +97,7 @@ jobs:
       env:
         PYOPENCL_COMPILER_OUTPUT: 1
       run: |
+        # make sure the test will fail if the gui test doesn't work
+        set -e -o pipefail
         cd src/sas/qtgui
-        xvfb-run -a --server-args="-screen 0 1024x768x24" python GUITests.py || true
```
llimeht, line 135: These test failures are suppressed with `|| true` because they fail badly. They are worth running because they're fast enough and to track the (very slow) progress towards fixing them.

.github/workflows/test.yml
```diff
@@ -85,28 +68,11 @@ jobs:
...
-
-    - name: Build and install sasmodels
-      run: |
-        cd ../sasmodels
-        rm -rf build
-        rm -rf dist
-        python setup.py clean
-        python setup.py build
-        python -m pip install .
+        pip install git+https://github.com/bumps/bumps.git
```
llimeht, line 73: This would be nice if we were able to use `pip install git+https:..` everywhere and not just in this job. Can it also be used in `installers.yml` and `docs.yml`? I suspect not as those jobs need the unpacked sources for the `bumps` and `sasmodels` present to be able to build the `sasview` documentation.

The follow-on question is then: do we want to use different methods in different jobs?

.github/workflows/test.yml
```diff
@@ -28,32 +28,15 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     ### Caching of pip downloads and local wheel builds
+    - name: Get pip cache dir
+      id: pip-cache
+      run: |
```
llimeht, line 33: Nice change to remove some annoying duplication here. Thanks!

BTW this could be a single line for `run` without the `|` multiline indication.

.github/workflows/test.yml
```diff
@@ -28,32 +28,15 @@ jobs:
...
-        restore-keys: |
-          ${{ runner.os }}-pip-${{ matrix.python-version }}-
-          ${{ runner.os }}-pip-
-
-    - name: Obtain pip cache (Windows)
-      uses: actions/cache@v2
-      if: startsWith(runner.os, 'Windows')
       with:
-        path: ~\AppData\Local\pip\Cache
+        path: ${{ steps.pip-cache.outputs.dir }}
```
llimeht, line 39: These changes are probably also worth making over in `docs.yml`. 

(Even better would be to split these jobs so that there was a `build` stage that was common and then `test`, `guitest`, `docs` tasks that could run in parallel; at present that parallelisation is faked by using the separate `docs.yml` which duplicates code and compilation CPU usage in order to provide faster CI completion. Maybe those are changes for another day, however.)

llimeht (changes requested):

.github/workflows/test.yml
andyfaff: In order to track progress towards fixing them wouldn't it be better to mark them as known fails?

llimeht: That would be sensible too. I'm not sure how `@unittest.expectedFailure` interacts with the custom test collection and runner that is used in `GUITests.py`. (It's also not clear to me that there is any real value in the custom runner over what can be achieved fairly easily with pytest.)

wpotrzebowski:
As discussed at the call we are closing the PR as it has became obsolete

PR #1912, opened 2021-09-08
Allow output of scattering calculator to be used as a theory curve

rjbourne:
The current code already sends the output of the generic scattering calculator to the data explorer window where it can be used for fitting in the same way as any data loaded from a file, a large part of the functionality requested by this ticket. The results of this fitting could not be plotted however. This was due to a mismatch in the name of the data and the corresponding plot. This pull request ensures the name of the plot matches the name of the data rather than being left as the default `"Data1D"`/`"Data2D"`, and hence allows models and residuals to be plotted as would be expected.

### DIFF
--- a/src/sas/qtgui/Utilities/GuiUtils.py
+++ b/src/sas/qtgui/Utilities/GuiUtils.py
@@ -411,6 +411,7 @@ def createModelItemWithPlot(update_data, name=""):
     Adds 'update_data' to that row.
     """
     py_update_data = update_data
+    py_update_data.name = name # name must match title due to how plots are referenced
 
     checkbox_item = HashableStandardItem()
     checkbox_item.setCheckable(True)


wpotrzebowski:
It needs to be tested across entire saview where plotting is involved

src/sas/qtgui/Utilities/GuiUtils.py
```diff
@@ -411,6 +411,7 @@ def createModelItemWithPlot(update_data, name=""):
     Adds 'update_data' to that row.
     """
     py_update_data = update_data
+    py_update_data.name = name # name must match title due to how plots are referenced
```
rozyczko, line 414: As mentioned during the call, this updates `name` attribute on all plots generated using this utility call.
Those live in:
- Data Operation Utility panel
- Generic Scattering calculator
- Fitting widget
- Slicer parameters widget
- Image viewer
- File loader

I **think** this change is safe, but it would probably be prudent to assure the above functionality remains unchanged (including unit tests).

rjbourne: I have performed an initial test of the functionality compared to version 5.0.4:

1) The Data Operation Utility Panel appears to work as before - datasets can be combined and then plotted or passed to the fitting panel. 
I have noticed a couple of things in version 5.0.4 that feel odd, I'm not sure of the intended functionality but these are potentially new tickets:
    - The title of the combined data in the data explorer does not match its name but the name of the second dataset in the combination:
image
   - After deleting one of the generated datasets e.g. MyNewDataName above using the 'Delete Data' button in the data explorer, the name still cannot be re-used in the Data Operation Panel - it is highlighted with a red background as still existing.

2) The generic scattering calculator functionality is changed in this ticket as expected. Plots can now be generated in the fitting calculator after creation by the scattering calculator.

3) The fitting widget uses this function only when creating a new theory index. While I haven't tested all possible routes into this function the basic use of creating a theory from a model appears to work correctly - I can plot it and append it to plots.
   - I notice that in 5.0.4 when saving a project and re-opening it all theories become standard datasets. Again I'm not sure of the intended functionality but this seems odd - should it become a ticket?

4) The slicer parameters widget appears to work as in 5.0.4

5) The Image Viewer appears to have the same issue as the generic scattering calculator in 5.0.4: Once an image has been loaded as data - it cannot be plotted by the fitting panel. This is also fixed by the change in this PR. Another potential issue in 5.0.4 which this PR does not address:
   - The title of the loaded image data is not displayed in the data explorer
image

6) The FileLoader uses this method in two places.
    a) DataExplorer.py > updateModelFromData() which is fundamentally used when loading in projects. I can save a project made in this branch and reload it as expected. In sasview 5.0.4 saving and reloading a project would fix any issues of plots not be able to be viewed from the fitting calculator - although at the cost of some datasets being renamed to generic names such as `Data1D` and `Data1D [1]`. When re-opening 5.0.4 projects in this branch the projects load as they would have in 5.0.4.
    b) DataExplorer.py > freezeDataToItem(). I cannot find where this method is being used - but given it passes in a `new_name` argument as a new identifier I see no reason that this would cause issues.
 
I am not confident that I know how the code for these widgets works well enough to write unit tests for them, and now that university has restarted I doubt I will have the time to learn this many different areas of the codebase; hopefully this is a reasonable starting point.

butlerpd:
there are a number of strange behaviors with linearized plots but it is hard to tell whether they are due to this PR or from the state of main at the time this branch was created. Can we merge main back into this PR so that it is clear what new behaviors are due to the new code in this PR?

butlerpd:
ready for testing on Win

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -60,6 +60,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Expose the loaded perspectives for easy linking between perspective methods/inputs and plots
+LOADED_PERSPECTIVES = {}
```
pkienzle, line 64: Global variables are contraindicated, especially in this case where they lead to circular imports.

I do not know the qt gui code well enough to code it, but it seems to me that you must be able to find the graph associated with the model (otherwise you wouldn't be able to update the theory plot when parameters change in the fitting perspective). Can you use this mechanism to find the plot associated with the perspective that needs a Q range? Then you can call the Q range selector on the specific plot with a callback to adjust low/high values when the interactors are moved. In particular, your Q range slider code won't need to rummage through the perspective tabs to find the associated perspective and it won't need to know anything about the structure of the perspective to set up the associated getter/setter/input. It should also work well with a linear fit tool, which wants to use the data for selecting the range but not the associated perspective.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -172,49 +300,273 @@ def selectedshapechange(self):
...
+        """Verifies that enabled files are compatible and can be combined
+
+        When the user wishes to combine two different files for nuclear and magnetic
+        data they must have the same 3D data points in real-space. This function
+        decides whther verification of this is necessary and if so carries it out.
+        In the case that the two files have the same real-space data points in different
+        orders this function re-orders the stored data within the MagSLD objects to make
+        them align. The full verification is only carried out once for any pair of loaded
+        files.
+        """
```
pkienzle, line 343: Move what code that doesn't depend on gui to sascalc. You may want to mix and match files when running a remote batch job for example.

dehoni: This issue will be addressed with [PR1898](https://github.com/SasView/sasview/pull/1898).

butlerpd:
ready for testing on Win

butlerpd:
The github actions installer suffers the problem describe in issue #1981 so I was unable to test this simple fix.  Moreover it is built against python 3.9 and does not seem to have used the latest actions setup.  That however will probably not fix the #1981 issue since that issue exists in the latest PRs of a number of branches.

As I recall, this branch requires matplotlib 3 so may not run from the Jenkins build but will try that next.

butlerpd (approved):
Works as advertized

Issue #1914, opened 2021-09-16
Simultaneous fitting returning incorrect fit


Caddy-Jones:
When attempting to use the “constrained or simultaneous fit” functionality with the master branch SasView version, the fit isn’t correct. 

For example, the attached plot shows the fit the three tutorial files: AOT_Microemulsion-Core_Contrast.xml, AOT_Microemulsion-Shell_Contrast.xml, and AOT_Microemulsion-Drop_Contrast.xml; using SasView on the Master branch. This seems incorrect and is different to the plots obtained using the current released version of SasView, version 5.0.4, and the demonstration plots in the tutorial. Also the scale and background parameters seem to be constrained without being asked to.
![shell_contrast_AOT_me_SANS_incorrect](https://user-images.githubusercontent.com/88383842/133593734-70681661-cb47-4bac-a875-ef36f5938009.png)
![core_contrast_AOT_me_SANS_incorrect](https://user-images.githubusercontent.com/88383842/133593741-a7037d87-b520-47c9-a7b2-eb351ccc0389.png)
![drop_contrast_AOT_me_SANS_incorrect](https://user-images.githubusercontent.com/88383842/133593744-a1e53fc6-6e80-4b3a-84cc-e30c42a0ca92.png)

When attempting to plot two sets of experimental data, plots also attaches, SasView seems to only use the data from the first fit page for fitting. This behaviour is different form the behaviour of the released version of SasView, version 5.0.4. Again all selected fitting parameters, in this case sld_core, sld_shell and sld_solvent, appear to be constrained with the same values between the datasets. This is without any constraints being added.
![SAXS_incorrect_plot](https://user-images.githubusercontent.com/88383842/133593764-0434443f-f541-4416-b5c0-6d2beb56a1be.png)
![SANS_incorrect_plot](https://user-images.githubusercontent.com/88383842/133593767-616120fc-84ef-487a-a61f-57e7c47ac4e4.png)

smk78:
I concur @Caddy-Jones . But I think it's more serious than that; I think fitting is totally broken! You can't even get a single fit to converge. And it's affecting multiple optimisers. I'm making this a blocker for the moment.
But the question is, what could have caused this?

smk78:
Does _main_ now require a later version of bumps? I currently have 0.7.11.

Caddy-Jones:
[...]

PR #1916, opened 2021-09-22
Using the uncertainties module to propagate errors

Caddy-Jones:
As discussed in issue #1914, the pull request from #1682 lead to fitting error in SasView. This pull request attempts to accomplish the same goal of propagating errors on constrained parameters by using code inspired by #1682, while avoiding the fitting issues that lead to #1914.

### DIFF
--- a/src/sas/sascalc/fit/BumpsFitting.py
+++ b/src/sas/sascalc/fit/BumpsFitting.py
@@ -322,25 +322,23 @@ class BumpsFit(FitEngine):
         # collect the results
         all_results = []
 
-        for M in problem.models:
-            fitness = M.fitness
+        for fitting_module in problem.models:
+            fitness = fitting_module.fitness
             pars = fitness.fitted_pars + fitness.computed_pars
             par_names = fitness.fitted_par_names + fitness.computed_par_names
 
-            R = FResult(model=fitness.model,
-                        data=fitness.data,
-                        param_list=par_names)
-            R.theory = fitness.theory()
-            R.residuals = fitness.residuals()
-            R.index = fitness.data.idx
-            R.fitter_id = self.fitter_id
+            fitting_result = FResult(model=fitness.model, data=fitness.data, param_list=par_names)
+            fitting_result.theory = fitness.theory()
+            fitting_result.residuals = fitness.residuals()
+            fitting_result.index = fitness.data.idx
+            fitting_result.fitter_id = self.fitter_id
             # TODO: should scale stderr by sqrt(chisq/DOF) if dy is unknown
-            R.success = result['success']
-            R.convergence = result['convergence']
+            fitting_result.success = result['success']
+            fitting_result.convergence = result['convergence']
             if result['uncertainty'] is not None:
-                R.uncertainty_state = result['uncertainty']
+                fitting_result.uncertainty_state = result['uncertainty']
 
-            if R.success:
+            if fitting_result.success:
                 pvec = list()
                 stderr = list()
                 for p in pars:
@@ -391,16 +389,16 @@ class BumpsFit(FitEngine):
                             pvec.append(p.value)
                             stderr.append(0)
 
-                R.pvec = (np.array(pvec))
-                R.stderr = (np.array(stderr))
+                fitting_result.pvec = (np.array(pvec))
+                fitting_result.stderr = (np.array(stderr))
                 DOF = max(1, fitness.numpoints() - len(fitness.fitted_pars))
-                R.fitness = np.sum(R.residuals ** 2) / DOF
+                fitting_result.fitness = np.sum(fitting_result.residuals ** 2) / DOF
             else:
-                R.pvec = np.asarray([p.value for p in pars])
-                R.stderr = np.NaN * np.ones(len(pars))
-                R.fitness = np.NaN
+                fitting_result.pvec = np.asarray([p.value for p in pars])
+                fitting_result.stderr = np.NaN * np.ones(len(pars))
+                fitting_result.fitness = np.NaN
 
-            all_results.append(R)
+            all_results.append(fitting_result)
         all_results[0].mesg = result['errors']
 
         if q is not None:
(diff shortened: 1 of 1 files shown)


src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -282,77 +282,112 @@ def fit(self, msg_q=None,
...
+                        # Details of p
+                        param_model, param_name = p.name.split(".")[0], p.name.split(".")[1]
+                        # Constraints applied on p, list comprehension most efficient method, will always return a
+                        # list with 1 entry
+                        constraints = [model.constraints for model in models if model.name == param_model][0]
+                        # Parameters p is constrained on.
+                        reference_params = [v for v in varying if str(v.name) in str(constraints[param_name])]
+                        err_exp = str(constraints[param_name])
+                        # Convert string entries into variable names within the code.
+                        for i in enumerate(reference_params):
```
rozyczko, line 359: this looks a bit awkward, maybe use something like:
```
for i, p in enumerate(reference_params):
    err_exp = err_exp.replace(p.name, f"reference_params[{i}].value")
```

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -282,77 +282,112 @@ def fit(self, msg_q=None,
...
+                            index = i[0]
+                            err_exp = err_exp.replace(reference_params[index].name, f"reference_params[{index}].value")
+                        try:
+                            # Evaluate a string containing constraints as if it where a line of code
+                            pvec.append(eval(err_exp).n)
+                            stderr.append(eval(err_exp).s)
+                        except NameError as e:
+                            # Get model causing error
```
rozyczko, line 367: I think `eval` returns all sorts of exceptions, also e.g. `SyntaxError` - are you sure these can't be triggered by `err_exp`?
Maybe we need a blanket `except Exception as e:` instead?

rozyczko (approved):
Code looks good, minor stylistic issues raised

src/sas/sascalc/fit/BumpsFitting.py
Caddy-Jones: That's a good point, I'm pretty sure Syntax error will not occur, as you are limited when writing a constraint manually. But I will add a second exception catcher in the case of an unexpected error that will affect the error propagation, but not the plotting

This Exception is caused when a user has a constraint on for a parameter that hasn't been calculated. For example:
M1.radius = M2.radius * 2
M3.radius = M2.radius + M1.radius 
is fine but
M3.radius = M2.radius + M1.radius 
M1.radius = M2.radius * 2
will lead to an error, as m3 is constrained based on the value for m1, but this hasn't been calculated yet as there is a constraint on m1.

This error could be fixed by looping over all the constraints multiple times, until the prerequisites for each parameter has been computed. However this hasn't been added as it's easy for the user to avoid the issue, by reordering the constraints.

RichardHeenan: Are we allowing iterative, stacked up, constraints at all in the constraints set up? I thought not ????
( I have made much use of them myself in FISH, but you do have to be careful as to the order in which they are applied, e.g I suspect sorting the error bars out might require running through the list of constraints in reverse order?)

smk78 (approved):
Tested functionality on W10/x64 comparing against release 5.0.4.
Tried single fit (using LM) and 3 dataset simultaneous fit (using DREAM).
Behaviour seems 'normal'. Uncertainties on optimised parameters look sensible. Residuals look sensible. DREAM plots looked sensible. DREAM output in Log Explorer looked sensible and parameter values were correctly assigned to the correct parameters (something that didn't happen in PR1682).
Parameter/uncertainty values matched v5.0.4 to many dp when using LM. Correspondence using DREAM was obviously not as good but still acceptable (IMHO); eg:

In 5.0.4:
scale,True,0.068696,0.00058734,0.0,inf,():
background,True,-0.057794,0.061883,-inf,inf,():
radius,True,23.969,0.10756,0.0,inf,():
thickness,True,9.0774,0.06878,0.0,inf,():

In this branch:
scale,True,0.068957,0.00060114,0.0,inf,():
background,True,-0.058371,0.0679,-inf,inf,():
radius,True,23.915,0.11556,0.0,inf,():
thickness,True,9.0682,0.089137,0.0,inf,():

So I approve this PR.

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -282,77 +282,124 @@ def fit(self, msg_q=None,
...
+                fitted = uncertainties.correlated_values(values, cov)
+                for param, val in zip(varying, fitted):
+                    param.value = val
+
+        # Propagate correlated uncertainty through constraints.
+        problem.setp_hook()
 
         # collect the results
         all_results = []
+
```
wpotrzebowski, line 324: Stylistic issue (and not introduced by @Caddy-Jones) but I think we try to avoid one letter variables (M, R).

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -282,77 +282,124 @@ def fit(self, msg_q=None,
...
+                        # Parameters p is constrained on.
+                        reference_params = [v for v in varying if str(v.name) in str(constraints[param_name])]
+                        err_exp = str(constraints[param_name])
+                        # Convert string entries into variable names within the code.
+                        for i, index in enumerate(reference_params):
+                            err_exp = err_exp.replace(reference_params[index].name, f"reference_params[{index}].value")
+                        try:
+                            # Evaluate a string containing constraints as if it where a line of code
+                            pvec.append(eval(err_exp).n)
+                            stderr.append(eval(err_exp).s)
```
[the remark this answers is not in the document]
Caddy-Jones: Hi Wojciech. You are correct, the constraints are definitely applied. The try clause is for the eval expression, which can fail is the constraint is based on an undefined parameter.

PR #1919, opened 2021-10-04
Slicer extension 1344

butlerpd:
This PR attempts to allow the box slicers to take slices whose center is not 0,0. if using the full q range of the long dimension this would be the same as giving slices at constant qx yielding I(qy|qx=constant) for example.


butlerpd:
ready for testing on Win

kmothander:
I've tested this locally on Linux.

The errors that occurred previously are fixed, sometimes I get the error "ValueError: Average Error: No points inside ROI to average...", but not 100% reproducible, and only happens at -Qx. I'm not really sure where this error comes from or how to fix it...

I've also fixed so that the box is placed in the middle of the plot not just at Qx = Qy = 0. However, if the plot only contains data at negative Qx (and negative Qy) then I get the same error as above. 

I've added a x_center and y_center to the parameter list. The parameters can be edited through the Edit slicer parameter window. 

Question:
As @butlerpd mentioned, should there be other parameters then x_max, y_max, x_center and y_center? I think it should be possible to have x/y_center and then an sizes of the box, in the same way as for the BoxSum. Or should the size and placement of the box be defined in some other way?

butlerpd:
LOL .. I see that it looks like these commits break all the unit tests .. which makes sense since the behavior has been altered.  Those will need to be adjust clearly to reflect the new correct behavior.

Regarding the question about how to present the width variables that is a question of what might be easier/more intuitive for a user. My thinking was that for the purpose at hand my guess is that the user will more likely be thinking in terms of delta Q around the center rather than qmax?

Another usability question that has come to mind is whether for the boxslicer as opposed to boxsum, it would not make much more sense (and be less confusing to a new user) if the default box was not drawn as a square from -.5 imagedQ to +.5 imagedQ in both directions.  I would think something like -3/4 imagedQ to + 3/4 imagedQ (or even -1 to 1) in the direction of the q being scanned and maybe -.2 to +.2 in the direction of q being integrated.

Note that by imagedQ I mean qmax-qcenter of the displayed data (image)

kmothander:
@butlerpd
I have no idea how to change the unit test...

To your other questions I can change and add the new functionality. Changing the parameter list and see if I can change the initial size of the box!

wpotrzebowski:
@butlerpd will look at it

butlerpd:
There is a difference between very limited resources on a complex project and a stale project :-)

lucas-wilkins:
This branch is [9 commits ahead](https://github.com/SasView/sasview/compare/main...SlicerExtension_1344), [1374 commits behind](https://github.com/SasView/sasview/compare/SlicerExtension_1344...main) main.

lucas-wilkins:
> There is a difference between very limited resources on a complex project and a stale project :-)

Unlike issues, branches get behind main and merging them gets harder and harder. If you don't want a stale label, please keep it up to date with main, preferably with log10(commit_difference) < 3.

lucas-wilkins:
@butlerpd I'm feeling generous, have a star instead.

butlerpd:
OOpps.  a few problems with waiting so long:
1. We foolishly clicked on the github pullrequest "update branch" which merges in the commits from the parent .. but the parent here is listed as `release_5.0.5` instead of `main`
2. `release_5.0.5` has 3 commits not in `main` (though should be ok since they are commits and reverse commits), but we can no longer do a reset to undo the changes.
3. More importantly the calculation part of the slicers (`manipulations.py`) were in the dataloader folder (not really the directory they should have been in) which has now been moved to a different repo: `sasdata`.  Thus the changes to `manipulations.py` will be lost in merging.  So far the only change to that was on line 1016 (now 1021) when class SectorQ returns`self._avg(data2D, 'q2')` , the second term was changed from 'q2' to 'q'.  Apparently q2 folds the negative on top of the positive q whereas q would not. so again the non folding was apparently implemented in the calculation part of the sectors but not exposed to the GUI.  This is meant to serve as a record of that before it gets wiped in the update to come.

butlerpd:
This PR is getting close to be pulled off a draft.  Remaining issues:

-  [ ] The ROI issue noted by @rozyczko remains an issue. It is due to the ROI being so small as to not have an actual data in them so the calculator cannot operate. Oddly, this error does not seem to occur when using main even when the ROI cannot possibly have any data?
- [ ] The issue of the markers moving after releasing noted above. Still jumping issue
- [ ] Should implement the key move function (defined in the base interactor) so that slabx and slab y can be moved along the scanning direction using the arrow keys.
- [ ] The f`old=False` option needs to be implemented for the sector average just as it is for the box interactor
- [ ] The xaxis scale of the resulting plots should default to linear scale when fold is False
- [ ] Move the Validation of parameters to the Validate section and make sure all appropriate validation gets done

krzywon:
> that last commit message is nonsense

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message

krzywon (approved):

JoGaudet (approved):
I tested the new functionalities for the Box Slicer on Mac OS 11.2.3 and look all good.

astellhorn:
Review for MacOS 13.6.3:
1) changing x_width in the 2D window sometimes not fixes the position, one needs to several times click to fix new x_width. This happens to x_width in both Qx and Qy slicers. 
2) Same sometimes happens to Q_center position changes in the 2D view
3) y_width visually not updated when changing in the 2D window. Only after any other value is changed, then y_width gets updated. Also the value in the "Edit slicer parameters window" is not updated (also not after it visually changes in the 2D view when x_width is changed again - then in "Edit slicer" it still is the wrong value). Changing y_width in the "Edit box slicer" window works
4) Can SOMETIMES not set x_width. Error message "No points inside ROI to average" even though ROI is still inside 2D view and intensity>0. See attached image:
image
5) Right click in 2D window to edit plot options does not work (no window pops up). I.e., cannot change from log to lin scale etc. Maybe related to point (4) - not sure.

jack-rooks:
The merge addresses problem 3 in @astellhorn 's comment

jack-rooks:
@astellhorn there was an error every time you right clicked the 2D plot to pull up the context menu (the 5th problem you mentioned), can you post that error message please?

Additionally, regarding problems 1 and 2: are you saying that the response of the program was slow, and you needed to click multiple times to get it to stop moving the box when you move your cursor?

jack-rooks:
Pull request [SasData counterpart to SasView SlicerExtension_1344](https://github.com/SasView/sasdata/pull/61) will correct @astellhorn 's problem 4.

timsnow:
Screenshot 2024-01-22 at 09 33 48

butlerpd:
NOTE: I believe the lack of context menu (@astellhorn problem 5) was addressed only in release_6.0.0 branch by @rozyczko in PR #2670 (mainly because the ticket reporting the problem stated that it was a 6.0.0 problem but I've since realized it was also in main.  The merge of 6.0.0 in to main, also done by @rozyczko, should fix the problem once that PR (#2808) is merged

butlerpd:
@krzywon will give final code review check

src/sas/qtgui/Plotting/Plotter2D.py
```diff
@@ -396,6 +422,47 @@ def setSlicer(self, slicer, reset=True):
...
+        # Clear the old slicer plots so they don't reappear later
+        if hasattr(self, '_item'):
+            item = self._item
+            if self._item.parent() is not None:
+                item = self._item.parent()
+
+            # Go through all items and see if they are a plot. The checks done here are not as thorough
+            # as GuiUtils.deleteRedundantPlots (which this takes a lot from). Will this cause problems?
+            # Primary concern is the check (plot_data.plot_role == DataRole.ROLE_DELETABLE) as I don't
+            # know what it does. The other checks seem to be related to keeping the new plots for that function
```
krzywon, line 435: Why reinvent the wheel? What advantage does this have over the existing method?

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
```diff
@@ -269,6 +272,7 @@ def setParams(self, params):
             params["Delta_Phi [deg]"] = MIN_PHI
 
         self.nbins = int(params["nbins"])
+        self.fold =params["fold"]
```
krzywon, line 275: Assuming `params` is a dictionary (no typing in the method signature...), this should probably be a `get`, with some default value, and bool coercion. I'm not sure if params is set from an external source (like save states), but existing methods might not have the `fold` key, resulting in a `KeyError`.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -19,50 +19,78 @@ class BoxInteractor(BaseInteractor, SlicerModel):
...
-                           numpy.fabs(self.data.xmin))
-        # when reach qmax reset the graph
-        self.qmax = max(self.data.xmax, self.data.xmin,
-                        self.data.ymax, self.data.ymin)
+        if self.direction == "Y":
+            self.xwidth = 0.1 * (self.data.xmax - self.data.xmin) / 2
+            self.ywidth = 1.0 * (self.data.ymax - self.data.ymin) / 2
+            # when reach qmax reset the graph
+            self.qmax = max(numpy.fabs(self.data.ymax), numpy.fabs(self.data.ymin))
+        else:
```
krzywon, line 39: This case holds where `self.direction != "Y"`. Is this what we really want? Does case (y vs. Y) matter? What should happen if something other than X or Y is passed?

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -236,9 +289,11 @@ def getParams(self):
 
         """
         params = {}
-        params["x_max"] = numpy.fabs(self.vertical_lines.x)
-        params["y_max"] = numpy.fabs(self.horizontal_lines.y)
+        params["x_width"] = self.xwidth
+        params["y_width"] = self.ywidth
```
krzywon, line 293: The other `params` keys match their respective class dictionary keys. If this were true for `(x|y)width`, this could allow for much more straightforward (de)serialization of the slicer objects.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -250,14 +305,24 @@ def setParams(self, params):
         :param params: a dictionary containing name of slicer parameters and
             values the user assigned to the slicer.
         """
-        self.x = float(numpy.fabs(params["x_max"]))
-        self.y = float(numpy.fabs(params["y_max"]))
+        self.xwidth = float(numpy.fabs(params["x_width"]))
+        self.ywidth = float(numpy.fabs(params["y_width"]))
```
krzywon, line 309: Why the need take the absolute value of the serialized value(s)?

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -250,14 +305,24 @@ def setParams(self, params):
...
+        self.horizontal_lines.update(center=self.center,
+                                     width=self.xwidth, height=self.ywidth)
+        self.vertical_lines.update(center=self.center,
+                                   width=self.xwidth, height=self.ywidth)
+        # compute the new error and sum given values of params
         self._post_data()
+
+        #self.horizontal_lines.update(x=self.x, y=self.y)
+        #self.vertical_lines.update(x=self.x, y=self.y)
+        #self.post_data(nbins=None)
```
krzywon, line 325: Please remove any commented code that is no longer needed.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -477,18 +577,197 @@ def restore(self, ev):
...
+        self.move(x, y, None)
+        self.update()
+
+class HorizontalDoubleLine(BaseInteractor):
+    """
+    Draw 2 vertical lines that can move symmetrically in opposite directions in y and centered on
+    a point (PointInteractor). It also defines the left and right x positions of a box.
+    """
+    def __init__(self, base, axes, color='black', zorder=5, x=0.5, y=0.5,
+                 center_x=0.0, center_y=0.0):
```
krzywon, line 613: Nitpick: `HorizontalDoubleLine` and `VerticalDoubleLine` signatures are almost exactly the same. A meta class that encompasses this overlap would be useful to minimize code repetition.

.github/workflows/ci.yml
```diff
@@ -109,7 +109,7 @@ jobs:
 
       - name: Fetch sources for sibling projects
         run: |
-          git clone --depth=50 --branch=master https://github.com/SasView/sasdata.git ../sasdata
+          git clone --depth=50 --branch=SasData-Counterpart-to-Sasview-SlicerExtension_1344 https://github.com/SasView/sasdata.git ../sasdata
```
krzywon, line 112: For record-keeping purposes: This line needs to be reset to `master` once the associated sasdata PR is merged.

src/sas/qtgui/Plotting/Plotter2D.py
```diff
@@ -396,6 +422,47 @@ def setSlicer(self, slicer, reset=True):
...
+                    # First take care of this item, then we'll take care of its children
+                    if hasattr(item.child(itemIndex).data(), 'type_id'):
+                        if slicer_type_id in item.child(itemIndex).data().type_id:
+                            # Store this plot to be removed later. Removing now
+                            # will cause the next plot to be skipped
+                            tempPlotsToRemove.append(item.child(itemIndex))
+                # It looks like the slicers are children of items that do not have data of instance Data1D or Data2D.
+                # Now do the children (1 level deep as is done in GuiUtils.plotsFromModel). Note that the slicers always
+                # seem to be the first entry (index2 == 0)
+                for itemIndex2 in range(item.child(itemIndex).rowCount()):
```
krzywon, line 451: What happens if the child has children? How the existing data explorer system works, that shouldn't be possible, but this limits our ability to search through larger tree structures.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -18,41 +18,66 @@ def __init__(self, base, axes, item=None, color='black', zorder=3):
...
         # when reach qmax reset the graph
         self.qmax = max(self.data.xmax, self.data.xmin,
                         self.data.ymax, self.data.ymin)
+
+        # center of the box
+        # puts the center of box at the middle of the q-range (??)
+
+        self.center_x = 0.0002
+        self.center_y = 0.0003
+
```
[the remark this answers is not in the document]
krzywon: Agreed. Please update the code with comment on why the hard-coded values are used.

krzywon (commented):
I would suggest addressing the comments not noted as nitpicky. A couple of them could result in errors.

Side note - There are changes to sas_gen.py and associated unit tests in here. Is this a merge-related issue? I think @dehoni should take a look to be sure we don't remove necessary functionality.

src/sas/qtgui/Plotting/Plotter2D.py
jack-rooks: I did it this way because I would have to provide GuiUtils.deleteRedundantPlots a list of all of the plots to keep. I was unsure of a way to reliably include all other plots to keep and exclude only the old slicers. If someone has a good idea of what all of the plots that could possibly be a child of the 2D data then they could make use of GuiUtils.deleteRedundantPlots and delete the addition here.

butlerpd:
> Side note - There are changes to sas_gen.py and associated unit tests in here. Is this a merge-related issue? I think @dehoni should take a look to be sure we don't remove necessary functionality.

Indeed these were spurious commits that came in from a couple of poorly executed merging/reverting/munging 2 years ago. This was fixed by my last commit cd3e7fda097c69c43712ecf85f92c7ef4135c4f9 (see commit notes for more details). The files mentioned are no longer part of the changes in this PR

.github/workflows/ci.yml
butlerpd: The two associated PR should be merged together. But agreed this file should NOT be part of the merge. It will probably get cleaned up in the rebasing

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
butlerpd: good point. I suspect not necessary. It was probably initially necessary when the logic was for a slab always centered on 0,0 and using xmax instead of xwidth. The person who first changed that logic kept the fabs and it has been carried through sense but from the code I think xmax is always positive.

We can try to remove that and see if it breaks.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
butlerpd: Agreed @krzywon  - but that is part of a bigger slicer refactor (which I have aspirations for?)  so not for this PR.

BTW - these two interactors classes are also used by boxsum. And when I say "also used" I mean rewritten completely (copy pasted?). 

There are really only a small number of base element interactors (radial line, arc, center, horizontal and vertical lines).  That to say that I have a whole aspirational map drawn out. but that will be a lot of work.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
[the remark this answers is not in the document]
butlerpd: Humm! This seems way out of date. It was wrong and is defined properly as 
```python
        self.center_x = (self.data.xmin + self.data.xmax) /2
        self.center_y = (self.data.ymin + self.data.ymax) /2
```
But why does this not show up as `outdated` then? I am confused.

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
butlerpd: I think we should keep this as is. `params` is how the slicer parameters that you set in the slicer editor widget (`QTableView`) are passed to the slicer.  This was set up by @rozyczko years ago.  The magic (as far as I'm concerned even after looking at it for a while) is in `SlicerParameters.py` and `SlicerModel.py`.
as stated in `SlicerParameter`
```python
class SlicerParameters(QtWidgets.QDialog, Ui_SlicerParametersUI):
    """
    Interaction between the QTableView and the underlying model,
    passed from a slicer instance.
    """
Unless I completely missed the point being made?

```

src/sas/qtgui/Plotting/Plotter2D.py
krzywon: @jack-rooks, that seems like a good option.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -18,41 +18,66 @@ def __init__(self, base, axes, item=None, color='black', zorder=3):
         self.markers = []
         self.axes = axes
         self._item = item
-        # connecting artist
+        #connecting artist
```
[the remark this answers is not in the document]
butlerpd: done (reverted)

src/sas/qtgui/Plotting/Plotter2D.py
jack-rooks: The most recent commit does this in a slightly different way since it doesn't seem "info" is actually a child of the data (checks to see if it has any text for use in the data explorer headings, if none of the children do then it deletes it).

src/sas/qtgui/Plotting/Plotter2D.py
jack-rooks: Talked with @butlerpd and I won't have time to get this change in for the pull request. I've created #2838 to address this at some point.

dehoni:
> > Side note - There are changes to sas_gen.py and associated unit tests in here. Is this a merge-related issue? I think @dehoni should take a look to be sure we don't remove necessary functionality.
> 
> Indeed these were spurious commits that came in from a couple of poorly executed merging/reverting/munging 2 years ago. This was fixed by my last commit [cd3e7fd](https://github.com/SasView/sasview/commit/cd3e7fda097c69c43712ecf85f92c7ef4135c4f9) (see commit notes for more details). The files mentioned are no longer part of the changes in this PR

These files are part of main. It must have been part of an update of the branch. Anyway, the files are unrelated to this PR feature and only work on the general scattering calculator.

butlerpd:
All comments and concerns should now have been addressed. Also in testing a few new bugs were discovered and fixed and all user inputs now checked to avoid failures as documented in each of the commit messages.

One question: The tests now log warning when the user tries to provide an invalid ROI. However, Warnings do not trigger opening the GUI console. Only Error level logging does. However the warning is meant as an aid to the user. Can we have warnings bring up the console? or should these messages be labelled as error even though I would argue they technically do not rise to the error level at this point?

src/sas/qtgui/Plotting/Slicers/BaseInteractor.py
```diff
@@ -139,7 +141,8 @@ def onDrag(self, ev):
             self.clickx, self.clicky = ev.xdata, ev.ydata
             self.move(ev.xdata, ev.ydata, ev)
         else:
-            self.restore(ev)
+            logging.warning("the handle cannot be moved outside the data")
+            #self.restore(ev)
```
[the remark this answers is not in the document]
butlerpd: Three reasons.
  * The primary reason is that it is completely unnecessary. since the illegal move is caught BEFORE it actually moves, there is nothing to restore. The cascade of changes has not taken place and **all** the elements are therefore in a perfectly legal state.  NOTE: I removed quite a few of those instances, many of which were introduced early in this branch to "copy" the existing structure.
  *  Secondly, jumping from what is a totally legal state all the way back to when the mouse click was first engaged can be annoying depending on how much effort the user is taking in moving the mouse "just so" (and again ... did I mention "unnecessary"
  * It makes the code cleaner and easier to follow. Originally (back in the dawn of time when slicers were first introduced) all save and restores were done by the `BaseInteractor`. That is no longer the case. Besides making it unnecessarily complicated to find all the occurrences of `restore` there were edge cases when you ended up with 2 restores or a restore when it really should not have restored.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -435,59 +610,296 @@ def clear(self):
...
+        self.axes = axes
+        # Center coordinates
+        self.center_x = center_x
+        self.center_y = center_y
+        # Box half width and height and  horizontal and vertical limits
+        self.half_height = half_height
+        self.save_half_height = self.half_height
+        self.half_width = half_width
+        self.save_half_width = self.half_width
+        self.y1 = self.center_y + half_height
```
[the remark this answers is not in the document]
butlerpd: Humm... I think you mean that only the half_height is required? in principle, if moving horizontal lines neither the center nor the half_width should change. Regardless `restore` should restore everything that was saved in `save` which is what is done here.  That said I did look at this for this reason. It is however how these slicers were written back in the dawn of time and is exactly how `BoxSum` handles it as well (also since the dawn of time). Given all the surprising interactions encountered while working on this "small change" I opted not to go down that rabbit hole in this PR.  I think it would be worth placing a comment in the code for future work.

In terms of restoring, that is correct. The way this entire slicer module was designed (and I agree that it could definitely do with a refactor, but that is not what this PR is doing so we stay within the framework given) seems to be that restore function **only** changes the values back (restore) locally but that has to be followed with an update (which makes sure all the different actors are updated with those values, followed by a draw to actually render that on the screen for the user.  Remember that changing the horizontal line position (changing the height of the box) also mush change the length of the vertical lines.

src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
```diff
@@ -250,101 +346,158 @@ def setParams(self, params):
...
+          at that point.
+        """
+        isValid = True
+
+        if param_name =='half_width':
+            # Can't be negative for sure. Also, it should not be so small that
+            # there remains no points to average in the ROI. We leave this
+            # second check to manipulations.py
+            if param_value <= 0:
+                logging.warning("The box width is too small. Please adjust.")
```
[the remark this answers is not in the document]
butlerpd: Actually not true in this particular case since that is in the Validate function which only applies to text entries.  It also is never true I don't think for the "no points in ROI" which is captured in a try except block. However it is the case with all other illegal mouse drags.  I did find that annoying but not sure how to fix. I will need to think a bit more about this. I'm not sure that "unselecting" will not be more annoying to the user. Suspending the error till `isValid == True` maybe? but have to make sure I capture it everywhere it can be called....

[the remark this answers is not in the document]
butlerpd: OK ... This looks like it might work. See the commit message for explanation of `BaseInteractor.py` changes.  What do you think?

src/sas/qtgui/Plotting/Slicers/BaseInteractor.py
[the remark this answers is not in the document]
butlerpd: comment now removed

krzywon (approved):
Minor nitpick at this point: The graphical box interactors won't allow me to extend/move to the edge of the data anymore, but that is mostly a visual issue.

krzywon:
I just tested the latest changes (from code, not the installer) and the user experience is much better. The box sides are coerced to the data limits and no warnings.

Assuming the installers work, I think this can be merged.

butlerpd:
It still needs to have the ci.yml reverted. before merging. Will do after my meeting.

We probably should also say something in the docs. The coercion to the edge worked a bit better for me the old way but the big advantage here that I think trumps all minor differences in usage is the fact that now you can slide the initial box up and down (or sideways depending on which slicer it is) without shrinking the box.

Note I plan to update the docs in another PR. A lot of the changes will need documentation updates as well.

Issue #1826, opened 2021-04-24
Load and Process files for General Scattering Calculator


dehoni:
Explore ways to speed up or even allow to load larger (not abnormal) simulation files like this ".omf" file in plain text: 
[fe_40nm_saturation_omf.txt](https://github.com/SasView/sasview/files/7437889/fe_40nm_saturation_omf.txt)

The Direct Fourier Transform in the code is a further bottleneck and should be replaced by a parallelized version if possible.

PR #1930, opened 2021-10-22
File loader enhancement of the General Scattering Calculator

dehoni:
This PR improves the data load routine that is used to read omf, and pdb files in the general scattering calculator.

### DIFF
--- a/src/sas/sascalc/calculator/sas_gen.py
+++ b/src/sas/sascalc/calculator/sas_gen.py
@@ -369,9 +369,9 @@ class OMFReader(object):
         :return: x, y, z, sld_n, sld_mx, sld_my, sld_mz
         """
         desc = ""
-        mx = np.zeros(0)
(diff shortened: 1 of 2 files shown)


rozyczko:
> Sorry, the file was missing the magnetisation unit value. It is fixed and should work now.

Thanks! Can you please attach the file? I can't find its location anymore :/

dehoni:
The test file is attached to the ticket https://github.com/SasView/sasview/issues/1826 .

rozyczko (approved):
This indeed helps A LOT with loading large OMF files. As in, order of magnitude.

The code, as mentioned, looks good.

dehoni:
The error message "We accept only Text format OMF file" was rather cryptic. The file is a text but the physical units are missing. The commit returns now a more sensible error.

wpotrzebowski:
ready for testing on Win

Issue #1839, opened 2021-05-07
Tidy up how FitPage parameter table reflects array distributions


smk78:
When polydispersity is enabled in a model, selected parameters can take on distributions of values. Array distributions (basically a simple file of weight values as a function of the chosen parameter, such as a radius) allow the User greater flexibility over the form of the distribution to be applied than simply using the built-in dispersity functions (Gaussian, Lognormal, etc), without the complexity of custom-defined distribution functions which require some coding.

The caveat is that, unlike the built-in or custom distributions, array distributions are utilised 'as supplied'; sasmodels does not shift or scale the distribution, meaning it cannot be fit/optimised. This is stated in the help docs. However, this is _not_ reflected in the FitPage.

The unobservant User could easily miss that the parameter using an array distribution is not changing and, worse, believe that that parameter displayed actually has some meaning. It does not! This needs rectifying.

Discussions with @butlerpd and @pkienzle have suggested two options:

- the quick fix would be to simply clear any values and 'grey out' the parameter that is using an array distribution.

- a better fix would be to return some meaningful value characterising the chosen array distribution to the parameter table (though it would not be a fittable parameter for the reason above); the _mean_ of the distribution would be one such parameter.

On this latter point @pkienzle comments:
![image](https://user-images.githubusercontent.com/10938679/117441917-f7c1e280-af2d-11eb-8e7a-684a812e492a.png)

The help docs for array distributions would also need updating to reflect any change made in the process of dealing with this issue.

smk78:
@pkienzle also comments:
![image](https://user-images.githubusercontent.com/10938679/117442582-e200ed00-af2e-11eb-834a-5702c0a62498.png)

smk78:
Also see https://github.com/SasView/sasview/issues/1213

smk78:
Doc updates are complete, but the wider issue mentioned remains.

PR #1938, opened 2021-10-28
Doc Update: Polydispersity

smk78:
This PR attempts to address some of the documentation deficiencies in SasView raised in https://github.com/SasView/sasview/issues/1839. There is a wider reporting issue that would require coding changes which is not addressed here.

Work so far:

- actually mention polydispersity under Single Fit in fitting_help.rst (Yup, we really didn't tell Users they could have polydisperse parameters!)
- add some images
- mention GPU


butlerpd (approved):
OMG!!! this `rst` is **HUUUGE** 😄 with a number of things that need updating by the looks of it from just a quick skim.  It is hard to keep these uptodate sometimes. I am wondering if we need to think about breaking this up somehow into several sub pages, both for the benefit of the user and for maintenance?

That said the PD addition seems correct and complete so I approve the PR for merging

butlerpd:
I note however that this has conflicting changes in fitting_help.rst blocking an actual merge.  This will have to be resolved before we can merge it.

src/sas/qtgui/Perspectives/Fitting/media/fitting_help.rst
```diff
@@ -650,6 +650,41 @@ The DREAM optimiser is the most sophisticated, but may not necessarily be the be
...
+(the *PD[ratio]*) to be adjusted. If necessary the 'step size' (*Npts*) and
+'range' (*Nsigs*) of the function can also be adjusted.
+
+.. image:: pd_tab.png
+
+For more information, see the descriptions of :ref:`polydispersityhelp` . In
+particular, pay attention to the Suggested Applications and Usage Notes therein.
+Note that SasView defaults to Gaussian distributions, but these will not always
+be the best choice. Also, the definition of *PD[ratio]* varies depending on
+the chosen distribution!
```
pkienzle, line 671: PD for orientation is absolute; PD for volume parameters are relative to the current value. It doesn't depend on the chosen distribution, though the meaning of center and width does (first and second moments was a good analytic choice, but it sure was confusing for rectangular distributions).

src/sas/qtgui/Perspectives/Fitting/media/fitting_help.rst
```diff
@@ -650,6 +650,41 @@ The DREAM optimiser is the most sophisticated, but may not necessarily be the be
...
+.. image:: enable_pd.png
+
+Clicking on the *Polydispersity* tab then provides access to these polydisperse
+parameters and allows the type (i.e. the *function* to be used) and 'width'
+(the *PD[ratio]*) to be adjusted. If necessary the 'step size' (*Npts*) and
+'range' (*Nsigs*) of the function can also be adjusted.
+
+.. image:: pd_tab.png
+
+For more information, see the descriptions of :ref:`polydispersityhelp` . In
```
pkienzle, line 667: Neither here nor the polydispersity docs mention that the distribution is a number density distribution, not a particle volume distribution. This is discussed in [fitting_sq.rst](https://raw.githubusercontent.com/SasView/sasmodels/master/doc/guide/fitting_sq.rst) and can be linked as `PStheory`.

src/sas/qtgui/Perspectives/Fitting/media/fitting_help.rst
smk78: Thanks @pkienzle . Have changed it to read now:

image

src/sas/qtgui/Perspectives/Fitting/media/fitting_help.rst
smk78: Actually, it does say in polydispersity.rst:

image

I've now also added a note to the same effect in fitting_help.rst.

pkienzle: Probably better to phrase this as `These are all implemented as number densities. The resulting scattering is the number average from the different-sized particles.`, but that would be a sasmodels PR.

Within the current text you may want to mention the ability to reparameterize the particle so that you can control polydispersity according to within-particle constraints (link to plugin.rst). For example, if the process which generates the particles constrains the aspect ratio but not the volume, or if it is a squishy particle that preserves volume but allows a range of aspect ratios for each volume. You may need to define your own distribution function to fully describe the model (link to polydispersity.rst user defined functions).

pkienzle: I have fielded several questions from users about the details of the polydispersity calculation, which is why I suggest linking to PStheory in fitting_sq.rst.

This file also discusses the relationship between scale, volume fraction, effective radius, effective radius mode and structure factor mode, which is useful information for the discussion of interaction and mixture models. I think there was an attempt to link to it in on line 105 of this file:
```
Also see :ref:`Interaction_and_Mixture_Models` for further information.
```
but it is self-referential!  Please change it to ref `Interaction_Models`, which is the label used in fitting_sq.rst.

pkienzle: Still tweaking the "number-average" phrase. The distributions are implemented as simple distributions, but they are being used to compute the number average. Something like:
```
These distributions define the number density of particles. The resulting scattering is the number average over the distribution.
```
Probably not worth changing.

smk78: Ok, several changes to polydispersity.rst:

- Have added a link to the end of the opening sentence (that resolves to PStheory): 
image
  
- Have changed the 'number distribution' sentence to read:
image

- In array distribution have changed the existing text after the example array to a Note (the existing text had grammatical & typo errors anyhow) and added a new sentence to the bottom:
image

smk78: And in fitting_help.rst:

- Have changed the link from :ref:`Interaction_and_Mixture_Models` to :ref:`Interaction_Models` as requested.

- Have also added a link to :ref:`PStheory`: from the Polydisperse Parameters sub-section:
image

- Have revised the wording of the note regarding number density to echo that now in polydispersity.rst:
image

- Added a new note:
image

- And have added a section flagging that you can reparameterize models:
image

pkienzle (approved):

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
Hmm, Jenkins is not happy about this: 
`/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/docs/sphinx-docs/source-temp/user/qtgui/Perspectives/Fitting/fitting_help.rst:709: WARNING: undefined label: reparameterized_models (if the link has no caption the label must precede a section header`

smk78:
@wpotrzebowski That's odd. The link target was added to Sasmodels /doc/guide/plugin.rst in this commit https://github.com/SasView/sasmodels/pull/484/commits/3965bee7b591adf97fda00fd61020faacfea01ad which has been merged. Is it possible Jenkins was trying to build against an outdated Sasmodels?

smk78:
ready for testing on Windows

smk78:
I've just built this branch locally against the current Sasmodels master and the link resolves fine. I think it's a Jenkins issue.

wpotrzebowski:
ready for testing on OSX

PR #1947, opened 2021-11-01
Use version parsing package for MPL version check

krzywon:
This uses a version parsing package bundled with setuptools (no extra dependencies) to help fix the 'nonposy' issue in MPL version > 3.3 that is used for the Github actions build.

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -7,6 +7,7 @@ import sys
 import matplotlib as mpl
 import numpy as np
 from matplotlib.font_manager import FontProperties
+from packaging import version
 
 from sas.qtgui.Plotting.PlotterData import Data1D
 from sas.qtgui.Plotting.PlotterBase import PlotterBase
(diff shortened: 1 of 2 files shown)


krzywon:
I realize `main` is already working in this regard, but this will help future proof the change, just in case this temporary solution isn't so temporary.

`version.parse("10.0.0") > version.parse("9.9.9")`

*but* 

`"10.0.0" < "9.9.9"`

rozyczko (commented):
Yes, this is a more flexible and likely more error-proof solution.
However, I don't see any other use of this module (`packaging`) and it isn't specified in the yml, at least not on windows.
I think this is a non-standard module so it would need to be added, if not already pulled by some other dependency.
Let me check the installer then.

rozyczko:
ready for testing on Win

krzywon:
The `packaging` module comes bundled with `setuptools` so no extra dependencies should be required.

rozyczko:
Interesting. Installing `setuptools` with conda does indeed add `packaging` even though the list of modules listed for download/installation doesn't include it...
In any case, we seem to have this package installed.

rozyczko (approved):

Issue #1948, opened 2021-11-02
Check that using current 'master' bumps version does not break sasview


bmaranville:
A new release of bumps is planned soon (less than 2 weeks).  The requirements for sasview seem to set a lower bound on the bumps version, but not an upper bound.

There are some notes on the bumps README.rst about what is changing.

krzywon:
I just pulled and installed bumps master and ran SasView main with it as part of my environment. I haven't looked at the bumps changes yet, but we should remain on [bumps v0.8.0](https://github.com/bumps/bumps/releases/tag/v0.8.0) for now.

Workflow:
- Load `apoferritin.txt`
- Send to fitting
- Select `Sphere` -> `sphere` for the model
- Check `scale`, `background`, and `radius`
- Fit using `Levenberg-Marquardt`: **No errors**
- Switch to `Nelder-Mead` and Fit: **Error [1]**
- Switch to `Differential Evolution` and Fit: **Error [1]**
- Switch to `DREAM` (unable to fit): **Error [2]**
- Switch to `Quasi-Newton` and Fit: **No errors**
- Switch to 'MPFit': **Error [3]**

[1] Results window is blank and the following error is thrown
> ERROR: Traceback (most recent call last):
  File "...\sasview\src\sas\qtgui\MainWindow\GuiManager.py", line 1007, in showFitResults
    self.results_panel.onPlotResults(output_data, optimizer=self.perspective().optimizer)
  File "...\sasview\src\sas\qtgui\Utilities\ResultPanel.py", line 72, in onPlotResults
    self.convergenceView.update(best, pop)
  File "...\qt5_win\lib\site-packages\bumps-0.8.0-py3.6.egg\bumps\gui\convergence_view.py", line 63, in update
    self.plot()
  File "...\qt5_win\lib\site-packages\bumps-0.8.0-py3.6.egg\bumps\gui\convergence_view.py", line 60, in plot
    pylab.draw()
  File "...\sasview\src\sas\qtgui\Utilities\PlotView.py", line 68, in __exit__
    Gcf._activeQue = [f for f in Gcf._activeQue if f is not self.fm]
AttributeError: type object 'Gcf' has no attribute '_activeQue'

[2] Unable to fit with DREAM engine and the following error is thrown when the algorithm is selected
> ERROR: Traceback (most recent call last):
  File "...\sasview\src\sas\qtgui\Perspectives\Fitting\FittingOptions.py", line 149, in onAlgorithmChange
[...]

PR #1951, opened 2021-11-03
Fix for removed _activeQue in newer Matplotlib

bmaranville:
handle newer versions (>3.3.1) of matplotlib, where _activeQue is not defined

### DIFF
--- a/src/sas/qtgui/Utilities/PlotView.py
+++ b/src/sas/qtgui/Utilities/PlotView.py
@@ -204,7 +204,7 @@ class PlotView(QtWidgets.QWidget, _PlotViewShared):
             self.title = kw['title']
 
         # Instantiate a figure object that will contain our plots.
-        figure = Figure(figsize=(1,1), dpi=72)
+        figure = Figure(figsize=(10,10), dpi=72)
 
         # Initialize the figure canvas, mapping the figure object to the plot
         # engine backend.
--- a/src/sas/qtgui/Utilities/PlotView.py
+++ b/src/sas/qtgui/Utilities/PlotView.py
@@ -65,12 +65,14 @@ class EmbeddedPylab(object):
     def __exit__(self, *args, **kw):
         # delay loading pylab until matplotlib.use() is called
         from matplotlib._pylab_helpers import Gcf
-        Gcf._activeQue = [f for f in Gcf._activeQue if f is not self.fm]
-        try:
-            del Gcf.figs[-1]
-        except KeyError:
-            pass
-
+        if hasattr(Gcf, '_activeQue'):  # CRUFT: MPL < 3.3.1
+            Gcf._activeQue = [f for f in Gcf._activeQue if f is not self.fm]
+            try:
+                del Gcf.figs[-1]
+            except KeyError:
+                pass
+        else:
+            Gcf.figs.pop(self.fm.num, None)
 
 class _PlotViewShared(object):
     title = 'Plot'


bmaranville:
This change is copied from https://github.com/bumps/bumps/blob/master/bumps/gui/util.py#L71

krzywon:
This corrects the original error, but I'm seeing a new error when using the DREAM engine. The `Convergence`, `Correlation`, and `Parameter Trace` views are shown in the `Results panel`, but the `Uncertainty` results are throwing the error.

`matplotlib` is defaulting to a size of 1.0" x 1.0" for new figures in [bumps/varplot.py line 43](https://github.com/bumps/bumps/blob/4a33419dbe8e6d93f0b554fd10abd6c93e142e5e/bumps/dream/varplot.py#L43). The sum of the margins used for the DREAM results are larger than 1.0", so negative sizes are calculated. No size is specified when the figure is generated but the default should be [6.4" x 4.8"](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.figure.html) according to the mpl documentation.

I'm can't say if this is a sasview or bumps issue at this point.

> ERROR: Traceback (most recent call last):
  File "...\sasview\src\sas\qtgui\MainWindow\GuiManager.py", line 1017, in showFitResults
    self.results_panel.onPlotResults(output_data, optimizer=self.perspective().optimizer)
  File "...\sasview\src\sas\qtgui\Utilities\ResultPanel.py", line 84, in onPlotResults
    self.uncertaintyView.update((result.uncertainty_state, stats))
  File "...\bumps\gui\uncertainty_view.py", line 24, in update
    self.plot()
  File "...\bumps\gui\uncertainty_view.py", line 19, in plot
    dream_varplot.plot_vars(history.draw(), stats)
  File "...\bumps\dream\varplot.py", line 101, in plot_vars
    fig = _make_var_axes(n)
  File "...\bumps\dream\varplot.py", line 79, in _make_var_axes
    ax = fig.add_axes(dims)
  File "...\matplotlib\figure.py", line 654, in add_axes
    a = projection_class(self, rect, **pkw)
  File "...\matplotlib\_api\deprecation.py", line 471, in wrapper
    return func(*args, **kwargs)
  File "\matplotlib\axes\_base.py", line 606, in __init__
    raise ValueError('Width and height specified must be non-negative')
ValueError: Width and height specified must be non-negative

bmaranville:
Certainly those files, bumps/dream/varplot.py and bumps/gui/uncertainty_view.py have changed since the last release, so it merits deeper investigation.

bmaranville:
In bumps, the figure is created by bumps/dream/views.py like this:
```python
figure(figsize=varplot.var_plot_size(len(all_vstats)))
```
and then used in bumps/dream/varplot.py by calling gcf().  The size coming back from var_plot_size() is a multiple of (3" by 2")
It seems that in sasview, _make_var_axes is being called without first creating a figure of a size set by var_plot_size.

krzywon:
Sasview is creating an `UncertaintyView` instance (bumps/gui/uncertainty_view.py) and then calls the `update()` method which generates the plot using `plot_vars()` in bumps/dream/varplot.py. plot_vars calls the `_make_var_axes()` method you mentioned. Should we be calling `plot_all()` from bumps/dream/views.py instead?

Either way, I think the `UncertaintyView` plotting routine needs some attention.

pkienzle:
You can't be use plot_all() since that won't put the plots on the figure tabs.

I can't imagine that SasView could be using UncertaintyView directly: It imports plot_view which imports wx. … or maybe you are talking about the wx version?

bmaranville:
These discussions are getting somewhat afield of the original pull request - maybe we need another issue or PR in which to discuss them?

krzywon:
I am (unfortunately?) talking about the Qt version.

https://github.com/SasView/sasview/blob/9c7a451ff8b5d789fc01f7f7711c4900ec59ca7a/src/sas/qtgui/Utilities/ResultPanel.py#L40

@bmaranville - agreed. Further discussion should probably continue in the original ticket #1948, but might need a separate ticket.

pkienzle:
Okay, I see the horrible hack: https://github.com/SasView/sasview/blob/9c7a451ff8b5d789fc01f7f7711c4900ec59ca7a/src/sas/qtgui/Utilities/ResultPanel.py#L55

We do realize that we can modify bumps as a sister DANSE project do we not?

pkienzle:
What's going on with varplot is that we changed from using a fixed size figure with shrunken variable histograms to fixed sized histograms with an expanded figure, at least in the command line version. As of three years ago I adapted this to work with the wx GUI. The problem is that the width and height of an empty figure is being returned as 1" x 1" on my mac, and this is not big enough. This is the size given to the figure in the PlotView object when it is first created.

Changing the intitial size to 10x10 lets things run and looks okay: https://github.com/SasView/sasview/blob/9c7a451ff8b5d789fc01f7f7711c4900ec59ca7a/src/sas/qtgui/Utilities/PlotView.py#L29

This is not a proper fix. Instead we need to root around in matplotlib to figure out why qt backend is not reporting the correct figure size.

We should also modify varplot to not crash when the figure size is too small. We could just make the subplots have zero space in between when more than 20% of the figure space is gaps and margins, but it might be better to throw scroll bars on the canvas and make the plot big enough to show the variable histograms clearly. Code is here: https://github.com/bumps/bumps/blob/b247986cb02fcc6c7326ce53c55f508528832072/bumps/dream/varplot.py#L49-L61

pkienzle:
> These discussions are getting somewhat afield of the original pull request - maybe we need another issue or PR in which to discuss them?

I suggest doing the `figsize=(10,10)` hack so that the plots run but leave the real fix in another ticket.

krzywon:
I saw the same yesterday, using `(3,2)` instead of `(10,10)`, but had to run off before I could comment. For now, this should be the way we go. I'll push the change here and create a test build.

> I suggest doing the `figsize=(10,10)` hack so that the plots run but leave the real fix in another ticket.

krzywon:
ready for testing on OSX

krzywon:
ready for testing on Win

krzywon (approved):
This fixes the intended problem and all builds are passing.

I think this fixes everything in #1948.

wpotrzebowski:
ready for testing on OSX

PR #1958, opened 2021-11-04
Revert "1869 easily accessible dependencies"

wpotrzebowski:
Reverts SasView/sasview#1903
It seems that it breaks the build...


wpotrzebowski:
ready for testing on OSX

PR #1961, opened 2021-11-04
Second attempt for easily accessible dependencies

wpotrzebowski:
Follow up on #1903 that broke build and was reverted


wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
@Caddy-Jones: can you please take a look on these warnings apparently causing Jenkins build failure. Note that you need to checkout new branch becuse of (re)reverting...
```
/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer:4: WARNING: Unexpected section title.

Methods
-------
/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer.format_unattainable_modules_list:10: WARNING: Unexpected section title.

Parameters
----------
/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer.format_unattainable_modules_list:17: WARNING: Unexpected section title.

Returns
-------
/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer.get_imported_modules:8: WARNING: Unexpected section title.

Returns
-------
/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer.log_imported_modules:7: WARNING: Unexpected section title.

Returns
-------
/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer.log_installed_modules:7: WARNING: Unexpected section title.

Returns
-------
/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer.remove_duplicate_modules:10: WARNING: Unexpected section title.

Parameters
----------
/Users/sasview/Jenkins/workspace/SasView-OSX10.13-ghprb/sasview/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer.remove_duplicate_modules:15: WARNING: Unexpected section title.
```

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+            #     pass
+            # else:
+            #     continue
+
+            # Modules version number could not be attained by any of the previous methods
+            unattainable_modules.append(mod)
+
+        # Clean up
+        module_versions_dict = self.remove_duplicate_modules(module_versions_dict)
+        unattainable_modules = self.format_unattainable_modules_list(module_versions_dict, unattainable_modules)
```
pkienzle, line 163: Don't know why you would want to list unknown packages separately. My suggestions above produce `package_name: "unknown"` or `package_name: "unknown: import exception"`, which seems like it will be more useful.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+        as the values. Only packages available through pip will be included.
+
+        Returns
+        -------
+        None
+        """
+        installed_modules = {'python': sys.version}
+
+        # Get python modules installed locally
+        installed_modules_json = json.loads(subprocess.check_output("pip list -l --format=json"))
```
pkienzle, line 48: Fails for installed application.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+        respective version numbers as the values. There may be some packages whose version number is unattainable.
+
+        Returns
+        -------
+        None
+        """
+        imported_modules = self.get_imported_modules()
+
+        logging.info(f"Imported modules\n"
+                     f"{imported_modules}")
```
pkienzle, line 69: It would be polite to sort and format these.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        installed_modules = {'python': sys.version}
+
+        # Get python modules installed locally
+        installed_modules_json = json.loads(subprocess.check_output("pip list -l --format=json"))
+        for mod in installed_modules_json:
+            installed_modules[mod['name']] = mod['version']
+
+        logging.info(f"Installed modules\n"
+                     f"{installed_modules}")
```
pkienzle, line 53: It would be polite to sort these, and maybe format them a little nicer as:
```python
logging.info("Installed modules:\n  " + "\n".join(f"{k}: {v}" for k, v in sorted(installed_modules.items())))
```
If log level is warning or error then this message won't show. This will cause user confusion if for some reason we lower the log level but leave a menu item active that appears to do nothing. Do we have a mechanism other than logging for dumping things to the SasView console?

pkienzle (changes requested):

Caddy-Jones:
Hi @wpotrzebowski, The errors appear to come from the titles I used in the docstrings, causing “WARNING: Unexpected section title.” I used the NumPy/SciPy Docstring format for the docstings in the PackageGatherer class. Is there a different docstring standard used by SasView?

src/sas/qtgui/MainWindow/PackageGatherer.py
Caddy-Jones: I see what you’re saying, and I get that it may be confusing to users if the logging level is set to say error.

But isn’t the primary use of logging to debug, and this tool is primarily a debugging tool, so I would argue that using logging is the most consistent way to show this information. However it may be useful to add something that will always show in the console, like “modules successfully logged”

Also maybe set the logging level to debug in the __init__ section of the class, since this is a debugging tool?

src/sas/qtgui/MainWindow/PackageGatherer.py
Caddy-Jones: Added

pkienzle:
> Hi @wpotrzebowski, The errors appear to come from the titles I used in the docstrings, causing “WARNING: Unexpected section title.” I used the NumPy/SciPy Docstring format for the docstings in the PackageGatherer class. Is there a different docstring standard used by SasView?

Coding standards for docstrings are [here](https://github.com/SasView/sasview/wiki/DevNotes_DevGuide_CodingRules#coding-standards). We don't suggest (or follow) any particular style and there is wide variation across the project. So long as it works with sphinx it's okay.

Instead of underlined `Section heading` I've used `**Section heading**` to get a similar effect.

src/sas/qtgui/MainWindow/PackageGatherer.py
pkienzle: Okay, not worth worrying about log level.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+
+        Returns
+        -------
+        module_versions_dict : dict
+            A dictionary with the module names as the key, with their respective version numbers as the value.
+        """
+        module_versions_dict = {'python': sys.version, 'SasView': sas.sasview.__version__}
+        unattainable_modules = []
+        # Generate a list of standard modules by looking at the local python library
+        standard_lib = [path.stem.split('.')[0] for path in pathlib.Path(pathlib.__file__).parent.absolute().glob('*')]
```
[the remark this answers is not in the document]
Caddy-Jones: Hi, so I've included your standard_path code in a try except block, with a hard coded list of bultin modules in the event that pathlib.__file__ causes an error. 

Could someone check if pathlib.__file__ does indedd fail in the shipped application, as I would not know how to check.

If it does fail, then the list of builtin modules will probably have to be checked each time the python version SasView uses is updated.

May be worth having this list in an external file, making the job of updating the list easier?

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+            # Modules that require specific methods to get the version number
+            if "PyQt5" in mod:
+                try:
+                    from PyQt5.QtCore import QT_VERSION_STR, PYQT_VERSION_STR
+                except:
+                    unattainable_modules.append(["Qt", "PyQt"])
+                else:
+                    module_versions_dict["Qt"] = QT_VERSION_STR
+                    module_versions_dict["PyQt"] = PYQT_VERSION_STR
+                continue
```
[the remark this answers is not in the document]
Caddy-Jones: To my knoledge PySlide is not used anymore

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+
+            # Unreliable, so last option
+            try:
+                module_versions_dict[mod] = pkg_resources.get_distribution(mod).version
+            except:
+                # Modules that cannot be found by pkg_resources
+                pass
+            else:
+                continue
```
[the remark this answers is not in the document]
Caddy-Jones: Are you saying that pkg_resources fails in the distributed app?

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+            #     logging.error(f"{x} when attempting to get the version of {mod} through pip")
+            #     pass
+            # else:
+            #     continue
+
+            # Modules version number could not be attained by any of the previous methods
+            unattainable_modules.append(mod)
+
+        # Clean up
+        module_versions_dict = self.remove_duplicate_modules(module_versions_dict)
```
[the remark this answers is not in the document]
Caddy-Jones: True after some refactoring, but since the code is there I see no harm in leaving it

src/sas/qtgui/MainWindow/PackageGatherer.py
Caddy-Jones: Pip fails?

Caddy-Jones:
Is there anyway to add a new branch to this pull request for the code changes? If not i can set up a new PR

wpotrzebowski:
@Caddy-Jones I think the easiest is to checkout this branch `revert-1958-revert-1903-1869_WIP_easily_accessible_dependencie` and add code there.

src/sas/qtgui/MainWindow/PackageGatherer.py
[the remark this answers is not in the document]
pkienzle: I haven't checked recently, but I believe pyinstaller creates a zip file with the package tree and doesn't use site packages and the associated egg info for the installed application. I don't know for sure that pkg_resource is using egg info but it seems likely. Without the egg info, the pkg_resources may trigger an exception or return None so that the `.version` reference triggers an exception. This will cause a fall-through to the "package version not found" condition.

So not a crash just a useless result. Adding two menu items to the user interface that do not work in the shipped application does not feel like a good design choice, even if it doesn't cause a crash.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,281 @@
...
+                                                     f"version using .version"
+                    pass
+
+                continue
+
+            # Unreliable, so last option
+            try:
+                package_versions_dict[package_name] = pkg_resources.get_distribution(package_name).version
+
+            except:
```
pkienzle, line 180: Don't use bare `except` statements. These capture interrupt and shutdown signals. At a minimum use `except Exception:`.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,281 @@
...
+        standard_lib.extend(sys.builtin_module_names)
+        standard_lib.append("sas")
+
+        for module_name in sys.modules.keys():
+
+            package_name = module_name.split('.')[0]
+
+            # A built in python module or a local file, which have no version, only the python/SasView version
+            if module_name in standard_lib or module_name in package_versions_dict:
+                continue
```
pkienzle, line 120: I don't know that you need to check for `module_name` separately from `package_name` in you dict. If `module_name` does not contain dot then `package_name` will be `module_name`. Conversely, if `module_name` does contain dot then it won't match anything because you never enter any dotted names in the dict.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,281 @@
...
+            if hasattr(package, '__version__'):
+                # Module has __version__ attribute
+                try:
+                    package_versions_dict[package_name] = package.__version__
+                except Exception as x:
+                    # Unable to access module
+                    err_version_dict[package_name] = f"Unknown: {x} when attempting to access {package_name} " \
+                                                     f"version using .__version__"
+                    pass
+                continue
```
pkienzle, line 162: This code is testing that `package.__version__` exists before referencing it so it will not throw an exception.

Also, `Exception as x` is unusual. Here's what is used in the standard library:
```sh
conda/envs/py38/lib/python3.8: $ grep "except .* as [a-zA-Z_.]* *:" *.py | sed -e"s/.* as //" | sort | uniq -c | sort -r
  65 e:
  46 err:
  25 exc:
  18 msg:
  12 why:
  10 ex:
   6 val:
   5 error:
   4 f:
   3 value:
   2 x:
   2 var:
   1 v:
   1 stopinst:
   1 nag:
   1 _err:
   1 _:
```

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,247 @@
...
+                module_versions_dict[mod] = __import__(mod).version
+            except AttributeError:
+                # Module has no .version attribute
+                pass
+            except Exception as x:
+                # Unable to access module
+                logging.error(f"{x} when attempting to access {mod} version using .version")
+                pass
+            else:
+                continue
```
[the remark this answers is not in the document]
pkienzle: Same comment as above re: exception handling.

And to repeat, this will probably not do what you want.

Example
```
>>> h5py.version
<module 'h5py.version' from '/Users/pkienzle/conda/envs/py38/lib/python3.8/site-packages/h5py/version.py'>
```
Though in this case `h5py.__version__` exists so you won't get here.

If you insist on keeping this clause at least check `if isinstance(package.version, str)`.

Caddy-Jones:
Hi @pkienzle . 

Thanks for your feedback, the reason i included the .version check is because some modules, do not have a __version__ attribute, and instead .version returns the version number. But i will include a check to see if the version number is a string or a float.

src/sas/qtgui/MainWindow/PackageGatherer.py
Caddy-Jones: True

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,278 @@
...
+                try:
+                    package_versions_dict[package_name] = package.__version__
+                except Exception as e:
+                    # Unable to access module
+                    err_version_dict[package_name] = f"Unknown: {e} when attempting to access {package_name} " \
+                                                     f"version using .__version__"
+                    pass
+                continue
+
+            if hasattr(package, '.version'):
```
pkienzle, line 161: no '.', just 'version'

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,278 @@
...
+                # Module has .version attribute
+                try:
+                    if isinstance(package.version, str) or isinstance(package.version, float):
```
pkienzle, line 164: 'float' is very unlikely. Do you have a counterexample? A tuple of integers and strings is much more likely.

src/sas/qtgui/MainWindow/PackageGatherer.py
```diff
@@ -0,0 +1,278 @@
...
+                        package_versions_dict[package_name] = package.version
```
pkienzle, line 165: If the tests fail then the code falls through to "continue", without listing version for package as "unknown". This will happen for known occurrences of package.version being either a module or a function.

You are presumably coding for conditions that don't exist, otherwise your tests would have resolved these issues before committing. I suggestion removing everything after `package.__version__`, and only add it back in if you know that you need it for an existing sasview dependencies. Since python packages seem to have settled on `__ version__` as the place to store version numbers any new dependencies will probably just work.

You still need a fall-through in case `__version__` isn't found in some future dependency. In those cases just tag the version as unknown. Since this is only for debugging you don't need to go to extraordinary lengths to get versions for every package.

Caddy-Jones: Hi @pkienzle 

The: sip, qt5reactor and webencodings packages' version number can only be accessed through pkg_resources, so I think it's worthwhile keeping that code block in, instead of writing specific code for those 3 modules.

I have commented out the .version code block. When I was originally writing the code, there were some modules where .version was required to get their version number. However I cannot remember which modules that was for. Upon testing it seems that not a single SasView imported module triggers this block, so it is redundant. But I can leave the code in just in case as a comment

wpotrzebowski:
@wpotrzebowski will take a look on it

wpotrzebowski:
ready for testing on OSX

wpotrzebowski:
@Caddy-Jones you may consider fixing this one before we merge:
/src/sas/qtgui/MainWindow/PackageGatherer.py:docstring of sas.qtgui.MainWindow.PackageGatherer.PackageGatherer.remove_duplicate_modules:15: WARNING: Field list ends without a blank line; unexpected unindent

Caddy-Jones:
Hi @wpotrzebowski , is there a way to test that the docstring is formated correctly?

wpotrzebowski:
> Hi @wpotrzebowski , is there a way to test that the docstring is formated correctly?

@Caddy-Jones I've just tested it by running docs build locally and looking for warnings (you can '-w' oprtion to build_sphinx.py to output warnings to file).

Caddy-Jones:
Hi @wpotrzebowski , where am i supposed to run build_sphinx.py from, i keep getting errors with the script unable to find sasmodels.

smk78:
@Caddy-Jones you should be able to do python setup.py build docs from the sasview top level. Otherwise, look in \docs\sphinx-docs. But you have to build the sasmodels docs first. Go to sasmodels\doc and make clean html.

Caddy-Jones:
Hi guys, I just cannot recreate the results @wpotrzebowski  has gotten, can you give me the command line input you used please

smk78:
@Caddy-Jones I've pulled your branch and built the docs locally and I'm seeing the exact same doc error that @wpotrzebowski reports. I'm running Sphinx 1.8.4 (according to pip list). But nothing I've tried so far has fixed the error! :-(

Caddy-Jones:
Apologies but I can’t seem to generate the sphinx documentation, anything I try gives me an error. Could somebody give me a step by step explanation of how to generate the documents, or point me to the location of such a thing. For example trying to do any make command gives me 
`'make' is not recognized as an internal or external command,operable program or batch file.`
Trying to run the sphinx-build.py script gives me 
```
=== Sasmodels Docs ===
  File "build_sphinx.py", line 294, in <module>
    rebuild()
  File "build_sphinx.py", line 280, in rebuild
    retrieve_user_docs()
  File "build_sphinx.py", line 134, in retrieve_user_docs
    copy_tree(SASMODELS_MODEL_SOURCE, SASMODELS_MODEL_TARGET)
  File "C:\Code_base\Anaconda3\envs\qt5_win\lib\distutils\dir_util.py", line 124, in copy_tree
    "cannot copy tree '%s': not a directory" % src)
distutils.errors.DistutilsFileError: cannot copy tree 'C:\Local_repo\sasview\docs\sphinx-docs\..\..\..\sasmodels\doc\model': not a directory
```
setup.py build docs Isn’t an option for sasmodels, and doesn’t give me the results wojeck had for sasview, instead i get an error:
```
copying src\sas\sasview\test\upcoming_formats\1000A_sphere_sm.xml -> build\lib\sas\sasview\test\upcoming_formats
running docs
========= check for sasmodels at C:\Local_repo\sasmodels\doc ============
============= Building sasmodels model documentation ===============
error: [WinError 2] The system cannot find the file specified
```

smk78:
@Caddy-Jones : Go to your sasmodels repo. cd doc. make clean html. If make gives you an error then presumably you don't have it installed? Once the sasmodels docs are built, cd to your sasview branch and python setup.py build docs. The sasview doc build copies the built sasmodels docs to the sasview\docs\sphinx-docs\build folder. If all has worked you should then be able to cd \html and open index.html and navigate around the docs. If you're using an Anaconda environment, remember you have to activate the environment before doing any of this! :-)

Caddy-Jones:
Thanks @smk78 , turns out make wasn't installed and there was an issue with my conda environment which prevented me from installing it. I've fixed the issue with the docstring so there hopefully should be no further issues

smk78:
Yes, I've just pulled your branch and re-made the docs and they pass compilation now. Good!

wpotrzebowski:
ready for testing on OSX

PR #1962, opened 2021-11-05
Fix for loader modules not found error

### LINKED ISSUE #1879: Unable to find/load NXcanSAS reader using Development Builds
Greg Smith and Jan Ilavsky have both reported an issue with the NXcanSAS reader in MacOS. The release version works in Mojave, but not in BigSur and possibly not in Catalina.

Both had similar tracebacks to this, suggesting the `cansas_reader_HDF5.py` reader cannot be found.
> 13:27:23 - INFO: --- SasView session started ---
> 13:27:23 - INFO: Python: 3.6.6 | packaged by conda-forge | (default, Jul 26 2018, 09:55:02)
> [GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]
> 13:27:35 - ERROR: No module named 'sas.sascalc.dataloader.readers.cansas_reader_HDF5'No module named 'sas.sascalc.dataloader.readers.cansas_reader_HDF5'
> 13:27:35 - ERROR: The data file you selected could not be loaded.
> Make sure the content of your file is properly formatted.

krzywon:
This should now fix #1879, where certain data reader modules cannot be imported. The issue is related to a change in associations.py that swapped `exec` calls for `importlib`. Instead, the loader modules are now explicitly imported when associations.py is loaded.


krzywon:
ready for testing on Win

krzywon:
ready for testing on OSX

krzywon:
ready for testing on Win

krzywon:
ready for testing on Win

krzywon:
ready for testing on Win

krzywon:
ready for testing on Win

krzywon:
ready for testing on OSX

wpotrzebowski (approved):
I've tested functionality and it works fine. I've also looked at the code and it looks fine to me (but I am not that familiar with code for loaders). The only question I have is about removing `GENERIC_ASSOCIATIONS`. Isn't better (easier to maintain) to keep dictionary or is `get_generic_readers ` only place in the code where these are used?

krzywon:
> The only question I have is about removing `GENERIC_ASSOCIATIONS`.

`GENERIC_ASSOCIATIONS` has only ever been called from `associations.py` and it should never have been a dictionary. I replaced it with a list and updated the documentation to better explain what the methods are doing.

krzywon:
Thanks for the review, @pkienzle. I've made some updates based on your comments.

> I looked up how you use `get_generic_readers` and found it is called from exactly one place with no parameters, which begs the question of why you need this function at all rather than using GENERIC_READERS directly.

This is for a future addition I have in mind that would allow users to define which readers to fall back on.

pkienzle (approved):

llimeht:
@wpotrzebowski:  With this change, it might be possible to remove all the `hiddenimports` from the pyinstaller `.spec` files.

wpotrzebowski:
ready for testing on OSX

PR #1969, opened 2021-11-17
Gh actions artifacts

wpotrzebowski:
Addition to gh actions installer to store artifacts on https://github.com/SasView/sasview/releases
Drafts are avaialable only for logged in users (and I believe bellonging to SasView organization). Regular users will see releases and pre-releases as before. Artifacts will be stored for each branch and will be updated once new change is pushed to the code, which could keep the list of managable size. As discussed with @butlerpd after merging branch entry on the list should be removed (currently plan to do it manualy but can probably be also done through gh actions).

### DIFF
--- a/.github/workflows/installers.yml
+++ b/.github/workflows/installers.yml
@@ -170,7 +170,6 @@ jobs:
         if-no-files-found: error
 
     - name: Upload zipped offline app installer to GitHub releases (non-master branch)
-      if: github.event_name == 'push' && env.BRANCH_NAME != 'main'
       uses: ncipollo/release-action@v1
       with:
         draft: true
@@ -179,5 +178,3 @@ jobs:
         replacesArtifacts: true
         token: ${{ secrets.GITHUB_TOKEN }}
         artifacts: "SasView-Installer-*.zip"
-        name: ${{ env.BRANCH_NAME }}
-        tag: ${{ env.BRANCH_NAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/installers.yml
```diff
@@ -168,3 +168,25 @@ jobs:
...
+      if: github.event_name == 'push'
+      shell: bash
+      run: echo "BRANCH_NAME=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+
+    - name: Declare env variables on pull_request only
+      if: github.event_name == 'pull_request'
+      shell: bash
+      run: echo "BRANCH_NAME=$GITHUB_HEAD_REF" >> $GITHUB_ENV
+
+    - name: Upload zipped offline app installer to GitHub releases
```
rozyczko, line 182: `.exe ` nor `.dmg` are zipped (at least not implicitly... I think this is leftover from copying the easyD yml :)

.github/workflows/installers.yml
```diff
@@ -168,3 +168,25 @@ jobs:
...
+      uses: ncipollo/release-action@v1
+      with:
+        draft: true
+        prerelease: true
+        allowUpdates: true
+        replacesArtifacts: true
+        token: ${{ secrets.GITHUB_TOKEN }}
+        artifacts: "installers/dist/SasView5.dmg, installers/Output/setupSasView.exe"
```
rozyczko, line 190: this is osx/windows specific but I think it is fine, since we won't be releasing linux installers any time soon

rozyczko (approved):
looks good!

PR #1971, opened 2021-11-22
Call magnetic angles up_theta 

dehoni:
Orientation angles are θ, φ, ψ  but magnetic orientation uses up_angle and up_phi. Rename up_angle to up_theta for consistency

Verify that θ, φ use the same orientation system as the particle axis, both for applied field and for orientation of the magnetism within the particle.


pkienzle:
Is there a "save perspective" for the generic scattering calculator, and does it store `up_angle`? You may need to add code to update the perspective parameters on reload.

dehoni:
> Is there a "save perspective" for the generic scattering calculator, and does it store `up_angle`? You may need to add code to update the perspective parameters on reload.

The Scattering Calculator can only save nuclear and magnetc SLD data in the sample coordinate frame with no information on environment coordinates that include the angles up_theta and up_phi setting the polarisation vector. The user will provide these parameters as input before computation of the scattering pattern.

PR #1975, opened 2021-11-23
New plotting option iq

Caddy-Jones:
This pull request allows the user to specify what is plotted on the y-axis. For example they may have Q against I(Q^2) data. The user can specify this in the “Fit Options” tab.
image
Currently there are 4 default options, [I, I^2, I(Q^2), I(Q^4)], but there is an option for users to define their own equation using I and Q, along with python mathematical operators. 
Requires https://github.com/SasView/sasmodels/pull/491 to function

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -359,6 +359,10 @@ class GuiManager(object):
                 self._current_perspective.name] = self._current_perspective
             self._workspace.workspace.removeSubWindow(self._current_perspective)
             self._workspace.workspace.removeSubWindow(self.subwindow)
+        import PyQt5
+        PyQt5.QtCore.pyqtRemoveInputHook()
+        import pdb;
+        pdb.set_trace()
         # Get new perspective
         self._current_perspective = self.loadedPerspectives[str(perspective_name)]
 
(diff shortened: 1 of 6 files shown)


rozyczko:
This branch requires having `new_plotting_option_IQ` branch in sasmodels as well.

The UI element is good - one minor quibble is about enablement of custom Data/Title line edits. With default size of the control, enabling the edits causes visible shrinking of the `Weighting` groupbox. I think the layouts are a bit mismatched.

Also, I can see no effect of any change in this widget on the charts. Should I be doing something special to see the new labels?

Caddy-Jones:
Hi @rozyczko 

With regard to the UI element, is there a way to change the behaviour so a scroll bar appears, instead of the `Weighting` group box contracting?

With regard to the “Custom Y-axis” text box. I’m unsure whether it’s worth keeping this as an option. I included this option before I was aware that the “Change Scales” functionality, and I was just hoping to be able to change the chary Y-axis tittle to reflect the data plotted.
There is scope to integrate the Y-axis plotting functionality with the “change scales” functionality for example if the user want to plot I*(Q**2), then the scales are automatically changes to ln(I*(Q**2)). The custom “Y-axis title” will may still need to be an option to accommodate for a user wishing to plot I*(Q**3), in which case the label should probably be defined as “Y-axis scale” instead of “Y-axis title”. But this is just brainstorming.

wpotrzebowski:
@butlerpd and @gonzalezma to look at it again.

butlerpd:
After discussions with Lionel and at the fortnightly meeting, it was agreed that this is not ready to merge. In fact we should separate this into two separate issues. The first addressing the ability to fit data in rescaled plots (e.g. I*Q^4 vs Q^2) so that the fit line is also plotted in that scale.  The second would be to add more weighting functions beyond I, root I etc to include things like Q and Q^2 etc. Alternatively change the weighting choice from a finite selection to a free form?

In the meantime we should close this request but keep the branch which has started the work.

PR #1976, opened 2021-11-24
fix handedness of magnetic structures in mag SLD

dehoni:
The calculation of the magnetic scattering length density does not conserve handedness. The magnetic scattering length assumes always a positive magnetisation in the magnetic scattering component for the SF channel. This is fine for collinear structures, but might run into issue for some non-symmetric cases, when M(-x)=-M(x).

### DIFF
--- a/src/sas/sascalc/calculator/geni.py
+++ b/src/sas/sascalc/calculator/geni.py
@@ -223,6 +223,12 @@ def _calc_Iqxy_magnetic_helper(
         dd, du, ud, uu):
     # Process each qx, qy
     # Note: enumerating a pair is slower than direct indexing in numba
+
+    p_hat = np.array([sin_spin * cos_phi, sin_spin * sin_phi, cos_spin ])
+    #two unit vectors spanning up the plane perpendicular to polarisation for SF scattering
+    perpy_hat = np.array([-sin_phi, cos_phi, 0 ])
+    perpz_hat = np.array([-cos_spin * cos_phi, -cos_spin * sin_phi, sin_spin ])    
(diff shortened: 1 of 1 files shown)


src/sas/sascalc/calculator/geni.py
```diff
@@ -216,6 +216,12 @@ def _calc_Iqxy_magnetic(
 def orth(A, b): # A = 3 x n, and b_hat unit vector
     return A - np.outer(b, b)@A
 
+
+@njit
+def basis_vector(A, B): # A = 3 x n, and B = 3 x n
+    vec = np.cross(A, B)
+    return vec / np.sqrt(np.inner(vec, vec))
```
pkienzle, line 223: `np.inner` is not listed as a [supported numpy function](https://numba.readthedocs.io/en/stable/reference/numpysupported.html) in the current release of numba. This may interfere with JIT performance.

dehoni: It is changes now to np.dot, which should be supported.

dehoni:
Instead of refering to the scattering vector to calculate the orthogonal set of vectors, a better choice would be magnetisation. This would avoid to obtain qat x phat / | qhat x phat | = 0/0 if qhat and phat are parallel. The normalization of perpz could also create problems since P*q = 1 and so perpz -> 0/0. 
For the magnetisation, we know that components along the polarisation direction is always non-spin-flip and if a perpendicular component exist it will contribute to SF scattering making it easy to check if SF is worth to calculate and avoiding ill-behaviour.

PR #1985, opened 2021-12-02
Update of the github workflows

wpotrzebowski:
Fixing release branches (sasview release uses sasmodels 1.0.6 release branch). It currently uses fixed/harcoded branch name, which means that it will need to be updated manually.  Likewise it clones bumps 0.8.1. Not sure if there is better mechanism to do this.  Bumps can probably be pip installed, however I am not sure if documentation works then (we build bumps docs on the fly).

Pinning to mpl version 3.4.3.


.github/workflows/installers.yml
```diff
@@ -168,3 +174,27 @@ jobs:
...
+      run: echo "BRANCH_NAME=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+
+    - name: Declare env variables on pull_request only
+      if: github.event_name == 'pull_request'
+      shell: bash
+      run: echo "BRANCH_NAME=$GITHUB_HEAD_REF" >> $GITHUB_ENV
+
+    - name: Upload installer to GitHub releases
+      #Release and main branch will be treated differently
+      if: env.BRANCH_NAME != 'main' && env.BRANCH_NAME != 'release*'
```
rozyczko, line 190: The second part of the conditional is essentially not needed, since this workflow is running on non-release builds anyway.

.github/workflows/release.yml
```diff
@@ -205,3 +211,26 @@ jobs:
...
+      shell: bash
+      run: echo "BRANCH_NAME=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+
+    - name: Declare env variables on pull_request only
+      if: github.event_name == 'pull_request'
+      shell: bash
+      run: echo "BRANCH_NAME=$GITHUB_HEAD_REF" >> $GITHUB_ENV
+
+    - name: Upload installer to GitHub releases
+      if: env.BRANCH_NAME == 'release*'
```
rozyczko, line 226: Like above, this `if` isn't needed since this runs only on release

rozyczko (dismissed):
Looks good.
Mentioned issues can be dealt with later

PR #1994, opened 2021-12-20
Save project: Properly handle scalars

krzywon:
When saving projects, some scalar quantities (e.g. `numpy.float32` and `numpy.float64`) were unable to set a default when saving the value, so a message was pasted into the console log saying `data cannot be serialized to json`. This fixes that issue.

No associated ticket. Originally noted in an email by @smk78 during testing of 5.0.5rc1.

### DIFF
--- a/src/sas/qtgui/Utilities/GuiUtils.py
+++ b/src/sas/qtgui/Utilities/GuiUtils.py
@@ -2,7 +2,7 @@
 """
 Global defaults and various utility functions usable by the general GUI
 """
-
+import numbers
 import os
 import re
 import sys
@@ -1266,9 +1266,11 @@ def saveData(fp, data):
             content = o.__dict__.copy()
             return add_type(content, type(o))
 
-        if np.isscalar(o):
-            # scalar quantity - no default necessary
-            return None
+        if isinstance(o, numbers.Integral):
+            return int(o)
+
+        if isinstance(o, numbers.Real):
+            return float(o)
 
         # not supported
         logging.info("data cannot be serialized to json: %s" % type(o))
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Utilities/GuiUtils.py
```diff
@@ -1266,6 +1266,10 @@ def jdefault(o):
             content = o.__dict__.copy()
             return add_type(content, type(o))
 
+        if np.isscalar(o):
+            # scalar quantity - no default necessary
+            return None
```
pkienzle, line 1271: `jdefault` should return the serialized version of the value not `None`.

The docs for `np.isscalar` suggest using `isinstance(o, numbers.Number)` if you want to restrict your test to numeric values (note that 'np.isscalar("str")' returns True):
```
import numbers
if isinstance(o, numbers.Number):
    return float(o)
```
However, this will convert all values (int16, int32, int64, float32, float64) to float when they are reloaded.

To preserve int-ness you can use:
```
import numbers
if isinstance(o, numbers.Integral):
    return int(o)
if isinstance(o, numbers.Real):
    return float(o)
```

Note that tuples are treated as lists by json.dump and np.float is deprecated so the first clause should only be checking for sets.

You might be able to dump/load tuples with a specialized JSONEncoder if you care.

pkienzle (approved):
Looks okay. Didn't test. You may still want to remove `np.float` from the first clause: https://github.com/SasView/sasview/blob/e5e9dc04476759586226d413f9eac61aeaf4720b/src/sas/qtgui/Utilities/GuiUtils.py#L1233

butlerpd:
can @smk78 verify this fixes the error he reported above?

smk78 (approved):
Just built locally on W10/x64. I can confirm that this branch is no longer throwing the
```
INFO: data cannot be serialized to json: <class 'numpy.float32'>
```
error when saving the VTMA.h5 file (or any others I've tried).
Fix approved!

butlerpd:
looks like we can merge unless @krzywon wants to respond to the `np.float` comment?

krzywon:
@butlerpd, I don't plan on removing the `np.float` for this release.

PR #2005, opened 2022-01-18
Fixing high CPU consumption on MacOSX

wpotrzebowski:
Also applicable for other platforms.  
(PyQt5 warning fix was reverted to keep this PR simple and not introduce any potential issues for ISIS school)


krzywon (approved):
Running from the build, the CPU usage on windows is now 0% when idle. Approved.

smk78:
@krzywon said:
> Running from the build, the CPU usage on windows is now 0% when idle. Approved.

Just did the same test on W10/x64 and I concur; CPU usage now 0% instead of 15% when SasView running but idle.

Issue #1597, opened 2020-06-16
ESS_GUI: plugin editor not updating model


pkienzle:
After modifying a model in the plugin editor and replotting, the fitpage appears to show the old version of the model.

Perhaps the models are not being properly cleared from memory by the editor?

butlerpd:
Actually, if the model is currently running, SasView does not notice that a change has been made and reload before computing again. However switching to a different plugin and back DOES reload the model including the changes.

Likewise, if Plugin Models category is chosen, adding a new model does not update the model dropdown menu. However, choosing a different category then switching back to Plugin Models DOES reload the list of models including the newly created one.

**Question:** Do we really want SasView to be constantly scanning the plugin directory for changes ... for all open fitpages?   While it is a bit ugly, I'm not convinced that the current behavior is not the correct one? However, even if we leave this behavior, it would be good to document better AND to include that in a tutorial on model writing?

rozyczko:
It's fixed. I originally didn't want the model to be directly notified about a new plugin, since this can occasionally present visual artefacts with the list briefly vanishing and showing again but it seems this isn't happening in the newest Qt?

butlerpd:
Hummm.... I had the same concern hence the question. I guess you are answering it? 😄 Lets see if it is robust I guess.

butlerpd:
This did not close as the merge was not to the default branch (`main`) and linked keywords therefore ignored. Will close this manually now.

PR #2006, opened 2022-01-18
Keyerror fix 1996

rozyczko:
Sometimes the main category dict does not contain custom structure
factors so a check needs to be added.

Issue #1443, opened 2020-01-16
Error with custom models in 5.0


butlerpd:
A user from Lund University has found two bug in the 5.X custom models package (4.2.2 works correctly).

1) New models that are dropped into the plugin folder while SasView is running do not get picked up on the fly. This is true of 4.x as well which is why the option to "load plugin" under " Plugin Model Operations" was added. This option seems to have been lost in the transition to 5.x and clearly needs to be restored.

2) 5.x incorrectly assumes the python file name is the same as the UI name given within the model file via the "name" attribute, appended with a .py.   This of course is not the case. If it were, the name property would be pointless.

rozyczko:
1. The automatic loading of plugins happens on any GUI invoked change in the plugins. Indeed, manually placing a plugin in the directory will not make it discoverable until the application restart. Adding the explicit 'load plugins' action might help but since the workaround is obvious **I don't really agree with the "blocker" status.**
2. Can't reproduce this - if I create a plugin with name test.py and later modify its "name" attribute inside the file to say "test 2", this modified plugin loads in just fine in 5.0, displaying correctly "test 2" in the list of models. Is this what you say is working incorrectly? If so, please let me know all the steps required.

butlerpd:
Sorry ... I was being lazy.  For number 1 I would have put critical (still higher than major because we added it early on to 4.x because of confusion and figure it will only be worse now that people are used to it? but maybe not), but indeed not blocker.
 
The blocker was because of point 2 which I am very surprised didn't reproduce.  Actually we did reproduce on 2 different windows 10 machines.  Further it was failing in 5.0 release and on 5.0.1 nightly jenkins build #317 as well as probably 5.0.1 release candidate 1 (though could have been a later version). I eventually realized the reason the student could not load the plugin by noticing that an error was being generated "cannot find file xxxx" where xxx  was the name attribute value not the file name.  When we renamed the files to match the name attribute it worked on both computers. 

I will try to investigate a bit more, starting by seeing what happens for me when creating a test.py

pkienzle:
Perhaps a dialog to add models to the plugin directory?

That way we don't have to document ~/.sasview/plugins as the path, which can be difficult to navigate to on Mac since dot files are hidden from the finder, and the GUI will know to reset the plugin list.

rozyczko:
Added plugin file loader to the Plugin Manager, under "Add file...".

smk78:
[...]

PR #2013, opened 2022-02-01
Enable model update on adding new plugin. #1597

rozyczko:


butlerpd:
Testing I got a key error which may be fixed by the keyerror fix in PR #2006. Once that is merged, I will merge release branch back into this and try again.

butlerpd (approved):
works like a charm ... even if I change the "name" attribute which currently makes the model no longer loadable (see issue #1443). In that case the model dropdown goes to its default model for the category and updates the fitpanel with the relevant parameters.

Code review is trivial so this can be merged

Issue #1551, opened 2020-05-27
Cancelling Save Project dialog throws


rozyczko:
A simple case of missing early return on cancelling the Save Project dialog.
Exception is caught and displayed and no other effects are present, just looks bad in the Log Explorer.

rozyczko:
Similarly, Edit/Save Params to File dialog upon cancelling

krzywon:
I think I fixed this issue in #1775 with commit https://github.com/SasView/sasview/pull/1775/commits/a1596b1c954e536b121be091c6d86c7ce919448c

butlerpd:
@krzywon and @rozyczko. My test suggests that the first issue (save project) is fixed but that the second one " Edit>Save Params to File" Still has the issue.  Let me know if I'm wrong and thi should be closed. Meanwhile I will move it to 5.1. If I am correct should this be labeled a good first issue? or is it more complicated?

rozyczko:
You are right, @butlerpd - it has not been fixed for this action (Save Params to File). Considering this involves just looking up the error handling from "Save Params" and copy/pasting it, I would conider it a great first issue bug. Assuming we want to postpone it to the next release, rather than fix on the spot?

butlerpd:
Ok .. moving it back to 5.0.5 and we can discuss whether to fix "on the spot" or not.

butlerpd:
The merge should have closed this but it didn't :-( so closing now

PR #2014, opened 2022-02-01
Added a simple check so cancelling the dialog doesn't throw. #1551

rozyczko:


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -3913,20 +3913,20 @@ def onCopyToClipboard(self, format=None):
...
-                Type_output = Excel_output
-                filename = '.'.join((filename, 'xls'))
-            elif file_path[1] == 'Latex (*.log)':
-                Type_output = Latex_output
-                filename = '.'.join((filename, 'log'))
-
-            file_open = open(filename, 'w')
-            with file_open:
-                file_open.write(Type_output)
+            if filename:
```
krzywon, line 3916: My suggestion would be to immediately return if no filename is given `if not filename: return`. The offered implementation suggests potential use cases and code execution if no filename is given.

krzywon (approved):

Issue #1917, opened 2021-09-22
File loading lag in generic scattering calculator


rjbourne:
When loading files into the generic scattering calculator the file selection dialogue appears to lag, for example a delay in displaying which file is being hovered over with the mouse. This does not appear to be related to the recent changes to the generic scattering calculator, as I can replicate the issue on branches derived from older commits prior to the merge of their PRs. The oldest commit I can replicate the issue on is commit https://github.com/SasView/sasview/commit/4cda4d5415bcc2f4635cd67d9c10842e57039b3d, and reverting this commit also appears to prevent the issue. Additionally this lag did not appear on my old laptop - suggesting it may be dependent on the computer or setup of the environment in some way.

Other file loading dialogues in Sasview pass the additional argument `options=QtWidgets.QFileDialog.DontUseNativeDialog | QtWidgets.QFileDialog.DontUseCustomDirectoryIcons`, and adding this argument to the generic scattering calculator file dialog (see branch https://github.com/SasView/sasview/tree/file-loading-lag) also appears to resolve the issue. Exactly what is occurring here to cause the lag I am unsure.

rjbourne:
The issue also effects the Image Viewer tool - which also opens a file dialogue without the additional options. Again either reverting commit 4cda4d5415bcc2f4635cd67d9c10842e57039b3d or adding the additional options resolves the issue. Since the option which has an effect is `QtWidgets.QFileDialog.DontUseNativeDialog` I would guess that some option/setting in the default windows dialogue box causes the issue - and using the widget-based alternative removes this. Exactly how this is affected by the changes in 4cda4d5415bcc2f4635cd67d9c10842e57039b3d I do not know.

smk78:
Could this be related to https://github.com/SasView/sasview/issues/1866?

rozyczko:
Could indeed be the same issue, at least the symptoms look similar.

butlerpd:
Indeed and the fix was exactly the fix that rjbourne says works. Has anybody actually applied those options to the two areas discussed here?

rozyczko:
https://github.com/SasView/sasview/pull/2015

butlerpd:
Specifically, the code snippet used for the general scattering calculator in #2015 should also be applied appropriately in Image Viewer and File Converter.

PR #2015, opened 2022-02-02
Change default options on file load widget to not display dir icons.

rozyczko:
Same solution as elsewhere in code - don't interpret/display directory icons in the file load widget.
Apparently this leads to serious performance issues for very large directories.


butlerpd:
This should resolve the Gen Scat Calculator issue. However I note that the ticket also mentions two other areas, namely `Image Viewer` and `File Converter` tools, that have the same problem? Quoting from #1917 
> The issue also effects the Image Viewer and File Converter tools - which also open a file dialogue without the additional options. Again either reverting commit 4cda4d5 or adding the additional options resolves the issue.

Should we fix those at the same time? or should we just fix this, which would be what affects one of the major new functionalities in 5.0.5?

Issue #1987, opened 2021-12-06
New Dependency Code broken again


butlerpd:
Downloading the Dec Release_5.0.5 Github build, Plotting still seems to be an issue.  Attempting to verify which version of matplotlib was being used I tried running both of the new functions: `Log installed packages` and `Log imported packages`.  Both failed and produced the following errors:
```python
18:03:47 - ERROR: Traceback (most recent call last):
  File "sas\qtgui\MainWindow\GuiManager.py", line 1181, in actionLog_installed_packages
  File "sas\qtgui\MainWindow\GuiManager.py", line 486, in log_installed_packages
AttributeError: 'PackageGatherer' object has no attribute 'log_installed_packages'
18:04:22 - ERROR: Traceback (most recent call last):
  File "sas\qtgui\MainWindow\GuiManager.py", line 1187, in actionLog_imported_packages
  File "sas\qtgui\MainWindow\GuiManager.py", line 492, in log_imported_packages
  File "sas\qtgui\MainWindow\PackageGatherer.py", line 51, in log_imported_packages
  File "sas\qtgui\MainWindow\PackageGatherer.py", line 114, in get_imported_packages
RuntimeError: dictionary changed size during iteration
``` 
Not sure if this is a problem in the Jenkins builds as well (I suspect not?) but if we are to release from github this will need to be addressed.

rozyczko:
When the jenkins build is used, the plotting issues disappear - this is  some issue with mpl 3.x likely.
However, the dependency checker is failing with the same error(s) as posted by Paul.
This will have to be fixed.

wpotrzebowski:
I beleive that reported error has been fixed https://github.com/SasView/sasview/commit/ec1e04ba1d46902f0d19e8248638e7adad0f4186 as it seems to work locally, however running from installer returns:
`08:31:12 - ERROR: Traceback (most recent call last):
  File "sas/qtgui/MainWindow/GuiManager.py", line 1181, in actionLog_installed_packages
  File "sas/qtgui/MainWindow/GuiManager.py", line 486, in log_installed_packages
  File "sas/qtgui/MainWindow/PackageGatherer.py", line 34, in log_installed_modules
  File "subprocess.py", line 415, in check_output
  File "subprocess.py", line 493, in run
  File "subprocess.py", line 858, in __init__
  File "subprocess.py", line 1704, in _execute_child
FileNotFoundError: [Errno 2] No such file or directory: 'pip'`
suggesting that we may need to ship pip with installer

[...]

PR #2016, opened 2022-02-03
Disabling dependency checker from GUI 

wpotrzebowski:
As described in https://github.com/SasView/sasview/issues/1987, dependency checker has been temporally disabled from GUI.


butlerpd (commented):
I thought this was only supposed to disable the two menu items. The changes seem much more substantial here. Am I missing something?

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -651,9 +651,7 @@ def addTriggers(self):
         self._workspace.actionAbout.triggered.connect(self.actionAbout)
         self._workspace.actionWelcomeWidget.triggered.connect(self.actionWelcome)
         self._workspace.actionCheck_for_update.triggered.connect(self.actionCheck_for_update)
-        self._workspace.actionLog_installed_packages.triggered.connect(self.actionLog_installed_packages)
-        self._workspace.actionLog_imported_packages.triggered.connect(self.actionLog_imported_packages)
-
```
rozyczko, line 656: This can be just commented out, so we know we need to uncomment it when fixed.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -1174,18 +1172,6 @@ def actionCheck_for_update(self):
...
-        Log version number of locally installed python packages
-        """
-        self.log_installed_packages()
-
-    def actionLog_imported_packages(self):
-        """
-        Log version number of python packages imported in this instance of SasView.
-        """
-        self.log_imported_packages()
-
```
rozyczko, line 1188: same as above

rozyczko:
> I thought this was only supposed to disable the two menu items. The changes seem much more substantial here. Am I missing something?

While looking for a solution we switched from querying `pip` to using `pkg_resources` module. This gives the list quicker and without spawning a separate process for pip. This would be the preferred solution if it only worked for pyinstaller created virtenvs :(

src/sas/qtgui/MainWindow/GuiManager.py
wpotrzebowski: @rozyczko I was contemplating whether s better to remove or comment out. At the end I decided to remove and added note on commits to the issue https://github.com/SasView/sasview/issues/1987#issuecomment-1029373374.  But I am ok with commenting out if you think it is better.

wpotrzebowski:
> 

The original code was failing (even on local build), so I think it is better to do at least this fix?

src/sas/qtgui/MainWindow/GuiManager.py
rozyczko: @wpotrzebowski I guess this is OK, since we can't really comment out the relevant code in the .ui file (it'll get overwritten on any change). 
Adding a new menu item is trivial and can be done later.
Let's have it removed then.

wpotrzebowski:
> I thought this was only supposed to disable the two menu items. The changes seem much more substantial here. Am I missing something?

As @rozyczko mentiones this is not fully working, however original solution based on `pip` seemed  to be more problematic. I can revert this part of the code and keep it in the branch, however I don't think it affects any other part of code (outside PackageGatherer).

butlerpd:
I vaguely remember that the idea of using `pkg_resources` was in fact the first solution but that it didn't fully work or that it worked in a way that people complained about? It may be that it was not being used properly and that the move away from `pkg_resources` was not the right answer ... or that I'm remembering wrong. I'll try to see if I can figure out what I'm remembering.

However, my main concern was that we needed a better code review than I could do given the changes?.. but I guess the fact that both of you have looked now at the code now we can say that review is good enough?  I'll try running the installer later today for a bit of a functionality test (and for broken things? :-)

butlerpd (approved):
Pip seems to have been used from the first pull request but I do remember a discussion about using `pkg_resources` and it not working?  I'm guessing that was related to not working in the installer version which indeed would have been a non-starter. I also vaguely recall a suggestion on using pip instead as a way around the problem.  But I think these may have all been oral conversations that would have occured before the fist pull request?

That said, if `pkg_resources` can in fact be made to work instead that would be a fine solution so leaving the changes as a starter for finishing this feature seems reasonable enough, but wanted to document what little I remember.

My tests indicate that the version number is still displayed in the log file at startup and on the title bar of the application while the offending menu items are no longer available. Some quick testing of graphs and fitting did not reveal any new issues and the code changes seem to be well contained so really should not affect anything.  Thus I think we can merge this pull request as is.

Issue #1986, opened 2021-12-03
Version tagging for dependencies in release builds


pkienzle:
Looking at the [installers](https://github.com/SasView/sasview/blob/main/.github/workflows/installers.yml) and [release](https://github.com/SasView/sasview/blob/main/.github/workflows/release.yml) workflows, it appears that  SasView is not using any version tagging on third-party dependencies. This will make reproducible builds impossible since it is pulling the latest versions of all packages.

_Originally posted by @pkienzle in https://github.com/SasView/sasview/issues/1973#issuecomment-985758590_

lucas-wilkins:
This seems to be superseded by many of the recent discussions on requriements.txt etc

butlerpd:
Partly. I think the "informal" agreement from Tuesday meetings is that the release.txt should fix version numbers though not clear which packages or all (which would be a real chore).  A discussion with an eventual final decision should probably be had with said final decision documented?

The other part of this is the work done by @Caddy-Jones to add a button in the help menu that prints out a list (a json file actually) of each package and its version number being used.  That code is in 5.0.5 (and main) but the menu button removed as it was still glitchy when he had to move on.

lucas-wilkins:
I'm closing this because I've made a more emphatic plea for it in #2177.

PR #2021, opened 2022-02-14
Switching to PyQT 5.13 

wpotrzebowski:
The latest version of PyQt (5.15) is not statble on Mac OSX. Version 5.13 seems to be stable (tested by few users). 
As discussed at the call this PR also fixes scipy to 1.7.3 (for all platforms).

### DIFF
--- a/.github/workflows/installers.yml
+++ b/.github/workflows/installers.yml
@@ -72,7 +72,12 @@ jobs:
         python -m pip install numpy scipy docutils "pytest<6" sphinx unittest-xml-reporting
         python -m pip install tinycc h5py sphinx pyparsing html5lib reportlab pybind11 appdirs
         python -m pip install six numba mako ipython qtconsole xhtml2pdf unittest-xml-reporting pylint
-        python -m pip install qt5reactor periodictable PyQt5 uncertainties debugpy
+        python -m pip install qt5reactor periodictable uncertainties debugpy
+
+    - name: Install PyQt (Windows + Linux)
+      if: ${{ matrix.os == 'macos-latest' }}
+      run: |
+        python -m pip install
 
     - name: Install lxml, matplotlib and PyQt (OSX)
       if: ${{ matrix.os == 'macos-latest' }}
--- a/.github/workflows/installers.yml
+++ b/.github/workflows/installers.yml
@@ -77,6 +77,7 @@ jobs:
     - name: Install lxml, matplotlib and PyQt (OSX)
       if: ${{ matrix.os == 'macos-latest' }}
       run: |
+        python -m pip install PyQt5==5.13
         python -m pip install matplotlib==3.4.3
         python -m pip install --no-binary lxml lxml
 
--- a/.github/workflows/installers.yml
+++ b/.github/workflows/installers.yml
@@ -69,7 +69,7 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install wheel setuptools
 
-        python -m pip install numpy scipy=1.7.3 docutils "pytest<6" sphinx unittest-xml-reporting
+        python -m pip install numpy scipy==1.7.3 docutils "pytest<6" sphinx unittest-xml-reporting
         python -m pip install tinycc h5py sphinx pyparsing html5lib reportlab pybind11 appdirs
         python -m pip install six numba mako ipython qtconsole xhtml2pdf unittest-xml-reporting pylint
         python -m pip install qt5reactor periodictable uncertainties debugpy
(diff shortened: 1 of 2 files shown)


wpotrzebowski:
@butlerpd: did you have a chance to take look at this pull request?

butlerpd:
I saw that one coming @wpotrzebowski 😄I got caught in paracrystal hell. Good news is I think I've cracked it but items on todo list got backed up. But I'll try to get on it this evening - It should be fairly straightforward as discussed.

wpotrzebowski:
@butlerpd sounds good! @rozyczko can probably also review it?

.github/workflows/installers.yml
```diff
@@ -69,14 +69,20 @@ jobs:
...
 
-        python -m pip install numpy scipy docutils "pytest<6" sphinx unittest-xml-reporting
+        python -m pip install numpy scipy==1.7.3 docutils "pytest<6" sphinx unittest-xml-reporting
         python -m pip install tinycc h5py sphinx pyparsing html5lib reportlab pybind11 appdirs
         python -m pip install six numba mako ipython qtconsole xhtml2pdf unittest-xml-reporting pylint
-        python -m pip install qt5reactor periodictable PyQt5 uncertainties debugpy
+        python -m pip install qt5reactor periodictable uncertainties debugpy
 
-    - name: Install lxml and MPL (OSX)
+    - name: Install PyQt (Windows + Linux)
```
butlerpd, line 77: You are installing PyQt twice on mac and never on win and linux -- thus SasView installer fails to load.  I think line 77 should not change and the changes below moved here then add the windows/linux PyQt .. just make sure that the command is `!=` rather than `== masos-latest` :smile

Same comment applies to the release yml changes

butlerpd (changes requested):

.github/workflows/installers.yml
wpotrzebowski: Good catch @butlerpd! Indeed it was not what I inteded to do... It's been fixed now.

wpotrzebowski:
The builds in prinicple work but SasView fails to start both on Mac and Windows. The reason is latest release of xhtml2pdf: https://github.com/xhtml2pdf/xhtml2pdf/issues/589. It seems that it has already been fixed but haven't been pushed to pypi. It will hopefully be done soon and we will be able to merge this PR. Alternative is to fix version for yet another package...

butlerpd:
reading the xhtml2pdf ticket isn't it the version of reportlab that is broken?  What is needed is to pin reportlab to 3.6.6.... or wait for reportlab to push a fix to pypy? I don't think there is any evidence that this has been fixed though looking at the release history on pypi (https://pypi.org/project/reportlab/#history) there seem to be extremely frequent releases (on average ~ monthly?). Also looking at the commits it seems like only one person?  We may want to pin this?

I think this is part of the issue mentioned by @pkienzle in #1986 - we have lots of dependencies now and just pulling the latest from each will likely cause increasing headaches.  The other side of that coin is that by pinning to versions we don't have the resources to keep them all updated manually -- a true conundrum.

Keeping the `installer.yml` pinned to the latest version might be a way to help mitigate, but then, as soon as a release branch is created all the versions of dependencies should be pinned to the latest version at the time of the branch creation and all PR from branches targeting the release branch (not just commits to the release branch) would need to be building using the installer.yml otherwise the testing becomes problematic....

rozyczko:
> we have lots of dependencies now and just pulling the latest from each will likely cause increasing headaches. 

Yes, this was going to bite us sooner or later.
With the old Jenkins builds, we gradually realized all major modules had to be version-frozen, otherwise we would have to spend a lot of time maintaining updates and assuring things work. Even with this, occasionally we would get issues, when Conda channels dropped the older versions.

This is a serious issue - we can (in general, not for 5.0.5):
1. fix all modules for a given release to have the release stable wrt. dependencies. Then, after the release is out, we un-fix the versions and assure the most recent modules work. Once we update the modules, we fix them in yaml again for the next release etc.
2. test the installers more actively, in order to see the module compatibility issues sooner and address them on the spot.

Any other options?

butlerpd:
That seems like the options. I cannot think of anything else other than "do nothing special" and struggle at the last minute on a regular basis :-).  Unfortunately I'm not sure how to do anything but the last given the resources/bandwidth we have?

Actually .. maybe we can us a script like we were trying to get into 5.0.5 that lists the version of each dependency used at the point of making a release then automatically edit the GHA yml file to pin to those versions (not sure that is possible)? If such were possible  1 above might be the best?

butlerpd (approved):

butlerpd (commented):
the code now makes sense and a quick running on windows doesn't show any issues with plotting or calculations. also reports seem to work OK

Issue #2009, opened 2022-01-20
5.0.5rc2: radius_effective_mode appears broken in P*S fits


smk78:
Testing 5.05rc2 on W10/x64. This issue is a change from 5.0.4.

Fitting a P*S model in 5.0.5rc2 in radius_effective_mode = radius I find that the constrained radius for S is not updated in the FitPage:
![image](https://user-images.githubusercontent.com/10938679/150380891-4df4b37f-528b-40fb-8f81-bf185fe69dc3.png)

It _does_ update in 5.0.4:
![image](https://user-images.githubusercontent.com/10938679/150381023-80575d5d-bde2-47c2-a2fe-8f45d2231641.png)

Worse, setting radius_effective_mode = unconstrained and checking both radii in 5.0.5rc2 only appears to adjust the P radius:
![image](https://user-images.githubusercontent.com/10938679/150381798-00025c88-a2cf-4e5b-a2b6-da1da96a2016.png)
Notice that the chi-squared did not change from the earlier fit either; this suggests that radius_effective_mode is actually broken.

Whereas in 5.0.4 both radii are optimized and the chi-squared changes:
![image](https://user-images.githubusercontent.com/10938679/150381971-70f2ae4d-5012-49f9-a4fc-10402c68c582.png)

smk78:
Having re-installed 5.0.5rc1, it seems to behave like rc2.

butlerpd:
Seem comments posted to #1837 and the discussion in #1525. this may be more subtle than it seems

rozyczko:
Using current `release_5.0.5` branch.

I loaded `P123_D2O_10_percent.xml` and chose `sphere/sphere/hayter_msa`
Left constrained radius_effective and fitted radius and background.
Both values seem the same (barred incorrect formatting of `radius_effective`.

Switched to `unconstrained` and fitted again. Got two different values with Chi2 also different.

I probably did something silly, maybe used a wrong data. Can you please share your `SDS_2_2_SANS` dataset?

RichardHeenan:
Note that Steve's original example has "custom pinhole smear" included.
See https://github.com/SasView/sasview/issues/1966
https://github.com/SasView/sasmodels/issues/488

RichardHeenan:
> Using current `release_5.0.5` branch.
> 
[...]

PR #2024, opened 2022-02-24
radius_effective update for smeared data

rozyczko:
Bring intermediate results back when smearing is on and display radius_effective. #2009

After all this was a SasView issue, although it did look like it originated in SasModels...

This assumes SasModels returns properly smeared P(Q) and S(Q) so we don't have to do it in `ModelThread` as it is currently done. Quick test shows this is the case but more careful check should be done before merge.

`100nmPinholeSphere` file, sphere/sphere/hardsphere, fitted `radius` only.

left: existing 5.0.5rc2, right: from this PR
smeared_intermediates

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -3066,7 +3066,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             return
         # ensure the model does not recompute when updating the value
         self._model_model.blockSignals(True)
-        self._model_model.item(ER_row, 1).setText(str(ER_value))
+        self._model_model.item(ER_row, 1).setText(GuiUtils.formatNumber(ER_value, high=True))
         self._model_model.blockSignals(False)
         # ensure the view is updated immediately
         self._model_model.layoutChanged.emit()
--- a/src/sas/qtgui/Perspectives/Fitting/ModelThread.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ModelThread.py
@@ -180,7 +180,7 @@ class Calc1D(CalcThread):
             if isinstance(return_data, tuple):
                 # see sasmodels beta_approx: SasviewModel.calculate_Iq
                 # TODO: implement intermediate results in smearers
-                return_data, _ = return_data
+                return_data, intermediate_results = return_data
             unsmeared_output[first_bin:last_bin+1] = return_data
             output = self.smearer(unsmeared_output, first_bin, last_bin)
 


gonzalezma:
I repeated my test (sphere * hayter_msa) with and without smearing and got the expected results (same as in version 4..2.2), and now with radius_effective being correctly updated and formatted, so it looks good to me.

smk78 (approved):
Functionality review building this branch on W10/x64. Repeated my original test using the hSDS_D2O_2p0_percent.xml and fitting to sphere*hayter_msa without & with smearing, _radius_effective_ now updates as expected, and the precision reported has been sanitised too! Thumbs up from me!

butlerpd:
Do we know why the mac OS installers are failing on this PR?

Question based on previous conversations, mostly around the fact that the value was properly returned without smearing but not with smearing: Is this a hack just for the smeared case that we should tag needs to be revisited to address properly? or is this the correct fix that duplicates what was being done for the unsmeared case?

NOTE: this should be merged to release either way. The question is simply about whether we should consider the case closed or need to submit a follow on ticket for 5.1?

Also I note that line 182 in ModelThread.py (just above the line 183 which is the one changed for this PR) is a comment 
```
     # TODO: implement intermediate results in smearers
```
which seems to be what this is doing? so should that be removed now?

rozyczko:
> Do we know why the mac OS installers are failing on this PR?

I think it had to do with the Qt version and it will/would be fixed when @wpotrzebowski merges #2021 

> Question based on previous conversations, mostly around the fact that the value was properly returned without smearing but > not with smearing: Is this a hack just for the smeared case that we should tag needs to be revisited to address properly? or is > this the correct fix that duplicates what was being done for the unsmeared case?

The latter. This isn't a hack but extension of what Torin coded originally, 3 years ago.

> Also I note that line 182 in ModelThread.py (just above the line 183 which is the one changed for this PR) is a comment
> which seems to be what this is doing? so should that be removed now?

If this indeed works as we expected for all cases, then yes, this can be removed and the code simplified. I am not familiar with possible use/edge cases so I'll leave proper testing to others.

butlerpd (approved):
Code changes are minimal and two reviewers have approved that the functionality is as expected. The macOS issues should now have been fixed and merged.  The changes to release_5.0.5 could be merged into this branch to verify before merging this back into the release branch but given the actual changes I do not see why that would be necessary and recommend we just merge now.

wpotrzebowski:
I've just merged it with release_5.0.5 branch. I will test Mac OSX installer once it is build and I have time to do so (hopefully today).

wpotrzebowski:
Tested on Mac. Works fine.

RichardHeenan:
Testing local build of 5.0.5_pre_release_setup and 1.0.6_pre_release_setup on Windows, all qualitatively seems to do as expected trying core_shell_sphere times hardsphere with various combinations of polydispersity, radius_effective choice, smearing and beta(Q),  

(Glad to see that Torin had left a "TODO" in the appropriate place to fix this issue, Sorry that the tester likely never tried anything with smearing included, or maybe he forgot to progress the addition of smearing )

However after creating Graph3 with the separate P(Q), S(Q), beta(Q), and Seff(Q), I then appended the fit to Graph3. After a short while the fit on Graph1 stopped updating on subsequent fits. After this first happened I restarted sasview, but then it occurred again. Have not yet worked out how to reproducibly demonstrate this, as I was turning on/off various features here.

Also noted that the separate P(Q) and S(Q) are not smeared, presume this is what we expect???

butlerpd:
Interesting point about P and S not being smeared.  That is I think what was envisaged but now that you mention it @RichardHeenan, one could ask the question. I think I would want to keep the pure, unsmeared versions, but would it also be useful to have the smeared version which would then give rise to the actual smeared curve? I would think yes... but then I start worrying about confusing people with the infinite list of subplots?  Not sure what the best answer is but access to a smeared version of P and S seems like a good idea? Maybe we should add this to one of the discussion tickets on plotting?

As for the fit question, I'm pretty sure I've seen similar things before.  As I recall the updating happens on the new "Graph3" rather than the original fit plot which seems to be "orphaned?"  I thought there was a ticket about that at one point but I' not finding it right away.... I suppose if it does not exist (except in my imagination 😄 ) we should create one.

Issue #2022, opened 2022-02-15
Problem with RPA model in ESS_GUI 5.x


smk78:
This issue is a pointer to a _sasmodels_ issue, the cause and resolution of which is unclear but which could actually be (probably is) a SasView GUI problem.

See https://github.com/SasView/sasmodels/issues/396

The issue itself, however, is serious.

rozyczko:
This is a SasView GUI issue and needs quite a lot of work. Given the problem was around for several 5.0.x releases it probably would be safer to address it in the next release cycle.

butlerpd:
Oops ... I thought it was agreed that the fix is related to completely redoing how sasview handles the control parameters from sasmodels which is clearly beyond anything that can be achieved for 5.0.5.  The confusion maybe that for 5.0.5, or more precisely sasmodels 1.0.6, a note was added to the model in sasmodels. Perhaps this is in 5.0.5 in case there is some note that can be placed in the GUI for when that model, and only that model, is chosen?  But I don't think the GUI structure allows that in any reasonable way?

Otherwise, I second the motion that this should move to 5.1

rozyczko:
We can remove rpa from the gui altogether?
This would be a pretty simple, easily reversible fix.

butlerpd:
?? Really @rozyczko ? basically so it does not show up in the dropdown even though it is part of the sasmodels package?  If so we should probably do that. That would be awesome. I thought that would be nearly impossible. I guess I don't know how that list is loaded and categorized?

smk78:
This issue is fixed by the merge above so closing this ticket.

PR #2025, opened 2022-02-25
Don't show RPA in the list of models #2022

rozyczko:
Added a simple conditional in the list comprehension for filling the model combobox with model names.

This can be easily removed once rpa works again.

Note: this only removes RPA from the model list on the fitting widget. It is still present in the Category Manager.


butlerpd (approved):
Very clean solution with only one line of code slightly modified.  Tested with windows and works as intended. While it does show up in category manager as mentioned, it doesn't matter as adding the model to an existing category or even creating a new one does not allow it to showup due to this line of code.

Someone may want to double check on Mac OS but I recommend it be merged

Issue #2036, opened 2022-04-26
Unable to select LM (scipy:leastsq) optimizer despite it is default optimizer


wpotrzebowski:
I am not able to select LM (scipy:leastsq) optimizer from Fit Algorithm menu:

<img width="736" alt="Screenshot 2022-04-26 at 08 14 30" src="https://user-images.githubusercontent.com/4189922/165234060-9d7530c3-e8fd-47c1-b4b4-b6d484e1eacd.png">

even tough it is default optimizer

wpotrzebowski:
I observed this on Mac but I guess it may be the same issue on other platforms too.

rozyczko:
This requires adding a new page for `scipy.leastsq` minimizer (Perspectives/UI/FittingOptions.ui) identical to that for `page_lm`.
Also, the name will have to be internally changed to `scipy_leastsq` so the widget can be properly called.

wpotrzebowski:
This issue came out after merging bumps=0.9.0 and it is not vissible in 5.0.5rc2 (which uses standard LM).

smk78:
> I observed this on Mac but I guess it may be the same issue on other platforms too.

Yes, same on Windows.

krzywon:
Places that need changing for the default to show up properly on the fitting perspective:

> sas.sascalc.fit.BumbsFitting.py, line 17 -> change `fitters.LevenbergMarquardtFit.id` to `fitters.MPFit.id`
> sas.qtgui.Perspectives.Fitting.FittingPerspective.py, line 80 -> change `fitters.LevenbergMarquardtFit.id` to `fitters.MPFit.id`

butlerpd:
fixed by merging of #2037

PR #2037, opened 2022-04-26
Use MPFit fitter as the default fit engine

krzywon:
This uses the id for the MPFit engine as the default name when generating the fitting window since this optimizer is now the default. If we want to remove the old `scipy.leastsq` option from the Fit Options combobox we could add something like the following to src/sas/sascalc/fit/BumpsFitting.py at line 17 (credit @pkienzle).

>`try:`
>`FIT_ACTIVE_IDS.remove(fitters.LevenbergMarquardtFit.id)`
>`except ValueError:`
>`pass`

refs #2036

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
@@ -77,7 +77,7 @@ class FittingWindow(QtWidgets.QTabWidget):
         # Fit options - uniform for all tabs
         self.fit_options = options.FIT_CONFIG
         self.fit_options_widget = FittingOptions(self, config=self.fit_options)
-        self.fit_options.selected_id = fitters.LevenbergMarquardtFit.id
+        self.fit_options.selected_id = fitters.MPFit.id
 
         # Listen to GUI Manager signal updating fit options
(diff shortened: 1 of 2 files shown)


rozyczko (approved):
Code looks good.
Functionality is as expected, the new MPFit based minimizer is used as seen in the Log Explorer.
A new Convergence plot shows up, coming from Bumps, which is a change from previous behaviour but this commit has nothing to do with it.
Correct string is displayed in the fitting widget.
Project save/load work.

butlerpd:
After discussion with @krzywon he agrees that @pkienzle comments are correct but that at this extremely late stage in the release process, the PR as is would be the least invasive/risky.  He recommends we take the comments and create a high priority issue for after the release.  Given @rozyczko comments this sounds like the right answer?

pkienzle:
The proposed change is not a high priority issue. The initialization is the same in both locations so there is little risk of a bug.

I checked whether the GUI could change the choice of fitter before sascalc fitting is imported but I don't think it can. You could mitigate that risk with the following at the top of FittingPerspective:
```python
# Force initialization of default fitter.
import sas.sascalc.fit.BumpsFitting  # pylint: disable=unused-import
```

Issue #2045, opened 2022-05-21
INFO and DEBUG logging doesn't work with 5.0.5 (Mac)


wpotrzebowski:
As a part of final testing for 5.0.5 release I discovered that logging of INFO and DEBUG doesn't work on Mac. 
Diagnostics: SasView doesn't output anything to Log Explorer window on start up and e.g model generation. 
The issue seems to be related to problem with finding logging.ini file as 
`/Applications/SasView5.app/Contents/MacOS$ ./sasview` works fine
` /Applications/SasView5.app/Contents/MacOS/sasview` gives error `ERROR: Logging.ini not found...`
![Screenshot 2022-05-21 at 07 26 19](https://user-images.githubusercontent.com/4189922/169637128-02fa7f09-2536-4996-aa2f-9f1edb6c1836.png)

ERROR logging works fine
![Screenshot 2022-05-21 at 07 33 20](https://user-images.githubusercontent.com/4189922/169637205-e14fe7d5-8859-4c6a-8449-28924c439ec3.png)

Strangely enough 5.0.4 bundle has the same setup for folders and it works fine...

PR #2050, opened 2022-05-24
Fixing INFO logging on Mac OSX

wpotrzebowski:
Addressing https://github.com/SasView/sasview/issues/2045. 
The change in code is rather small and shouldn't really affect other platforms but it has to be tested on Widnows (and ideally Linux too).


rozyczko (approved):
the change looks OK, works fine on windows
needs testing on Linux and macos

wpotrzebowski:
Tested on Linux from developers env and works fine.

timsnow:
Can confirm that this works on Mac as wel
Screenshot 2022-05-26 at 09 32 17
l

PR #2053, opened 2022-05-25
Minor refactoring of corefunc code, removed python2 stuff, added type…

lucas-wilkins:
Addresses #1365, which is essentially a "wont do".

The current smoothing seems perfectly reasonable, and the Savitzky-Golay filter is not particularly appropriate for this task.

However, I've done some refactoring and movement towards more idiomatic python3 and added some type hints.

### DIFF
--- a/src/sas/sascalc/corfunc/smoothing.py
+++ b/src/sas/sascalc/corfunc/smoothing.py
@@ -64,8 +64,11 @@ class SmoothJoin:
 
             # Use np.divide
             x_mid = x[mid_indices]
-            h = 1 / (1 + (x_mid - self.stop) ** 2 / (self.start - x_mid) ** 2)
 
+            # This shouldn't have any divide by zero, as x_mid is strictly bigger than start
+            # and strictly less than stop
+            h = 1.0 / (1.0 + ((x_mid - self.stop) / (self.start - x_mid)) ** 2)
+            
             y[mid_indices] = h * self.right_function(x_mid) + \
                              (1 - h) * self.left_function(x_mid)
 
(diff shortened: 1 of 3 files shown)


smk78 (commented):
I get the gist of what's been done here, but don't consider my python strong enough to approve these changes.

src/sas/sascalc/corfunc/smoothing.py
```diff
@@ -0,0 +1,98 @@
...
+def show_plots():
+    """ Local sanity-check plots for the joiner"""
+    import matplotlib.pyplot as plt
+
+    # Sin example
+    xs = np.linspace(0, 10, 100)
+    # Symmetry check
+
+    plt.plot(xs, SmoothJoin(lambda x: 0.0, lambda x: 1.0, 2,8)(xs))
+    #plt.plot(xs[::-1], SmoothJoin(lambda x: 1.0, lambda x: 0.0, 2, 8)(xs))
```
rprospero, line 88: Can we drop this dead code?

src/sas/sascalc/corfunc/smoothing.py
```diff
@@ -0,0 +1,98 @@
...
+
+    :param left_hand_function: The first curve to interpolate
+    :param right_hand_function: The second curve to interpolate
+    :param start: The value at which to start the interpolation
+    :param stop: The value at which to stop the interpolation
+    """
+
+    def __init__(self,
+                 left_hand_function: Callable[[ArrayLike], ArrayLike],
+                 right_hand_function: Callable[[ArrayLike], ArrayLike],
```
rprospero, line 26: Adding types here was absolutely the right idea.

src/sas/sascalc/corfunc/vectorisation.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+from typing import Union, Callable, Any
+
+# TODO: Proper typing when newer numpy (>1.20) is supported
+ArrayLike = Union[np.ndarray, float, int]
+
+def assure_vectorised_argument(f: Callable):
```
rprospero, line 7: I imagine that there will be many places in the code where this could be useful.  Maybe we could move it to a more general location than just within Corfunc?

rprospero (changes requested):
Everything looks great.  I just think that `assure_vectorised_argument` needs to be put in a more prominent location.

src/sas/sascalc/corfunc/vectorisation.py
lucas-wilkins: Rough plan is to generalise it (or find the code where I've done it before) and to move it somewhere better, but the util section needs tidying in general. I've not round to doing this yet, will create a separate issue for it.

lucas-wilkins: Created issue #2054

src/sas/sascalc/corfunc/smoothing.py
lucas-wilkins: maybe, I always like to have that kind of thing hanging around, but its perhaps not appropriate here

lucas-wilkins: Ah, you were referring to the outdated comments, rather than the plotting tool.

Fixed in latest commit

rprospero (approved):

Issue #1540, opened 2020-05-17
Corfunc GUI plot initialization idiosyncrasies


butlerpd:
As of 5.0.2, when choosing corfunc as the perspective for the next `Send data to` operation it brings up not only the parameter and input panel but two blank plot pages.  However, while these have default scales on the axes they neither have their plot name (`Scattering data` for the top plot and `Real Space Correlations` for the bottom) nor the axis labels [e.g.  `I(Q)` vs `Q` for the top]

Two other things about these plots that are problematic:
1. They are clearly using a different plotting routine as the rest of SasView as can be seen from the (rather ugly) standard matplotlib toolbar at the bottom of the plot.  That said, there have been other request for better toolbar functionality on regular SasView plots, so this may be a case of needing to have a discussion on GUI design and user needs?
2. Corfunc also is "violating" the standard way perspectives behaves in SasView which is part of what leads to the initial comment.  Basically it seems to be creating a panel with inputs and two plots as a single panel so the plots cannot move around or be stretched independently of each other or the rest of the panel.  it does however have the advantage of preventing the multiplication of random floating plots everywhere which is a huge issue when working on big projects.   Again this may be an opportunity to open a discussion on GUI design and user needs.  In the end though we should strive to maintain some sort of uniformity.

Noted on both PC and mac

lucas-wilkins:
Can we discuss this at meeting, in particular how this would fit with any upcoming changes of the plotting system.

lucas-wilkins:
Corfunc has been changed significantly since this issue. In summary, the plots are now better, they do "violate" the floating plots system, this is perhaps not a bad thing though. There are currently some things to bring up to date still, but there are separate issues for them.

PR #2055, opened 2022-05-26
1936 unable to save extrapolation in corfunc in 5x

lucas-wilkins:
Added a button to save the extrapolated data in corefunc. Although this is apparently for feature parity, I couldn't find the original feature.

Currently, you can use the original q values, or choose a new **linear** sampling. It would be a bit weird to use just the original q values for exporting an extrapolation. But I'm also aware that using geometrically spaced Q values is quite common, so maybe this linear version isn't the right thing - depends what its used for ultimately, I guess.

It currently doesn't have any tests, as the unit tests are currently broken and disabled.


rprospero (approved):

smk78:
The ability to save the extrapolation in SasView 4.x was through the plotting context menu.
image
This ability was lost when Corfunc was added to SasView 5.x because a different approach to plotting was implemented.

The way that plotting is currently handled in Corfunc in 5.x actually needs to change to be consistent with the rest of plotting in SasView (ie, like Fitting) because the current implementation of the Corfunc perspective is restrictive for some screen resolutions. From my quick look, it looks like an issue for this requirement never got created. But was discussed.

lucas-wilkins:
Perhaps we create a new issue then?

smk78:
> Perhaps we create a new issue then?

Actually, see https://github.com/SasView/sasview/issues/1540

PR #2061, opened 2022-06-06
add cases in OMFReader

dehoni:

Issue #2060, opened 2022-06-06
read OOMMF V2 files


dehoni:
omv files created with OOMMF are shlightly different to what the GSC expects as input for magnetic files, e.g. the entry "valuemultiplier", "valuerangeminmag", "valuerangemaxmag" are optional, and "valueunit" can provide a value for each dimension.

The OMFReader should be able to handle the different cases.

dehoni:
The entries "valuerangeminmag", "valuerangemaxmag" are actually never used later in the GCS. I commented them out for the moment. The other issues are handled such that the reader accepts files in the omf format seen so far.

dehoni:
[isolated_sk_T-Oxs_MinDriver-Magnetization-00-0000217.omf.txt](https://github.com/SasView/sasview/files/8854247/isolated_sk_T-Oxs_MinDriver-Magnetization-00-0000217.omf.txt)
[isolated_skyrmion.omf.txt](https://github.com/SasView/sasview/files/8854248/isolated_skyrmion.omf.txt)
The two files are identical except of the header, which is OMF V2.0 for the first file and a V1 version called isolated_skyrmion.omf.

dehoni:
The PR should provide a unit test, and a more meaningful message stating Version loaded. For failure the error message "We accept only Text format OMF file." is not specific enough, test if it is a omf file and where it fails (e.g. missing variable).

PR #2076, opened 2022-06-15
provide test data for GSC in test folder and update documentation 

dehoni:


wpotrzebowski:
I will test the files. Quick question. I guess we want to merge to` main` not `env_setup`?

wpotrzebowski (approved):
We should probably fix the typo (and maybe tooltip) but otherwsie good to go.

dehoni:
> @dehoni, are these oomf v2 files, in which case, this would be related to issue #2060 and PR #2061? In either case, this is an innocuous change that adds more example data and lays the groundwork for adding unit tests.
> 
> I haven't tested loading the files using the scattering calculator and would suggest someone do that before merging.

The omf files are a separate issue (with a bit of discussion how to go forward). Here it is only sld and vtk files.

PR #2079, opened 2022-06-15
Added feature request label to template

lucas-wilkins:
I was finding it hard to see which issues were feature requests, I've added the label to a few, add it to the template too


wpotrzebowski (approved):

krzywon (approved):
I think #2086 is a bit better, making it obvious we can provide a list of pre-defined tags, but, as-is, the two PRs should do exactly the same thing. I'm going to approve both. @lucas-wilkins feel free to choose which one you like the best and close the other.

lucas-wilkins:
This appears to have used the wrong syntax

Issue #1979, opened 2021-11-25
update check points to http instead of https


butlerpd:
To check for the latest version the application checks a json file on github io called latestversion.json. however the link currently coded in `src/sas/sasview/local_config.py` is http://www.sasview.org/latestversion.json it should instead point to https://www.sasview.org/latestversion.json.  This is a minor issue since in most cases the address will just be redirected to the secure version. There may be occasions however where the client computer will refuse the redirect a problem that is likely to increase a security becomes a bigger and bigger concern.

smk78:
Would it not be sensible to get this into 5.0.5?

butlerpd:
I agree the case could certainly be made - and in fact was how I first assigned it.  However, I then remembered the rules of the road agreed to just recently and as a problem that has existed since the beginning (going back as far as 5.0.1 or before) it was not clear to me that it should be pushed to the release branch, particularly since it is unlikely to be causing a problem to anyone.  

That said the change is so trivial it should not pose any problem whatsoever to include it I would think.

lucas-wilkins:
This appears to have been fixed but not closed. Closing.

PR #2087, opened 2022-06-16
Update latestversion to https

smk78:
This PR addresses https://github.com/SasView/sasview/issues/1979, updating the location of latestversion.json from http to https.

It also similarly updates the location of www.sasview.org.


wpotrzebowski (approved):

butlerpd:
The only thing I might suggest, as long as we are changing sasview.org as well, is that all the http:// get replaced with  https://... all EXCEPT of course for the marketplace which for now is still on an old unencrypted server?

Otherwise it is ready to merge

smk78:
> The only thing I might suggest, as long as we are changing sasview.org as well, is that all the http:// get replaced with https://... all EXCEPT of course for the marketplace which for now is still on an old unencrypted server?

I had wondered about doing that @butlerpd but was a little concerned there might have been knock-on implications somewhere else. :-) But with your encouragement I have now done so! And have updated a couple of broken links too.

Issue #2098, opened 2022-06-20
Using the Data Operation Tool adversely affects Save/Load Project


smk78:
**Describe the bug**
Responding to a help query from User Marite C I have discovered that if you have any datasets in the Data Explorer that have been created in the Data Operation tool, any project you save will be void; it will be a 2-line file containing:
```
{
  
```

This happens whether you _only_ use the Data Operation tool, _or_ if you use any of the main perspectives _and then_ use the Data Operation tool.

I find this behaviour exists in versions 5.0.5, 5.0.4, 5.0.3 and 4.2.2 and thus it would appear to be 'forever behaviour'.

**To Reproduce**
1. Start SasView
2. Load some data
3. Send it to Fitting, and optimise a a parameter or two
4. Save Project (a valid .json file will be written)
5. Now create a dataset in the Data Operation tool
6. Save project (an invalid .json file will be written)

**Expected behavior**
I think there are two strings to this issue and thus I have labelled it as both a defect and a feature request:

The **defect**: using the Data Operation tool in a SasView session that has been using the normal perspectives (Fitting, Inversion, P(r), Corfunc) should not then stop a User from saving a valid project file with those analyses.

The **feature request**: datasets created with the Data Opearation tool should be able to be saved to a project file.

**SasView version (please complete the following information):**
 - certainly affects 4.2.2 and later

**Operating system (please complete the following information):**
 - W10 & Mac affected for certain, likely affects all platforms

rozyczko:
Nice find!
I've no idea how this issue managed to hide for so long, with the data operation tool being rewritten in 5.0...
Anyway, it seems fixed.

smk78:
Fixed by https://github.com/SasView/sasview/pull/2099

PR #2099, opened 2022-06-20
Add id to the generated plot. #2098

rozyczko:
Plots generated by the Data Operation tool did not have the `id` attribute set. This attribute is used as a dict key when saving plots to the project json file.
Added the id attribute.

A more correct way would be to have this (and other, required) attributes defaulted in the Plotter initializer, so they are always present. This should be a part of the plot rewrite story, mentioned elsewhere


lucas-wilkins:
Looks good as a bug fix

lucas-wilkins (approved):

wpotrzebowski:
I can confirm the fix works on Mac. Ready to nerge.

smk78:
Confirm the fix also works on W10/x64

PR #2114, opened 2022-07-01
Updating developers list for zenodo

wpotrzebowski:
Adding Nicolas and Lucas to the developers list. We can also add more contributors.


smk78 (approved):
Changes look good.

PR #2117, opened 2022-07-08
Resource and UI compilation is now path-independent.

### LINKED ISSUE #1060: use sasview installation to run batch scripts (Trac #989)
Migrated from http://trac.sasview.org/ticket/989
```json
{
    "status": "new",
    "changetime": "2017-10-27T16:12:50",
    "_ts": "2017-10-27 16:12:50.689931+00:00",
    "description": "Should be able to use the installed sasview GUI package to run fitting scripts such as sasmodels/example/batch_fit.py.\n\nMake sure this works and provide documentation for doing so.",
    "reporter": "pkienzle",
    "cc": "",
    "resolution": "",
    "time": "2017-08-09T20:41:18",
    "component": "SasView",
    "summary": "use sasview installation to run batch scripts",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView 4.3.0",
    "owner": "",
    "type": "defect"
}
```

rozyczko:
This change is for a small improvement to the resource and UI building step in `run.py`.
Instead of requiring `pyrcc` and `pyuic` scripts be present in the PATH, the main run script is now importing necessary modules.
This also avoids using os.system() calls which are potential security issue.
Patch supplied by @pkienzle 
This fixes #1060

### DIFF
--- a/src/sas/qtgui/convertUI.py
+++ b/src/sas/qtgui/convertUI.py
@@ -2,15 +2,33 @@
 import os
 import sys
 
+def run(main, name, *args):
+    saved_argv = sys.argv
+    sys.argv = [name, *args]
+    try:
+        main()
+    except SystemExit as exc:
+        if exc.code != 0:
+            raise RuntimeError(f"\"{name} {' '.join(args)}\" exited with {exc.code}")
+        return exc.code
+    finally:
+        sys.argv = saved_argv
+        pass
+    return 0
+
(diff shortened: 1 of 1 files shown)


lucas-wilkins (approved):
I think this is an improvement, it's a much better way of running things.

As a side note though, I'm not sure this does much for security when we can execute any bit of code we like anyway by just dropping them in the right directory (e.g. loaders, plugins etc).

PR #2125, opened 2022-07-21
2108 display average soft block thickness in corfunc

lucas-wilkins:
Soft block thickness now present in corfunc. @smk78 to verify it does the right thing.


lucas-wilkins:
Looks like I need to fix the tests. Will do that tomorrow.

smk78 (approved):
Functionality review on W10/x64.

Does what's expected. New parameter is also incorporated in the new report. Thumbs up.

Issue #2126, opened 2022-07-22
UI compiler not found


lucas-wilkins:
**Describe the bug**

The packages PyQt5.uic.pyuic doesn't appear to exist with the current requirements.txt. Note that this did work before - it's probably a missing package or something path related. Looking through my Qt installation, the file does not appear to be there.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new conda environment, python 3.8.13
2. pip install -r requirements.txt
3. Try and run

**Expected behavior**
Finds package

**SasView version (please complete the following information):**
Current main branch

**Operating system (please complete the following information):**
 Windows 10

**Additional context**

The problematic imports are from the recent #2117 PR

rozyczko:
Seems that the requirements file needs to have this added: `pip install pyqt5-tools`

lucas-wilkins:
Seems it was the missing its also missing bumps, html2text, and the conditional stuff didn't work

butlerpd:
Hummm... I thought we worked on fine tuning that at the Lund workshop?  How did this get missed? or were the changes made in a branch that was not merged. @wpotrzebowski would be the one to ask I guess?

wpotrzebowski:
Somehow older version of requirements file got merged to main... What we did in Lund at the end is here https://github.com/SasView/sasview/blob/env_setup/build_tools/requirements.txt. Please note that platform specific conditionals didn't work so they were removed but it seemed to work on both Mac and Windows.

PR #2127, opened 2022-07-22
Added missing dependencies to requirements.txt

lucas-wilkins:
Resolves recent issue #2126

### DIFF
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -30,7 +30,5 @@ cffi
 dominate
 html5lib
 importlib-resources
-importlib_resources
 bumps
 html2text
-pyqt5-tools
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -22,7 +22,7 @@ qt5reactor
 periodictable 
 uncertainties 
 debugpy
-matplotlib==2.2.5
+matplotlib==3.4.3
 PyQt5
 lxml
 pytools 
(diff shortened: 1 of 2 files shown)


butlerpd:
I think we should add @wpotrzebowski to the review since he was the one last working on fine tuning requirements.txt.  Just in case there was something special we are missing here?

lucas-wilkins:
> I think we should add @wpotrzebowski to the review since he was the one last working on fine tuning requirements.txt. Just in case there was something special we are missing here?

I meant to.

lucas-wilkins:
@wpotrzebowski I've removed the OS dependency - can you check I used the right versions (I picked the windows/linux associated ones)

wpotrzebowski (commented):
Do we need "pyqt5-tools" if we already have PyQt5?

wpotrzebowski (changes requested):
importlib_resources seems to refer to importlib-resources i.e. does it appear twice on the list?

wpotrzebowski:
Probablty not as a part of this pull request but it may be worth to set up GH actions that "tests" requirements.txt file when change is made in it. By testing I mean pip installing on workers.

lucas-wilkins:
importlib_resources seems to refer to importlib-resources i.e. does it appear twice on the list?

They're for different python versions - though as we discussed recently, there is no need to support python < 3.7, so the former is not really needed.

wpotrzebowski (approved):
Tested on Mac. It works fine

Issue #2107, opened 2022-06-24
IDF plot selection


lucas-wilkins:
Migrated from #1502

implement a way of displaying the IDF but not on top of the Correlation Functions (y axes are too different). Maybe by using a "Display IDF" checkbox that is unchecked by default?

Issue #2103, opened 2022-06-24
Copy parameters from corfunc gives wrong values


lucas-wilkins:
**Describe the bug**

Edit -> copy parameters gives fitting parameters not corfunc parameters

**To Reproduce**
1. Corfunc
3. Menu -> Edit -> copy parameters
4. ...
5. See error

**Expected behavior**
Copys corfun parameters

See issues #1992 and #1502

lucas-wilkins:
Need to have a chat about what level to implement the copy/paste functionality in other perspectives.

Issue #2043, opened 2022-05-11
Font scaling becomes irregular when using multiple screens on windows


lucas-wilkins:
When using multiple monitors the GUI gets confused about font size, and the dialogs become unreadable.

![image](https://user-images.githubusercontent.com/9784906/167890267-b2344f88-66cf-4c71-a485-cdd9a5c4dca7.png)

Its currently unclear to me what extent this is a general Qt issue, and how much is specific to SasView.

V5.0.4, Tested on windows 10 enterprise.

butlerpd:
This has been noted for OSX but not for windows before. Wonder if it is related?

lucas-wilkins:
I've just had the same issue with Qt designer. I take it that this means that it's a general Qt problem, and therefore something that is not something that we'll fix within sasview.

Close as wont fix

PR #2135, opened 2022-08-05
2133 resize and and rearrange corfunc panel

lucas-wilkins:
I've done some rearranging of the CorFunc panel to make it more compact.

@smk78 and @rozyczko - I would like your opinion

I also have an outstanding bug that I'm struggling with. The left hand part seems to load at a strange size on my machine. I'm on my week off now, so I thought I'd get this PR up and give you some time to look at it, despite this issue.

The issue is this: In Qt Designer its has a fixed horizontal size and looks like this:

image

But when I load it it looks like this

image

Note: the fixedness of the size doesn't change, just the size it gets fixed to. Any thoughts on what's happening would be appreciated.


smk78:
Running this branch on W10/x64 I get this console message as the SasView starts:
```
(dev) D:\SasViewDevelopment\sasview-pr2135>python run.py
Attribute Qt::AA_EnableHighDpiScaling must be set before QCoreApplication is created.
```
Then, when it does start the screen looks like this (on a 1680x1050 monitor):
image
Is it possible the Qt error is related to the weird scaling?
In any event, I don't see why I should be expected to use the lateral scroll to use the Corfunc perspective on a widescreen monitor.

smk78 (approved):
I do, however, like this layout more than the previous!

- As commented on https://github.com/SasView/sasview/pull/2132, I would much prefer to have draggable range bars on the Q-space plot, rather than the 'Adjust' bar;
- Could we display the experimental data as markers+error bars, rather than joined datapoints, please?;
- Could we have a third plot tab to display the IDF, please? This might require a change to the plot tab labels (since the IDF is also real space);
- I note that 'Save Extrapolated' writes a 2-column _CSV_ file whilst 'Save Transformed' writes a 4-column _space-delimited_ file with a Corfunc-specific extension: there should be some consistency! I would suggest we standardise on CSV for both for now (ideally, of course, we should also offer the transformed as CanSAS1D and NXcanSAS with separate SASdata blocks for the 1D, 3D & IDF);
- Could we add a 3rd column to Save Extrapolate - the intensity minus the background (ie, what is being transformed; at least, what I assume is being transformed???!!!) - please?;
- We need a way to save the Extracted Parameters to file (Edit > Save Params to File doesn't work), but I think this is/was a separate issue?

But overall, I give this PR qualified approval.

lucas-wilkins:
> Attribute Qt::AA_EnableHighDpiScaling must be set before QCoreApplication is created.

Hmmm, that might be something to look at in general, for all my weird Qt issues. e.g. #2043

lucas-wilkins:
> As commented on https://github.com/SasView/sasview/pull/2132, I would much prefer to have draggable range bars on the Q-space plot, rather than the 'Adjust' bar;

In due time.

> Could we display the experimental data as markers+error bars, rather than joined datapoints, please?;

Please make feature request to this effect.

> Could we have a third plot tab to display the IDF, please? This might require a change to the plot tab labels (since the IDF is also real space);

Isn't this already a feature request - #2107

> I note that 'Save Extrapolated' writes a 2-column CSV file whilst 'Save Transformed' writes a 4-column space-delimited file with a Corfunc-specific extension: there should be some consistency! I would suggest we standardise on CSV for both for now (ideally, of course, we should also offer the transformed as CanSAS1D and NXcanSAS with separate SASdata blocks for the 1D, 3D & IDF);

Feature request

> Could we add a 3rd column to Save Extrapolate - the intensity minus the background (ie, what is being transformed; at least, what I assume is being transformed???!!!) - please?;

Sounds like you want to open a new feature request ;)

> We need a way to save the Extracted Parameters to file (Edit > Save Params to File doesn't work), but I think this is/was a separate issue?

I've been working on this. #2103 - Same underlying functionality, work on this stalled because I needed to chat with people about it and I started on other bits instead.

PR #2139, opened 2022-08-11
Include sas.qtgui.Utilities.Reports in build output

krzywon:
During some testing, I found a couple of things missed by PR #2065. The Reports package was never included in setup.py so builds would likely not include the reports dialog (untested). Also, a unit test package was named incorrectly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -195,6 +195,13 @@ package_dir["sas.qtgui.UtilitiesUI"] = os.path.join(
     "src", "sas", "qtgui", "Utilities","UI")
 packages.append("sas.qtgui.Utilities.UI")
 
+package_dir["sas.qtgui.Utilities.Reports"] = os.path.join(
+    "src", "sas", "qtgui", "Utilities", "Reports")
+packages.append("sas.qtgui.Utilities.Reports")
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -195,6 +195,13 @@ def run(self):
     "src", "sas", "qtgui", "Utilities","UI")
 packages.append("sas.qtgui.Utilities.UI")
 
+package_dir["sas.qtgui.Utilities.Reports"] = os.path.join(
+    "src", "sas", "qtgui", "Utilities", "Reports")
+packages.append("sas.qtgui.Utilities.Reports")
+package_dir["sas.qtgui.Utilities.UI"] = os.path.join(
```
lucas-wilkins, line 201: Should this be "sas.qtgui.Utilities.Reports.UI"

lucas-wilkins (approved):
Just that directory that needs to be checked.

wpotrzebowski (commented):
@krzywon do we need to add all this dependencies to docs.yml and test.yml? I think initial idea was to have these jobs light-weight. I understand that for consistency it may be worth to keep the same set of requiremenets for all ymls.

PR #2141, opened 2022-08-12
Use sasdata package in place of sas.sascalc.dataloader

krzywon:
This PR replaces the `sas.sascalc.dataloader` subpackage with the stand-alone [sasdata](https://github.com/SasView/sasdata) package. This is a step toward a complete separation between the calculation and GUI.

All data import/export capabilities, including unit tests, are now held within the sasdata package and are removed from sasview in this PR. Github workflows are updated to build and bundle the package into the installers. 

This is a parallel PR to [sasmodels 517](https://github.com/SasView/sasmodels/pull/517). This PR also includes all work done in #2139.


lucas-wilkins:
I expect this will have quite a lot of collisions with #2145

smk78 (approved):
Functionality test on W10/x64 (once I'd figured out how to test this PR!).

Tested loading individual and multiple (ctrl-click) 1D files of different types from 'Load data' button. OK.
Tested File > Load Data File(s) with 1D data. OK.
Tested File > Load Data Folder with 1D data. OK.
In some instances the following appeared in the Log Explorer:
```
WARNING: ascii_reader: could not load file
```
It would be more helpful if the filename that threw the warning was included.

Tested loading individual and multiple (ctrl-click) 2D files of different types from 'Load data' button. OK.
Tested File > Load Data File(s) with 2D data. OK.
Tested File > Load Data Folder with 2D data. OK.

Rejects coordinate and convertible files as it should with a sensible message in the Log Explorer. OK.

Also rejects image files (which should only load through the Image Viewer) but populates the Log Explorer with a whole lot of junk! It would be nice if this could be sanitised.

Image Viewer loads image files. OK.

Generic Scattering Calculator loaded coordinate files. OK.

Workspaces loaded in the Data Operation tool. OK.

So in functionality terms this PR seems fine (hence I'm approving), but a little tidying up on the messaging would be nice.

butlerpd:
Need to prioritize this PR due to the number of changes that might affect other work.

lucas-wilkins:
Seems to work with the pycharm workflow I have

lucas-wilkins (approved):

wpotrzebowski (commented):
I've been wondering where `sasdata` is included in the installer file and couldn't find it. Interestingly `sasdata.__file__` run from the shell returns:
`/Volumes/SasView5/SasView5.app/Contents/MacOS/sasdata/__init__.pyc`, which doesn't exist. 

It seems to work, so it is fine. It is just puzzling how it is imported and whether we need to anything for pyinstaller (like we did for sasmodels) to make sure is always running. 

Otherwise looks good.

rozyczko:
Looks and behaves good on Windows.
`python setup.py build install` in `sasdata` properly installs the module.
All references in sasview seem to work correctly.
I guess python examples and jupyter notebooks in other repos will need updating?

wpotrzebowski:
> Looks and behaves good on Windows. `python setup.py build install` in `sasdata` properly installs the module. All references in sasview seem to work correctly. I guess python examples and jupyter notebooks in other repos will need updating?

Yes, I was thinking the same but shouldn't that produce a folder/file in the installer? On Mac I would expect some trace of it in:
 
Screenshot 2022-08-18 at 12 45 22

rozyczko:
> > Looks and behaves good on Windows. `python setup.py build install` in `sasdata` properly installs the module. All references in sasview seem to work correctly. I guess python examples and jupyter notebooks in other repos will need updating?
> 
> Yes, I was thinking the same but shouldn't that produce a folder/file in the installer? On Mac I would expect some trace of it in:
> 
> Screenshot 2022-08-18 at 12 45 22

Oh, you mean the installed binary. Then it probably got compiled in into the executable?

krzywon:
Should I add it to the spec? It wouldn't take much to make the change.

wpotrzebowski:
> Should I add it to the spec? It wouldn't take much to make the change.

If the `sasdata` imports correctly I don't think it is necessary to add it explicetly to installer. I would even consider importing/installing `sasmodels` same way, so we can also remove it from installer. I don't really know why we need to add `sasmodels` and `sasmodels-data` explicetly. Maybe for docs?

Issue #2134, opened 2022-08-05
Corfunc plot functionality


lucas-wilkins:
My recent PR #2106 has made the Corfunc behave a bit more normally, however, they are still quite different from the other plots.

There are ongoing questions about how they should behave,
1) Discuss what should be done at this stage:
  a) how do we want it to behave
  b) what are the plans for other plotting functionality
2) Implement

lucas-wilkins:
I've not really got any answer to this, will be leaving things as they are for now.

butlerpd:
@lucas-wilkins I personally think these are great questions we **should** address but not sure about enough "community bandwidth" a the moment? Maybe this would be a good discussion topic (or subtopic ) for the code camp?

PR #2142, opened 2022-08-15
Changed appearance of Corfunc q space

lucas-wilkins:
I've changed the appearance, though there is still the question of how plots are to work in general, and how corfunc plots should work (#2134)


smk78 (approved):
Experimental data in Corfunc plots now displays the same as data in other perspectives.
image

PR #2143, opened 2022-08-15
2137 change file format for save transformed in corfunc

lucas-wilkins:
Changed output to .csv format


smk78 (approved):
Does what it says!

Issue #2140, opened 2022-08-11
A more maintainable setup.py


lucas-wilkins:
After forgetting to add a new directory I created to setup.py, and looking at the issues it has if package names are not properly entered, I wonder if would it make more sense to just recursively search the source tree and add everything automatically, rather than having to explicitly maintain a list of the packages we want to include?

Is this something that should be done? Perhaps there a reason why it is currently done the way it is.

krzywon:
In the same regard, we should talk about bundling our workflow ymls so we don't miss dependencies. The html2text dependency (adding it to #2139 shortly) is currently only in the release yml.

krzywon:
To add to my previous comment, most of the workflow steps overlap each other. We should be able to combine installers, test, and docs into a single workflow.

lucas-wilkins:
> The html2text dependency (adding it to https://github.com/SasView/sasview/pull/2139 shortly) is currently only in the release yml.

I thought we decided to only maintain release.yml  - am I wrong about this? - anyway, I think this highlights the issue quite well.

krzywon:
> I thought we decided to only maintain release.yml - am I wrong about this? - anyway, I think this highlights the issue quite well.

I don't know, TBH. What I do know is installers.yml is what generates the test executables for branches and PRs. release.yml is what is used to generate the executables attached to releases. Releases typically use a pinned version of sasmodels, and often for bumps as well, instead of master branches.

lucas-wilkins:
> I don't know, TBH. What I do know is installers.yml is what generates the test executables for branches and PRs. 

o_O 

> release.yml is what is used to generate the executables attached to releases. Releases typically use a pinned version of sasmodels, and often for bumps as well, instead of master branches.

Yeah

lucas-wilkins:
This has been done via #2419

PR #2152, opened 2022-08-17
Moved example data files

lucas-wilkins:
moved example data files, I don't think this has broken anything - but maybe there's some things I have forgotten to check.


krzywon (commented):
setup.py bundles these files on lines 260-272 and should be updated so they aren't missed. Related to #2140

lucas-wilkins:
@krzywon Thanks. Definitely need to be able to move stuff around without breaking things, its a real barrier to cleaning anything up.

lucas-wilkins:
Installer failed error fixed in #2153 - merging this into that.

PR #2158, opened 2022-08-18
Removed and moved multiple .spec files

lucas-wilkins:


rozyczko (commented):
you removed `sasview.spec` entirely?

wpotrzebowski:
> you removed `sasview.spec` entirely?

I think it is the one in build_tools and we have another one in installers?

rozyczko:
> > you removed `sasview.spec` entirely?
> 
> I think it is the one in build_tools and we have another one in installers?

Ah! Ok, I just looked at the changes and assumed this was the only occurrence of that file.

lucas-wilkins:
Yeah, it was an old one that built version 4

rprospero (approved):
These spec files haven't been kept up to date.  Even if they were needed, someone would be better off modifying the main spec file than trying to update these files.

Issue #2164, opened 2022-08-19
Update config from list of variables to class


lucas-wilkins:
Use a class to specify the config, so its clear what variable should and shouldn't be there. It's hard to keep things tidy when you can't easily trace usages.

lucas-wilkins:
Step 1: Move config stuff (custom_config, local_config, LocalConfig and their copies) into a single class that can be inspected
Step 2: Make config IO work with yaml, NO OUTPUTTING PYTHON CODE
Step 3: Remove any usused config items
Step 4: Relocate any items that don't belong
Step 5: Make config more of a data container by removing any methods and dynamic stuff
Step 6: Rename config items in a consistent way
Step 7: Add schema for validating config files

lucas-wilkins:
TODO: images should be located using importlib, not through ad-hoc dynamic paths in config

PR #2167, opened 2022-08-19
Add a preferences window

### LINKED ISSUE #2166: Create Centralized Preferences Window
Many options and defaults are changed and set in different places throughout SasView.

There should be a single, cohesive window that handles all preferences and options. This panel should tie into the config file(s) to set defaults as needed.

**Additional context**
This is related to #258, #268, and #1726. There are likely more issues that I wasn't able to find in a 2-minute search.

krzywon:
This adds a preferences panel that can be opened from the File menu. There is a helper class for adding property widgets, a help file, and some basic unit tests.

I'm opening this as a draft because @lucas-wilkins is actively working on the config files for #2164 and we plan to combine our efforts once both are ready.

Fixes #2166


lucas-wilkins:
See https://github.com/SasView/sasview/issues/2176 for why your tests are failing, merge the associated branch for fix, or merge main when https://github.com/SasView/sasview/pull/2178 is merged in.

lucas-wilkins:
Hey, I'm pretty close to having the config stuff updated. There's a lot of changes, but it should be pretty simple to use. Just ironing out bugs that are only detected when building the installers.

src/sas/qtgui/Utilities/PreferencesPanel.py
```diff
@@ -0,0 +1,170 @@
...
+        super(PreferencesPanel, self).__init__(parent)
+        self.setupUi(self)
+        self.parent = parent
+        self.setWindowTitle("Preferences")
+        self.warning = None
+        # A list of callables used to restore the default values for each item in StackedWidget
+        self.restoreDefaultMethods = []
+        # Set defaults values for the list and stacked widgets
+        self.stackedWidget.setCurrentIndex(0)
+        self.listWidget.setCurrentRow(0)
```
rozyczko, line 61: why is this necessary in the `__init__` method? Aren't current indices/rows set to 0 anyway on instantiation?

src/sas/qtgui/Utilities/PreferencesPanel.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def onClick(self, btn: QPushButton):
+        """Handle button click events in one area"""
+        # Reset to the default preferences
+        if btn.text() == 'Restore Defaults':
+            self.restoreDefaultPreferences()
+        elif btn.text() == 'OK':
+            self.close()
+        elif btn.text() == 'Help':
+            self.help()
```
rozyczko, line 79: it's probably better to create separate slots for all the buttons rather than compare with the `text` property. Those texts can change and updating the handlers is not an obvious follow up.

src/sas/qtgui/Utilities/PreferencesPanel.py
```diff
@@ -0,0 +1,170 @@
...
+        for method in self.restoreDefaultMethods:
+            if callable(method):
+                method()
+            else:
+                logger.warning(f'While restoring defaults, {str(method)} of type {type(method)}'
+                               + ' was given. A callable object was expected.')
+
+    def close(self):
+        """Save the configuration values when the preferences window is closed"""
+        if hasattr(self.parent, 'guiManager'):
```
rozyczko, line 92: Why is this check necessary? Isn't `guiManager` always present on the mainWindow?

src/sas/qtgui/Utilities/PreferencesPanel.py
krzywon: It's not necessary, but a good practice to ensure the two widgets are at the same point.

src/sas/qtgui/Utilities/PreferencesPanel.py
krzywon: Agreed. This was a placeholder.

src/sas/qtgui/Utilities/PreferencesPanel.py
krzywon: For unit testing, I create a dummy parent with a minimal implementation of the main window that doesn't have `guiManager`. When #2168 is finished, how the custom config is written will change so the check will not be required.

wpotrzebowski:
@krzywon is this PR still in DRAFT state?

krzywon:
> @krzywon is this PR still in DRAFT state?

This is waiting on #2168 to be finalized. Once it is, I will need to make a few changes here to account for the configuration changes.

krzywon:
I have updated the window to account for changes made to the configuration system in PR #2168. There are a few outstanding modifications needed, as noted by @rozyczko, otherwise, this is ready for review.

**Question:** There are at least three ways forward from here - which one is preferred? I plan to work on 2 and 3 (and beyond) during the code camp if that makes a difference:

1. I can keep this PR as bare as possible with just the blank window. Other settings can then be moved into the window by submitting separate PRs.
2. I can add to this PR by creating a basic widget that houses the existing configuration values. (default file extensions, open folders, toolbar visibility, etc.)
3. Along with (2), I can also move certain preferences into this window, including the fitting options, and the fitting optimizer selection tool.

lucas-wilkins:
Regarding option 2:

> I can add to this PR by creating a basic widget that houses the existing configuration values. (default file extensions, open folders, toolbar visibility, etc.)

I'm not sure how many of these need to be in a preferences panel. I fact, I'm not sure some of them belong in the config, I just wasn't sure about removing them from it at the time.

wpotrzebowski (approved):
I think it is good enough to give basis for Preferences windows. I've vote for merging it and move on with remaining functionality later

rozyczko:
One potential issue is with multiple config files design. Should we not expect tab-per-config file view rather than have all config items thrown into a single window/view?
On the other hand, this GUI should offer a view on the json file generated by SasView rather than on the python files with definitions/defaults, so it's probably moot.

krzywon:
I merged main, updated my unit tests, and added tests for the preferences widget generator. Assuming CI succeeds, this should now be ready to merge

PR #2171, opened 2022-08-30
1911 - Fix issue copying custom config file

### LINKED ISSUE #1911: sasview.exe cannot create the custom_config.py file
As mentioned in @1672 and other places, if the `sasview.exe` program tries to copy the `default_config.py` to `custom_config.py` in the users `.sasview\config` folder it fails.  Fortunately this bit of code should in principle never be needed as the `custom_config.py` file is normally installed by the installer on a clean install (barring some unusual and as yet undocumented situation) and from there on it should always be there with whatever user modifications get made.  However, in the case that it accidentally gets deleted or moved, the following function `src/sas/_config.py` is activated:

```python
def setup_custom_config(app_dir, user_dir):
    path = make_custom_config_path(user_dir)
    if not os.path.isfile(path):
        try:
            # if the custom config file does not exist, copy the default from
            # the app dir
            shutil.copyfile(os.path.join(app_dir, "custom_config.py"), path)
        except Exception:
            logger.error("Could not copy default custom config.")
```
I have not yet seen any case where this ever works. In all cases I have been able to ascertain this function **fails** producing the error in logger.error and the program then stops with a message `Failed to execute script sasview`. Since it is so rare that this would actually be called, the easiest way to test it is to manually delete the `.sasview\config\custom_config.py` file and then try running the program again.

It is unlikely to be a permission problem since the program can write the the `.sasview` directory (for example when saving a custom model). Further, even when changing the `HOME` directory variable so that the program tries to write to a different directory (e.g. c:\temp), it will still fail.

In order to keep tickets from getting convoluted with many issues, I recommend we only add comments here related to this copying and not to any other failure to run sasview issues.  Those more general issues can be added to #1672 which remains open and now contains quite a few reports.

krzywon:
This is a test to see if the installer generated during the merge with the main branch is able to find and copy `custom_config.py` from the installation directory to the user directory.

I will test the windows installer. @celinedurniak or @wpotrzebowski, could one or both of you test the mac installer?

Steps to test:
- Delete (or move) `custom_config.py` from your `~/.sasview/config` directory
- Install the [build](https://github.com/SasView/sasview/releases/tag/untagged-a93138cafa80d1785765)

Fixes #1911

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,6 +54,9 @@ if os.path.isdir(sas_dir):
     f_path = os.path.join(sas_dir, 'config', "custom_config.py")
     if os.path.isfile(f_path):
         os.remove(f_path)
+    f_path = os.path.join(sas_dir, 'config', "custom_config.pyc")
+    if os.path.isfile(f_path):
+        os.remove(f_path)
     #f_path = os.path.join(sas_dir, 'plugin_models')
     # if os.path.isdir(f_path):
     #     for f in os.listdir(f_path):
--- a/src/sas/_config.py
+++ b/src/sas/_config.py
@@ -1,6 +1,7 @@
 # Setup and find Custom config dir
 from __future__ import print_function
 
+import glob
 import sys
 import os
 import traceback
@@ -90,7 +91,7 @@ def setup_custom_config(app_dir, user_dir):
         try:
             # if the custom config file does not exist, copy the default from
             # the app dir
-            shutil.copyfile(os.path.join(app_dir, "custom_config.py"), path)
+            shutil.copyfile(glob.glob(os.path.join(app_dir, "custom_config.py"))[0], path)
         except Exception:
             logger.error(traceback.format_exc())
             logger.error("Could not copy default custom config.")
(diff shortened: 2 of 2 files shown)


celinedurniak:
@krzywon The link to the new build given above does not work: https://github.com/SasView/sasview/releases/tag/untagged-a93138cafa80d1785765

rozyczko:
These should work
https://github.com/SasView/sasview/releases/download/untagged-b40208b2345b22730c20/SasView5.dmg
https://github.com/SasView/sasview/releases/download/untagged-b40208b2345b22730c20/setupSasView.exe

celinedurniak:
Thanks @rozyczko I managed to download and install this test version of SasView. But I still have the issue with the absence of custom_config.py.

lucas-wilkins:
This will probably collide with my current PR, which deletes local_config and custom_config

krzywon:
> This will probably collide with my current PR, which deletes local_config and custom_config

Then this work isn't needed. Closing.

PR #2174, opened 2022-08-31
Update feature_request.md

lucas-wilkins:
I merged the enhancement and feature request labels. Feature request no longer exists.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,7 +2,7 @@
 name: Feature request
 about: Suggest an idea for this project
 title: ''
-labels: ["Feature Request"]
+labels: ["Enhancement"]
 assignees: ''
 
 ---


smk78 (commented):
Hmm. My tuppence worth is that enhancements and features are different beasts, maybe 'defined' depending on whether the change is incremental or significant? For example, a User might point out that there is a better way to code a model or that some text in the documentation needs updating - I would call these enhancements - or they might request a completely new model, or a new type of analysis- which I would call feature requests.

What do others think?

lucas-wilkins:
@smk78 The feature request label has only been a thing for about 2 months, since I added it when the feature request template was made. Adding it was a mistake. Most of the time its not clear whether something is an enhancement or a feature request, and having two different labels makes it hard to look through the requests - in practice what happened was that all the new [enhancement/feature requests] were labelled Feature Request, and all the old ones were labelled Enhancement. 

Of course, viewed from a more theoretically oriented, semantic perspective the terms "Enhancement" and "Feature Request" have different referents. But this really isn't about having labels with analytically precise meanings, its about splitting issues into useful categories (currently Defect, Enchancement, Housekeeping and no job type label if it doesn't really fit those).

In any case, I have already deleted the Feature Request label. This PR is just to fix the feature request template. Not merging will just keep the incorrect template in place, it wont bring the Feature Request label back.

smk78:
Oh I see. Sorry, I misunderstood. In that case I guess I'd better approve this!

smk78 (approved):
The Feature Request label has already been deleted. This PR is just to fix the feature request template.

PR #2181, opened 2022-09-07
Tabs, Batch Processing and 2D data slicing and processing for pr

ru4en:
# Tabs, Batch Processing, Plots, Tables and 2D data slicing & processing for Pr

This is the work I’ve done during my summer placement. Most of the changes I have made are in the Inversion Perspective directory. I’ve attached A short document describing the changes/feature I’ve implemented.

## Features: 
-	[X] Tabs
-	[X] 1D Batch Processing
-	[X] 2D slicing and Batch Processing of slices
-	[X] 2D plot and 1D slice Plots
-	[X] Batch Table
-	[X] CSV exporting and Plotting on the Batch Table window
-	[x] Minor changes such as fixing the Dropdown menu and remembering Parameters

Find more Information in the PDF bellow.
[Notes and Reference Document for Pr interface DRAFT.pdf](https://github.com/SasView/sasview/files/9506431/Notes.and.Reference.Document.for.Pr.interface.DRAFT.pdf)


krzywon:
The tests are failing because the fix applied in #2178 is not incorporated here. Your branch is 200+ commits behind so I suggest you merge `main` it into your branch, fix the conflicts, and push the result. If you need help with this process, let me know.

ru4en:
Hi and thanks @krzywon I just merged main to 2D-Data-Processing-for-Pr (Hopefully i didn't break anything). However im still getting a unit-test error. From my understanding its something to do with installing openCL not sure how to fix this to be honest. 

image

krzywon:
> From my understanding its something to do with installing openCL

This was a temporary issue accessing the package for download. Rerunning the action solved the issue.

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
```diff
@@ -111,21 +85,83 @@ def __init__(self, parent=None, data=None):
...
+            self.removeTab(index)
+            del self.tabs[index]
+            self.tabsModifiedSignal.emit()
+        except IndexError:
+            # The tab might have already been deleted previously
+            pass
+
+    ######################################################################
+
+    def serializeAll(self):
```
krzywon, line 143: This will only save the active inversion tab into project files. A list of active indices should be stored in the perspective, iterating through the indices here, serializing each tab, and then combining the serialized data into a single dictionary.

According to `GuiManager.py` line 761:
`# Analysis should return {data-id: serialized-state}`

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
```diff
@@ -111,21 +85,83 @@ def __init__(self, parent=None, data=None):
...
+        # The tabs need to be movable
+        self.setMovable(True)
+
+        # Initialize the first tab
+        self.addData(None)
+
+    ######################################################################
+    # Batch Mode and Tab Functions
+
+    def calculate(self):
```
krzywon, line 104: I don't see any usages of this method.

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
```diff
@@ -0,0 +1,1333 @@
...
+        # else add to the tabs.
+        if self.batchResultsWindow is None:
+            self.batchResultsWindow = BatchInversionOutputPanel(
+                parent=self,
+                output_data=self.batchResults)
+        else:
+            self.batchResultsWindow.newTableTab(data=self.batchResults)
+        self.batchResultsWindow.show()
+
+    def stopCalculation(self):
```
krzywon, line 526: Clicking the stop button gives the following error:

> ERROR: Traceback (most recent call last):
  File "\sasview\src\sas\qtgui\Perspectives\Inversion\InversionWidget.py", line 537, in stopCalculation
    self.updateGuiValues()
  File "\sasview\src\sas\qtgui\Perspectives\Inversion\InversionWidget.py", line 746, in updateGuiValues
    GuiUtils.updateModelItemWithPlot(self._data, self.prPlot, title)
  File "\sasview\src\sas\qtgui\Utilities\GuiUtils.py", line 325, in updateModelItemWithPlot
    assert isinstance(item, QtGui.QStandardItem)
AssertionError

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
```diff
@@ -0,0 +1,1333 @@
...
+        self.communicate = self.parent.communicate
+        self.communicate.dataDeletedSignal.connect(self.removeData)
+        self.batchResults = {}
+
+        self.plot2D = Plotter2D(quickplot=True)
+        self.plot1D = Plotter(quickplot=True)
+        self.logic = InversionLogic()
+
+        # Allow Tabs to close
+        self._allowClose = True
```
krzywon, line 112: Closing the invariant window no longer allows any perspective to be opened. All other perspectives are minimized when the red X is clicked. This should default to False and only change to True when switching perspectives.

> ERROR: Traceback (most recent call last):
  File "\sasview\src\sas\qtgui\MainWindow\GuiManager.py", line 1131, in actionInversion
    self.perspectiveChanged("Inversion")
  File "\sasview\src\sas\qtgui\MainWindow\GuiManager.py", line 378, in perspectiveChanged
    self._workspace.workspace.removeSubWindow(self._current_perspective)
RuntimeError: wrapped C/C++ object of type InversionWindow has been deleted

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
```diff
@@ -171,10 +172,10 @@ def _post_data(self, nbins=None):
         item = self._item
         if self._item.parent() is not None:
             item = self._item.parent()
-        GuiUtils.updateModelItemWithPlot(item, new_plot, new_plot.id)
+        # GuiUtils.updateModelItemWithPlot(item, new_plot, new_plot.id)
 
-        self.base.manager.communicator.plotUpdateSignal.emit([new_plot])
-        self.base.manager.communicator.forcePlotDisplaySignal.emit([item, new_plot])
+        # self.base.manager.communicator.plotUpdateSignal.emit([new_plot])
+        # self.base.manager.communicator.forcePlotDisplaySignal.emit([item, new_plot])
```
krzywon, line 178: This breaks the base Sector interactor used in 2D plotting. No sector plots are generated, or, if they are, they are no longer accessible to the user. Maybe add a new variable to the method call called something like `generate_plots` that defaults to True and your method(s) could override that? These signals need to be emitted by default.

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
```diff
@@ -273,6 +274,55 @@ def draw(self):
...
 
+    def getSlice(self, nbins=None):
+        """
+        """
+        data = self.data
+        # If we have no data, just return
+        if data is None:
+            return
+        # Averaging
+        from sas.sascalc.dataloader.manipulations import SectorQ
```
krzywon, line 285: This should now import from the `sasdata` package at `sasdata.data_util.manipulations`

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
```diff
@@ -273,6 +274,55 @@ def draw(self):
         """
         self.base.draw()
 
+    def getSlice(self, nbins=None):
```
krzywon, line 277: This overlaps the `_post_data` method on line 121 quite a lot. Did you just want to expose the private method in some meaningful way?

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
```diff
@@ -0,0 +1,1333 @@
...
+    def currentTabDataId(self):
+        """
+        Returns the data ID of the current tab
+        """
+        tab_id = []
+        if self.logic.data_is_loaded:
+            tab_id.append(str(self.logic.data.id))
+        return tab_id
+
+    def updateFromParameters(self, params):
```
krzywon, line 819: Tab index will need to be added to the parameters dictionary so all tabs can be saved/loaded appropriately.

krzywon (changes requested):
I haven't had a chance to fully review the code or any of the 2D functionality but will come back to finish soon. See my individual notes for required changes.

lucas-wilkins:
> What is self.supports_reports and self.supports_fitting_menu used for and why do we need it for Inversion? These 2 attributes are new after the merge from main and are required for the interface to load. I’m assuming self.supports_reports is something to do with @lucas-wilkins new reporting/exporting functionality if so that would be a great addition to inversion and I would like to know how I could connect the output to it so reporting would work for Pr.

`supports_reports` is a flag that should be set to true if the perspective can make reports (its used by the main gui to work out whether to give the an option to make a report).

`supports_fitting_menu` is a similar flag that determines whether the fitting menu should be displayed.

I should add doc strings to perspecive.py

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
ru4en: I haven't worked with any of the serialising, exporting and importing sides of things so this slipped under the radar. IDing things is something that I will  needed to add as tabs or data currently is not IDed, Working on getting this fixed.

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
ru4en: I've removed that function since it was something I was trying out but didn't end up using at the end. Sorry about that it wasn't supposed to end up in the final version.

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
ru4en: Could you please elaborate on how I could reproduce this? I'm unable to replicate this error.

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
ru4en: moved closeEvent() from InversionWidget.py to InversionPerspective.py in the new commit. The perspective does not close now but minimises like it used to.

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
ru4en: I've tried to get the slicer to work with the uncommented code but I cant seem to get to work. I'll try again tomorrow.

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
ru4en: Made the changes: `from sas.sascalc.dataloader.manipulations import SectorQ` to `from sasdata.data_util.manipulations import SectorQ`

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
ru4en: Yes that's exactly the function I wanted but I needed it to be public and output the data. Also I was afraid if I made any changes to it id end up breaking it. Ill try this again by creating a new function that calls this and then returns the data.

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
ru4en: Yes, Ill try to assign IDs for each of the tab I believe this should also fix [this](https://github.com/SasView/sasview/pull/2181#discussion_r966385534).

ru4en:
Thanks @lucas-wilkins, I’ll make sure to set those as true for the Inversion Perspective.

ru4en:
Thank you @krzywon for reviewing and providing feedback. Ive made a few changes as per the comments however Im still working on the output for _post_data and base Sector interactor bits.

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
krzywon: I'm suggesting something like this and when you call `getSlice()`

```
def _post_data(self, nbins=None, show_plots=True):
    ...
    if show_plots:
        GuiUtils.updateModelItemWithPlot(item, new_plot, new_plot.id)
        self.base.manager.communicator.plotUpdateSignal.emit([new_plot])
        self.base.manager.communicator.forcePlotDisplaySignal.emit([item, new_plot])
```

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
krzywon: If this is to capture the slice data and use it elsewhere, I'd suggest you couple my above suggestion with:

```
def captureSlice(self, nbins=None):
    new_plot = self._post_data(nbins, show_plots=False)
    # Do anything else you need to do with the plot that isn't done in _post_data
    return new_plot
```

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
krzywon: If you assign a unique id to each tab (similar to M1, M2, etc for the fitting perspective), you can capture that for each serialized page in `serializeCurrentPage`.
```
def serializeAll(self):
    state = {}
    for index in self.tab_ids:  # tab_ids would need to be populated somewhere
        state.update(self.serializeCurrentPage(index))
    return state
```

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
krzywon: I can't seem to recreate this issue right now. Let me get back to you on this.

krzywon:
One more issue I just found - the number of terms button text is never populated. Clicking it clears the input box. I haven't been able to trace the issue, but that will also need to be fixed.

lucas-wilkins:
This branch is [33 commits ahead](https://github.com/SasView/sasview/compare/main...2D-Data-Processing-for-Pr), [612 commits behind](https://github.com/SasView/sasview/compare/2D-Data-Processing-for-Pr...main) main.

lucas-wilkins:
This branch has got very behind and now has some conflicts that are not obvious to resolve. I'm going to mark this as stale, hopefully that will prompt people, e.g. @dehoni and @krzywon, to do something about merging it before it gets even worse.

lucas-wilkins:
I've merged in main, this has the following consequences:
1) @rozyczko's tab ordering has been lost because of the deep conflict there
2) I noticed that P(r) didn't inherit from perspective, it does now
3) I also noticed some problems with the updating of the window doing single fitting:
  a) The empty window wasn't replaced when I put in the first data
  b) The values didn't update in the window until I selected different tabs

lucas-wilkins:
@dehoni If it happens again I won't fix it, but suggest that its closed instead.

butlerpd:
@dehoni cannot look at it just now but will. Keep on the list for updates.

butlerpd:
Now being looked at by @dehoni

butlerpd:
@dehoni working on it. Progress being made not ready yet

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
dehoni: Right now it gives the file name as tab_name. Is there any convention on it?

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
dehoni: Tried to reproduce it, but stop seems to works. At least it for 1D

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
dehoni: Rewritten taking into account next comment exposing _post_data with captureSlicer method.

src/sas/qtgui/Plotting/Slicers/SectorSlicer.py
dehoni: Rewritten as proposed

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
dehoni: Done

wpotrzebowski:
@dehoni still works on it

dehoni:
I have one issue with the sliders, when moving them the values for qmin or qmax are not updated. If I change the value qmin and qmax in the tab, the changes are forwarded to the slicer only after calculating. Issue 1), I am not sure where to get the actual slicer values to update, and on reverse side, i.e. when the text fields are changed, a signal should be triggered...

wpotrzebowski:
@dehoni should we be aiming for 6.0.0 with this PR?

dehoni:
Definitely, but I will need help to embed the graphs into the Sasview with Guiutils... Tried and failed.

dehoni:
Thank you Lucas for showing me one or two very helpful IDE tricks. Makes life much easier. Left to do linking in the SectorInteractor correctly.

dehoni:
Mandatory features included, open for testing.

wpotrzebowski:
@butlerpd and @wpotrzebowski will try to look at

wpotrzebowski:
I've discovered a few issues while doing a functionality test
1) The "suggestion" for the parameter/parameters becomes empty when running PR in batch mode
Screenshot 2023-12-04 at 06 08 51
2) Legends on the plot look bit odd. (I guess there is some fixed number of characters. Would it possible to follow logic from fitting?)
Screenshot 2023-12-04 at 06 06 03
3) Help button returns an error: 
"""
06:18:52 - ERROR: Traceback (most recent call last): File "sas/qtgui/Perspectives/Inversion/InversionWidget.py", line 569, in help AttributeError: 'InversionWidget' object has no attribute '_manager'
"""
4) Remove button doesn't seem to work

src/sas/qtgui/Utilities/GridPanel.py
```diff
@@ -573,6 +573,7 @@ def newTableTab(self, tab_name=None, data=None):
 
     @classmethod
     def onHelp(cls):
```
lucas-wilkins, line 575: This should not be a class method if it needs to access self. What's going on?

dehoni: Rewrote and included regeneration of documentation with the latest commit.

dehoni:
All request from reviewers have been addressed.

lucas-wilkins (approved):
I've not re-reviewed, but my changes have been addressed

butlerpd (changes requested):
I have done a quick functionality test with windows 10. I tested the old PR functionality reasonably extensively but do not know how to use the new 2D functionality nor have I tried testing too much adjacent functionality. 

I must say I **love** the new tabbed PR window which now behaves like the fitting page allowing multiple curves to be fit with PR at the same time. This is **GREAT**. However there are two major issues that need to be addressed (breaking old functionality)
* On loading a 1D data the parameter window does not calculate the suggested values for the number of terms or the regularization constant as it used to.  After the first calculation it will do so. However from there on it is remains one step behind, only suggesting the best value after the calculation.  I suspect that is because someone accidentally deleted the call to the calculation as soon as a value in the parameter box is changed (or intentionally not realizing why that was being done?).
* An even worse issue is that the Q range is fixed in stone: the min and max display but cannot be changed (well you can change them but as soon as `calculate` is hit the values revert to the original). Further, while the slider lines activate, turning red when you mouse over them, the mouse **cannot** make them move.
* A minor point is the dropdown box that was meant to allow several different 1D plots at once in the inversion (but does not work very well for me in 5.0.6) should be removed (i.e. it should no longer be a dropdown as there is never any choice there). In fact, given we are tabbing it like the fit panel, we should treat it like the fit page where the data set is just a text label not a text box of any sort?

Finally there is NO documentation accompanying this massive change which is part of why I could not test the 2D (there is no information on that). Moreover there are signivicant changes to how to use the 1DThis however could be added during the beta phase but should not be forgotten.

butlerpd (changes requested):
I noticed there were some significant changes to `SectorSlicer.py` so have now tested that functionality, which is also not working properly.  As I know that code base pretty well now I'll try to review the code soon to make suggestions. However, before I forget I wanted to document the failing functionality:
1. Good news is that all the other slicers seem to be unaffected (not surprising given what was changed but good to know?)
2. when invoking the sector slicer nothing happens until you try moving (graphically) the sliders.
3. the slicer editor changes never lead to an updated plot either, but they do change the values internally. Even clicking the `Apply` button does not cause any new 1D plot. By then manually moving a slider the new plot will appear.

I have not tested the batch slicing which I don't think should be affected as that code does not seem to have been touched?

butlerpd:
Just to dot the i here, the change to `CorFuncSlider` does nothing to fix the functionality based on my retesting of the installer. I.e. the range setting still does not work at all.

Looking at the code more carefully I suspect that the changes to `CorFuncSlider` were an attempt to fix the sliders but not realizing that corfunc is completely separate? Large chunks of code seem to have been copied and pasted from `InversionPerspective.py` and `InversionLogic.py` to `InversionWidget` which I'm pretty sure still uses the standard `src/sas/qtgui/Plotting/QRangeSlider.py` and suspect that something got lost in the move (possibly some initialization?). I suspect therefore that all the changes to `CoreFuncSlider` could (should?) be reverted?

This "losing something in the move" is likely also what has broken the autosuggestion part of the 1D PR? I don't think much of that code was actually changed otherwise? Just a guess.

As for the slicer issue, the present code completely breaks the slicer design (for better or worse) and I'm not sure what the best fix is. We probably need a meeting to understand the goals?  Actually it is not clear to me that the 2D PR shouldn't maybe be integral to the slicer, using the batch slicer mode (which would probably need to be updated/modified)? The actual PR calculation still seems to be all 1D not 2D - just based on a series of cuts rather than the circular average used by default? But again maybe I'm not fully understandting

krzywon (commented):
This still needs work. Here is a list of items that I've found are not working that will need to be addressed before this can be merged.

Non P(r) functionality that requires attention:

- Modifying the sector slicer from the `Slicer Parameters` window no longer updates the 1D slice. The slicer on the 2D plot is updated, but the 1D plot update is not triggered. 
- Plotting of batch Fit results is no longer working with the error `sas\qtgui\Utilities\GridPanel.py", line 209, in onPlot name = data['Filename'][row] ~~~~^^^^^^^^^^^^ KeyError: 'Filename'`
- Save project always throws an error, regardless of what is being saved

P(r) functionality that requires attention:

- After sending data to the perspective, the `number of terms` and `Reg. constant` suggested values are not calculated correctly. For a single P(r), the values are set to 10 and 0, respectively. For batch P(r), the buttons are blank.
- `Calculate All` does not make sense for individual P(r) fits and `Calculate` does not make as much sense for batch P(r). There should only be one button that acts differently for Batch P(r) vs. single P(r) fits.
- Closing the 2D plot that is generated when sending 2D data to the P(r) window breaks a lot of functionality and clicking the `Slice` button will throw errors.
- `Remove` does not remove the data from the combo box and throws an error.
- Sending 2D data to P(r) when Batch mode is selected throws an error, but still creates a P(r) page.
- Attempting to send 1D and 2D data to a batch page throws the inexplicable error  `'<=' not supported between instances of 'NoneType' and 'int'`
- Report results is enabled for P(r), but is not implemented.

src/sas/qtgui/Perspectives/Inversion/DMaxExplorerWidget.py
```diff
@@ -146,6 +146,7 @@ def modelChanged(self, item):
                 # This inversion failed, skip this D_max value
                 msg = "ExploreDialog: inversion failed "
                 msg += "for D_max=%s\n%s" % (str(x), ex)
+                msg = "Please adapt Inversion parameters"
```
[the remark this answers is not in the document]
lucas-wilkins: Done

src/sas/qtgui/Utilities/GridPanel.py
```diff
@@ -201,9 +206,10 @@ def onPlot(self):
         # look for the 'Data' column and extract the filename
         for row in rows:
             try:
-                name = data['Data'][row]
+                name = data['Filename'][row]
```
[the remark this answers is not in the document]
lucas-wilkins: I have no idea why this was changed, nothing ever references a dictionary with that key in any of the files in inversion perspective, as far as I can tell

src/sas/sascalc/pr/invertor.py
```diff
@@ -496,7 +504,7 @@ def lstsq(self, nfunc=5, nr=20):
         # Compute the reg term size for the output
         sum_sig, sum_reg = self._get_reg_size(nfunc, nr, a)
 
-        if math.fabs(self.alpha) > 0:
+        if self.alpha > 0:
```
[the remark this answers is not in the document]
lucas-wilkins: Think negative alphas are incorrect, and the suggested value should never be negative. This is right, I think, @dehoni ?

src/sas/sascalc/pr/invertor.py
```diff
@@ -377,7 +382,7 @@ def pr_fit(self, nfunc=5):
         # Compute chi^2
         res = self.pr_residuals(out)
         chisqr = 0
-        chisq = np.sum(res)
+        chisq = np.sum(math.fabs(res))
 
         self.chisqr = chisqr
```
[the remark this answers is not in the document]
lucas-wilkins: Removed this, because it does nothing anyway!

lucas-wilkins:
@krzywon should all be fixed.

krzywon:
This might be superseded by https://github.com/SasView/sasview/pull/3137. Keeping this open for now, but likely will be closed next week.

rozyczko:
One issue (?) found with the range sliders: moving them on the I(q) chart doesn't update the values in the `Options` tab for Total Q range`.
The other way - when modified in the `Max:` textbox, the range change propagates to the chart

rozyczko:
Another: the widget should not be resizable. There is nothing to be gained by resizing. It only looks bizarre when you maximize this widget.
It's different than the Fitting Perspective, where resizing can help access and view parameters, when the table is large.

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -616,8 +616,22 @@ def updatePerspectiveWithProperties(self, key, value):
         if 'pr_params' in value:
             self.cbFitting.setCurrentIndex(self.cbFitting.findText('Inversion'))
             params = value['pr_params']
-            self.sendItemToPerspective(items[0])
-            self._perspective().updateFromParameters(params)
+            # Make the perspective read the rest of the read data
```
rozyczko, line 619: this conditional contains the same content as the branch above (for fitting) and the common code should be merged

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
```diff
@@ -12,135 +14,213 @@
...
     ######################################################################
     # Base Perspective Class Definitions
 
+
+    @property
+    def supports_reports(self):
+        return True
+        
+    @property
+    def supports_fitting(self):
```
rozyczko, line 206: this should probably be `supports_fitting_menu`

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
```diff
@@ -12,135 +14,213 @@
...
+        return False
+
     def communicator(self):
-        return self.communicate
+        return self.filesWidget.communicate
```
rozyczko, line 210: there's only one instance of `Communicate` - created in `GuiManager`. Both `self.communicate` (from the parent, which is GuiManager) or from `filesWidget` should be the same object. Why change it here?

src/sas/qtgui/Utilities/GridPanel.py
```diff
@@ -215,8 +228,13 @@ def dataFromTable(cls, table):
         assert(isinstance(table, QtWidgets.QTableWidget))
         params = {}
         for column in range(table.columnCount()):
-            value = [table.item(row, column).data(0) for row in range(table.rowCount())]
+            value = list()
             key = table.horizontalHeaderItem(column).data(0)
+            for row in range(table.rowCount()):
```
rozyczko, line 233: avoid lazy try..except.

```python
        for column in range(table.columnCount()):
            value = []
            key = table.horizontalHeaderItem(column).data(0)
            for row in range(table.rowCount()):
                item = table.item(row, column)
                if item is not None:
                    value.append(item.data(0))
                else:
                    value.append(" ")
```

rozyczko (changes requested):
small changes requested for consideration

krzywon (commented):
Code wise, this is getting closer to finished.

Functionality wise, I'm seeing the following:

- the save state is still broken - attempting to save with two tabs throws an error
- the regularization constant and number of terms never seems to be calculated. Maybe it is slow, but that needs to get fixed.
- Some of the sector slicer changes are still present. Updating the slicer parameter window updates the view, but not the 1D data. The 2D P(r), which is now disabled, should really be added to the batch slicer window as a fitting option.

@dehoni - Do you need/want help with any of this? I should be able to find time early next calendar year to work on this.

lucas-wilkins:
Closed as rebased PR now exists: #3137

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
```diff
@@ -12,135 +14,213 @@
...
+                state[data_id] = {'pr_params': tab_data}
+            return state
+        except:
+             return
+            
+
+    def _calculateUpdate(self, params: dict, tab_name):
+        """ Update the perspective using a dictionary of parameters
+        e.g. those loaded via open project or open analysis menu items"""
+        raise NotImplementedError("Update from parameters not implemented yet.")
```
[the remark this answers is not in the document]
lucas-wilkins: can be deleted

src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
```diff
@@ -0,0 +1,1365 @@
...
+    """Converts text input values to floats. Empty strings throw ValueError"""
+    try:
+        return float(value)
+    except ValueError:
+        return 0.0
+
+
+# Default Values for inputs
+NUMBER_OF_TERMS = 10
+REGULARIZATION = 0.0
```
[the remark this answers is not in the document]
lucas-wilkins: 0 "always works" - DH, "SANS" "Scientist"

src/sas/qtgui/Utilities/GridPanel.py
```diff
@@ -51,16 +52,20 @@ def __init__(self, parent=None, output_data=None):
...
         # Fill in the table from input data
         self.setupTable(widget=self.tblParams, data=output_data)
-        if output_data is not None:
-            # Set a table tooltip describing the model
-            model_name = output_data[0][0].model.id
-            self.tabWidget.setTabToolTip(0, model_name)
+        # if output_data is not None:
+        #     # Set a table tooltip describing the model
+        #     model_name = output_data[0][0].model.id
+        #     self.tabWidget.setTabToolTip(0, model_name)
```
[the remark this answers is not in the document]
lucas-wilkins: delete

src/sas/sascalc/pr/calc.py
```diff
@@ -333,8 +335,10 @@ def positive_errors(pars, err, d_max, nslice):
 
     index = pr_val > pr_val_err
     total_pos = np.sum(pr_val[index])
-
-    return total_pos / total  # dx cancels
+    if total == 0:
+        return 0
+    else:
+        return total_pos / total  # dx cancels
```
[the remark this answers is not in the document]
lucas-wilkins: meh

src/sas/sascalc/pr/invertor.py
```diff
@@ -161,21 +161,26 @@ def __setattr__(self, name, value):
             return self.set_y(value)
         elif name == 'err':
             value2 = abs(value)
+            if None in value:
```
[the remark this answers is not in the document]
lucas-wilkins: Fixed in rebase

src/sas/qtgui/Utilities/GridPanel.py
lucas-wilkins: agreed

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
lucas-wilkins: probably changed because parent not set somewhere and was giving errors. Don't know if it is fixed elsewhere. Not sure why this function exists anymore anyway... it works though

src/sas/qtgui/Perspectives/Inversion/InversionPerspective.py
lucas-wilkins: yes, get rid of this.

src/sas/sascalc/pr/calc.py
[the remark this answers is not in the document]
jamescrake-merani: This is a matter of opinion but I'm inclined to agree with the concise version that Jeff has suggested so I'm going to make that change.

PR #2194, opened 2022-09-14
Deleted unused sketch of PanDoc implementation

lucas-wilkins:
The pandoc-docs directory contained very little and appears to be stale. Delete.


krzywon (approved):
I don't see any reason to keep the files. They don't seem to do anything. At worst, they can be recreated if someone wants to attempt what @smk78 was suggesting.

Issue #2182, opened 2022-09-08
check_packages bonks


dsclassen:
I'm trying to install sasview into an existing conda env.

```
(ssp-2.1) [user@machine sasview]# python check_packages.py 
Traceback (most recent call last):
  File "check_packages.py", line 37, in <module>
    with open(file_location, 'r') as stream:
FileNotFoundError: [Errno 2] No such file or directory: './build_tools/conda_qt5_ubuntu.yml'
```

This machine is running Rocky Linux 8.6

Looks like ./build_tools/conda_qt5_ubuntu.yml is missing?

or maybe the check_packages script knows nothing about Rocky Linux?

lucas-wilkins:
Hi, conda_qt5_ubuntu.yml has been removed, and it looks like check_packages.py needs to be updated or removing outright.

Packages are now managed through pip and requirements.txt, i.e. by running
```
pip install -r requirements.txt
```

PR #2195, opened 2022-09-14
Removed check_packages.py

lucas-wilkins:
check_packages.py was associated with the conda system.

It could be updated to work with requirements.txt, however, as it stands it currently just doesn't work. Better to remove than have a broken diagnostic tool. See #2182.

Perhaps there is a reason to keep it, putting this PR is a suggestion really.


krzywon (approved):
This file is superseded by `sas.qtgui.MainWindow.PackageGatherer.py` so deleting it shouldn't be a problem. The work done integrating the `PackageGatherer` class into SasView was removed from release v5.0.5 because of a few issues, so this is a tentative approval and I am asking others to weigh in on this.

wpotrzebowski:
Indeed, there were some issues with PackageGatherer and it didn't make for the release. We probably need to take a look on it again. I am fine with removing check_packages.py as it indeed not useful and probably it doesn't make sense to spend time fixing it (PackageGatherer should be probably prioritized).

wpotrzebowski (approved):

butlerpd:
Indeed if not useful ... there is no point but it was a very useful tool (for me anyway) when it worked and I agree we should prioritize fixing PacakgeGather! Bottom line I also will approve removal though I have not checked to make sure the removal is not breaking anything else.

Issue #2196, opened 2022-09-14
Delete unused files in base directory


lucas-wilkins:
Plan: remove `Vagrant*`, `.pydevproject`, `pytest.ini`, `.project`

Also remove `mac_setup*`, but do it last in case it needs to be rolled back.

Unclear about whether anyone uses `ez_setup.py` these days, @rozyczko - you seem to know about these things.

PR #2197, opened 2022-09-14
2196 delete unused files in base directory

lucas-wilkins:
Remove the files detailed in #2196


krzywon (approved):
I can't speak for the mac homebrew script, but the other changes look fine to me. This is a conditional approval but would need someone else to make a call on homebrew.

wpotrzebowski:
Homebrew script was added in 2017 by one of SNS develepors and I haven't heard that it's been used. I guess it is pretty safe to remove but probably should check with someone from SNS. @mdoucet do you know if https://github.com/SasView/sasview/blob/main/mac_setup_homebrew.sh is still used?

mdoucet:
@wpotrzebowski This can safely be removed. Thanks for checking.

PR #2210, opened 2022-09-16
2103 copy parameters from corfunc gives wrong values

lucas-wilkins:
The highlight of this pull request is that copy parameters from corfunc no longer gives "wrong" values.

I have disabled copy and paste for any perspectives that don't support it, this has required refactoring the copy and paste functionality in fitting. I have also done some tidying up of FittingUtilities but there is still plenty to do there.

As Corfunc never did, or attempted to support copy/paste, having enabled copy buttons was an oversight. 

I'm not implementing anything new in this PR. For a proper implementation I will require a clear specification of what the right values should be. I'm not sure that there is a use case for copying and pasting everything like there is in fitting (note that individual fields can be copy and pasted). This enhancement should be a new issue / discussion.

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -830,7 +830,7 @@ class GuiManager:
         can be saved to the clipboard
         """
         if self._current_perspective is not None:
-            self._current_perspective.on_copy()
+            self._current_perspective.copy()
 
     def actionPaste(self):
         """
@@ -838,7 +838,7 @@ class GuiManager:
         from the clipboard can be used to modify the fit state
         """
         if self._current_perspective is not None:
-            self._current_perspective.on_paste()
+            self._current_perspective.paste()
 
     def actionReport(self):
         """
@@ -869,7 +869,7 @@ class GuiManager:
         can be saved to the clipboard
         """
         if self._current_perspective is not None:
-            self._current_perspective.on_excel_copy()
+            self._current_perspective.excel_copy()
 
     def actionLatex(self):
         """
@@ -877,7 +877,7 @@ class GuiManager:
         can be saved to the clipboard
         """
         if self._current_perspective is not None:
-            self._current_perspective.on_latex_copy()
+            self._current_perspective.latex_copy()
 
     def actionSaveParamsAs(self):
         """
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -5,7 +5,7 @@ import json
 import webbrowser
 import traceback
 
-from typing import Optional
+from typing import Optional, Dict
 
 from PyQt5.QtWidgets import *
 from PyQt5.QtGui import *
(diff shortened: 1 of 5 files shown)


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -3878,60 +3878,95 @@ def getParameterDict(self):
                 line_dict[content[0]] = content[1:]
         return line_dict
 
-    def onCopyToClipboard(self, format=None):
+    def copy(self):
```
rozyczko, line 3881: I don't think this naming is explanatory.
If I see `widget.copy()` I expect this to return me a shallow copy of the widget object.
Maybe `copyToClipbard` is a better name, after all

rozyczko: Likewise, other methods, which names you modified.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -382,7 +380,16 @@ def perspectiveChanged(self, new_perspective_name: str):
         # but new_perspective is of type Perspective, thus call to Perspective members are safe
         new_perspective = self.loadedPerspectives[new_perspective_name]
 
+        # Report options
         self._workspace.actionReport.setEnabled(new_perspective.supports_reports)
+
+        # Copy paste options
+        self._workspace.actionCopy.setEnabled(new_perspective.supports_copy)
+        self._workspace.actionLatex.setEnabled(new_perspective.supports_copy_latex)
```
rozyczko, line 388: Aren't the specialised options redundant? If you can copy parameters, you can surely convert them into latex and excel.
Using only `perspective.supports_copy` is enough to toggle the state of all these actions, I think.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -821,16 +829,16 @@ def actionCopy(self):
         Send a signal to the fitting perspective so parameters
         can be saved to the clipboard
         """
-        self.communicate.copyFitParamsSignal.emit("")
-        #self._workspace.actionPaste.setEnabled(True)
-        pass
+        if self._current_perspective is not None:
```
rozyczko, line 832: good idea to replace signaling with direct call. Please update the method comment, otherwise it is confusing.
Same with the other methods below.

rozyczko: Also, please remove the signal definitions from GUI Utilities

rozyczko (approved):
Good to go.
A few stylistic comments.

src/sas/qtgui/MainWindow/GuiManager.py
lucas-wilkins: They have significantly different content, and its very contextual. Over generalising at this level will cause issues, exactly the kind of issues that I've spent time fixing in this PR. Need to treat things that are different as different.

src/sas/qtgui/MainWindow/GuiManager.py
lucas-wilkins: Will do.

Issue #2170, opened 2022-08-24
Compatibility with newer matplotlib (3.5.x)


llimeht:
**Describe the bug**
SasView is not compatible with currently supported versions of matplotlib such as the current 3.5.4 release. One symptom of this is seen in #1981 where the GUI closes during initialisation. There are others. Some of these might be picked up by `GUITests.py` but as they don't pass (#1732) their failures on CI don't help spot these issues in advance.

**To Reproduce**
1. Install matplotlib ≥ 3.5 and sasview from source (version 5.0.5 or current git main)
2. Start SasView
3. It crashes

SasView's attempts to handle the exceptions obfuscate where it is from:
```
14:06:41 - ERROR: Application failed with: 'Arrow3D' object has no attribute 'do_3d_projection'
14:06:41 - ERROR: Traceback (most recent call last):
  File "/usr/bin/sasview", line 33, in <module>
    sys.exit(load_entry_point('sasview==5.0.5', 'console_scripts', 'sasview')())
  File "/usr/lib/python3/dist-packages/sas/qtgui/MainWindow/MainWindow.py", line 119, in run_sasview
    mainwindow.guiManager.showWelcomeMessage()
AttributeError: 'MainSasViewWindow' object has no attribute 'guiManager'
```
While the error looks to be in `MainWindow.py` the issue is actually that `sas/qtgui/Plotting/Arrow3D.py` needs to be updated to the new 3.5 API.

Poking matplotlib compatibility further, `sas/qtgui/Plotting/Plotter2D.py` also seems to need an update for the matplotlib API as `colorbar.update_bruteforce` [has now been removed](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.5.0.html) (and had been deprecated for some years). 

**SasView version (please complete the following information):**
 - Version: 5.0.5; relevant files are the same in git main.

**Operating system (please complete the following information):**
 - OS: above tests done on Linux, but also an issue on Windows as seen in #1981.

**Additional context**

The following patch at least allows sasview to run with mpl 3.5.4. I can't attest to the correctness of what it is doing to the Arrow3D class beyond (a) it let the main GUI actually render, and (b) the general scattering calculator, which is where this class is used, still displayed some arrows on the plots.
```
diff --git a/src/sas/qtgui/Plotting/Arrow3D.py b/src/sas/qtgui/Plotting/Arrow3D.py
index 2c58bc2..2adf919 100644
[...]

PR #2213, opened 2022-09-21
Fix compatibility with matplotlib 3.5.x

llimeht:
This patch attempts to fix compatibility with matplotlib 3.5.x by updating the code to avoid functions that were deprecated (in some cases for years) and have now been removed.

Merging this patch at least allows the sasview GUI to start with matplotlib 3.5; additional places where code needs to be updated may well be uncovered by automated or manual testing.

Closes: #2170

### DIFF
--- a/src/sas/qtgui/Plotting/Arrow3D.py
+++ b/src/sas/qtgui/Plotting/Arrow3D.py
@@ -2,6 +2,7 @@
 Module that draws multiple arrows in 3D coordinates
 """
 
+import numpy as np
 from matplotlib.patches import FancyArrowPatch
 from mpl_toolkits.mplot3d import proj3d
 import time
@@ -21,7 +22,7 @@ class Arrow3D(FancyArrowPatch):
         :Params colors: [[R0, G0, B0], [R1, G1, B1], ...]
         where R, G, B ranges (0,1)
         """
-        FancyArrowPatch.__init__(self, (0, 0), (0, 0), *args, **kwargs)
+        super().__init__((0, 0), (0, 0), *args, **kwargs)
         self.leftdown = False
         self.realtime = False
         self.t_click = 0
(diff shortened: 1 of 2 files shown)


lucas-wilkins:
In case you've not noticed, the tests for all your PRs are failing, not sure why this is.

```Error: Error 403: Resource not accessible by integration```

llimeht:
> In case you've not noticed, the tests for all your PRs are failing, not sure why this is.
> 
> `Error: Error 403: Resource not accessible by integration`

Yes, I saw that. I _think_ it's due to the PRs coming from my own fork of sasview and so the upload of release artifacts of the installer to SasView can't happen via `ncipollo/release-action@v1`. I have no idea why we'd be making a _release_ out of every single git push, however... all that is doing is polluting the [releases page](https://github.com/SasView/sasview/releases) with lots of weird stuff to which users should never be subjected.

Building an installer to test can (and does!) work via `actions/upload-artifact@v2`. Those installers are already available in the artifacts summary of the actions for each run, without making a release. For example: [https://github.com/SasView/sasview/actions/runs/3113602635](https://github.com/SasView/sasview/actions/runs/3113602635)

I suspect the correct solution is to remove the creation of releases from everything other than the "release" workflow, and spend quite some time deleting all those fake releases that have appeared in the list.

lucas-wilkins (approved):
Looks OK on the face of it, though slightly concerned with the failing tests.

llimeht:
> Looks OK on the face of it, though slightly concerned with the failing tests.

No **_tests_** failing, just the upload of the installer image to the sasview release page.

In poking the GUI tests, I suspect that they were picking up the incompatibility with matplotlib 3.5 but since they were already broken and not gating CI, no-one had noticed.

lucas-wilkins:
@llimeht We merged in the change that should allow this to work, but you'll need to merge our main (or your delta) into each of your PRs.

llimeht:
@lucas-wilkins it's not actually merged yet.

rozyczko (commented):
Code looks good to me.
Works well on windows with mpl=3.5.2

wpotrzebowski:
Closing because changes were part of #2233

PR #2217, opened 2022-09-25
Fix incorrect escape sequences

llimeht:
Embedded latex ends up with lots of backslashes that need to be protected as raw strings rather than (mis)interpreted as standard C-style escape sequences.


pkienzle (approved):
Code looks fine but I didn't run it to verify. It is tempting to replace these with f-strings but that would be scope creep.

rozyczko:
Also seems to behave properly, at least on Windows.

PR #2219, opened 2022-09-26
Stop making a new release on every push

llimeht:
As described in #2213, we are in the very strange situation of making a new release out of every single git push, and they are all published on the releases page. That is polluting the [releases page](https://github.com/SasView/sasview/releases) with lots of weird stuff ~~to which users should never be subjected~~ (Edit: not visible to not logged in users, fortunately; but still not pleasant!).

Building an installer to test can (and does!) work via the actions. They are published as artifacts of the particular action run and are visible under the "Summary" of the action once it has run. Making a release isn't needed to make those installers available For example: [https://github.com/SasView/sasview/actions/runs/3113602635](https://github.com/SasView/sasview/actions/runs/3113602635)

I don't think we really want to have releases made for every single git push so this PR removes that step from the standard GitHub Actions. I think someone also needs to do the manual work of deleting all the old ones out of the releases page... I don't know if that can be automated or even made less tedious.

### DIFF
--- a/.github/workflows/installers.yml
+++ b/.github/workflows/installers.yml
@@ -194,27 +194,3 @@ jobs:
         name: SasView-Installer-${{ matrix.os }}-${{ matrix.python-version }}-${{ github.sha }}
         path: installers/dist/SasView5.dmg
         if-no-files-found: error
-
-    - name: Declare env variables on push only
-      if: github.event_name == 'push'
-      shell: bash
-      run: echo "BRANCH_NAME=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-
-    - name: Declare env variables on pull_request only
-      if: github.event_name == 'pull_request'
-      shell: bash
-      run: echo "BRANCH_NAME=$GITHUB_HEAD_REF" >> $GITHUB_ENV
-
-    - name: Upload installer to GitHub releases
-      #Release and main branch will be treated differently
-      if: env.BRANCH_NAME != 'main'
-      uses: ncipollo/release-action@v1
-      with:
-        draft: true
-        prerelease: true
-        allowUpdates: true
-        replacesArtifacts: true
-        token: ${{ secrets.GITHUB_TOKEN }}
-        artifacts: "installers/dist/SasView5.dmg, installers/Output/setupSasView.exe"
-        name: ${{ env.BRANCH_NAME }}
-        tag: ${{ env.BRANCH_NAME }}


lucas-wilkins:
The `gh` command line tool can be used to do batch things on github: https://cli.github.com/

llimeht:
> The `gh` command line tool can be used to do batch things on github: https://cli.github.com/

Indeed... the following would do it, without the obvious thing in there that disables it being used accidentally...:

```gh release list -R SasView/sasview --limit 1000 | grep Draft | cut -f1 | xargs -n1 echo gh release delete```

lucas-wilkins:
> They are published as artifacts of the particular action run and are visible under the "Summary" of the action once it has run

Where are they? I can't find them except through the releases page.

krzywon:
> > They are published as artifacts of the particular action run and are visible under the "Summary" of the action once it has run
> 
> Where are they? I can't find them except through the releases page.

[Start here](https://github.com/SasView/sasview/actions/workflows/installers.yml) and click on the particular push/merge request you are interested in. At the bottom of the screen are compressed files, one for mac and one for windows.

butlerpd:
LOLOL  ... Indeed --- that was the reason we specifically added these to the release page in the first place. The problem was that the default place the builds are located turns out to be far too obscure for most of the folks doing the functionality testing.  It was admittedly ugly but the best solution that was found at the time. 

My understanding is that these are never made to be actual releases but are pushed to the release page in a way that is **only** visible to logged in users? Otherwise they are hidden?  I also thought it was only for pushes to a branch with a pull request in, but I could be wrong (that is really where it was needed)?  The idea was to also remove older installers from a branch when a new one was created but, if memory serves, nobody could figure out a way to do that in the action scripts? So maybe relying aspirationally on "someone" going through every so often and deleting?

A more elegant solution would be great. Ideally one would like to push pull request installers to a web page accessible to the general public for testing in a way that clearly separates those from different pull requests.  something like the old Jenkins functionality I guess?

In the meantime, there probably needs to be a bit more discussion with the broader "contributor" community before merging this?

llimeht:
> In the meantime, there probably needs to be a bit more discussion with the broader "contributor" community before merging this?

Definitely!

I suspect the answer is education on how to find the artifacts, not throwing things into the release page. It's the same number of clicks to get to relevant installer either way (click on the "installer" test details for the installer, then click on "summary"), just need to know which clicks. 

The current situation of trying to make a release on every PR has the unfortunate effect of requiring every PR to come from within the SasView/sasview git repository and therefore only from people added to the SasView project. Anyone making a PR from their own fork will get failing CI and therefore get [comments about being unsure whether to merge because the CI is failing](https://github.com/SasView/sasview/pull/2213#pullrequestreview-1120204674). Requiring that PRs come only from within the SasView/sasview repo is a very unusual workflow; requiring that PRs come only from SasView project members isn't the desired messaging to potential new contributors.

wpotrzebowski:
Yes, as @butlerpd  mentioned managing releases on the webpage requires some human maintainace, which I do from time to time but it is not optimal. It would be great if someone can come with better mechanism for this. We need something that is cross-platform and something that goess through all steps (including pyinstaller). As I experienced many times installer files fail even if all tests pass.

llimeht:
An interesting internship project (Google Summer of Code?) would be to create something that installs using the artifacts and tests they work with a GUI testing framework...

wpotrzebowski (approved):

PR #2221, opened 2022-09-26
2140 A more maintainable setup.py (WIP)

lucas-wilkins:


llimeht:
I was curious to see the `setup.py` changes but I see they're entangled in rather a lot of other changes. I assume the plan is to land some other PRs and then rebase this work?

(The reason I was keen to look at the `setup.py` changes was that I was rather hoping that it would actually go away to be instead replaced by a PEP517 builder with a declarative config file rather than an executable one...)

lucas-wilkins:
> I was curious to see the `setup.py` changes but I see they're entangled in rather a lot of other changes. I assume the plan is to land some other PRs and then rebase this work?

Yeah, its based on #2164 which is not yet merged.

> (The reason I was keen to look at the `setup.py` changes was that I was rather hoping that it would actually go away to be instead replaced by a PEP517 builder with a declarative config file rather than an executable one...)

I was just going to make it so that you didn't have to manually change entries in it whenever you rename a file. There are reasons why the executable will need to remain, for now. Maybe I'll raise this today at the call.

lucas-wilkins:
Closing this for now. Going to make a discussion topic about it.

Issue #2218, opened 2022-09-26
Remove duplicate images dir


lucas-wilkins:
We have two copies of `images`, really should only have one.

lucas-wilkins:
Done

PR #2227, opened 2022-09-30
Removed images dir, addresses #2218

rozyczko:
The only location using the old path was in the pyinstaller spec file (from what I found).
Fixed that.


lucas-wilkins:
This should close #2218

Issue #2249, opened 2022-10-12
Weird Scaling on Some Windows Installations


lucas-wilkins:
![image](https://user-images.githubusercontent.com/9784906/195374443-fafec342-e6c5-4134-88cb-f703800f8193.png)

This ^

smk78:
This is what I just got on a x64/W10 local build (pulling repos at 16:15 on CodeCamp Thu):
![image](https://user-images.githubusercontent.com/10938679/198309463-2bfb43cf-1783-4683-a919-c52bd8242ff6.png)
The **Load data**, **Send data to** and **Fitting** buttons seem to have abnormal fonts?

The display resolution is 1920x1080.
pip list shows: matplotlib 3.4.3, PyQt5 5.15.7, QtPy 2.1.0; python 3.9.12

lucas-wilkins:
As fixed as it is ever going to be in Qt5

PR #2250, opened 2022-10-12
Fixed scaling issues

lucas-wilkins:
Fixed scaling issues by setting a default font size.

### DIFF
--- a/src/sas/qtgui/MainWindow/MainWindow.py
+++ b/src/sas/qtgui/MainWindow/MainWindow.py
@@ -91,7 +91,7 @@ def run_sasview():
     splash = SplashScreen()
     splash.show()
     app.setAttribute(Qt.AA_EnableHighDpiScaling)
-    app.setStyleSheet("* {font-size: 11pt}")
+    app.setStyleSheet("* {font-size: 9pt}")
     # Main application style.
     #app.setStyle('Fusion')
 
--- a/src/sas/qtgui/MainWindow/MainWindow.py
+++ b/src/sas/qtgui/MainWindow/MainWindow.py
@@ -91,6 +91,7 @@ def run_sasview():
     splash = SplashScreen()
     splash.show()
     app.setAttribute(Qt.AA_EnableHighDpiScaling)
+    app.setStyleSheet("* {font-size: 11pt}")
     # Main application style.
     #app.setStyle('Fusion')
 


wpotrzebowski:
@lucas-wilkins is it supposed to be changing all text except plots? This is at least what I see on Mac when testing with Fitting perspective.

lucas-wilkins:
@wpotrzebowski it sets the default font size, so that for any component that doesn't explicitly set the font size will have a font size of 11pt.

It looks right on my machine, I guess the question that needs to be answered by you or another mac person, is whether it looks wrong on MacOS?

wpotrzebowski:
The size of general font is fine but it looks quite strange depending on what screen do you work. See images bellow

wpotrzebowski:
On laptop
Screenshot 2022-10-13 at 12 53 09

wpotrzebowski:
On external screen:
Screenshot 2022-10-13 at 10 33 37

wpotrzebowski:
We've been having issues with this on Mac for long while... There is also extreme case when one works with external screen and lid open. Then legend on plots can become huge. Unfortunatelly I am not able to test it now.

lucas-wilkins:
@wpotrzebowski Ah, I see. That's a different issue to the one this PR is supposed to fix (see the image #2249), which is purely Qt problem that needs a workaround.

Your problem is probably relevant to discussion #2244 

If it doesn't break anything for you, I think its good to go.

butlerpd:
I have not had a chance to look at this but I have been thinking for a while that our default font size seems to be a bit large which may be a good thing from the point of accessibility but does make it hard to work with on a laptop screen. 11 point (depending on the font) sounds reasonable. If we figure out it is too small or too big we can change the one line that is added here to tune it?

I note it is also a single line change to the style sheet so don't see there is much risk?

lucas-wilkins:
@butlerpd Also, I took the 11pt value from some of the buttons where it is explicitly set.

Annoyingly, if you want a different font size, you'll have to go through all the widgets that have their font size set, e.g. the load and send buttons.

rozyczko:
maybe then font size should be a user-settable property? In the config?

butlerpd:
I like the idea of user settable size a lot, though that sounds like a big project.  To @lucas-wilkins point, as I recall things like the "LOAD" button were deliberately set in a larger font (though as I recall there was some discussion about how big it should be).  Perhaps what is needed is for all fonts to be defined in % increments from a base, user settable, font?  At this point though this probably should be moved to a discussion?

lucas-wilkins:
> At this point though this probably should be moved to a discussion?

Sure. Will someone accept this PR though, please.

butlerpd:
Ok. I downloaded the installer and tested on windows and see some really bizare behavior in fitting that will need addressing, but I don't think is linked to this PR. so I will not worry about it here.

I am however now really confused.

First off, I don't understand how you can get the image you posted in #2249 @lucas-wilkins?!!  both directions of the fit panel (and the data explorer panel) on my windows machine have a minimum size that absolutely does not allow scrunching down to anything that small. Secondly the fonts are opposite of what you see: `Load data`, `Send data to` and `Fitting` are actually BIGGER than all other text not smaller as in the image you post.  This is true of 5.0.4 and all computers I've seen it on.

What I see that your PR does compared to 5.0.5 and before is make the text from the three buttons above smaller (not bigger as would be presumably needed from the image you post) and now the same size as all other fonts? ..... EXCEPT this weird thing in the Fitting perspective where the `Value` column in the Model tab. That font is quite a bit smaller than the rest now (including the text in the adjacent `Paramater' and `Error` (or `Min`) columns. When clicking on the Value to change it, the font increases to match the rest of the fonts but then goes back to it small size when hitting enter.

The only other thing I see, which may be the same in 5.0.5 is that the Corefunc perspective, under "Lamellar Parameters" the top line are truncated on the left!! (both columns) as is the 2nd line (1rst column). So, for example line 1 column1 reads:

> Avg. Thicknesses: rd Block: (text box).

 actually part of an "a" is showing in front of "rd"  Again.... I suspect this is not to do with this PR but with a previous on that maybe @smk78 did not notice? (the whole layout is different from 5.0.5).

All this to say, I'm not comfortable merging this from my testing alone without understanding what is supposed to be happening?  Can anybody else test this?

rozyczko:
This is what I get. 1920x1080, 100% font size settings. Visibly different font on charts and in fitting parameter values. 
Also, some widget labels overflow.
Things get worse when switched to 125% or 150% font size.

fonts

lucas-wilkins:
Arguably the font size should be 9, not 11.

lucas-wilkins:
> I don't understand how you can get the image you posted in https://github.com/SasView/sasview/issues/2249 @lucas-wilkins?!!

I just run sasview, that's how it has always looked on my machine.

I say always, maybe 50% of the time it looks like that. Sometimes its normal, sometimes the buttons are so small you can bearly read them. By always, I mean it is a problem that has existed since I started working on SasView - I did put an issue for it at the time, though it appears to have disappeared. It's just got to the point where I hate it so much I want to do something about it.

> Secondly the fonts are opposite of what you see: Load data, Send data to and Fitting are actually BIGGER than all other text not smaller as in the image you post. 

I see, that's why 11pt is the wrong size. They're the only ones with a font size set and I got it from there, thinking that's the size it is supposed to be. Should probably be 9.

> on my windows machine have a minimum size that absolutely does not allow scrunching down to anything that small.

Nothing is small, everything is giant, except the load/send buttons, which stay the same size.

lucas-wilkins:
@butlerpd How about now?

lucas-wilkins:
@rozyczko I have no idea why it would be affecting the charts. Is it possible that they have always been different, and its just that nobody has drawn attention to it until now?

rozyczko:
This looks much better now. At 100% and 125% I can see almost no issues, at 150% some label clipping occurs but this might have already happened with the current code.
I would propose to merge it in its current state but add another task to expose font size in the config file and make sure the sizing is consistent everywhere (labels, tables etc.)

rozyczko (approved):

wpotrzebowski:
9pt for me is too small:
Screenshot 2022-10-14 at 09 07 39
However I am fine with merging it and making font size configurable (as discussed).

wpotrzebowski (approved):

lucas-wilkins:
@wpotrzebowski @butlerpd Perhaps you guys could measure the font size that you think it should be (i.e. in main on your machines) and we can take a data driven approach.

wpotrzebowski:
I would prefer 11 for my setup on Mac but this seems to contradicts windows requierments

lucas-wilkins:
> I would prefer 11 for my setup on Mac but this seems to contradicts windows requierments

I also prefer 11, but it is a bit bigger than some of the widgets were layed out for...

*sigh* Thought this would be an easy fix!

rozyczko:
`if macos then fontsize=11 else fontsize=9` :) 

(no, please don't)

krzywon:
With the range of fonts wanted and the cascading of sizes to other elements, this sounds like a good use of our new configuration system (#2168) and (upcoming) preferences panel (#2167). This could become a project where we fully move to a user-modifiable stylesheet but this will likely require removing inline styles and ensuring all panels have the sheet applied.

wpotrzebowski:
Should we put it on hold until we have a proper solution?

krzywon:
> Should we put it on hold until we have a proper solution?

I'm working on a proper solution in #2258

lucas-wilkins:
Closing this because its got out of hand.

PR #2258, opened 2022-10-24
2249: Create a global stylesheet and allow user-built themes

krzywon:
This PR is creates a centralized style sheet for all GUI widgets, has the start of two other themes that are currently disabled (Dark and Large Font), and allows for user-created themes.

Current functionality:
- Sets a single style sheet for all GUI elements and alternates to modify background colors, fonts, etc.
- Built-in style sheets are not (easily) accessible by the user
- Replaced all `setStyleSheet`, `setFont` instances with a standardized GuiUtil method
- Added the option for users to add their own themes
- Added a preference to select a theme from the available options

Future features:
- Document the system
- User documentation for creating themes
- Finish dark and large-font themes (and any others?)


src/sas/system/style.py
```diff
@@ -0,0 +1,73 @@
...
+            pass
+
+        for filepath in places_to_look_for_conf_file:
+            if os.path.exists(filepath):
+                self.style_sheet = filepath
+                return
+        print(f"'{filename}' not found.", file=sys.stderr)
+        self.style_sheet = None
+
+    def __call__(self, *args, **kwargs):
```
lucas-wilkins, line 67: Why use __call__ like this, its a pretty non-standard use of this dunder

src/sas/system/style.py
```diff
@@ -0,0 +1,73 @@
+import os
+import pkg_resources
+import sys
+
+from sas.system import user
+
+DEFAULT_STYLE_SHEET_NAME = 'sasview.css'
```
lucas-wilkins, line 7: put in config?

src/sas/system/style.py
```diff
@@ -0,0 +1,73 @@
...
+class StyleSheet:
+    """
+    The directory where the per-user style sheet is stored.
+
+    Returns ~/.sasview/sasview.css, creating it if it does not already exist.
+    """
+    def __init__(self):
+        self.style_sheet = None
+        self._find_style_sheet()
+        self.save()
```
lucas-wilkins, line 19: .save contains ._find_style_sheet

src/sas/system/style.py
lucas-wilkins: also just call STYLE_SHEET if its there?

src/sas/system/style.py
```diff
@@ -0,0 +1,73 @@
...
+import sys
+
+from sas.system import user
+
+DEFAULT_STYLE_SHEET_NAME = 'sasview.css'
+
+
+class StyleSheet:
+    """
+    The directory where the per-user style sheet is stored.
```
lucas-wilkins, line 12: Text needs updating

src/sas/system/style.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Returns ~/.sasview/sasview.css, creating it if it does not already exist.
+    """
+    def __init__(self):
+        self.style_sheet = None
+        self._find_style_sheet()
+        self.save()
+
+    def style_sheet_filename(self):
```
lucas-wilkins, line 21: make private?

src/sas/system/style.py
```diff
@@ -0,0 +1,73 @@
...
+        self.style_sheet = os.path.join(user_dir, DEFAULT_STYLE_SHEET_NAME)
+
+    def save(self):
+        self._find_style_sheet()
+        sheet = self()
+        self.style_sheet_filename()
+        with open(self.style_sheet, 'a') as file:
+            file.write(sheet)
+
+    def _find_style_sheet(self, filename=DEFAULT_STYLE_SHEET_NAME):
```
lucas-wilkins, line 33: if this does come from config, this would have to be replaced with an optional because it keyword defaults are evaluated at compile time

src/sas/system/style.py
```diff
@@ -0,0 +1,73 @@
...
+        '''
+        The style sheet is in:
+        User directory ~/.sasview/
+        Debug .
+        Packaging: sas/sasview/
+        Packaging / production does not work well with absolute paths
+        thus the multiple paths below
+        '''
+        self.style_sheet_filename()
+        places_to_look_for_conf_file = [
```
lucas-wilkins, line 43: can we look in less places?

lucas-wilkins (commented):
I think there needs to be a bit of thought about how this will work if we want, say, the ability to change the font size as a user-friendly setting.

pkienzle (commented):
I don't think you want a user-controlled style sheet. I expect that old .css files will be a support headache down the line.

Instead I suggest putting the default style sheet as a resource much like an image file. We can document its location so that interested users can override it, but their changes will be gone with the next install.

butlerpd:
Should this be a draft?

krzywon:
> Should this be a draft?

No.

butlerpd:
OK - So looking for a final positive review then.

krzywon:
There are now 4 themes. This should be ready for a full review.

butlerpd:
Actually could be part of 6.0 .. but maybe a lower priority? Should take a look at

rozyczko:
hmmm these look like a downgrade from the existing style.
The default stylesheet looks pretty bad now.
Checkboxes look less intuitive.
Font clarity is much worse.

NEW:
sasview_sheets

OLD:
sasview_sheets_old


NEW:
sasview_sheets2

OLD:
sasview_sheets2_old

lucas-wilkins:
Pyside6 now natively supports dark and light themes, inherited from the OS.

PR #2259, opened 2022-10-25
Drop debugpy from installer

llimeht:
debugpy isn't used anywhere in the sasview code or in the documentation/examples; bundling it into the installer requires additional fragile workarounds, so if it's not needed, can we drop it?


wpotrzebowski (approved):
Tested on Mac installer and it worls fine

PR #2261, opened 2022-10-25
Add cli option to installer to skip licence dialogue

llimeht:
A tiny bit of code within the installer configuration allows the installer to be run as a completely non-interactive installer allowing for later automation:

```
./setupSasView.exe /SP- /VERYSILENT /SUPPRESSMSGBOXES /NORESTART /ALLUSERS /SKIPLICENSE=true /LANG=english /NOICONS /TASKS='' /LOG=install.log /DIR=C:/somewhere/
```

Automated testing of the Windows installer needs this to land first.


rozyczko (approved):
Code looks good, needs checking in the working installer.

pbeaucage:
It works in a Windows installer, but worth noting that the flag has to be ` .\setupSasView.exe /skiplicense=true` with the `/`, simply `.\setupSasView.exe skiplicense=true` does not work.

llimeht:
Yes, all options need a / in front of them as is the windows style.

the /SKIPLICENSE option won't be in that list at all I suspect - that's all the generic command line options for the installer  I don't think there's a way of changing that list at all form what `iscc` provides.

The default set of command line options are documented at: https://jrsoftware.org/ishelp/index.php?topic=compilercmdline

PR #2263, opened 2022-10-25
Rework CI to be non-repetitive, simpler, test more things

llimeht:
This is a preview of some work to update the GitHub Actions, aiming to:

- make them less repetitive
- hopefully easier to maintain
- updated to the latest versions of the helper-actions they call
- able to be extended to do other testing more easily

This is only in a draft state, as this work is blocked by the following things:

- merging #2252 to check GUI tests
- merging #2259 to unbreak the installer on Linux
- merging #2260 to unbreak logging on Linux
- merging #2261 to allow the Windows installer to be automated

(There's also another PR for pyinstaller that is queued behind #2259)

In the last steps of these workflows, I started to experiment with actual functional testing of the GUI and taking screenshots of the GUI in action. That's not working at the moment. I intend to back out that work and return to it later, in a separate PR.


llimeht:
I've split the functional GUI testing off. It's now ready for some review and discussion (but #2260, #2261, #2281 should go in before merging)

installers/sasview.spec
```diff
@@ -4,25 +4,25 @@ import sys
 from pathlib import Path
 import warnings
 import platform
-import sys
+import sysconfig
 
 block_cipher = None
-PYTHON_LOC = sys.exec_prefix
+PYTHON_PACKAGES = sysconfig.get_path('platlib')
```
pkienzle, line 10: Maybe `site_package_path = ...`? The name `PYTHON_PACKAGES` sounds like a list of package names. `PYTHON_PACKAGE_PATH` would also work.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,482 @@
...
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install wheel setuptools
+
+          python -m pip install numpy scipy==1.7.3 pytest pytest-mock pytest_qt
+          python -m pip install tinycc h5py pyparsing html5lib reportlab==3.6.6 pybind11 appdirs
+          python -m pip install six numba mako ipython qtconsole xhtml2pdf pylint
+          python -m pip install qt5reactor periodictable uncertainties dominate importlib_resources
+          python -m pip install html2text
```
pkienzle, line 92: Perhaps out of scope but reproducible builds should pin package version numbers. Because we do significant manual testing prior to release we probably don't want packages changing out from under us. Automated testing should happen against the pinned versions and against the latest versions so we don't get too out of date.

You could do this with a pair of `requirements.txt`. These even allow os-dependent installs ([stackoverflow link](https://stackoverflow.com/questions/29222269/is-there-a-way-to-have-a-conditional-requirements-txt-file-for-my-python-applica/35614580#35614580)).

llimeht: Yes, that particular can of worms is out of scope for this. The small amount of pinning that is already here has pin conflicts that need to be forced/ignored.

installers/sasview.spec
llimeht: That might be a better name yes, but that was a local copy of #2281 that I needed to hold on to deal with platform specific goo; since #2281 has been merged it will actually disappear from this PR.

.github/workflows/matrix.py
```diff
@@ -0,0 +1,89 @@
...
+#   dynamically linked by pyinstaller.
+#   https://pyinstaller.readthedocs.io/en/stable/usage.html#making-gnu-linux-apps-forward-compatible
+os_release_list = [
+    'ubuntu-18.04',
+    'windows-latest',
+    'macos-latest',
+]
+
+# List of OS images to use for release tests
+os_list = os_release_list + [
```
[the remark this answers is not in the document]
llimeht: That's reasonable; have updated. 

I also figured it was reasonable to explain a bit more what `matrix.py` was doing and why, so I've added some more comments in there to help!

pkienzle (approved):
Code looks fine. Artifacts from a recent build are [here](https://github.com/SasView/sasview/actions/runs/3333616518). The Mac installer runs sasview 5.0.5 as expected.

I don't know why the dmg is wrapped in a zip file. Presumably this reproduces the existing behaviour. Feel free to skip the zip or add a ticket to suggest doing so.

pkienzle:
Re: zip file, I suspect it is created by github to bundle the build artifacts from a workflow, so no need to change anything.

wpotrzebowski (approved):
Good job!
The code looks good. One questions/suggestion I have is related to how we install packages and whether it can be done through `pip install -r requirements.txt` rather then listing them explicetly?

The mac installer works fine, however linux installer failed on ubuntu-20.04

```
  File "sasview.py", line 15, in <module>
  File "sas/qtgui/MainWindow/MainWindow.py", line 129, in run_sasview
  File "sas/qtgui/MainWindow/MainWindow.py", line 44, in __init__
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "sas/qtgui/MainWindow/GuiManager.py", line 27, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "sas/qtgui/Utilities/GuiUtils.py", line 45, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "sasmodels/sasview_model.py", line 30, in <module>
    from . import weights
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "sasmodels/weights.py", line 11, in <module>
    from scipy.special import gammaln  # type: ignore
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "scipy/__init__.py", line 153, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "scipy/_lib/_ccallback.py", line 1, in <module>
  File "_ccallback_c.pyx", line 210, in init scipy._lib._ccallback_c
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "ctypes/__init__.py", line 7, in <module>
ImportError: libffi.so.6: cannot open shared object file: No such file or directory
[29967] Failed to execute script 'sasview' due to unhandled exception!

```

wpotrzebowski:
I accidently approved. I guess it may still require some work on linux side?

butlerpd:
Windows installer (#5) works well for me. Tested my standard 1D fitting which worked perfectly. The only issue was that I could not run the python shell which gave the following error:

> FileNotFoundError: [WinError 3] The system cannot find the path specified: 'c:\\SasView-5.0.6a\\debugpy\\_vendored'

I suspect that is not related to this PR but more to something done earlier (and now reversed)?

llimeht:
> The mac installer works fine, however linux installer failed on ubuntu-20.04

Yes, that's a new failure that has appeared during code camp. I'm not sure if that's due to lack of CI or some other external change.

llimeht:
@wpotrzebowski 
> Good job! The code looks good. One questions/suggestion I have is related to how we install packages and whether it can be done through `pip install -r requirements.txt` rather then listing them explicetly?

Perhaps. That's a can of worms in itself and can be a separate PR. The objective of this is to do what we already doing but in a simpler fashion. I'm trying avoid scope creep and this exploding into much larger and longer discussions.

llimeht:
@wpotrzebowski: linux installer updated; looks like some library changes that pyinstaller wasn't quite catching up with, so a little tweak with a sover symlink farm there solves it.

@butlerpd: added in the `debugpy._vendored` hidden import again; that _might_ be enough to the console working, but I have a recollection that it wasn't quite enough in the past. 

(And now I've got to include #2326 in commits for this PR otherwise the test suite fails)

wpotrzebowski:
@llimeht thanks for fixes!  

I will test the installer.

I think we don't need `debugpy` installed explicitily via pip (it is pulled by `qt_consoloe`). It is probably enough to add debugpy._vendored to the spec file but I am fine with the current version too.

butlerpd:
Not sure I understand (OK .. I am sure I don't understand:-)  but testing the installer on windows I still see the same error and indeed the path to: `C:\SasVIew\debugpy` does not exist (i.e. there is no folder `debugpy` in the SasVIew folder.  Afraid I cannot be of much more help. Maybe @wpotrzebowski has a thought?

wpotrzebowski:
The linux install installer works now besides python shell editor (debugpy._vendored error as reported by @butlerpd ). Not sure if it helps in this case but should we tell pyinstaller to explicetly add it to installer: ` datas.append((os.path.join(PYTHON_PACKAGES, 'debugpy'), 'debugpy'))` as we do it currently for Mac?

wpotrzebowski:
MacOSX installer seems to work fine (in sasme cases even better) with latest PyQt version. No need to bind it to 5.13 anymore.

llimeht:
Updated once more (but perhaps this is a topic for bug fixes for a separate PR)

wpotrzebowski:
I've tested OSX and linux installers and they seem to work fine up to the point when one starts python shell editor. The editor itself works fine however the error occurs: 
`5:34:41 - ERROR: Traceback (most recent call last):
  File "sas/sascalc/data_util/calcthread.py", line 272, in _run
  File "sas/qtgui/Perspectives/Fitting/ModelThread.py", line 179, in compute
  File "sasmodels/sasview_model.py", line 713, in calculate_Iq
    return self._calculate_Iq(qx, qy)
  File "sasmodels/sasview_model.py", line 721, in _calculate_Iq
    self.__class__._model = core.build_model(self._model_info)
  File "sasmodels/core.py", line 335, in build_model
    source = generate.make_source(model_info)
  File "sasmodels/generate.py", line 1025, in make_source
    lineno = getframeinfo(currentframe()).lineno + 2
  File "inspect.py", line 1473, in getframeinfo
  File "PyInstaller/hooks/rthooks/pyi_rth_inspect.py", line 25, in _pyi_getsourcefile
AttributeError: module '__main__' has no attribute '__file__'
15:34:41 - ERROR: Traceback (most recent call last):
  File "sas/sascalc/data_util/calcthread.py", line 272, in _run
  File "sas/qtgui/Perspectives/Fitting/ModelThread.py", line 179, in compute
  File "sasmodels/sasview_model.py", line 713, in calculate_Iq
    return self._calculate_Iq(qx, qy)
  File "sasmodels/sasview_model.py", line 721, in _calculate_Iq
    self.__class__._model = core.build_model(self._model_info)
  File "sasmodels/core.py", line 335, in build_model
    source = generate.make_source(model_info)
  File "sasmodels/generate.py", line 1025, in make_source
    lineno = getframeinfo(currentframe()).lineno + 2
  File "inspect.py", line 1473, in getframeinfo
  File "PyInstaller/hooks/rthooks/pyi_rth_inspect.py", line 25, in _pyi_getsourcefile
AttributeError: module '__main__' has no attribute '__file__'`

Not sure how difficult to fix it is, however I would vote for merging it and addressing this issue in a separate PR.

wpotrzebowski:
I would vote for merging it, however I would wait until end of the code camp (tomorrow).

pkienzle:
I suspect the error is a result of the python app builder putting modules in a zip file. IIRC the import hook which loads the module from the zip does not define `__file__`. That doesn't explain why the problem is showing up now.

The code which is failing is not critical. It is telling the model compiler that the code is coming from the current line in `generate.py` so that it'll be easier to interpret compiler errors. As a short term fix we can wrap this in an exception handler and use a hard-coded line number as the default. It doesn't matter if the offset drifts a little as the code is modified; just knowing that the code is coming from generate.py should be enough to debug any problems.

llimeht:
Updated once more for merge conflicts.

wpotrzebowski (approved):

wpotrzebowski:
Tested latest OSX installer. The convoluted issue with shell editor still persists but I am merging it as it is.

Issue #1337, opened 2019-05-20
non-resizing plot legend on OSX


rozyczko:
Found in #690 by Paul K.
Plots are hard to read on mac (legends are huge, data is pushed to one side, they keep popping above the fit like whack-a-mole)

This is on OSX on small screen laptops. Matplotlib version = 2.2.2
The same version on Windows behaves properly (legend resized, doesn't go beyond canvas boundary)

ajj:
Not just legend text - axis labels don't scale either and are too large, and marker in legend doesn't scale and is too large. Suspect issue with how MacOS reports resolution when there is retina scaling and conflict with what AA_EnableHighDpiScaling in Qt5 is doing.

ajj:
![Screenshot 2019-07-23 at 12 07 04](https://user-images.githubusercontent.com/871392/61703703-70988000-ad42-11e9-93c1-e38847f2dc1e.png)


Screenshot from my external monitor running sasview with c789273 - Apple Thunderbolt Display 27" 2560x1440

ajj:
<img width="1440" alt="Screenshot 2019-07-23 at 12 08 50" src="https://user-images.githubusercontent.com/871392/61703969-00d6c500-ad43-11e9-9bc7-c0ea320b1fbd.png">

Screenshot from my laptop display with no external display connected running sasview with c789273 (Retina display 2560x1600 real resolution set to default scaling - "looks like 1440x900")

ajj:
![Screenshot 2019-07-23 at 12 21 02](https://user-images.githubusercontent.com/871392/61704733-ab9bb300-ad44-11e9-8315-a25f73a4c646.png)


Screenshot from my external display when the laptop is closed. This is more well-behaved. I suspect we need to do something in Qt to determine the active screen and set the scaling accordingly. This does extend the ticket somewhat - it now seems to behave OK with c789273 when only using the laptop screen, but if you attach your laptop to an external display then it is still wonky. Also still need to fix other labels.

wpotrzebowski:
@ajj: It seems you run it from the local environment. Does it look the same when you run it from installer, e.g: https://jenkins.esss.dk/sasview-beta/job/SasView-OSX10.13-ghprb/411/? 
We enable NSHighResolutionCapable option for the bundle, so it may behave differently.

butlerpd:
As per @wpotrzebowski keeping this for now but moving to 5.0.2 given imminent release of 5.0.1

ajj:
[...]

Issue #690, opened 2019-03-30
set reasonable min/max on polydispersity values for fit (Trac #558)


pkienzle:
polydispersity cannot be negative, so set the min to 0.

relative polydispersity doesn't make a lot of sense above 1 (thickness of 30 +/- 200%?)

Migrated from http://trac.sasview.org/ticket/558
```json
{
    "status": "new",
    "changetime": "2017-10-27T10:49:46",
    "_ts": "2017-10-27 10:49:46.234557+00:00",
    "description": "polydispersity cannot be negative, so set the min to 0.\n\nrelative polydispersity doesn't make a lot of sense above 1 (thickness of 30 +/- 200%?)\n\n",
    "reporter": "pkienzle",
    "cc": "",
    "resolution": "",
    "workpackage": "SasView Bug Fixing",
    "time": "2016-04-12T15:20:28",
    "component": "SasView",
    "summary": "set reasonable min/max on polydispersity values for fit",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView 5.0.0",
    "owner": "",
    "type": "defect"
}
```

butlerpd:
Trac update at `2016/08/16 14:30:21`: 
* **butler** commented:

sasmodels apparently has default limits but GUI does not pick those up.  This would be very useful but changing the GUI at this point is probably not realistic for release 4.0.  Moving to next +1


* **butler** changed milestone from "SasView 4.0.0" to "SasView Next Release +1"

butlerpd:
Trac update at `2017/10/27 10:49:46`: **butler** changed milestone from "SasView Next Release +1" to "SasView 5.0.0"

rozyczko:
5.0 uses param.limits[0] and param.limits[1] to display parameter limits. Is this the correct attribute for polydispersity values as well?

pkienzle:
polydispersity limits will be different from param limits.  For shape parameters, [0, 1] are reasonable default limits.  For orientation parameters [0, 90] would be more appropriate defaults.

butlerpd:
but aren't those defined in sasmodels? if so then the GUI just needs to read them and display?

pkienzle:
Here's the code in the sasview wrapper which sets the defaults:
    https://github.com/SasView/sasmodels/blob/ed3ea824726b2ec0c941f6c7dedd5895bc052570/sasmodels/sasview_model.py#L421-L430

[...]

PR #2264, opened 2022-10-25
Fix unusable legend size with long filenames on Mac

pbeaucage:
As described in #1337; the fix from @krzywon both works on Mac and dramatically simplifies the legend code.  I'll test on Windows shortly.


pbeaucage:
image

This is working on Windows as well, and while the behavior with long datanames is a little wacky, it is usable.

pbeaucage:
I added text-wrapping and a floating legend width; these combine to make this a fairly robust fix for #1337.  Unfortunately some plot geometries still have the issue but they are largely when the plot is tiny.  Making the legend fixed- and full-width fixes the issue reliably but looks bad, see above.  Maybe we could make this a user preference?

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -253,14 +230,18 @@ def plot(self, data=None, color=None, marker=None, hide_error=False, transform=T
...
         # Now add the legend with some customizations.
         if self.showLegend:
-            width=_legendResize(self.canvas.size().width(), self.parent)
-            if width is not None:
-                self.legend = ax.legend(loc='upper right', shadow=True, prop={'size':width})
-            else:
-                self.legend = ax.legend(loc='upper right', shadow=True)
-            if self.legend:
-                self.legend.set_picker(True)
+            max_legend_width = 30
```
wpotrzebowski, line 233: I am wondering if we should have this is as setting or is it enough to keep it at 30.

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -301,16 +282,6 @@ def plot(self, data=None, color=None, marker=None, hide_error=False, transform=T
         # refresh canvas
         self.canvas.draw_idle()
 
-    def onResize(self, event):
```
wpotrzebowski, line 304: I am not entirely sure if we completely should get rid of it. Could there is still be the case e.g. on low resolutions screen that this may be useful

pbeaucage: The underlying _legendResize is gone in this, originally done by @krzywon.  The gist is that matplotlib now provides its own handler for this and we don't need to duplicate that functionality any more.

src/sas/qtgui/Plotting/Plotter.py
pbeaucage: I definitely think this could be a setting.  I was pondering setting it dynamically, but 30 works pretty well on most screen sizes / for most reasonable data name lengths.  Another option is to forego the text-wrapping behavior; there is a flag to  `ax.legend()`, something like `expand=True` that makes the legend full width.  It's sort of a question of config bloat vs something that works most of the time.  My feeling was that static-30 is good enough for something that's blocking usability on some platforms.

gonzalezma:
It does what it says, but with very long titles the legend can still become annoying, e.g.

legend1

What about "cutting" the title, at the price of losing information in the legend, e.g.

legend2

rozyczko:
Miguel's suggestion is good, I would go even further - have names with ellipsis by default but expand them fully when the cursor hovers over the legend (or on click?).

pbeaucage:
I'll point out that this PR really was intended to have a quite narrow scope: the point of #1337 and thus this PR was to fix a font scaling issue that has made sasview 5 unusable on macs with certain hard-to-define, but fairly common monitor configurations for the last 3.5 years because the plot legends look like this:
image
or 
image
with no workarounds, not even hiding the legend (which only started working properly in #2266 yesterday).

this issue was first documented in March of 2019 #690 or Trac#558 in a side comment and I have heard informally from a large number of attempted sasview mac users that they have tried, and failed, to use 5.x because of this issue. They don't bother to email help because they assume that software that can't correctly plot data out of the box is not well supported or maintained.  So I would suggest that building a real solution for arbitrary length filename display is a larger-scale problem that shouldn't block merging this patch PR for a longstanding issue that critically impacts usability.
--- 
that said, 
In general, I feel like truncating the user's input is a fairly problematic thing to do.  Presumably they used a long name for a reason and if the plot legend is a little unwieldy (I point out that Miguel's demo above, while a bit ugly, is still _usable_ which the current solution **_is not_**).  Given that (a) we provide users with GUI tools to rename data if they wish, (b) users can rename their own files out on the filesystem, and (c) there is robust tooling in shells on every OS we support to do this renaming programmatically for users with large numbers of datasets, I don't see a good reason to override what the user has told us to call their data.  If it's unusable because of user input, and we provide the user with the tools to make it usable, what are we to do? You can only shield the novice user from their own incompetence so much before you start preventing power-users from doing what they want to do.

With the expand-on-hover / expand-on-click solution it becomes (a) tedious to hover/click each item, (b) difficult to save a plot with a complete legend for further re-use in another context.

I can see having 3 config parameters to let users really control this behavior:

FITTING_PLOT_FULL_WIDTH_LEGENDS default False, if True invokes the `plt.Legend` command with `mode='Expand'` so that the legend is always full-width and very little dynamic rescaling is performed; this will as far as I can tell **always** work but truncates legends at screen width and looks ugly.  Basically a debug setting should this fix turn out to fail in the wild.
FITTING_PLOT_LEGEND_TRUNCATE default False, if True truncates everything except the first and last (FITTING_PLOT_LEGEND_MAX_LENGTH/2 - 2) of the legend label and inserts ` .. ` in the center before display such that each legend line is a single line, basically Miguel's solution
FITTING_PLOT_LEGEND_MAX_LINE_LENGTH default 30, sets the text-wrap length or the truncation length depending on FITTING_PLOT_LEGEND_TRUNCATE, basically Wojciech's proposed parameter

Is this too many config parameters to add? or about the right level of flexibility?

pbeaucage:
OK, added the config items as above.  The result looks pretty good!  With max line length of 30 and truncate on:
image

Same, truncate off:
image

src/sas/qtgui/Plotting/Plotter.py
pbeaucage: This is now done.

pbeaucage:
Moving the preference panel GUI to a different branch/PR as it depends on updated tooling from @krzywon

pbeaucage:
This should now be ready for final review if tests pass.

wpotrzebowski (approved):

Issue #2215, opened 2022-09-23
Plot legend enable/disable does not persist between plot draws (e.g. fit runs)


pbeaucage:
**Describe the bug**
#1523 re-added a context menu toggle for visibility of the legend in a plot; however, that toggle just uses matplotlib methods to enable/disable the legend rather than using the internal Plotter class variables, which means that the visibility setting does not persist between re-draws of the plot during, for example, runs of a fit.

This tiles with other bugs such as #1337 on Mac and #2012 more generally to make 5.x versions often unusable with long filenames on Macs, since the legend obstructs the plot and one cannot turn it off without tediously clicking it off every time a fit variable is changed.

**To Reproduce**
1. Generate a plot from e.g. the fitting perspective
2. Toggle legend off using context menu
3. Perform an action (e.g. run a fit) that re-draws the plot.
4. Legend appearance will revert to on regardless of prior setting.  

**Expected behavior**
Legend appearance/disappearance should follow previous plot interactions.

**SasView version (please complete the following information):**
 - Version: 5.0.5

**Operating system (please complete the following information):**
 - OS: Mac 12.5.1

PR #2266, opened 2022-10-25
Add persistent legend visibility toggle

pbeaucage:
As discussed in #2215, adds a new instance variable to `PlotterWidget`, `self.legendVisible` which tracks the user's preference for having the legend visible/invisible (this is distinct from `self.showLegend` which tracks whether the plot has a legend at all).  The `onToggleLegend` and similar methods are updated to toggle `self.legendVisible` then trigger a redraw rather than just modifying the existing plot.  Tested and appears to work on Win10.

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -88,6 +88,8 @@ class PlotterWidget(PlotterBase):
         self.toolbar._actions['pan'].triggered.connect(self._pan)
         self.toolbar._actions['zoom'].triggered.connect(self._zoom)
 
+        self.legendVisible = True
+
         parent.geometry()
 
     @property
@@ -250,7 +252,7 @@ class PlotterWidget(PlotterBase):
         self.plot_lines[data.name] = line
 
         # Now add the legend with some customizations.
-        if self.showLegend:
+        if self.showLegend and self.legendVisible:
             width=_legendResize(self.canvas.size().width(), self.parent)
             if width is not None:
                 self.legend = ax.legend(loc='upper right', shadow=True, prop={'size':width})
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -250,7 +252,7 @@ def plot(self, data=None, color=None, marker=None, hide_error=False, transform=T
         self.plot_lines[data.name] = line
 
         # Now add the legend with some customizations.
-        if self.showLegend:
+        if self.showLegend and self.legendVisible:
```
krzywon, line 255: This is suppressing legend creation on redraw so the legend disappears with no way to get it back. Steps to repeat:
- Plot data
- Toggle legend to make it invisible
- Change the scale to force a redraw
- Try (and fail) to toggle the legend to make it visible again

Suggestion:
```
if self.showLegend:
    ...  # after legend is created
    self.legend.set_visible(self.legendVisible)
```

krzywon (changes requested):
I've noted one issue that needs to be fixed in a separate comment here.

Not surprisingly, on redraw, the location of the legend resets to the original position, but that is likely beyond the scope of this PR. I'm going to create a plotting discussion (if there isn't one already) about this. Plots are completely reconstructed on redraw, resetting many attributes.

wpotrzebowski (commented):
Works also on Mac as long as plot window is not closed (which should be as it is) and except bug @krzywon pointed out.

wpotrzebowski:
Also looked at the other perspectives and it works for all except Corfunc, which has differemt plotting mechanism. The intention is to make it also available for Corfunc, however this is not something that will be done for this PR.

src/sas/qtgui/Plotting/Plotter.py
pbeaucage: Fixed, it can now actually toggle, thanks!

krzywon (approved):
The most recent change fixes the issue I noted earlier.

Issue #2272, opened 2022-10-25
Main SV window is not maximized on start


rozyczko:
SV should maximize its main window on start. There even is a relevant `.showMaximized()` call but the window is not actually maximized, which leads to weird initial state, at least on windows.
The app window is small and squished but the maximize/restore button in the upper/right corner shows that the window is actually maximized!

pbeaucage:
Fixed in #2273

PR #2273, opened 2022-10-25
Force main window to maximize #2272

rozyczko:
The window now gracefully maximizes allowing the fitting perspective to be properly displayed with all the buttons visible.


pbeaucage (approved):
Does what it says on the box, works on Windows.  Needs testing on Mac.

pbeaucage:
Works on M1 Mac, as well.

PR #2275, opened 2022-10-25
A few more or less obvious fixes for speed of SasView startup

rozyczko:
From >30 sec down to ~5 sec.
numba related fixes suggested by @pkienzle + minor other import shuffling as suggested by startup profiles.

### DIFF
--- a/src/sas/system/log.py
+++ b/src/sas/system/log.py
@@ -29,6 +29,7 @@ class SetupLogger(object):
             logging.captureWarnings(True)
             logger = logging.getLogger(self.name)
         logging.getLogger('matplotlib').setLevel(logging.WARN)
+        logging.getLogger('numba').setLevel(logging.WARN)
         return logger
 
     def config_development(self):
@@ -39,6 +40,7 @@ class SetupLogger(object):
         self._update_all_logs_to_debug(logger)
         logging.captureWarnings(True)
         logging.getLogger('matplotlib').setLevel(logging.WARN)
+        logging.getLogger('numba').setLevel(logging.WARN)
         return logger
 
     def _read_config_file(self):
--- a/src/sas/qtgui/Utilities/Reports/ReportDialog.py
+++ b/src/sas/qtgui/Utilities/Reports/ReportDialog.py
@@ -3,7 +3,6 @@ import sys
 import re
 import logging
 import traceback
-from xhtml2pdf import pisa
 from typing import Optional
 
 from PyQt5 import QtWidgets, QtCore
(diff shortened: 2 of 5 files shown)


krzywon (approved):
Does what it says it does. I measured the time to open the main window and it decreased from 21.5 seconds running the main branch to 11.1 seconds for this one.

gonzalezma:
More modest gain in my Windows machine (from ~30" to ~18"), but still an improvement.

pkienzle:
Maybe [src/sas/sascalc/pr/p_invertor.py](https://github.com/SasView/sasview/pull/2278/files#diff-a60a53f7209ac88bbb00a2e90e9c05eee86cfefb0e79ead37b33297306262d5c) as well?

rozyczko:
> Maybe [src/sas/sascalc/pr/p_invertor.py](https://github.com/SasView/sasview/pull/2278/files#diff-a60a53f7209ac88bbb00a2e90e9c05eee86cfefb0e79ead37b33297306262d5c) as well?

In the startup profiling (Win10, dev build), importing the inversion perspective takes less than 1%. I'll add it anyway.

Other external imports, e.g. `from bumps import options` take significantly more time (>12%). I'm looking at making this import lazy as well.

rozyczko:
importing bumps options can't be made lazy, since we need to display the default bumps minimizer name, meaning importing and querying the module at start time.
We _could_ hard-code the initial value to LM but this is not pretty.
Leaving as is then.

src/sas/system/log.py
```diff
@@ -39,6 +40,7 @@ def config_development(self):
         self._update_all_logs_to_debug(logger)
         logging.captureWarnings(True)
         logging.getLogger('matplotlib').setLevel(logging.WARN)
+        logging.getLogger('numba').setLevel(logging.WARN)
```
pkienzle, line 43: Changing level to WARNING for handler_log_file and logger_root in log.ini seems like a better solution.

rozyczko: For releases it makes sense to change default logging level to WARNING.
For developer versions it would probably be better to leave it on DEBUG.
This means adding complexity to the logger logic and should be addressed in the config options PR or in a separate issue.

pkienzle: Line 40: (_update_all_logs_to_debug) changes all our libraries to DEBUG so that we don't have to set it as the default.

For third party libraries we should whitelist any that we want to debug not blacklist any that make too much noise.

rozyczko: Agreed.
Will be addressed in https://github.com/SasView/sasview/issues/2289

wpotrzebowski (approved):
The changes to the code seem harmless and it works fine on Mac

PR #2277, opened 2022-10-25
Improve output format for the fitting log (#2276)

### LINKED ISSUE #2276: Typo in BumpsFitting.py Progress info
**Describe the bug**
The first if to decide about the best format to print the remaining calculation time should read 18000 (seconds = 5 hours) instead of 1800.

gonzalezma:
Simple fix of a typo: 1800 (seconds) should be 18000 in the if block deciding which is the best format to print out the remaining calculation time

### DIFF
--- a/src/sas/sascalc/fit/BumpsFitting.py
+++ b/src/sas/sascalc/fit/BumpsFitting.py
@@ -38,22 +38,18 @@ from sas.sascalc.fit.expression import compile_constraints
 class Progress(object):
     def __init__(self, history, max_step, pars, dof):
         remaining_time = int(history.time[0]*(float(max_step)/history.step[0]-1))
-        # Completion time is displayed as a delta (e.g., 36s, 2m 36s or 3h 12m) if completion time is less
-        # than delta_cutoff hours, or as HH:MM if it is within 24 hours. Otherwise it is displayed using a
-        # full timestamp YYYY-mm-dd HH:MM.
-        delta_cutoff = 5  # hours before switching from duration to clock time
         if remaining_time < 60:
-            time = '%ds' % remaining_time
+            delta_time = '%ds' % remaining_time
         elif remaining_time < 3600:
-            time = '%dm %ds' % (remaining_time // 60, remaining_time % 60)
-        elif remaining_time < delta_cutoff * 3600:
-            time = '%dh %dm' % (remaining_time // 3600, (remaining_time % 3600) // 60)
+            delta_time = '%dm %ds' % (remaining_time // 60, remaining_time % 60)
+        elif remaining_time < 24 * 3600:
+            delta_time = '%dh %dm' % (remaining_time // 3600, (remaining_time % 3600) // 60)
         else:
-            completion_time = datetime.now() + timedelta(seconds=remaining_time)
-            time = completion_time.strftime('%Y-%m-%d %H:%M')
+            delta_time = '%dd %dh' % (remaining_time // (24 * 3600), (remaining_time % (24 * 3600)) // 3600)
+        finish_time = (datetime.now() + timedelta(seconds=remaining_time)).strftime('%Y-%m-%d %H:%M')
         chisq = "%.3g"%(2*history.value[0]/dof)
         step = "%d of %d"%(history.step[0], max_step)
-        header = "=== Steps: %s  chisq: %s  ETA: %s\n"%(step, chisq, time)
+        header = "=== Steps: %s  chisq: %s  ETA: %s (%s from now)\n"%(step, chisq, finish_time, delta_time)
         parameters = ["%15s: %-10.3g%s"%(k,v,("\n" if i%3==2 else " | "))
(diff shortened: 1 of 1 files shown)


src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, history, max_step, pars, dof):
         # Depending on the time remaining, either display the expected
         # time of completion, or the amount of time remaining.  Use precision
         # appropriate for the duration.
-        if remaining_time >= 1800:
+        if remaining_time >= 18000:
```
pkienzle, line 44: Maybe use the following:
```
# Completion is time is displayed as a delta (e.g., 36s, 2m 36s or 3h 12m) if completion time is less
# than delta_cutoff hours, or as HH:MM if it is within 10 hours, otherwise it is displayed using a
# full timestamp YYYY-mm-dd HH:MM.
delta_cutoff = 2 # hours before switching from duration to clock time
if remaining_time >= delta_cutoff*3600:
    ...
```
You could further add the following clause to the else if you wanted to display deltas more than one day away:
```
if remaining_time >= 24*3600:
    time = '%dd %dh'%(remaining_time//(24*3600), (remaining_time%(24*3600))//3600)
...
```

gonzalezma: Right. I was just lazy and did the minimum to fix the obvious typo. But I will revise the whole if block and reorganize it to put the time limits in increasing order, which is not the case now.
BTW, it has nothing to do with this, but if I spotted that is because I was trying to figure out why the reported chi2 in the log explorer is about twice higher than the real chi2 shown in the fit panel window when using mpfit. I explored why, but from what I could get this is coming from bumps/mpfit.py and bumps/fitters.py. I noticed that the number of dof is overestimated because the parameter_residuals are added to the standard residuals, but this does not explain the difference, so I guess there must be a factor of 2 somewhere else. This is only the case for mpfit, not for the other minimizers, so I'm not sure that I should open a sasview issue. Any suggestion?

pkienzle: I created ticket https://github.com/bumps/bumps/issues/111

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -38,22 +38,19 @@ def get_fitter():
...
+            time = '%dh %dm' % (remaining_time // 3600, (remaining_time % 3600) // 60)
         else:
-            if remaining_time >= 3600:
-                time = '%dh %dm'%(remaining_time//3600, (remaining_time%3600)//60)
-            elif remaining_time >= 60:
-                time = '%dm %ds'%(remaining_time//60, remaining_time%60)
-            else:
-                time = '%ds'%remaining_time
+            completion_time = datetime.now() + timedelta(seconds=remaining_time)
+            time = completion_time.strftime('%Y-%m-%d %H:%M')
```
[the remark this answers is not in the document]
gonzalezma: Not sure I understand you. I kept the same logic, but now starting to decide from below, i.e.:
remaining_time < 60 --> '%ds' , < 3600 --> '%dm %ds', < delta*3600 --> '%dh %dm' and finally if > delta*3600 --> show completion time as '%Y-%m-%d %H:%M'. I simply removed the option to show the completion time as just HH:MM, as I thought that it was clearer to show the date + time in all cases.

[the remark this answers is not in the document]
gonzalezma: Not really, unless one sets delta_cutoff < 1, which in this case would simply be ignored (so the changing of behaviour from delta to completion would happen at 3600 seconds). But showing both is even better. I'll go for that.

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -38,25 +38,18 @@ def get_fitter():
...
-            elif remaining_time >= 60:
-                time = '%dm %ds'%(remaining_time//60, remaining_time%60)
-            else:
-                time = '%ds'%remaining_time
+            delta_time = '%dd %dh' % (remaining_time // (24 * 3600), (remaining_time % (24 * 3600)) // 3600)
+        finish_time = (datetime.now() + timedelta(seconds=remaining_time)).strftime('%Y-%m-%d %H:%M')
         chisq = "%.3g"%(2*history.value[0]/dof)
         step = "%d of %d"%(history.step[0], max_step)
-        header = "=== Steps: %s  chisq: %s  ETA: %s\n"%(step, chisq, time)
+        header = "=== Steps: %s  chisq: %s  ETA: %s (%s from now)\n"%(step, chisq, finish_time, delta_time)
```
pkienzle, line 52: Could use f-strings if you find it more readable:
```python
chisq = 2*history.value[0]/dof
k, n = history.step[], max_step
header = f"=== Steps: {k} of {n}  chisq: {chisq:.3g}  ETA: {finish_time} ({delta_time} from now)\n"
```

pkienzle (approved):

PR #2278, opened 2022-10-25
Add command line interface to allow scripts to be run from sasview.exe

pkienzle:
Allow scripts from the command line:
```
sasview -m module [args...]
    Run module as main.
sasview -c "python statements"
    Execute python statements with sasview libraries available.
sasview -i 
    Start ipython interpreter.
sasview script [args...]
    Run script with sasview libraries available
sasview
    Start sasview gui
```

Partially addresses #2237


pkienzle:
We could also add a `--jupyter` command line option which opens up a server on localhost with everything setup. It'll increase the size of our installed package though, especially if we include jupyterhub with a nodejs distribution.

pkienzle:
I don't have a windows box handy so couldn't test the installed package on windows.

Do we need multiprocessing freeze support?

I don't know how modern qt apps handle console vs. gui support. Maybe need to do something like creating a separate sasviewcom.exe for the console.

pkienzle:
Replaced with #2280.

PR #2280, opened 2022-10-25
Add command line interface to allow scripts to be run from sasview.exe

pkienzle:
Allow scripts from the command line:
```
sasview -m module [args...]
    Run module as main.
sasview -c "python statements"
    Execute python statements with sasview libraries available.
sasview -i 
    Start ipython interpreter.
sasview script [args...]
    Run script with sasview libraries available
sasview
    Start sasview gui
```

Partially addresses #2237. Replaces #2278.

We could also add a --jupyter command line option which opens up a server on localhost with everything setup. It'll increase the size of our installed package though, especially if we include jupyterhub with a nodejs distribution.

I don't have a windows box handy so couldn't test the installed package on windows.

Do we need multiprocessing freeze support?

I don't know how modern qt apps handle console vs. gui support. Maybe need to do something like creating a separate sasviewcom.exe for the console.

### DIFF
--- a/run.py
+++ b/run.py
@@ -19,8 +19,7 @@ import sys
 import os
 from os.path import abspath, dirname, realpath, join as joinpath
 from contextlib import contextmanager
-
-PLUGIN_MODEL_DIR = 'plugin_models'
+from importlib import import_module
 
 def addpath(path):
     """
@@ -35,7 +34,6 @@ def addpath(path):
     os.environ['PYTHONPATH'] = PYTHONPATH
     sys.path.insert(0, path)
 
-
 @contextmanager
 def cd(path):
     """
@@ -46,62 +44,45 @@ def cd(path):
     yield
     os.chdir(old_dir)
 
-def setup_sasmodels():
-    """
-    Prepare sasmodels for running within sasview.
-    """
-    # Set SAS_MODELPATH so sasmodels can find our custom models
-
-    from sas.system.user import get_user_dir
-    plugin_dir = os.path.join(get_user_dir(), PLUGIN_MODEL_DIR)
-    os.environ['SAS_MODELPATH'] = plugin_dir
-
 def prepare():
     # Don't create *.pyc files
     sys.dont_write_bytecode = True
 
-    # find the directories for the source and build
-    from distutils.util import get_platform
+    # Turn numpy warnings into errors
+    #import numpy; numpy.seterr(all='raise')
+
+    # Find the directories for the source and build
     root = abspath(dirname(realpath(__file__)))
 
-    platform = '%s-%s' % (get_platform(), sys.version[:3])
-    build_path = joinpath(root, 'build', 'lib.' + platform)
-
-    # Notify the help menu that the Sphinx documentation is in a different
-    # place than it otherwise would be.
-    os.environ['SASVIEW_DOC_PATH'] = joinpath(build_path, "doc")
-
-    try:
-        import periodictable
-    except ImportError:
-        addpath(joinpath(root, '..', 'periodictable'))
-
-    try:
-        import bumps
-    except ImportError:
-        addpath(joinpath(root, '..', 'bumps'))
+    # TODO: Do we prioritize the sister repo or the installed package?
+    # TODO: Can we use sasview/run.py from a distributed sasview.exe?
+    # Put supporting packages on the path if they are not already available.
+    for sister in ('periodictable', 'bumps', 'sasdata', 'sasmodels'):
+        try:
+            import_module(sister)
+        except:
+            addpath(joinpath(root, '..', sister))
 
     # Put the source trees on the path
     addpath(joinpath(root, 'src'))
 
-    # sasmodels on the path
-    addpath(joinpath(root, '../sasmodels/'))
+    # == no more C sources so no need to build project to run it ==
+    # Leave this snippet around in case we add a compile step later.
+    #from distutils.util import get_platform
+    #platform = '%s-%s' % (get_platform(), sys.version[:3])
+    #build_path = joinpath(root, 'build', 'lib.' + platform)
+    ## Build project if the build directory does not already exist.
+    #if not os.path.exists(build_path):
+    #    import subprocess
+    #    with cd(root):
+    #        subprocess.call((sys.executable, "setup.py", "build"), shell=False)
 
-
+    # Notify the help menu that the Sphinx documentation is in a different
+    # place than it otherwise would be.
+    docpath = joinpath(root, 'docs', 'sphinx-docs', '_build', 'html')
+    os.environ['SASVIEW_DOC_PATH'] = docpath
 
 if __name__ == "__main__":
-    # Need to add absolute path before actual prepare call,
-    # so logging can be done during initialization process too
-    root = abspath(dirname(realpath(sys.argv[0])))
-
-    addpath(joinpath(root, 'src'))
     prepare()
-
-    # Run the UI conversion tool when executed from script.  This has to
-    # happen after prepare() so that sas.qtgui is on the path.
-    import sas.qtgui.convertUI
-    setup_sasmodels()
-
-    from sas.qtgui.MainWindow.MainWindow import run_sasview
-    run_sasview()
-    #logger.debug("Ending SASVIEW in debug mode.")
+    import sas.cli
+    sas.cli.main(logging="development")
--- a/src/sas/cli.py
+++ b/src/sas/cli.py
@@ -1,0 +1,90 @@
+"""
+SasView command line interface.
+
+sasview -m module [args...]
+    Run module as main.
+sasview -c "python statements"
+    Execute python statements with sasview libraries available.
+sasview -i 
+    Start ipython interpreter.
+sasview script [args...]
+    Run script with sasview libraries available
+sasview
+    Start sasview gui
+"""
+import sys
+
+# TODO: Support dropping datafiles onto .exe?
+# TODO: Maybe use the bumps cli with project file as model?
+
+import argparse
+parser = argparse.ArgumentParser()
+parser.add_argument("-m", "--module", type=str,
+    help="Run module as main")
+parser.add_argument("-c", "--command", type=str,
+    help="Execute command")
+parser.add_argument("-i", "--interactive", action='store_true', 
+    help="Run interactive command line")
+parser.add_argument("argv", nargs="*",
+    help="script followed by argv")
+
+def exclusive_error():
+    print("Use only one of -m module args, -c command, -i, or script.py args.", sys.stderr)
+    sys.exit(1)
+
+def run_interactive():
+    """Run sasview as an interactive python interpreter"""
+    try:
+        from IPython import start_ipython
+        sys.argv = ["ipython", "--pylab"]
+        sys.exit(start_ipython())
+    except ImportError:
+        import code
+        code.interact(local={'exit': sys.exit})
+
+def main(logging="production"):
+    from sas.system import log
+    from sas.system import lib
+
+    # Eventually argument processing might affect logger or config, so do it first
+    args = parser.parse_args()
+
+    # Setup logger and sasmodels
+    if logging == "production":
+        log.production()
+    elif logging == "development":
+        log.development()
+    else:
+        raise ValueError(f"Unknown logging mode \"{logging}\"")
+    lib.setup_sasmodels()
+
+    # Parse mutually exclusive command line options
+    # mutually exclusive (-m module args, -c command, -i, script args)
+    if args.argv and not args.module: # script [arg...]
+        import runpy
+        if args.command or args.module or args.interactive:
+            exclusive_error()
+        sys.argv = args.argv
+        runpy.run_path(args.argv[0], run_name="__main__")
+    elif args.module: # -m module [arg...]
+        import runpy
+        if args.command or args.interactive:
+            exclusive_error()
+        sys.argv = [args.module, *args.argv]
+        runpy.run_module(args.module, run_name="__main__")
+    elif args.command: # -c "command"
+        if args.argv or args.module or args.interactive:
+            exclusive_error()
+        exec(args.command)
+    elif args.interactive: # -i
+        if args.argv or args.module or args.command:
+            exclusive_error()
+        run_interactive()
+    else:
+        from sas.qtgui.MainWindow.MainWindow import run_sasview as run_gui
+        run_gui()
+
+if __name__ == "__main__":
+    from multiprocessing import freeze_support
+    freeze_support()
+    main()
--- a/src/sas/qtgui/MainWindow/MainWindow.py
+++ b/src/sas/qtgui/MainWindow/MainWindow.py
@@ -1,11 +1,10 @@
 # UNLESS EXEPTIONALLY REQUIRED TRY TO AVOID IMPORTING ANY MODULES HERE
 # ESPECIALLY ANYTHING IN SAS, SASMODELS NAMESPACE
-import logging
 import os
 import sys
+import logging
 
-from sas import config
-from sas.system import env, version
+from sas.system.version import __version__
 
 from PyQt5.QtWidgets import QMainWindow
 from PyQt5.QtWidgets import QMdiArea
@@ -17,6 +16,8 @@ from PyQt5.QtCore import Qt
 # Local UI
 from .UI.MainWindowUI import Ui_SasView
 
+logger = logging.getLogger(__name__)
+
 class MainSasViewWindow(QMainWindow, Ui_SasView):
     # Main window of the application
     def __init__(self, screen_resolution, parent=None):
@@ -25,7 +26,7 @@ class MainSasViewWindow(QMainWindow, Ui_SasView):
         self.setupUi(self)
 
         # Add the version number to window title
-        self.setWindowTitle(f"SasView {version.__version__}")
+        self.setWindowTitle(f"SasView {__version__}")
         # define workspace for dialogs.
         self.workspace = QMdiArea(self)
         # some perspectives are fixed size.
@@ -45,8 +46,7 @@ class MainSasViewWindow(QMainWindow, Ui_SasView):
         try:
             self.guiManager = GuiManager(self)
         except Exception as ex:
-            import logging
-            logging.error("Application failed with: "+str(ex))
+            logger.error("Application failed with: "+str(ex))
             raise ex
 
     def closeEvent(self, event):
@@ -68,30 +68,11 @@ def SplashScreen():
     return splashScreen
 
 def run_sasview():
-    app = QApplication([])
-
-    #Initialize logger
-    from sas.system.log import SetupLogger
-    SetupLogger(__name__).config_development()
-
-    # initialize sasmodels settings
-    from sas.system.user import get_user_dir
-    if "SAS_DLL_PATH" not in os.environ:
-        os.environ["SAS_DLL_PATH"] = os.path.join(
-            get_user_dir(), "compiled_models")
-
-    # Set open cl config from environment variable, if it is set
-
-    if env.sas_opencl is not None:
-        logging.getLogger(__name__).info("Getting OpenCL settings from environment variables")
-        config.SAS_OPENCL = env.sas_opencl
-    else:
-        logging.getLogger(__name__).info("Getting OpenCL settings from config")
-        env.sas_opencl = config.SAS_OPENCL
-
     # Make the event loop interruptable quickly
     import signal
     signal.signal(signal.SIGINT, signal.SIG_DFL)
+
+    app = QApplication([])
 
     # Main must have reference to the splash screen, so making it explicit
     splash = SplashScreen()
--- a/src/sas/system/__init__.py
+++ b/src/sas/system/__init__.py
@@ -1,3 +1,4 @@
+# TODO: Don't shadow module names; it makes debugging difficult.
 from .web import web
 from .legal import legal
 from .env import env
--- a/src/sas/system/env.py
+++ b/src/sas/system/env.py
@@ -1,3 +1,4 @@
+# ** DEPRECATED **
 """ Interface for environment variable access
 
 This is intended to handle any conversion from the environment variable string to more natural types.
--- a/installers/sasview.py
+++ b/installers/sasview.py
@@ -6,10 +6,10 @@ Run sasview from an installed bundle
 """
 
 import sys
-
-
 sys.dont_write_bytecode = True
 
-from sas.qtgui.MainWindow.MainWindow import run_sasview
-
-run_sasview()
+if __name__ == "__main__":
+    from multiprocessing import freeze_support
+    freeze_support()
+    from sas.cli import main
+    main()
--- a/src/sas/__init__.py
+++ b/src/sas/__init__.py
@@ -1,9 +1,9 @@
 from sas.system.version import __version__
-
-from sas.system import config, user
+from sas.system import config
 
 __all__ = ['config']
 
+# TODO: fix logger-config circular dependency
 # Load the config file
 config.load()
 
--- a/src/sas/qtgui/Perspectives/Fitting/GPUOptions.py
+++ b/src/sas/qtgui/Perspectives/Fitting/GPUOptions.py
@@ -18,8 +18,8 @@ from PyQt5 import QtGui, QtCore, QtWidgets
 from sas.qtgui.Perspectives.Fitting.UI.GPUOptionsUI import Ui_GPUOptions
 from sas.qtgui.Perspectives.Fitting.UI.GPUTestResultsUI import Ui_GPUTestResults
 
-from sas.system import env
 from sas import config
+from sas.system import lib
 
 try:
     _fromUtf8 = QtCore.QString.fromUtf8
@@ -107,11 +107,8 @@ class GPUOptions(QtWidgets.QDialog, Ui_GPUOptions):
 
         sas_open_cl = self.cl_options[str(checked.text())]
         no_opencl_msg = sas_open_cl.lower() == "none"
-        env.sas_opencl = sas_open_cl
-        config.SAS_OPENCL = sas_open_cl
+        lib.reset_sasmodels(sas_open_cl)
 
-        # CRUFT: next version of reset_environment() will return env
-        sasmodels.sasview_model.reset_environment()
         return no_opencl_msg
 
     def testButtonClicked(self):
--- a/src/sas/system/lib.py
+++ b/src/sas/system/lib.py
@@ -1,0 +1,41 @@
+# Setup third-party libraries (e.g., sasview, periodictable, bumps)
+import os
+
+# TODO: Add api to control sasmodels rather than using environment variables
+def setup_sasmodels():
+    """Initialize sasmodels settings"""
+    from .user import get_user_dir
+
+    # Don't need to set SAS_MODELPATH for gui because sascalc.fit uses the
+    # full paths to models, but when using the sasview package as a python
+    # distribution for running sasmodels scripts we need to set SAS_MODELPATH
+    # to the path used by SasView to store models.
+    from sas.sascalc.fit.models import find_plugins_dir
+    os.environ['SAS_MODELPATH'] = find_plugins_dir()
+
+    # TODO: Use same mechanism as OpenCL/CUDA to manage the cache file path
+    # Both scripts and gui need to know the stored DLL path.
+    if "SAS_DLL_PATH" not in os.environ:
+        os.environ["SAS_DLL_PATH"] = os.path.join(
+            get_user_dir(), "compiled_models")
+
+    # Set OpenCL config from environment variable if it is set otherwise
+    # use the value from the sas config file.
+    from sas import config
+    # Not using sas.system.env since it just adds a layer of confusion
+    SAS_OPENCL = os.environ.get("SAS_OPENCL", None)
+    if SAS_OPENCL is None:
+        # Let sasmodels know the value of the config variable
+        os.environ["SAS_OPENCL"] = config.SAS_OPENCL
+    else:
+        # Let config system know the value of the the environment variable
+        config.SAS_OPENCL = SAS_OPENCL
+
+def reset_sasmodels(sas_opencl):
+    from sasmodels.sasview_model import reset_environment
+    from sas import config
+
+    config.SAS_OPENCL = sas_opencl
+    os.environ["SAS_OPENCL"] = sas_opencl
+    # CRUFT: next version of reset_environment() will return env
+    reset_environment()
--- a/src/sas/system/log.py
+++ b/src/sas/system/log.py
@@ -84,3 +84,9 @@ class SetupLogger(object):
                 return
         print(f"'{filename}' not found.", file=sys.stderr)
         self.config_file = None
+
+def production():
+    return SetupLogger("sasview").config_production()
+
+def development():
+    return SetupLogger("sasview").config_development()


llimeht:
Making the shipped `installers/sasview.py` closer to `run.py` is a good thing. If it can approach having absolutely no difference that would be even better.

It would be nice to see:

- proper exit code handling - at the least differentiate between success and failure, but it's easy to think that different nonzero exit codes for "command made no sense", "command failed", "data file not found", "fit didn't converge" would be nice
- unittest of the command line handling module
- functional test that runs each of the new things (`sasview -m foo` etc) and ensures that they worked (exited cleanly, produced the right result); likewise running things that cannot succeed and checking that they exit non-zero. 

A functional test of the installed version is also needed - that can build on #2263 once merged. Running those same `sasview -m foo` style tests against the installed version both tests that this new interface works and tests that the installation works.

I'd suggest writing the functional tests as a series of tests with `pytest` that fork new processes using `subprocess` which then allows them to be easily run with the same test harness. That can be quite reusable against both `run.py` and the installed sasview. (I can find examples of this from other projects if that would be useful.)

wpotrzebowski:
This PR seems to be a part of bigger discussion that we plan to have tomorrow at the code camp.

smk78:
Depending on the outcome of a Windows review of this PR a small doc change may be needed to sasmodels/doc/guide/scripting.rst.

See https://github.com/SasView/sasmodels/pull/526#discussion_r1005112188 for details.

The doc is currently assuming the command will be 'sasview', not 'SasViewCom', or both.

wpotrzebowski (commented):
I've tested this PR for functionality and it works as exptected. I will look at the code later

pkienzle:
@smk78 thanks for the report.

When you ran the script how do you know that it ran? Did it produce any output on the console? To check the argument processing you can include: `import sys; print("Input arguments:", sys.argv)` in the script, or if the console output isn't working, use `import sys; print("Input arguments:", sys.argv, file=open('script.out','w'))` (untested).

If your `-c` print test did nothing then I'm guessing that the executable is redirecting standard out to NUL. This is what I expect from a windows gui app, and why I thought we might need a separate sasviewcon executable. I was hoping they might have fixed this sometime post windows 7, but I'm guessing not?

We may need to attach a windows console when running from the exe ([docs]( https://github.com/wuxc/pywin32doc/blob/master/md/win32console.md)). Something like (untested):
```python
import os, sys
if os.name == 'nt': # Make sure we are attached to a console
    from win32 import win32console as con
    if con.GetConsoleWindow() == 0: # No console attached
        con.AllocConsole()
        con.SetConsoleTitle('SasView console')
        sys.stdin = con.GetStdHandle(con.STD_INPUT_HANDLE) # -10
        sys.stdout = con.GetStdHandle(con.STD_OUTPUT_HANDLE) # -11
        sys.stderr = con.GetStdHandle(con.STD_ERROR_HANDLE) # -12
```

This creates a new console for the interaction. I don't know if we can attach to the existing console so that we can interact with shell pipelines.

pkienzle:
Once we have a working console you can try `sasview -m bumps.cli` to check that it prints the bumps help.

smk78 (commented):
@pkienzle , there was no output to the console when the script ran, but I knew that it had run because the script (from @wpotrzebowski) was designed to produce a plot and it did.

Similarly:
```
sasview -m sasmodels.compare sphere
```
generated:
image

With your sys.argv addition our script looks like this:
```
from matplotlib import pyplot as plt
from sasmodels.core import load_model
from sasmodels.direct_model import call_kernel
from numpy import logspace
import sasmodels
import sys

print("Input arguments:", sys.argv, file=open('script.out','w'))

model=load_model('cylinder')
q=logspace(-3,-1,200)
kernel=model.make_kernel([q])
Iq=call_kernel(kernel,dict(radius=200))
plt.loglog(q,Iq)
plt.xlabel('q (1/A)')
plt.ylabel('I(q)')
plt.title('Cylinder with radius 200')
plt.show()
```
and on running it _script.out_ contains:
```
Input arguments: ['wojtek4.py']
```

It then occurred to me that you can redirect STDOUT from the command line, so I tried it with your bumps help example:
```
sasview -m bumps.cli > bumps.out
```
for which _bumps.out_ contained (encouragingly):
```
Usage: bumps [options] modelfile [modelargs]

The modelfile is a Python script (i.e., a series of Python commands)
which sets up the data, the models, and the fittable parameters.
The model arguments are available in the modelfile as sys.argv[1:].
Model arguments may not start with '-'.

Options:

    --preview
        display model but do not perform a fitting operation
    --pars=filename or store path
        initial parameter values; fit results are saved as path/<modelname>.par
    --plot=log      [linear|log|residuals]
        type of plot to display
    --trim=true
        trim any remaining burn before displaying plots [dream only]
    --simulate
        simulate a dataset using the initial problem parameters
    --simrandom
        simulate a dataset using random problem parameters
    --shake
        set random parameters before fitting
    --noise=5%
        percent noise to add to the simulated data
    --seed=integer
        random number seed
    --err
        show uncertainty estimate from curvature at the minimum
    --cov
        show the covariance matrix for the model when done
    --entropy=gmm|mvn|wnn|llf
        compute entropy on posterior distribution [dream only]
    --staj
        output staj file when done [Refl1D only]
    --edit
        start the gui
    --view=linear|log
        one of the predefined problem views; reflectometry also has fresnel,
        logfresnel, q4 and residuals

    --store=path
        output directory for plots and models
    --overwrite
        if store already exists, replace it
    --resume=path    [dream]
        resume a fit from previous stored state; if path is '-' then use the
        path given by --store, if it exists
    --parallel=n
        run fit using multiprocessing for parallelism; use --parallel=0 for all cpus
    --mpi
        run fit using MPI for parallelism (use command "mpirun -n cpus ...")
    --batch
        batch mode; save output in .mon file and don't show plots after fit
    --noshow
        semi-batch; send output to console but don't show plots after fit
    --time=inf
        run for a maximum number of hours
    --checkpoint=0
        save fit state every n hours, or 0 for no checkpoints

    --fit=amoeba    [amoeba|de|dream|lm|newton|pt|scipy.leastsq]
        fitting engine to use; see manual for details
    --steps=0       [amoeba|de|dream|lm|newton|pt|scipy.leastsq]
        number of fit iterations after any burn-in time; use samples if steps=0
    --samples=1e4   [dream]
        set steps=samples/(pop*#pars) so the target number of samples is drawn
    --xtol=1e-4     [de, amoeba]
        minimum population diameter
    --ftol=1e-4     [de, amoeba]
        minimum population flatness
    --alpha=0.0     [dream]
        p-level for rejecting convergence; with fewer samples use a stricter
        stopping condition, such as --alpha=0.01 --samples=20000
    --outliers=none [dream]
        name of test used for removing outlier chains every N samples:
          none:   no outlier removal
          iqr:    use interquartile range on likelihood
          grubbs: use t-test on likelihood
          mahal:  use distance from parameter values on the best chain
    --pop=10        [dream, de, rl, ps]
        population size is pop times number of fitted parameters; if pop is
        negative, then set population size to -pop.
    --burn=100      [dream, pt]
        number of burn-in iterations before accumulating stats
    --thin=1        [dream]
        number of fit iterations between steps
    --nT=25
    --Tmin=0.1
    --Tmax=10       [pt]
        temperatures vector; use a higher maximum temperature and a larger
        nT if your fit is getting stuck in local minima
    --CR=0.9        [de, rl, pt]
        crossover ratio for population mixing
    --starts=1      [newton, rl, amoeba]
        number of times to run the fit from random starting points.
    --keep_best
        when running with multiple starts, restart from a point near the
        last minimum rather than using a completely random starting point.
    --init=eps      [dream]
        population initialization method:
          eps:    ball around initial parameter set
          lhs:    latin hypercube sampling
          cov:    normally distributed according to covariance matrix
          random: uniformly distributed within parameter ranges
    --stepmon
        show details for each step
    --resynth=0
        run resynthesis error analysis for n generations

    --time_model
        run the model --steps times in order to estimate total run time.
    --profile
        run the python profiler on the model; use --steps to run multiple
        models for better statistics
    --chisq
        print the model description and chisq value and exit
    -m/-c/-p command
        run the python interpreter with bumps on the path:
            m: command is a module such as bumps.cli, run as __main__
            c: command is a python one-line command
            p: command is the name of a python script
    -i
        start the interactive interpreter
    -?/-h/--help
        display this help
```

pkienzle:
@smk78 please try your modified wojtek4.py with `sasview wojtek4.py extra args` to verify that it produces `['wojtek4.py', 'extra', 'args']`.

I'm surprised that the `> bumps.out` redirection worked. I wonder if there are any tricks to allow input redirection from the console?

Meanwhile, can you try the following `test_cli.py`:
```python
import os, sys
if os.name == 'nt': # Make sure we are attached to a console
    from win32 import win32console as con
    if con.GetConsoleWindow() == 0: # No console attached
        con.AllocConsole()
        con.SetConsoleTitle('SasView console')
        sys.stdin = con.GetStdHandle(con.STD_INPUT_HANDLE) # -10
        sys.stdout = con.GetStdHandle(con.STD_OUTPUT_HANDLE) # -11
        sys.stderr = con.GetStdHandle(con.STD_ERROR_HANDLE) # -12

import code
code.interact(local={'exit': sys.exit})
```

This should start a new console with a python prompt which will I hope go away when you type `exit()`.

pkienzle:
If you don't need input you might be able to pipe to type:
```cmd
sasview -m bumps.cli | type
```

This might not work (I don't know if `type` accepts a stdin pipe), but you could use `| more` to put in a pager or within powershell `| tee` or perhaps `| tee -FilePath NUL` to echo it to the console while simultaneously saving in the great windows bit bucket.

smk78:
> @smk78 please try your modified wojtek4.py with `sasview wojtek4.py extra args` to verify that it produces `['wojtek4.py', 'extra', 'args']`.

Yes, it does!

> Meanwhile, can you try the following `test_cli.py`:
>
> ```python
> import os, sys
> if os.name == 'nt': # Make sure we are attached to a console
>     from win32 import win32console as con
>     if con.GetConsoleWindow() == 0: # No console attached
>         con.AllocConsole()
>         con.SetConsoleTitle('SasView console')
>         sys.stdin = con.GetStdHandle(con.STD_INPUT_HANDLE) # -10
>         sys.stdout = con.GetStdHandle(con.STD_OUTPUT_HANDLE) # -11
>         sys.stderr = con.GetStdHandle(con.STD_ERROR_HANDLE) # -12
> 
> import code
> code.interact(local={'exit': sys.exit})
> ```
> 
> This should start a new console with a python prompt which will I hope go away when you type `exit()`.

As expected, if I run this as _python test_cli.py_ it invokes Python (in the same console) and I can exit().

If I run this as _sasview test_cli.py_ this happens:
image
and I get this whether or not I invoke my Anaconda developers environment.

pkienzle:
The error says that win32 is not included in setupSasview.exe. I fiddled the .spec file so maybe it will be included now.

smk78:
@pkienzle , I just downloaded and installed the latest artefact for this branch. I get a slightly different error this time:
image

installers/sasview.spec
```diff
@@ -65,6 +66,10 @@ hiddenimports = [
     'uncertainties',
 ]
 
+if os.name == 'nt':
+    # Need win32 to run sasview from the command line.
+    hiddenimports.append('win32')
```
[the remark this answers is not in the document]
llimeht: btw reading https://github.com/pyinstaller/pyinstaller/issues/6244 makes me wonder if this is actually going to be a solvable problem on windows.

src/sas/cli.py
```diff
@@ -0,0 +1,90 @@
...
+        exec(args.command)
+    elif args.interactive: # -i
+        if args.argv or args.module or args.command:
+            exclusive_error()
+        run_interactive()
+    else:
+        from sas.qtgui.MainWindow.MainWindow import run_sasview as run_gui
+        run_gui()
+
+if __name__ == "__main__":
```
llimeht, line 87: What's the purpose of this entry point as opposed to `run.py` and `installers/sasview.py`. This seems like a good opportunity to reduce the number of entry points into the code; ideally there would be only 1 for ease of maintenance.

src/sas/cli.py
```diff
@@ -0,0 +1,90 @@
...
+    lib.setup_sasmodels()
+
+    # Parse mutually exclusive command line options
+    # mutually exclusive (-m module args, -c command, -i, script args)
+    if args.argv and not args.module: # script [arg...]
+        import runpy
+        if args.command or args.module or args.interactive:
+            exclusive_error()
+        sys.argv = args.argv
+        runpy.run_path(args.argv[0], run_name="__main__")
```
llimeht, line 68: Error handling from `runpy.run_*` functions is limited to exception handling. Raising an exception in the called code will rattle back and lead to this exiting non-zero. It's probably enough, although all failures will give the same exit code. If the function just returns `False` or `None` or any other common signal of failure, it will look like it has passed OK.

It probably needs to be documented that the custom script should raise an exception in the case of an error, not just return.

llimeht (changes requested):
Promising work :)

There's a few comments I've made inline. There's also three general areas to address:
- unittesting of the new code
- functional testing of the new commandline interface
- documentation of the commandline interface

smk78:
@llimeht commented: 
> There's a few comments I've made inline. There's also three general areas to address:
> 
> * documentation of the commandline interface

Documentation is being handled in https://github.com/SasView/sasmodels/pull/526

run.py
```diff
@@ -46,62 +44,45 @@ def cd(path):
...
-    except ImportError:
-        addpath(joinpath(root, '..', 'periodictable'))
+    # Find the directories for the source and build
+    root = abspath(dirname(realpath(__file__)))
 
-    try:
-        import bumps
-    except ImportError:
-        addpath(joinpath(root, '..', 'bumps'))
+    # TODO: Do we prioritize the sister repo or the installed package?
```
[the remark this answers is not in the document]
smk78: Out of curiosity, where/when is PYTHONPATH set. I don't have it set...

llimeht:
> @llimeht commented:
> 
> > There's a few comments I've made inline. There's also three general areas to address:
> > 
> > * documentation of the commandline interface
> 
> Documentation is being handled in [SasView/sasmodels#526](https://github.com/SasView/sasmodels/pull/526)

That obviously needs to grow to include the things other than `sasview foo.py` (i.e `-m`, `-c`, `-i`). These are handy options to have available.

run.py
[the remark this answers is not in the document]
llimeht: @smk78, `run.py` is full of calls to `addpath` which is a function that is fiddling with `PYTHONPATH` (and/or `sys,path`). The custom test runners used to have lots of that too and they have been slowly retired, which helps.

krzywon:
I'm getting a different error than you, @smk78. As a sanity check, I tested the latest build on main (as of commit [ba34b5b](https://github.com/SasView/sasview/commit/ba34b5be2433e1b06d66a06591e113b102ec8129)) and the app launches as expected.
image

src/sas/qtgui/MainWindow/MainWindow.py
```diff
@@ -1,11 +1,10 @@
 # UNLESS EXEPTIONALLY REQUIRED TRY TO AVOID IMPORTING ANY MODULES HERE
 # ESPECIALLY ANYTHING IN SAS, SASMODELS NAMESPACE
-import logging
 import os
 import sys
+import logging
 
-from sas import config
```
llimeht, line 7: there are still uses of `config` in this file; lines 77-80 for the `QT_*` variables.

src/sas/system/env.py
```diff
@@ -1,3 +1,4 @@
+# ** DEPRECATED **
```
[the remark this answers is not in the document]
pkienzle: This one is relevant. The PR removes the need for sas.system.env. I'm happier deleting rather than deprecating.

src/sas/system/__init__.py
```diff
@@ -1,3 +1,4 @@
+# TODO: Don't shadow module names; it makes debugging difficult.
```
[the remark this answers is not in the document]
pkienzle: You are right of course, and I already added a ticket. Having just wasted a bunch of time trying to figure out why my imports weren't working I was peeved and wanted to leave a mark in the code as well. I'll remove it.

installers/sasview.spec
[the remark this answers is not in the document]
pkienzle: Looking at the win32 api it seems we should be able to attach to an existing console, so it is just a matter of finding the right one. Worst case we need to create a SasViewCom.exe for windows to run scripts.

src/sas/cli.py
```diff
@@ -0,0 +1,90 @@
...
+    elif args.module: # -m module [arg...]
+        import runpy
+        if args.command or args.interactive:
+            exclusive_error()
```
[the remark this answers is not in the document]
pkienzle: Agreed. I was thinking that -c was exclusive of argv but it's not:
```
$ python -c "import sys; print(sys.argv)" extra args
['-c', 'extra', 'args']
```
Interesting that "-c" is the "program".

smk78:
@llimeht I've just added a section on the command line syntax to scripting.rst

src/sas/cli.py
pkienzle: This is the real entrypoint. `run.py` and `sasview.py` prepare the environment for running.

In future when we allow `pip install sasview` then we can start it with `python -m sas.cli`. This can be handy from shell scripts and make where you replace `python` with `$PYTHON` or `sys.executable` in python scripts. I've used this style for a number of applications such as sphinx, nosetest and ipython. It's also handy when "borrowing" an interpreter from an environment.

Maybe we should have `sas/__main__.py` so that we can use `python -m sas` to run sasview?

src/sas/cli.py
pkienzle: How is this behaviour different from `python script.py`? Scripts don't return error codes unless they include `sys.exit(nonzero)`.

src/sas/qtgui/MainWindow/MainWindow.py
pkienzle: Merge error. Fixing.

I might move the QT_* stuff to cli.py. Are we going to have scripts that invoke Qt that need these variables set?

run.py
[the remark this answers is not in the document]
pkienzle: The point of run.py is to set up the python when running in development mode. I often set up parallel directories for different branches and don't want to set up a complete python environment when I do so. This will be increasingly the case if we go to a fork rather than branch workflow for PRs.

I would think you will always want the equivalent for testing. How else do you run the tests before installing?

src/sas/cli.py
```diff
@@ -0,0 +1,90 @@
...
+    """Run sasview as an interactive python interpreter"""
+    try:
+        from IPython import start_ipython
+        sys.argv = ["ipython", "--pylab"]
+        sys.exit(start_ipython())
+    except ImportError:
+        import code
+        code.interact(local={'exit': sys.exit})
+
+def main(logging="production"):
```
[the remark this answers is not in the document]
pkienzle: This function needs to hack `sys.argv` to get things like bumps and ipython to work properly, so passing in args would be something of a fiction. I could write a context manager `with push_argv(args):` to restore on exit, but then so could the hypothetical test harness, so I'll leave it as is until needed.

pkienzle:
Beginnings of a test suite:
```sh
$ echo "print(f'{a=}')" | python run.py -iqV -c "a=15"
SasView 5.0.5
>>> a=15
>>> 
```

pkienzle:
@smk78 I rewrote things to be closer to the python interface, with -i for interactive, -q for quiet and -V for version number. Now when you run a script/module/command and have the -i option on the command line it will drop you into interactive mode with the variables defined in the script. With the -q it doesn't print the startup banner.

I've documented this in sas.cli. You should be able to link to it with something like:
```rst
SasView can be used from the command line.
... some examples here ...
For more details see :mod:`sas.cli`.
```

Feel free to edit the docs in sas/cli.py on this branch. Or alternatively move them somewhere in the sasview doc tree, and drop a reminder in the top of sas/cli.py that docs/guide/scripting.rst or whatever needs to be kept up to date.

smk78:
@pkienzle :
Functionality test of SasView-Installer-windows-latest-3.8-e75485ff5ce59b6e6b2c6f70d9917176665bd8ec built from this branch with your overnight commits.

```> sasview```
Invokes the GUI

```> sasview wojtek4.py```
Works

```> sasview -m sasmodels.compare sphere```
Works

```> sasview -m sasmodels.compare sphere > test.out```
Works and wrote to test.out

```> sasview -c "print("hello")"```
```> sasview -c "print("hello")" > test.out```
**Does not work**

```> sasview -V```
Invokes the GUI

```> sasview -V > test.out```
Invokes the GUI and writes 'SasView 5.0.5' to test.out

```> sasview -i```
Did this...
image

```> sasview test_cli.py```
(see https://github.com/SasView/sasview/pull/2280#issuecomment-1294821889 for script)
Did this...
image

**Comment by smk78** [comment]
> I've documented this in sas.cli.
> 
> Feel free to edit the docs in sas/cli.py on this branch. Or alternatively move them somewhere in the sasview doc tree, and drop a reminder in the top of sas/cli.py that docs/guide/scripting.rst or whatever needs to be kept up to date.

@pkienzle , I have linked to *sas/cli.py* from *scripting.rst* as you suggest but also modified the doc strings in *sas/cli.py*. BTW, the consensus in the room was that -x was a flag, not an option! :-)

**Comment by pkienzle** [comment]
> BTW, the consensus in the room was that -x was a flag, not an option!

For the current set of `-iqV`, but we could have options that are not flags such as `--config ~/.sasview/nogpu.json`

**Comment by pkienzle** [comment]
No gui if work given or interactive interactive is specified. Work includes:
* command `sasview -c "python statement"`
* module as main "sasview -m dotted.name" args...`
* script "sasview script args...`

It is interactive if '-i' included. That means after executing the commands, drop into a command prompt with whatever variables were defined by the command/module/script. If interactive and no work. When the interactive prompt starts it prints python version, etc. (the "banner"). The `-q` option suppresses the banner. Without `-i` the `-q` does nothing.
 The banner ought to show SasView version as well as python version.

So `sasview -q`, which specifies no work, should start the GUI. I considered treating `-q` as `-i` without banner but this is different from the behaviour of the python interperter.

`sasview -V` should exit after printing version. Starting the GUI is a bug.  Incidentally, if `-V` exits immediately this makes it an option rather than a flag ;-)

**Comment by smk78** [comment]
The last commit circumvents the sasmodels doc build breaking the Ubuntu CI builds.
See https://github.com/SasView/sasmodels/pull/532

**Comment by pkienzle** [comment]
Debugging the windows console issues will require somebody with a windows box and lots of time.

This PR already provides some utility even on windows. We can merge now and open another ticket to fix the console interaction. I already added the comment that interactive doesn't work on windows.

Or we abandon this approach and build a new installer based on embedded python.

For those wishing to continue this PR, there is a promising stackoverflow thread [here](https://stackoverflow.com/questions/5711291/get-the-handle-and-write-to-the-console-that-launched-our-process).

Another wrinkle: We need to check whether any I/O redirection is in effect before overriding the corresponding stdin/stdout/stderr.


Or

**Comment by wpotrzebowski** [comment]
I would vote for merging it after resolving the conflict

**Comment by butlerpd** [comment]
Need to discuss the yml files with @pkienzle - they no longer exist and should probably be deleted from his branch? Any changes needed should be put into ci.yml

**Comment by butlerpd** [comment]
looks like the conflicts are resolved. Can we merge now or are we waiting for @llimeht? As I understand the comments so far, particularly from @wpotrzebowski and @smk78,  the idea is to merge this to provide improved functionality and let another PR "finish" getting "full functionality?"  or is there more to do here?

**Comment by smk78** [comment]
Do we know how the core dump @pkienzle reports was triggered, and if it is now fixed?

If it is sorted, and there are no outstanding conflicts, then I think this can be merged. But I would then like to test again (in the wake of all the other changes that have been made all over the place). Some further furtling of scripting.rst may be required depending on what I find.

File: .github/workflows/ci.yml [hunk 10]
```diff
@@ -80,6 +80,11 @@ jobs:
           python -m pip install matplotlib~=3.5.2
           python -m pip install lxml
 
+      - name: Install pywin32 (Windows + Linux)
+        if: ${{ startsWith(matrix.os, 'windows') }}
+        run: |
```
llimeht, line 85: Step name does not match condition.

installers/sasview.spec
```diff
@@ -1,6 +1,7 @@
 # -*- mode: python ; coding: utf-8 -*-
 
 import sys
+import os
```
llimeht, line 4: Unneeded import?

run.py
```diff
@@ -46,62 +44,45 @@ def cd(path):
...
+    # TODO: Can we use sasview/run.py from a distributed sasview.exe?
+    # Put supporting packages on the path if they are not already available.
+    for sister in ('periodictable', 'bumps', 'sasdata', 'sasmodels'):
```
llimeht, line 60: Would prefer 'sibling' rather than 'sister'.

src/sas/cli.py
```diff
@@ -0,0 +1,171 @@
...
+#
+"""
+**SasView command line interface**
+
+This functionality is under development. Interactive sessions do not yet
+work in the Windows console.
+
+**Usage:**
+
+sasview [flags]
```
llimeht, line 12: What does this look like once rendered by sphinx? Would expect to have seen some more formatting on these to make a good looking output following the normal conventions for commands and options,  but perhaps it looks ok as it is.

src/sas/cli.py
```diff
@@ -0,0 +1,171 @@
...
+sasview -V
+    *Print sasview version and exit.*
+
+**Flags:**
+
+    -i, --interactive. *Enter an interactive session after command/module/script.*
+
+    -q, --quiet. *Suppress startup messages on interactive console.*
+
+Note: On Windows any console output gets written to NUL by default. If
```
llimeht, line 35: Is NUL the Windows equivalent of `/dev/null`? Will any user know that?

src/sas/cli.py
llimeht: Either this is an unneeded `__main__` that no-one should call, or it is yet another `__main__`. sasview.py and run.py aren't calling this entry point, so this is adding another entry point. Reducing the entry points is good as you say, but this isn't doing so... yet.

I'm cautious of scope creep - this PR is not about fixing the bug that there is a multiplicity of ways of starting sasview. I'm also cautious about adding yet another way of starting sasview to the set, which then needs testing and maintenance.

src/sas/qtgui/MainWindow/MainWindow.py
```diff
@@ -1,11 +1,10 @@
 # UNLESS EXEPTIONALLY REQUIRED TRY TO AVOID IMPORTING ANY MODULES HERE
 # ESPECIALLY ANYTHING IN SAS, SASMODELS NAMESPACE
-import logging
 import os
 import sys
+import logging
```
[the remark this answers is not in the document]
llimeht: Still just looks like reordering stdlib imports...

src/sas/system/__init__.py
```diff
@@ -1,6 +1,5 @@
 from .web import web
 from .legal import legal
-from .env import env
 from .config.config import config
 
-__all__ = ["web", "legal", "env", "config"]
\ No newline at end of file
+__all__ = ["web", "legal", "config"]
```
llimeht, line 5: Minor nit: missing line end on last line of file.

src/sas/cli.py
pkienzle: No idea. I can't get the docs to build on my machine.

wpotrzebowski:
@krzywon and @rozyczko will try to look at it

krzywon (approved):
For some reason, I can only get this to work from my IDE (PyCharm in my case), not the Windows command prompt. Running from the command line, the SasView GUI runs regardless of the arguments passed, and logging no longer works. Likely PEBCAK, so I'm approving.

.github/workflows/ci.yml
```diff
@@ -80,6 +80,11 @@ jobs:
           python -m pip install matplotlib~=3.5.2
           python -m pip install lxml
 
+      - name: Install pywin32 (Windows)
+        if: ${{ startsWith(matrix.os, 'windows') }}
+        run: |
+          python -m pip install pywin32
+
```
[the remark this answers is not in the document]
lucas-wilkins: We'll see what it looks like when #2393 is more complete

butlerpd:
@krzywon ok with this but would like a second look by someone else? would be good if @rozyczko could also take a look?

butlerpd:
@rozyczko will take the second look. If OK, @smk78 points out remember the doc change before merging

smk78:
Ohhhhh! I just discovered something relevant to this issue!
If you are building locally in a Conda environment on Windows, after checking out your _SasView_ branch, you **must** do
```
python setup.py build install
```
The 'install' is important. If you don't do this then _sasview-script.py_ and _sasview.exe_ in _<path>\Anaconda3x64\envs\dev\Scripts_ don't get updated and so when you type 'sasview' to invoke the program you may not be trying to run the version of _SasView_ you think you are!

smk78:
So as we are considering pushing this out with 5.0.6 I thought I would give this branch another walk around the block. On Windows I'm seeing exactly the same behaviour that @krzywon reported [above](https://github.com/SasView/sasview/pull/2280#pullrequestreview-1197655116), typing 'sasview' (or 'sasview' followed by any of the switches or parameters shown above) starts the program but does nothing else.

I can tell 'sasview' is not invoking one of my releases by mistake because the iconised toolbar is showing in the program when it starts, and to be on the safe side I deleted all other non-release copies of SasView on my machine except for this branch.
Thus, unfortunately, it seems that something somewhere has broken.

smk78:
PS I triggered a GitHub Action build to generate an installer in case the problem is with my Conda environment, but the build is hanging on the doc build on ubuntu-latest 3.10 for some reason...

butlerpd:
@butlerpd or @krzywon to ask @pkienzle about this

butlerpd:
Apparently the conflict resolution was done incorrectly -- SasView no longer starts due to missing `env` variable. 
`NameError: name 'evn' si not defined`

envErrorTraceback

src/sas/cli.py
```diff
@@ -0,0 +1,171 @@
...
+
+    context = {'exit': sys.exit}
+    if cli.module: # -m module [arg...]
+        import runpy
+        # TODO: argv[0] should be the path to the module file not the dotted name
+        sys.argv = [cli.module, *cli.args]
+        context = runpy.run_module(cli.module, run_name="__main__")
+    elif cli.command: # -c "command"
+        sys.argv = ["-c", *cli.args]
+        exec(cli.command, context)
```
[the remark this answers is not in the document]
pkienzle: From stackoverflow [here](https://stackoverflow.com/questions/334879/how-do-i-get-the-application-exit-code-from-a-windows-command-line/11476681#11476681) need to use "start /wait sasview" to capture the errorlevel to the windows console for windowed applications. For example:

```
>sasview -c "import sys; sys.exit(1)"
>echo %errorlevel%
1
```

pkienzle:
Running the following as `sasview test_cli.py` starts an interactive interpreter. It is still opening in a separate console.

test_cli.py:
```
import os, sys
def _attach_windows_console():
    import win32.win32console as con
    if con.GetConsoleWindow() == 0: # No console attached
        con.AllocConsole()
        con.SetConsoleTitle('SasView console')
def _windows_redirect(device, mode):
    _attach_windows_console()
    return open(device, mode)
class WindowsConsoleOut:
    def __init__(self):
        self.fd = None
    def write(self, *args, **kwargs):
        if self.fd is None:
            self.fd = _windows_redirect("CONOUT$", "w")
        out = self.fd.write(*args, **kwargs)
        self.fd.flush()
        return out
class WindowsConsoleIn:
    def __init__(self):
        self.fd = None
    def read(self, *args, **kwargs):
        if self.fd is None:
            self.fd = _windows_redirect("CONIN$", "r")
        return self.fd.read(*args, **kwargs)
    def readline(self, *args, **kwargs):
        if self.fd is None:
            self.fd = _windows_redirect("CONIN$", "r")
        return self.fd.readline(*args, **kwargs)
if os.name == 'nt': # Make sure we are attached to a console
    if sys.__stdin__ is None:
        sys.__stdin__ = sys.stdin = WindowsConsoleIn()
    if sys.__stdout__ is None:
        sys.__stdout__ = sys.stdout = WindowsConsoleOut()
    if sys.__stderr__ is None:
        sys.__stderr__ = sys.stderr = WindowsConsoleOut()

import code
code.interact(local={'exit': sys.exit})
```

pkienzle:
More progress. The popup console hangs around until the user dismisses it, giving them a chance to read the output. Code is in `src/sas/system/console.py`.

It is still not complete. The popup console is not recognized as the windows console so the python interactive loop doesn't work properly. In particular, `PyOS_InputHook` isn't called, so interactive graphs don't work properly from the console. You need to call `plt.pause(t)` for some number of seconds `t` so that plots render and you can interact with them.

Also, despite using lazy console creation (only on read/write from stdin) the console window was popping up when running sasview.exe as a GUI. It should only be appearing if the program reads from or writes to the console.

Because of the above problems I don't enable it by default. Instead scripts need to do the following:
```
from sas.system.console import setup_console
setup_console()
```

A better solution is to create a separate console app (sasbatch.exe?) so this hack won't be needed.

pkienzle:
Yes, this is expected behaviour because I'm not calling `setup_console` on startup. It still needs some work.

I added a new option `-o` to open the output console. Download the new version and try:
```
sasview -o -c "print('hello')"
sasview -o -m bumps.cli
sasview -o wojtek4.py extra args
sasview -o -V
sasview -i
```

Even with the new option it will still be unusable because plots don't work properly. You need to explicitly give them a chance to run using the `plt.pause(t)` command:
```
import matplotlib.pyplot as plt
plt.ion()
plt.plot([1, 2, 1])
plt.pause(0.1)
```

The fix will be to somehow trigger `PyOS_InputHook` in CPython `Parser/myreadline.c`, presumably by making `sys.stdin` into a value that is recognized as a windows console handle. This will presumably allow things like readline and ipython to work properly as well, so a worthwhile effort.

Since the `-o` option is a temporary hack I didn't document it.

pkienzle:
Looking at the CPython code for input we also need to define `fileno`, `encoding` and `errors` methods for the files. These is needed for stdin and stdout separately so we will need to split `sas.system.console.WindowsConsole` into an input console and an output console.

In particular, CPython checks that the fileno for python stdin matches that for the stdin from the C library. I'm betting that the windows app startup process leaves these empty so we may have to hack msvcrt to set them to the console buffer, but with luck it will be as easy as the following:
```
import sys
import msvcrt
from win32 import win32console as con
con.AllocConsole()
sys.__stdin__ = sys.stdin = open("CON:", "r")
sys.__stdout__ = sys.stdout = open("CON:", "w")
con.SetStdHandle(con.STD_INPUT_HANDLE, msvcrt.get_osfhandle(sys.stdin.fileno())) # -10
con.SetStdHandle(con.STD_OUTPUT_HANDLE, msvcrt.get_osfhandle(sys.stdout.fileno())) # -11
import code; code.interact(local={'exit': sys.exit})
```

---

CPython code for `input()`:

https://github.com/python/cpython/blob/f80db6cef075186f888a85258ccf2164bf148921/Python/bltinmodule.c#L2063-L2276

Docs for msvcrt: https://docs.python.org/3/library/msvcrt.html

run.py
```diff
@@ -46,62 +44,45 @@ def cd(path):
...
-    except ImportError:
-        addpath(joinpath(root, '..', 'periodictable'))
+    # Find the directories for the source and build
+    root = abspath(dirname(realpath(__file__)))
 
-    try:
-        import bumps
-    except ImportError:
-        addpath(joinpath(root, '..', 'bumps'))
+    # TODO: Do we prioritize the sibling repo or the installed package?
```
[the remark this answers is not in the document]
pkienzle: Currently it is prioritizing the installed module. We should change that...

Also, I'm using a bare except!

smk78:
> I added a new option `-o` to open the output console. Download the new version and try:
> 
> ```
> sasview -o -c "print('hello')"
> sasview -o -m bumps.cli
> sasview -o wojtek4.py extra args
> sasview -o -V
> sasview -i
> ```

Testing artefact 4297625493 of Feb 28 on W10/x64:
```
sasview -o -c "print('hello')"
```
Spawns a new console and prints 'hello'
image
```
sasview -o -m bumps.cli
```
Spawns a new console and prints the bumps cli help
image
```
sasview -o wojtek4.py extra args
```
Spawns a console and executes the script which includes generating a plot
image
```
sasview -o -V
```
Spawns a new console and prints the version of SasView.
image
```
sasview -i
```
Spawns an interactive Python console
image

pkienzle:
I added the `-o` option to the doc string. It works well enough and I'm not going to try to make it work without the option.

We wouldn't need the option if it only attached the console when something is written to it. I tried this but I found that it was popping up even when nothing was written. I didn't spend any time trying to debug it.

butlerpd:
Are we ready to merge this then?! @smk78 and others? It  sounds like it is working even better than at code camp which we said was "good enough" ?

smk78:
> Are we ready to merge this then?! @smk78 and others? It sounds like it is working even better than at code camp which we said was "good enough" ?

I would say from a functionality perspective, yes. But I guess we need to hear from @pkienzle .

Also @llimeht had quite a few comments. I don't know how many of those are still valid/current?

krzywon (approved):
Code looks fine. Works as expected.

PR #2281, opened 2022-10-26
Generalise pyinstaller spec paths 

llimeht:
The pyinstaller spec file sometimes needs to contain workarounds for packages to help pyinstaller find all the relevant imports or data files. If such introspection is still needed, we should avoid baking in platform-dependent and python-version-dependent paths. 

This PR makes use of the `sysconfig` module from the [Python stdlib](https://docs.python.org/3/library/sysconfig.html) that has information about where packages are located, so the details of `lib` vs `Lib`, `site-packages` vs `dist-packages`, `python3.8` vs `python3.10`,… all go away. This last one is very important as it means we can create installer artifacts with multiple different Python versions for testing purposes, without needing to change the spec file.

I tested an earlier version of this patch with Windows and Linux when it was still touching paths for those installers. With other changes, this code path is currently only used on macos which I can't test.

There's every chance that the actual code for `jedi` and `zmq` for Darwin can also be removed; that's a problem for a different PR and for a mac user to test. In any case, the stub of loading and using `sysconfig` may as well stay for the time being since it will be needed some time in the future, no doubt.

### DIFF
--- a/installers/sasview.spec
+++ b/installers/sasview.spec
@@ -4,10 +4,10 @@ import sys
 from pathlib import Path
 import warnings
 import platform
-import sys
+import sysconfig
 
 block_cipher = None
-PYTHON_LOC = sys.exec_prefix
+PYTHON_PACKAGES = sysconfig.get_path('platlib')
 
 datas = [
     ('../src/sas/qtgui/images', 'images'),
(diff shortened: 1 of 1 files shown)


installers/sasview.spec
```diff
@@ -4,10 +4,10 @@ import sys
 from pathlib import Path
 import warnings
 import platform
-import sys
+import sysconfig
 
 block_cipher = None
-PYTHON_LOC = sys.exec_prefix
+PYTHON_PACKAGES = sysconfig.get_path('platlib')
```
rozyczko, line 10: Querying a random package is a bit iffy, I guess, even if it is a core package.
It seems that `site.getsitepackages()` results are platform dependent, so probably `distutils` has the best query for `site_packages` location

```
from distutils.sysconfig import get_python_lib
PYTHON_PACKAGES  = get_python_lib()
```

rozyczko (approved):
Querying core module location is a good check but it is probably better to make a check which does not require specific module name.

wpotrzebowski (approved):

Issue #1753, opened 2021-01-18
Fix Confusing Custom Slit Resolution:


butlerpd:
### Background
The slit smear resolution function in SasView is designed for the “infinite slit” case of, for example, a Bonse Hart or Kratky camera.  It assumes the resolution is essentially a “flat top” over a q range that is much larger than the maximum q of the data (typically an order of magnitude or more) in the direction perpendicular to the “scan” direction (the slit height) while being negligible in the direction parallel to the scan direction.  A second order correction has been seen to be sometimes useful to account for a finite slit “width” compared to the slit height (see @ajj ).  This should however be small. Note that this is how it is defined in the NXcanSAS standard as well.

### Current SasView Status
SasView does not currently support resolution functions with non azimuthally symmetric components, and, in particular, does support the more complicated finite slits that some modern instruments are beginning to implement such as the NCNR VSANS.  It also does not support non Gaussian resolutions even if azimuthally symmetric such as is often encountered on TOF instruments. Note that at this time the NXcanSAS format does not provide for these either though there is a canSAS effort to do so.

### The Problem
There has been considerable confusion over the years as to what the SasView code does or should do witnessed by the use of qx and qy (leading to people confusing real space slit orientation in 2D with inverse space), q_radial and q_phi (not really correct except as it intersects the above definition), and q_parr and q_perp (correct if assuming parallel and perpendicular to the scan direction but not being defined is often confused with other things such as parallel and perpendicular to the q in 2D or parallel and perp to an alignment direction in real space)

More importantly, the users are not properly guided and can produce incorrect results by entering, for example, a large width vs height or entering a true rectangular resolution which is not appropriate. In the first case the result of course is a flat line Intensity vs Q corresponding to the average scattering intensity.

### Solution
The situation would be significantly improved as follows (particularly points 1,3 and 4): 

- [ ] capture user error when incorrectly providing a qwidth larger than qheight and
  - [ ] swap the two
  - [ ] provide a pop up error message to warn the user of the incorrect usage (refer them to the documentation) and the fact that the values have been swapped
- [ ] Ideally, check for rectangular apertures where q_height <10 q_width or q_height <qmax
  - [ ] convert to to a sqrt(q_width^2+q_height^2) and send to normal resolution function
  - [ ] pop up warning of what was done and why
- [ ] replace all nomenclature in the code of qx and qy, qradial an qphi etc with q_height and q_width
- [ ] Update the documentation
[...]

PR #2283, opened 2022-10-26
fix confusing slit resolution parameters and add guardrails

pbeaucage:
As discussed at length in #1753, this adds several guardrails to prevent incorrect use of the 1D slit-smearing function.  Specifically, adds LOUD warnings if a user specifies a slit length that is less than the width, a slit geometry with an aspect ratio less than 10:1, or a slit length less than q_max for the data.

This also changes the nomenclature throughout from the confusing qx_width and qy_width to the canSAS standard q_length and q_width, where length is the long axis of the slit and width the short axis.

This depends on and is linked to issue sasview/sasmodels#528 and PR sasview/sasmodels#530


lucas-wilkins (approved):
Seems fine

butlerpd (changes requested):
The severity of the log message (critical) is not really conveyed by the look of the popup. It there not a standard "dire" popup in QT? at least something with a red box with an x in it?  Also I'm wondering if the request in the issue that the length be larger than the qmax is not a bit aggressive?  Finally, I think that length is actually the half length of the slit? Even so, USANS would not actually qualify (and doubt USAXS would either?

I suggest we remove the requirement for length to be larger than qmax and merge this as better than before and then look more deeply into whether we should be limiting the qmax to smaller than length?  Also some more indication of the importance of the pop up would be good.

src/sas/qtgui/Perspectives/Fitting/SmearingWidget.py
```diff
@@ -384,16 +386,48 @@ def onSlitSmear(self):
...
+        elif d_length < 10 * d_width:
+            msg = f'Slit length specified which is less than 10 x slit width. ' \
+                  f'This is not slit-smearing (at least not in the form we implement). \n' \
+                  f'Use pinhole smearing instead.'
+            smearing_error_flag = True
+        elif d_length < q_max:
+            msg = f'Length specified which is less than q_max for the data. \n' \
+                  f'This is not covered by existing smearing model. \n' \
+                  f'Use pinhole smearing instead. '
+            smearing_error_flag = True
```
pbeaucage, line 416:
```suggestion
        # elif d_length < q_max:
        #     msg = f'Length specified which is less than q_max for the data. \n' \
        #           f'This is not covered by existing smearing model. \n' \
        #           f'Use pinhole smearing instead. '
        #    smearing_error_flag = True
```

butlerpd (approved):
works from development environment and cannot really test installer since it depends on the sasmodels as well so merging now.  We may want to look again at the popup later.

PR #2284, opened 2022-10-26
Fixing typo in build_tools/requirements.txt

wpotrzebowski:
Fixing typo in the pytest-mock. The bigger question is we can put this file under CI i.e. replace pip lines in ymls with `pip install -r build_tools/requirements.txt`?

### DIFF
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -3,7 +3,7 @@ scipy==1.7.3
 docutils 
 pytest
 pytest_qt
-pyteset-mock
+pytest-mock
 unittest-xml-reporting
 tinycc 
 h5py 


llimeht:
Err yes. Well spotted and good fix.

Yes, using `requirements.txt` from CI is a good idea (and unifying CI as I'm trying to do in #2263 will help with that step). 

However, some difficulties for the moment on that:
- Installing qt and opencl needs to be done in a different way and in a different order on each platform. 
- Overly aggressive version pinning in this project and its dependencies means pip can't actually install all the packages needed in one transaction; multiple pip runs are required to first install a package, then downgrade it while then violating some of the pins. To underline the full horror of the situation, sasmodels and sasview have conflicting version pins through transitive dependencies (unittest-xml-reporting, numba, matplotlib, setuptools)

So "yes" but I suspect not just yet :(

llimeht (approved):

PR #2285, opened 2022-10-26
Set high DPI settings earlier

lucas-wilkins:
This fixes some high DPI issues on my machine. Can we check it on some other machines (seems fine on mine and Piotr's machines).

Particularly Mac @wpotrzebowski


pbeaucage:
Unfortunately appears to not work well on my Win10 box, as a wild guess maybe the origin of the problem is intermediate scaling settings?  I'm using a 14" laptop screen driven at 1080p but scaled to 150% using windows scaling settings.  I suspect 200% would work well as that's about what I'm getting.  

Side by side screenshot of this build next to PyCharm:
image

Main for comparison:
image

wpotrzebowski:
The legends are tiny
Screenshot 2022-10-26 at 20 32 39

lucas-wilkins:
Resigned to this solution for now!

Scaling on plots may or may not be fixed by #1337

pbeaucage:
So this still breaks it on my 150% windows box.  Can we make QT_AUTO_SCALE_FACTOR a preference as well?

Also, given that preference panels are now ready, and the app is more or less unusable on a 14" laptop with 150% scaling by default, can we put a preference panel in this before merging?

lucas-wilkins:
> So this still breaks it on my 150% windows box. Can we make QT_AUTO_SCALE_FACTOR a preference as well?
> 
> Also, given that preference panels are now ready, and the app is more or less unusable on a 14" laptop with 150% scaling by default, can we put a preference panel in this before merging?
> 
> We should also merge #2264 , pull main in, and make sure legend scaling works on Mac before merging.

@pbeaucage done. Is it ok on mac @wpotrzebowski ?

wpotrzebowski (approved):

PR #2290, opened 2022-10-26
Preference Panel API Update

krzywon:
After some discussion with @pbeaucage, adding a widget to the preferences panel was not ready for production use. This helps to right that ship.

Changes:

- Moved the panel to a new package in `sas.qtgui.Utilities` called `Preferences`
- The various classes are now in separate python files
- The base widget layout was horizontal instead of vertical causing the label and interactor to stack instead of lay next to one antoher.
- Added the `Preferences` package to `setup.py`
- Added some documentation on adding new widgets (though one or two should be ready soon)


src/sas/qtgui/Utilities/Preferences/PreferencesWidget.py
```diff
@@ -112,16 +55,16 @@ def __init__(self, name: str, default_method: Optional[Callable] = None):
         super(PreferencesWidget, self).__init__()
         self.name = name
         self.resetDefaults = default_method
-        self.horizontalLayout = QHBoxLayout()
-        self.setLayout(self.horizontalLayout)
+        self.verticalLayout = QHBoxLayout()
```
pbeaucage, line 58: should be 
```suggestion
        self.verticalLayout = QVBoxLayout()
```

src/sas/qtgui/Utilities/Preferences/PreferencesWidget.py
```diff
@@ -1,12 +1,9 @@
...
+from typing import Optional, List, Union, Callable, Any
 
-from sas.system.config.config import config
-from sas.qtgui.Utilities.UI.PreferencesUI import Ui_preferencesUI
-
-logger = logging.getLogger(__name__)
+from sas.system import config
 
 
 def set_config_value(attr: str, value: Any):
```
pbeaucage, line 9: For this to work with the setter-callback-generator it also needs an optional dtype kwarg so that e.g. strings from textboxes can be cast to floats?  Unless there's another workaround for this.
```suggestion
def set_config_value(attr: str, value: Any, dtype: Optional[Any] = None):    
    """Helper method to set any config value, regardless if it exists or not
    :param attr: The configuration attribute that will be set
    :param value: The value the attribute will be set to. This could be a str, int, bool, a class instance, or any other
    :param dtype: The datatype to cast the input value to if casting is desired
    """
    if dtype is not None:
        config = dtype(config)
    setattr(config, attr, value)

```

pbeaucage (commented):
Looks good.  Two minor tweaks noted, after these it is working with my local plotting preference pane.

src/sas/qtgui/Utilities/Preferences/PreferencesWidget.py
```diff
@@ -34,15 +49,15 @@ def cb_replace_all_items_with_new(cb: QComboBox, new_items: List[str], default_i
     cb.setCurrentIndex(index)
 
 
-def config_value_setter_generator(attr: str,dtype: Optional[Any] = None):
+def config_value_setter_generator(attr: str, dtype: Optional[Any] = None):
```
pbeaucage, line 52:
```suggestion
def config_value_setter_generator(attr: str, dtype: Optional[Callable] = None):
```
to match signature above

pbeaucage (approved):
Approved after one exceedingly minor typing fix.

PR #2293, opened 2022-10-27
Support custom data names in plot legends

### LINKED ISSUE #2253: Custom data names are not displayed in legend
**Describe the bug**
Plot legends appear to default to the file name rather than a custom user-supplied name added using right-click > "Change Name"
**To Reproduce**
1. Add a model or datafile
2. Change name using the context menu to something custom
3. Plot the data
4. Legend reflects filename/model number, not data name
**Expected behavior**
Legend entry should reflect the custom name if the user goes to the effort to input one?

**Screenshots**
See attached 
 - Version: 5.0.5

 - OS: Win 10

image

pbeaucage:
Fixes #2253 by switching Plotter to use data.name as the default `label` for a plot item rather than `data._title`.  `data.name` changes with the "Change Name" option in the Data Explorer.

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -163,6 +163,7 @@ class PlotterWidget(PlotterBase):
         ax = self.ax
         x = data.view.x
         y = data.view.y
+        label = data.name # was self._title
 
         # Marker symbol. Passed marker is one of matplotlib.markers characters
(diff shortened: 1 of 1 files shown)


rozyczko (approved):
I swear we went through these changes (name->title, title->name) a few times already, with every change breaking something somewhere... 
I don't think we targeted the legends though, so I'd say it's kosher.

Code looks good.

PR #2294, opened 2022-10-27
2242 quit dont ask me again

lucas-wilkins:
Are you sure want to quit SasView?

I'm always sure.

----

An ask me again box on exit dialog


src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -517,15 +518,14 @@ def quitApplication(self):
...
-
-        # Exit if yes
-        if reply == QMessageBox.Yes:
+
+        answer = hidable_dialog("Exit SasView", quit_msg, config.SHOW_EXIT_MESSAGE)
+
+        if answer.result:
+
+            # Only set ask again on accepted quit, otherwise the user might
+            # never be able to leave SasView (@butlerpd might prefer this behaviour though)
```
pbeaucage, line 527: LOL

pbeaucage (approved):
Does what it says on the tin.  Working well on my machine.  Would be nice to have a GUI preference pane so that users can reset the dialog, but I don't see that as a blocker for this esp given that the preference API isn't merged yet.

pkienzle:
One can make a social engineering argument: The exit dialog is annoying but necessary since state isn't preserved on exit and reloaded on startup. By allowing developers to avoid the annoyance they are less motivated to do the correct fix. Therefore it may be in the best interest of the users to reject the PR.

We already have save/load project on the menu. How about we automatically save the project before quit and restore it on restart? Add a menu item to clear the project if you want to start afresh. Then we can eliminate the save dialog for everyone. 

If we have written the project save correctly then no information will be lost, and accidentally hitting quit will merely be an annoyance of resetting the plots. If we have not saving everything then that's a bug that needs to be addressed. For example, if the fit uncertainty table only appears in the console log then we should be sure to include the log in the project save file and restore it on load.

lucas-wilkins:
@pkienzle When that's done we can migrate the quit dialog with a don't ask me again, to a save changes on exit dialog with a don't ask me again.

pkienzle:
@lucas-wilkins not needing any dialog is the best state; configuration options should be few and far between

Disabling safety features because they are too annoying should be a last resort, not a temporary fix.

lucas-wilkins:
I, and I suspect many others, don't want things to save everything automatically when a program exists. Nor do I want it to hold me up every time it exits.

Usually, when I quit a program, I want it to forget everything I have not explicitly told it to.

This is why on many applications you have an "unsaved data" prompt, which can be disabled - because I know I have unsaved data, and its often why I'm exiting.

lucas-wilkins:
> Disabling safety features because they are too annoying should be a last resort

@pkienzle Is this your way of offering to implement the changes you want, because I don't have the bandwidth.

lucas-wilkins:
THIS DISABLES GUI TEST

PR #2296, opened 2022-10-27
Reverting debugpy package

wpotrzebowski:
We have to revert debugpy package as it turned out that is required to run python shell (qtconsole). 
Proper unit test for qtconsole will be of course very useful, however won't be addressed in this PR.


rozyczko (approved):
LGTM

PR #2297, opened 2022-10-27
Nightly build artifact storage

wpotrzebowski:
Experimenting with nightly builds

The artifacts now upload to release page and they should be updated at 2AM (UTC). The release is currently in draft stage and is not visible for people from outside. I think however that if we remove draft flag but still keep it as a pre-release it will be accessible for people from outside and won't paper on release page as latest. 

Since this is running as cronjob I hope this won't affact extrenal PR contributions. 
Eventually the plan is to add signing and notarization for Mac as a part of this workflow (as currently we are not doing it at all). 

Keeping in a draft mode to make sure cronjob actually works

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -1,0 +1,23 @@
+name: Last Succesful Build
+on:
+  schedule:
+    - cron: '0 2 * * *' # run at 2 AM UTC
+
+jobs:
+  nightly:
+    name: Deploy nightly
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+
+    - name: Find matching workflow
+      uses: SamhammerAG/last-successful-build-action@v2
+      with:
+          branch: "main"
+          workflow: "Installer"
+          verify: true
+
+
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -1,25 +1,29 @@
-name: Last Succesful Build
+name: Nightly Build Artifacts
 on:
-  schedule:
-    - cron: '0 2 * * *' # run at 2 AM UTC
+  [push, pull_request]
+  #schedule:
+  #  - cron: '0 0 * * *' # Eventuall will run at 2 AM UTC, for testimng 
 
 jobs:
   nightly:
     name: Deploy nightly
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [windows-latest, macOS-latest]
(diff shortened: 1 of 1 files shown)


pkienzle:
Why do this nightly rather than whenever there is a push to main?

That said, it appears to be working since artifacts are showing up on the pre-release.

To check that the cron is working it would be nice if the release were tagged with the build date and/or short hash. This can be done in the release description "Build: <DATE> <HASH> (last updated at <timestamp now>)" and/or the install filename "setupSasview-<DATE>-<HASH> .exe".

The workflow name will need to be coordinated with #2263

wpotrzebowski:
The reason why artifacts appear at the release page is that for testing I did them on the each push.
The main reason why is not every push is that we don't feel sigining and notarization needs to be done that often plus we want to avoid the problem reported recently with the PRs submitted outside of organization, which prevents from using upload artifacts action (github token required).

pkienzle:
I don't know anything about the cost of signing; I assumed it would be free once you have the key.

Re: PRs outside the organization, can you restrict the upload action to commits to main in the SasView/sasview repo?

Hmmm... I wonder what happens when a 3rd party PR modifies the workflow action to remove that restriction…

llimeht:
@pkienzle, uploads to the SasView/sasview/releases is restricted to runs that are run from within the SasView/sasview project.

For a PR coming from a fork, the CI for a PR is run within the submitting user's fork and they can't upload to SasView/sasview/releases; that then means that no build artifacts are created for testing and the CI turns red. People are then not prepared to merge the PR. That's not the desired approach.

@wpotrzebowski Do we know what the nightly cron will do in forks? I'm guessing forks don't really want it running and generating nightlies there.

pkienzle:
Can the "upload" step be made conditional on which fork is running? That sounds easier to manage than a cron job.

andyfaff:
Probably a good idea to limit the cron job (if not all CI runs) to the sasview/sasview repository. If it's limited to the main repository doing the job then the job will have access to repository secrets to do any uploads.

Have a look at https://github.com/scipy/scipy/blob/main/.github/workflows/wheels.yml for how this can be done.

pkienzle:
The release artifacts are accessible from action itself. For example: https://github.com/SasView/sasview/actions/runs/3333616518 comes from `Rework CI to be non-repetitive, simpler, test more things #3` using the fork llimeht:cicombinematrix

Kind of a pain to get there though: Click on https://github.com/SasView/sasview/actions and enter `actor:llimeht` into the search to get a list of work actions for llimeht, then click on one of them to get at the artifacts.

Maybe there is an action to link the latest run for the PR to the "changes approved/all checks passed" notifications box?

wpotrzebowski:
Closing this now and will come back after merging #2263

Issue #2118, opened 2022-07-08
Append To combo box only has one graph option for Theory data


smk78:
**Describe the bug**
When you have multiple datasets in the Data Explorer you have the option to select one or more and Create [a] New graph, _or_ to Append [them] To an existing graph.

It seems that the second of these two actions is only available to Data data and not Theory data.

**To Reproduce**
1. Run SasView
2. Create a FitPage and select any model. Calculate the model and then Compute/Plot the calculation.
3. Now create a second FitPage and repeat the process.

This creates a couple of Theory datasets and a graph of each calculation.
![image](https://user-images.githubusercontent.com/10938679/178048597-b23e1756-04c6-4b1c-b4f7-352cf8967d73.png)
![image](https://user-images.githubusercontent.com/10938679/178048122-9ca1c19b-ea67-4b8b-b3f1-176dc24d667e.png)
 
4. Now select one or other of the Theory datasets and try Append To the graph with the other dataset (ie, to put both datasets on the same graph).

There is only one graph option in the combo box.
![image](https://user-images.githubusercontent.com/10938679/178048550-26b00ba6-e3b2-48a4-836e-8477f2713b24.png)

**Expected behavior**
All of the generated graphs should be available in the combo box.

**SasView version (please complete the following information):**
 - Version: release 5.0.5, but affects all 5.x.

**Operating system (please complete the following information):**
 - OS: W10/x64

**Additional context**
The expected behaviour is available in 4.2.2. This issue was introduced in 5.x.

rozyczko:
Just did this with 5.0.5 and got the second theory plot correctly appended.
It would be good to see what is the different between what we both did

![Untitled](https://user-images.githubusercontent.com/8266281/178215886-d82d35e6-9104-4df7-bffb-442d89e546e5.png)

smk78:
@rozyczko , sorry for the delay in picking this up. But did you do that in the release version or your dev version?

I've just uninstalled release 5.0.5, re-downloaded & re-installed it, and I can re-create this issue at will.

butlerpd:
Looking at the examples ...I'm guessing that @rozyczko tried to append M2 to graph 1? That does work for me too. The problem is that if you cannot append M1 to Graph 2. In fact you cannot append anything to anything except graph 1. At least in my test of 5.0.5. Not sure if that helps?

smk78:
[...]

PR #2298, opened 2022-10-27
DIfferentiate which combobox to put the graphs in.

rozyczko:
This addresses #2118 - incorrectly populated Graph combobox for theory plots.


butlerpd (approved):
cannot say much about the code  but looks reasonable to me. I did test with the windows installer and the problem is indeed fixed. I tried as suggested in the issue but also then did some fits and added the residuals and fit data to one of the last theory graphs and that also worked.  No documentation updates should be needed.  I recommend merging

wpotrzebowski (approved):

PR #2300, opened 2022-10-27
Fit intermittent plot blanking (#2267)

### LINKED ISSUE #2267: Plot limits intermittently reset to (0,1) on all axes
**Describe the bug**
In some difficult-to-isolate circumstances, the limits for an active plot reset to (0,1) on both axes resulting in a blank graph: 

image

The data are *still present* within this plot, if you go to the context menu and "Reset Graph Range" the plot becomes normal: 

image

Also of note, the axes reset to linear when this happens, not log.

This was observed by several NIST summer school students without external monitor attached, also with external monitor.  I've observed in 1D fitting plots but also seen in P(r) and 2D fitting, maybe more, @butlerpd has notes.  @caitwolf also has been able to reproduce this, not reliably though.

**To Reproduce**

Working to develop steps to reproduce
Has been observed on Windows and Mac
Trigger is sometimes a mouse over the plot window, sometimes a plot redraw with no user action.

**Expected behavior**
Axes should not reset.

 - Version: 5.0.5 and current main

 - OS: Win10, mac

pbeaucage:
Fixes #2267 by commenting out calls to `matplotlib.pyplot.cla()` which were already paired with the more specific alternative calls to `self.ax.cla()` (at best a duplicate, at worst it causes arbitrary plots to blank themselves as seen here.

It would be worth eyes from @rozyczko and @krzywon who wrote this originally to see if there's some reason I'm missing for the `plt.cla()` call.  It seems to now work fine on my machine.  The tell, even without a way to reproduce, is that when data is "Compute/Plot"'d on Main the plot briefly will flash to a white background with default axes from 0 to 1, linear, on both axes.  That no longer happens on this branch, at least for me.  Or see reproduction instructions in #2267.

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -581,7 +581,6 @@ class PlotterWidget(PlotterBase):
         Resets the chart X and Y ranges to their original values
         """
         # Clear graph and plot everything again
-        # mpl.pyplot.cla()
         self.ax.cla()
         self.setRange = None
         for ids in self.plot_dict:
@@ -659,7 +658,6 @@ class PlotterWidget(PlotterBase):
         yl = self.ax.yaxis.label.get_text()
 
         self.ax.cla()
-        #mpl.pyplot.cla()
 
         # Recreate Artist bindings after plot clear
         self.connect = BindArtist(self.figure)
@@ -728,7 +726,6 @@ class PlotterWidget(PlotterBase):
         self.plot_dict = {}
 
         # Clean the canvas
-        # mpl.pyplot.cla()
         self.ax.cla()
 
         # Recreate the plots but reverse the error flag for the current
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -581,7 +581,7 @@ class PlotterWidget(PlotterBase):
         Resets the chart X and Y ranges to their original values
         """
         # Clear graph and plot everything again
-        mpl.pyplot.cla()
+        # mpl.pyplot.cla()
         self.ax.cla()
         self.setRange = None
         for ids in self.plot_dict:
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -581,7 +581,7 @@ def onResetGraphRange(self):
         Resets the chart X and Y ranges to their original values
         """
         # Clear graph and plot everything again
-        mpl.pyplot.cla()
+        # mpl.pyplot.cla()
```
rozyczko, line 584: These commented out calls can be safely removed.

rozyczko (approved):
Code looks good.
All three places behave correctly:
- on reset graph range
- on hide/show error bars
- on plot removal

I can't reproduce the original issue reliably, so can't comment on the applicability of the fix but it seems to be a step in the right direction.
A general improvement in chart responsiveness is also visible.

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -658,8 +658,8 @@ def removePlot(self, id):
         xl = self.ax.xaxis.label.get_text()
         yl = self.ax.yaxis.label.get_text()
 
-        mpl.pyplot.cla()
         self.ax.cla()
+        #mpl.pyplot.cla()
```
pbeaucage, line 662:
```suggestion 

```

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -728,7 +728,7 @@ def onToggleHideError(self, id):
         self.plot_dict = {}
 
         # Clean the canvas
-        mpl.pyplot.cla()
+        # mpl.pyplot.cla()
```
pbeaucage, line 731:
```suggestion 
```

src/sas/qtgui/Plotting/Plotter.py
pbeaucage:
```suggestion 
```

wpotrzebowski (approved):

PR #2304, opened 2022-10-27
Removed debug messages for h5py

lucas-wilkins:
Disabled the Converting from 5 to 7 etc messages via log

### DIFF
--- a/src/sas/system/log.ini
+++ b/src/sas/system/log.ini
@@ -41,7 +41,7 @@ args=(os.path.join(os.path.expanduser("~"),'sasview.log'),"a")
 # Loggers
 
 [loggers]
-keys=root,saspr,sasgui,sascalc,sasmodels
+keys=root,saspr,sasgui,sascalc,sasmodels,h5py
 
 [logger_root]
 level=DEBUG
@@ -72,3 +72,8 @@ qualname=sas.sascalc
 handlers=console,log_file
 propagate=0
 
+[logger_h5py]
+level=DEBUG
+qualname=h5py
+handlers=
+propagate=0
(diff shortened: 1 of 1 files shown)


src/sas/system/log.ini
```diff
@@ -41,7 +41,7 @@ args=(os.path.join(os.path.expanduser("~"),'sasview.log'),"a")
 # Loggers
 
 [loggers]
-keys=root,saspr,sasgui,sascalc,sasmodels
+keys=root,saspr,sasgui,sascalc,sasmodels,h5py
```
rozyczko, line 44: h5py is not "ours", so it has no place in these keys.
3rd party loggers should probably be all set by default to at least DEBUG if not WARNING, with those that need detailed logging whitelisted.
https://github.com/SasView/sasview/pull/2275#discussion_r1005979992

rozyczko (changes requested):

src/sas/system/log.ini
lucas-wilkins: @rozyczko If you know a better way to do it, please do. I'm definitely not sure the solution I proposed is good.

lucas-wilkins: I asked everyone downstairs if they knew the right way of disabling these messages, and nobody did.

pkienzle: Changing lines 36 and 47 from DEBUG to WARNING didn't do it?

lucas-wilkins: That works, is generally better.

rozyczko (approved):
Less junk in the log explorer/command prompt is always good.
Easy to change the .ini settings to increase the logging level, if required.

PR #2305, opened 2022-10-27
Modify copy of config.PLUGIN_STATE_EXTENSIONS value, not config value

lucas-wilkins:
Append additional runtime extensions to a copy of config.PLUGIN_STATE_EXTENSIONS value, not config value itself.

Fixes growing config file issue noticed in @wpotrzebowski's config.json


pbeaucage (approved):
This looks good.  I had the same issue on my machine.

pkienzle:
Red flag: shouldn't config items be immutable?

I suggest changing the config default to a tuple and putting a check in your configuration system initialization that all default values are immutable so they don't accidentally get modified in the rest of the program.

pkienzle:
Corollary: config loader should convert lists to tuples.

lucas-wilkins:
> Red flag: shouldn't config items be immutable?

They just shouldn't be added to like they were, just a teething problem that stems from it being a migration of the old config.

PR #2306, opened 2022-10-27
Move fitting optimizer selection to preferences panel

krzywon:
This PR moves the optimizer selector to the preferences window. This comes with a few improvements to the preferences panel and widget.

Fitting Options Changes:
- Converted the fitting options window into a preferences widget and moved all logic there.
- Added a default optimizer to the configuration and a way to change it in the preferences
- Allows the active optimizer to differ from the optimizer loaded when SasView starts
- Redirects 'Fitting->Fit Options' to the Fitting Preferences widget
- Deleted the previous fit options panel
- Updated fit option unit tests to conform to the existing panel

Preferences Changes:
- Added preferences widget methods to create a horizontal line and stand-alone text
- Added a helper method `PreferencesWidget._addAllWidgets()` that can be overridden by child classes, allowing preferences to be pushed to the top of the panel instead of vertically staggered
- Preferences widget element creators return the element to the child to allow name assignment
- The preference panel automatically assigns GuiManager as the parent to each preferences widget instance

Things that might need improvement:
- Allow each preferences widget to assign its own parent
- Documentation: create a preference panel help file that is called when the help button is pressed and incorporate the fit options into this new help file
- New unit tests for new element creators


rozyczko:
Starting sasview on `preferences-fitting-optimizer` gives me
```
 File "D:\projects\sasview_new\src\sas\qtgui\Utilities\Preferences\FittingOptionsWidget.py", line 44, in _addAllWidgets
    algorithm = self.config.names[default_id]
KeyError: ''
```

It looks lke `config.DEFAULT_FITTING_OPTIMIZER` returns an empty string rather than the default optimizer name?

krzywon:
> It looks lke `config.DEFAULT_FITTING_OPTIMIZER` returns an empty string rather than the default optimizer name?

Ugh, yes it is. Should be a quick fix.

krzywon:
@rozyczko Have another look?

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
```diff
@@ -87,10 +98,16 @@ def close(self):
...
 
     def addWidget(self, widget: QWidget, name: Optional[str] = None):
+        """Add a single widget to the panel"""
+        # Set the parent of the new widget to the parent of this window
+        widget.parent = self.parent
         self.stackedWidget.addWidget(widget)
+        # Set display name in the listWidget with the priority of
+        #  widget.name > name passed to method > "Unknown"
         name = widget.name if hasattr(widget, 'name') and widget.name else name
         name = "Unknown" if not name else name
```
rozyczko, line 108: if `widget.name` doesn't exist and the argument name is default, you should check for None rather than `not name`.
```
name = 'Unknown' if name is None else name
```

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
```diff
@@ -87,10 +98,16 @@ def close(self):
         super(PreferencesPanel, self).close()
 
     def addWidget(self, widget: QWidget, name: Optional[str] = None):
+        """Add a single widget to the panel"""
+        # Set the parent of the new widget to the parent of this window
+        widget.parent = self.parent
         self.stackedWidget.addWidget(widget)
+        # Set display name in the listWidget with the priority of
+        #  widget.name > name passed to method > "Unknown"
         name = widget.name if hasattr(widget, 'name') and widget.name else name
```
rozyczko, line 107: Shouldn't the passed argument name take precedence over widget.name here?

src/sas/qtgui/Utilities/Preferences/PreferencesWidget.py
```diff
@@ -56,25 +57,43 @@ def config_value_setter_generator(attr: str, dtype: Optional[Callable] = None):
...
+        self._addAllWidgets()
+        # Push all elements to the top of the window
+        self.verticalLayout.addStretch()
         self.adjustSize()
 
+    def _addAllWidgets(self):
+        """A private psuedo-abstract class that children should override. All widgets should be added here to force
+        elements to the top of the window.
+        """
+        pass
```
rozyczko, line 95: Since without this override no functionality of this widget exists, why not add a `NotImplementedError` exception thrown when this method is not implemented? Just make it more rigorously abstract.

src/sas/qtgui/Utilities/Preferences/FittingOptionsWidget.py
```diff
@@ -0,0 +1,547 @@
...
+        self.cbAlgorithm = self.addComboBox("Active Fit Algorithm",
+                                            active_names,
+                                            self.setActiveOptimizer,
+                                            algorithm)
+        self._addFittingOptionsWidgets()
+        self.verticalLayout.addWidget(self.stackedWidget)
+        self.onAlgorithmChange(self.cbAlgorithm.currentIndex())
+
+    def _addFittingOptionsWidgets(self):
+        """This is a direct copy from teh python file generated from the old UI file."""
```
rozyczko, line 63: Would it be very annoying to have the UI element definitions and layouts in a separate file, preferably a .ui file from the Designer? Or if not, maybe at least a separate .py file/class with those elements, so there's a clear distinction between UI and logic code?

rozyczko (approved):
No real showstopping issues, just a few questions/suggestions for consideration.
Especially about splitting UI definitions and logics into separate files for clarity.

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -458,7 +456,7 @@ def onFittingOptionsChange(self, fit_engine):
         """
         React to the fitting algorithm change by modifying window title
         """
-        fitter = [f.id for f in options.FITTERS if f.name == str(fit_engine)][0]
+        fitter = [f.id for f in fitters.FITTERS if f.name == str(fit_engine)][0]
```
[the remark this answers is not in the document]
rozyczko: No, scratch this. You are importing `bumps.options` elsewhere, anyway :(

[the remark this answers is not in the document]
krzywon: Looking closer, I think I can get rid of one of them.

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
krzywon: Yes, yes it should.

src/sas/qtgui/Utilities/Preferences/PreferencesWidget.py
krzywon: Good idea. I plan to do something similar for restoring defaults.

src/sas/qtgui/Utilities/Preferences/FittingOptionsWidget.py
krzywon: @pkienzle had a similar complaint with this change but more related to pulling the widget out of the fitting perspective, breaking the perspective as a plugin framework. I might merge this as-is, document all issues in #2316, and handle these issues one by one in separate branches.

pbeaucage:
When I try to set the default fitter, I get the following:

```
21:53:28 - DEBUG: Please add scipy.leastsq to the configuration and give it a sensible default value.
21:53:33 - DEBUG: Please add newton to the configuration and give it a sensible default value.
21:53:35 - DEBUG: Please add dream to the configuration and give it a sensible default value.
21:53:37 - DEBUG: Please add de to the configuration and give it a sensible default value.
21:53:39 - DEBUG: Please add amoeba to the configuration and give it a sensible default value.
```
something looks wrong with the default fitter config variable change?

butlerpd (changes requested):
Testing on windows, things seem to work fine. I switeched to DREAM and ran. Appreciate that the old menu item did not disappear but just sends to the new panel.  However, setting the default to a different optimizer does not persist. Setting to DREAM and then exiting then running again the default was set to LM again.  Shouldn't that be one of the things that this PR does?

The other thing I found is that the sizing preference is extremely dangerous. a value too small (dropping from 1.0 to 0.8 for me) renders the screen unusable and almost impossible to exit (specially since you have to be able to click on "yes I DO want to exit").  This will cause a LOT of pain to users I suspect. This needs to be hidden more deeply with lots of dire warnings I would think?

krzywon:
@pbeaucage and @butlerpd, the issues you both reported are the same bug and should be fixed with my latest commit.

In regards to the display properties, this is something we could disable/hide if enough people have issues with it.

butlerpd:
Regarding the display scaling, further discussion "round the fire" at code camp brought up the right answer I think: this should be treated like any display change. show the screen with the new changes then provide a popup (in the original scale) asking whether to keep/proceed with the change or revert/don't change.  This of course probably requires more work and outside the scope of this PR I guess -- though I blame the reviewer on the initial PR that let this in in the first place 😃

krzywon:
Having thought more about this PR, and the preferences panel in general, merging this will pollute the main branch with commits that will only be reverted in the future. Some of the work here will be reused, but the fitting optimizer move breaks too many of our established conventions. Closing.

PR #2309, opened 2022-10-28
Fixed file path for log.ini

lucas-wilkins:
File path error was causing log.ini not found error


rozyczko (approved):
No error anymore! Rejoice!
The code looks good.

PR #2311, opened 2022-10-28
Fixed showHelp reference

rozyczko:
showHelp is defined on the gui manager not on the main window.


lucas-wilkins:
This is now in brokenness parity with the rest of the help system

lucas-wilkins (approved):
looks good.

PR #2314, opened 2022-10-28
Adding installation path to about box

wpotrzebowski:
In order to facilitate finding path for running sasview from script: #2280 it has been added to about box. 
Some minor fixes (like correcting link to contributors list) has also been introduced.


smk78 (approved):
Functionality test on x64/W10.
Does what it says! Links all work, and the new path can be copied (allowing it to be easily appended to the Path environment variable, for example). The new path will also be helpful for identifying which version of multiple versions is actually in use!

Issue #2299, opened 2022-10-27
Sending single data to Fitting should automatically open the Data chart


rozyczko:
@butlerpd would like to see the original behaviour from 4.x, where the data sent from the data explorer to fitting perspective would automatically be displayed as a data-only plot.
Defining a model would then append the model curve to that data plot.

Issue #1685, opened 2020-10-12
Simultaneous fitting not on same graph


butlerpd:
This feature needs to be reinstated.

In SasView 4.x sending several datasets to fitting simultaneously would open multiple fit pages but plot them on one graph. The number that was allowed to go to one graph was limited as I recall to 4 originally but there were demands for more and I believe, but have not tested, that 4.2.2 allows more. 

The point is that, for many simultaneous fitting situations, seeing the fits on the same graph is very useful. In particular when fitting SANS and USANS or or two Q ranges where the incoherent background or the scale factor are allowed to float separately (background because of different wavelengths or scale because to allow for uncertainties in the different calibrations etc.) but also for example SAXS and SANS of the same sample.  A number of users liked to use it with contrast series as well.

5.x seems to have lost that feature altogether unless it is moved to a different place?  This is an important feature for many users and one which recently caused me to revert to 4.2.2 for a particular project.  A feature that allows plotting fitted data on a single graph, needs to be reinstated.

RichardHeenan:
The "sending simultaneously to fits" kick off for this feature seems a bit obscure unless you happen to know about it. Perhaps the simultaneous fit tab could have a check box for a master plot with all the data and fits? 

But would that only update if you hit fit on that tab? Could it be updated also on hitting fit on an individual fitpage?

Simply appending say the M2 and M3 fits and data to the Graph1 plot does not of course work, as the fits are not updated after the append operation. Since we do have a "freeze results" option in the data explorer, then the fits could perhaps dynamically update, but that might cause confusion elsewhere.

wpotrzebowski:
I will take a look on it but it doesn't sound like trivial thing to fix.

rozyczko:
Nothing is simple with SasView plotting, especially if it involves non-standard behaviour and multiple plots on one chart...
This difficulty stems from the requirement of having the state of all charts maintained and plots registering their opening/closing accordingly.
This will need a single Plotter instance initialized and all simfit pages updating this one plotter. Looks like a slightly more complex issue.

butlerpd:
We probably need a feature discussion (like we used to have at the old fashioned code camps :-).  So probably we should shelve this issue till we can have a proper discussion on what behavior would be most useful/desired. In the meantime, some clarification on the original "request" might be useful.

[...]

PR #2315, opened 2022-10-28
Sending data to fitting automatically plots the data (#2299)

pbeaucage:
Simple little implementation of #2299.  Added a config option for users who might prefer the earlier behavior of pre-selecting a model.

### DIFF
--- a/src/sas/qtgui/Utilities/Preferences/FittingPreferencesWidget.py
+++ b/src/sas/qtgui/Utilities/Preferences/FittingPreferencesWidget.py
@@ -1,0 +1,11 @@
+from sas.system import config
+
+from .PreferencesWidget import PreferencesWidget, config_value_setter_generator
+
+
+class FittingPreferencesWidget(PreferencesWidget):
+    def __init__(self):
+        super(FittingPreferencesWidget, self).__init__("Fitting Settings")
+        self.addCheckBox(title="Auto-plot data when sent to fitting perspective",
+                         callback=config_value_setter_generator('FITTING_PLOT_ON_SEND_DATA', dtype=bool),
+                         checked=config.FITTING_PLOT_ON_SEND_DATA)
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -778,8 +778,13 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         try:
             if self.chkSwap.isChecked():
                 self._perspective().swapData(selected_items[0])
+                if self._perspective().name == 'Fitting':
+                    self._perspective().tabs[0].onPlot() #todo sort out where the plot should go.
             else:
                 self._perspective().setData(data_item=selected_items, is_batch=self.chkBatch.isChecked())
+                if self._perspective().name == 'Fitting':
+                    self._perspective().tabs[0].onPlot()
+
         except Exception as ex:
             msg = "%s perspective returned the following message: \n%s\n" % (self._perspective().name, str(ex))
             logging.error(ex, exc_info=True)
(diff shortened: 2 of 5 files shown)


butlerpd (changes requested):
I think this is going to require some discussion.  However some first comments (ignoring the preferences issues since that really is another PR):
* The current implementation does not do what is really desired here. It basically implements a `Create New` call. Once I change a parameter and click compute a whole new set of 3 windows opens up.  The point of the immediate plotting is for that first plot to be the one where the model fit also appears. In other words it should be the plot window tied to the fitpage.
* This is attempting to implement the same for sending of multiple data and even for sending a batch fit.  There is a ticket somewhere to allow multiple data sets to be sent to a single fitting plot with the data and their models all showing up together, but again this just does the same as `Create New` and creates a plot with all the data that is divorced from fitting.
* Likewise the batch mode, which does not normally throw up plots (for obvious reasons I think), by default now gets a graph (not where fits go).  This may in fact be a useful feature but not sure it should be tied to the same preference?
I would suggest that maybe one should try to limit to use case 1 for starters as perhaps the biggest bang for the buck? So just implement for a **single** file sent to fitting. @rozyczko seems to think isolating to only single file sends to fit should be trivial but the tying to the fit panel may be trickier?

butlerpd:
and here is the issue I was thinking of ... interestingly it is a feature parity issue!   #1685

rozyczko (changes requested):
Adding a new attribute is not necessary. We can already differentiate between data and model/data plots.
Data plots have their `plot_role` property set to 0, while model plots have 1.
It's probably enough to add a `plot_role` check.

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -263,9 +269,27 @@ def addFit(self, data, is_batch=False, tab_index=None):
...
+        if tab.data_is_loaded and config.FITTING_PLOT_ON_SEND_DATA:
+            #First, create model and residuals inside this data.
+            def tidyup(self):
+                logger.info(msg='calculation complete, running tidy-up function')
+                #time.sleep(2)
+                #tab.onPlot()
+                #QtWidgets.QApplication.processEvents()
+                tab.showPlot()
+                #QtWidgets.QApplication.processEvents()
+                #time.sleep(2)
```
rozyczko, line 288: Playing with `processEvents` means there were issues with plot lifetimes? What did you observe?

src/sas/qtgui/Perspectives/Fitting/FittingPerspective.py
```diff
@@ -263,9 +269,27 @@ def addFit(self, data, is_batch=False, tab_index=None):
...
+                tab._model_model.clear()
+                tab.cmdPlot.setEnabled(False)
+            tab.SASModelToQModel('porod')
```
astellhorn, line 291: Why did you choose "porod" here, and not the currently active model?

src/sas/qtgui/Utilities/Preferences/FittingPreferencesWidget.py
```diff
@@ -0,0 +1,11 @@
...
+
+from .PreferencesWidget import PreferencesWidget, config_value_setter_generator
+
+
+class FittingPreferencesWidget(PreferencesWidget):
+    def __init__(self):
+        super(FittingPreferencesWidget, self).__init__("Fitting Settings")
+        self.addCheckBox(title="Auto-plot data when sent to fitting perspective",
+                         callback=config_value_setter_generator('FITTING_PLOT_ON_SEND_DATA', dtype=bool),
+                         checked=config.FITTING_PLOT_ON_SEND_DATA)
```
rozyczko, line 11: this needs porting to the new way of describing preferences

PR #2318, opened 2022-10-28
Sesans debugging

### LINKED ISSUE #1998: 5.x: Changing parameter in FitPage does not update plot when fitting SESANS data
5.0.4 on OSX Big Sur 11.6.2 & 5.0.5rc1 on W10/x64

Reported by User GregS. Verified by @smk78.

When fitting SESANS data, manually changing parameters in the Fit Page does _not_ update the plot as it does when fitting normal SAS data. The plot _does_ update when the SESANS data are fitted however.

Before:
Screenshot 2021-12-09 at 12 06 24

After changing radius and hitting <return>:
Screenshot 2021-12-09 at 12 06 32

caitwolf:
This removes the deep copying of the model that occurs during fitting and allows the plots to update with changes after fitting. This deepcopy was done by @rozyczko so he should probably check that it is fine to remove.

This should close #1998 - This was also a problem for non-sesans data but was more apparent when working in the sesans correlation space, making it seem that the plots were not updating after fitting for sesans data only.

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1879,7 +1879,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
 
         # Data going in
         data = self.logic.data
-        model = copy.deepcopy(self.kernel_module)
+        model = self.kernel_module
         qmin = self.q_range_min
         qmax = self.q_range_max
 


rozyczko:
This `deepcopy` goes back A LONG time, to 2018, where I was fixing an old bug in polydispersity behaviour.
This seems to have been a leftover from that exercise. The target bug does not occur anymore with the removed `deepcopy` so I assume it got resolved/hidden elsewhere over the years.
As such, I think removal of the `deepcopy` is benign and if it fixes the issue, should be merged.

rozyczko (approved):

rozyczko:
For the sake of tracking issue:
these two bugs are linked to the change from 2018 (viewable only to ESS staff)
https://jira.esss.lu.se/browse/SASVIEW-1008
https://jira.esss.lu.se/browse/SASVIEW-1013

butlerpd (approved):
seems to fix the problem and a quick check of my standard 1D fitting also still works and the fact that this may also have been a minor issue for 1D fitting, I think this should be merged post haste

PR #2326, opened 2022-10-29
Fix up some tests that were broken during codecamp

llimeht:
Some tests were broken during recent merges. This does some cleanup.


rozyczko (approved):
Good. Works as expected.

wpotrzebowski (approved):
Tested on OSX. Works fine

PR #2331, opened 2022-10-29
Fixes binning bug in implementation of sesans data

caitwolf:
This fixes a binning bug for sesans data.

Sesans is in real space and the Hankel transform from q space to real space (spin echo length) was implemented as a resolution function. For other resolution functions, the qmin and qmax range is compared to data.x which are all in q space. This is then passed to the calculator to return the unsmeared Iq. However, for sesans data, qmin, qmax, and data.x are all in real space while the calculator is expecting q space data to return the unsmeared Iq (which will then get passed to the sesans "smearer" which uses the hankel transform to convert Iq to real space).

Due to the number of points in q_calc (in q space) being much greater than the number of points in q (spin echo length in real space) in the sesans transform (resolution function), this resulted in a small portion of the data being improperly calculated at the incorrect q values (spin echo length being passed as q to the calculator returning Iq) with a error on the order of 1e-10.

The q, Iq data being passed to the Hankel transform before and after the fix for a sphere form factor with scale = 0.005, background = 0, sld = 1, sld_solvent = 6, and radius = 10000:
image

After the Hankel transform for both q, Iq datasets and compared to the analytical model for the same system: 
image

Error between the datasets in SESANS space: 
image

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/ModelThread.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ModelThread.py
@@ -172,9 +172,9 @@ class Calc1D(CalcThread):
         ##smearer the ouput of the plot
         if self.smearer is not None:
             if self.data.isSesans:
-                # For SESANS, data.x, qmin and qmax, and therefore get_bin_range are in correlation space, and
-                # the Hankel transform from q space to correlation space is set up as a resolution function, i.e.,
-                # the "unsmeared" data is in q space and the "smeared" data is in correlation space.
+                # For SESANS, data.x, qmin and qmax, and therefore get_bin_range are in real space, and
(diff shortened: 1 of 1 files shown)


lucas-wilkins (approved):
Looks like it should fix it. I take it that it looks right in the test script you have.

Also, you had some numbers quantifying the size of the error it produced. Can you put them here (in the PR description)?

butlerpd (approved):
clearly I'm not going to see any difference from my GUI tests, but running the installer doesn't show any obvious new issues. Code seems appropriate and minimalistic changes. Suggest we merge

wpotrzebowski (approved):
Played with on Mac OSX and it seems to work fine both on sesans and regular data sets

PR #2335, opened 2022-10-30
new color map code, based on a BSD-3 licensed module

### LINKED ISSUE #1944: Color range slider code needs to change
`src\sas\qtgui\Plotting\rangeSlider.py` is an external piece of code and it is GPL-2 licensed.
We strive to be BSD compliant so we can't ship this file. It got implemented back in 2017 as a temporary solution to the problem but reworking it properly has been, of course, forgotten.
Cursory check of similar functionality yields some interesting replacements, e.g.
https://pythonrepo.com/repo/tlambert03-QtRangeSlider-python-graphical-user-interface-applications
and
https://stackoverflow.com/questions/47342158/porting-range-slider-widget-to-pyqt5

This issue should probably be prioritized pretty high.

rozyczko:
## Description

Replaced QrangeSlider module with bad licensing (GPL) with similar module with good licensing (BSD).
Improved behaviour and visuals as well.

Fixes #1944

## How Has This Been Tested?

Tested on Win10

- [x] Code has been reviewed
- [X] Windows installer (GH artifact) has been tested (installed and worked) 
- [X] User documentation is available and complete (if required)
- [X] The introduced changes comply with SasView license (BSD 3-Clause)


src/sas/qtgui/Plotting/ColorMap.py
```diff
@@ -154,27 +154,33 @@ def initRangeSlider(self):
...
             self.updateMap()
         def set_vmax(value):
             self.vmax = value
-            self.txtMaxAmplitude.setText(str(value))
+            self.txtMaxAmplitude.setText(formatNumber(value))
             self.updateMap()
 
-        self.slider.lowValueChanged.connect(set_vmin)
-        self.slider.highValueChanged.connect(set_vmax)
+        def set_values(values):
```
lucas-wilkins, line 176: why not set_values(v1: int/float, v2: int/float), makes it easier to see what it needs

lucas-wilkins: or Tuple[int/float, int/float] if it needs to be a tuple

lucas-wilkins (approved):
Looks good

wpotrzebowski:
Looks good and works on Mac. These are not really related to this PR, however the list of color maps is unneceserally long and one can input min value higher than max (withouith reversing map). Probably this condition can be changed but I wouldn't worry about it. Also I am wondering what should interaction for reseting to very initial values (it now remembers values from previous setup). Should "Home" from navigation menu do that?

wpotrzebowski (approved):

rozyczko:
> the list of color maps is unneceserally long

I think it makes sense to enable users to use all the available color maps from the MPL dictionary.
Otherwise, we would have to do some categorisation and I don't really know which of those maps are common and which are not...
> one can input min value higher than max (withouith reversing map). 

This, actually, was a serious UI issue. Fixed.

> Also I am wondering what should interaction for reseting to very initial values (it now remembers values from previous setup). 

Without rewriting a large part of the code I don't see how this can be easily achieved. If you close the widget and then reopen it, the values come back to the original numbers, though.

PR #2336, opened 2022-10-30
Fix "Use dQ Data" bug that swtiched slit length and slit width in the gui

### LINKED ISSUE #2333: Use dQ Data values for slit length and width get switched after use of structure factor
**Describe the bug**
When loading slit smeared data with a slit length that can be extracted, the value is placed in the Slit length box of the Use dQ Data  within the resolution tab. However, the value gets moved to width after switching to and from a different resolution function when using a structure factor model. The model seems to still be correctly utilizing the larger number as the slit length.

**To Reproduce**
1. Load slit smeared data; I used the '1umSlitSmearSphere.ABS' dataset in the 1d directory of the example data. 
2. Send data to fitting and load a sphere form factor and any structure factor model.
3. Switch to the resolution tab; the slit length of 0.117 A-1 should be in the Slit length box.
4. Switch to a custom slit smear resolution function; you can enter a new length value or leave it empty.
5. Switch back to the Use dQ Data option and the 0.117 A-1 length will be in the Slit width box. 

**Expected behavior**
image

Unexpected behavior:
image

5.0.5

Windows 10

caitwolf:
## Description
This fixes a bug in the SasView gui that was switching the slit length and slit width values when using a structure factor for a slit smeared dataset after switching the resolution to "Custom Slit Smear" and then back to "Use dQ Data". 
This closes #2333. Recent changes were made to the slit smearing functions and nomenclature, and so @pbeaucage and @butlerpd should take a look.

## How Has This Been Tested?
Load a slit smeared dataset that includes the slit length, select a form factor and structure factor, go to the resolution tab, switch from "Use dQ Data" back to "Custom Slit Smear" and then back again. I've also tested for other datasets with different dQ data and also using different combinations of all resolution options and form factor models with and without structure factor.

- [x] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -4216,12 +4216,12 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
                 pass
         if 'smearing_min' in line_dict.keys():
             try:
-                self.smearing_widget.dq_l = float(line_dict['smearing_min'][0])
+                self.smearing_widget.dq_r = float(line_dict['smearing_min'][0])
             except ValueError:
                 pass
         if 'smearing_max' in line_dict.keys():
             try:
-                self.smearing_widget.dq_r = float(line_dict['smearing_max'][0])
+                self.smearing_widget.dq_l = float(line_dict['smearing_max'][0])
             except ValueError:
                 pass
 


butlerpd (commented):
what is dq_r?? should that not be dq_w?

pbeaucage (approved):
Verified that this bug existed in 5.0.5 and that this fixes it.  Very deep typo-level fix, nice job on finding it.

Tested on Win10 and working well.

pbeaucage:
@butlerpd I believe dq_r is used because this value could plausibly be _either_ the dq percentage (pinhole) or the slit length (formerly height) at this point in the code.  They go into the same field.

butlerpd (approved):
dq_r is appropriate I think as that section of code seems to be general and decides that it is slit smeared if it has a dl.  Finally reproduced the bug and verified that the windows installer works fine

PR #2348, opened 2022-10-31
Handling of constraints for polydisperse parameters (fix #1588)

### LINKED ISSUE #1588: 5.0 no constraint for polydispersity ?
In my local build of ESS_GUI today, the attached project (rename .txt to .json) has polydispersities fitting, but they do not appear in the constraints set up drop down lists. 

I thought that they might have been added now ? But may be this is still a work in progress?

[12June2020_3sets_test2.txt](https://github.com/SasView/sasview/files/4771872/12June2020_3sets_test2.txt)

gonzalezma:
## Description

Changes to handle correctly the constraints for polydisperse (PD) parameters. 
Now PD params will appear in the combo box menu of the Complex Constraint window, and they will behave as the main parameters (constraints can be added, selected/deselected, edited, etc.)

Fixes #1588

## How Has This Been Tested?

Tested on the 3 AOT data sets of the simultaneous 1D fitting tutorial, combining standard and PD constraints and trying the available options (check/uncheck, remove constraint, using a formula) and everything worked as expected.

## Known issue

A pending problem is that the PD constrains are not loaded with the project, so they have to be reintroduced every time one loads a project. This is not unexpected, as there are comments (TODO: add polidyspersity and magnetism) in FittingPage.readFitPage and FittingPage.saveToFitPage. The PD constraints are nevertheless written into the project json file, but not loaded. But I suggest to merge this PR and open a separate issue for this.

- [ ] Code has been reviewed


gonzalezma:
A few fixes/improvements to do after merging this PR, are described in issues #2352, #2353, #2355 and #2356.

wpotrzebowski:
Sorry, I've managed to break it. I am getting error: 
`21:45:41 - ERROR: Fitting failed: Traceback (most recent call last):
  File "sas/qtgui/Perspectives/Fitting/FitThread.py", line 79, in compute
  File "sas/qtgui/Perspectives/Fitting/FitThread.py", line 19, in map_apply
  File "sas/qtgui/Perspectives/Fitting/FitThread.py", line 16, in map_getattr
  File "sas/sascalc/fit/BumpsFitting.py", line 354, in fit
  File "sas/sascalc/fit/BumpsFitting.py", line 354, in <listcomp>
KeyError: 'radius'
21:45:41 - ERROR: 'radius' `
supposedly with this set of constraints:
Screenshot 2022-10-31 at 21 50 31

gonzalezma:
To be investigated. Error appears when the constraints are defined "cyclically": M2=M1 and M3=M2, and also with different behaviour depending if they are main or PD params.

gonzalezma:
The error did not come from the fit itself, but from the new procedure to compute the uncertainties of tied parameters. The call to problem.setp_hook() in BumpsFit.fit computes the uncertainty of the first parameter in the linked chain, i.e. if we have M2=M1 it will compute the uncertainty for M2 based on the bumps result for M1's uncertainty and will convert M2 to an uncertainty object. But it will not do the same for M3, as M3 is tied to M2 and M2 is not yet an uncertainty object (I guess). But a second call to problem.setp_hook() is able to resolve M3's uncertainty. Based on this finding, I implemented a solution by calling repeatedly problem.setp_hook() until all the parameters are uncertainty objects (or a maximum number of iterations, arbitrarily set to 10, are reached). I suppose that a more elegant solution addressing this problem directly in ParameterExpression or bumps is possible, but I could not find how to do it.

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -310,7 +326,19 @@ def fit(self, msg_q=None,
...
+        iter = 0
+        while not problem.setp_hook._allComputedParamsUncertaintiesDefined():
+            iter += 1
+            if iter > MAX_ITER:
+                logging.warn(f"Handling of uncertainties of constrained parameters "
+                            f"exited after {MAX_ITER} iterations.\n"
+                            f"This could be due to a too complex constraining scheme "
+                            f"or a very long chain of linked constraints.")
+                break
+            problem.setp_hook()
```
pkienzle, line 341: The code in `sascalc.fit.expression` determines an expression evaluation order required to propagate the constraints through the various expressions. It should work equally well for propagating the uncertainties.

The fact that you need this loop suggests a deeper problem with the expression evaluation order, meaning that the constraints will not be consistent within the model. Fix the evaluation order in `_compile_constraints` and the uncertainty propagation should happen automatically.

gonzalezma: Thanks for the hint. The order of the constraints seems to be correctly evaluated, but the order.reverse() command at the end of the function seems to spoil it. By commenting it out everything seems to work fine. I don't know if there could be some special case where this reverse command was needed. From my side, I did several tries linking the constraints in different ways and found that without the reverse everything works as expected and the uncertainties for all the parameters are well resolved.

butlerpd:
Think this is now fixed based on @pkienzle comments but need some thorough testing now.

src/sas/sascalc/fit/expression.py
```diff
@@ -379,10 +377,10 @@ def order_dependencies(pairs):
         else:
             left, right = [set(s) for s in zip(*pairs)]
             resolved = dependent - left
-        #print "independent", independent, "dependent", dependent, "resolvable", resolved
+
         order += resolved
-        #print "new order", order
-    order.reverse()
+
+    #order.reverse()
```
pkienzle, line 383: That doesn't seem right. I'm pulling the independent variables from the list until there are no more independent variables, so the last variables pulled must be the first ones evaluated. That's why the order is reversed.

Removing the `reverse()` breaks the tests in `expression.py`(not that you can tell because pytest isn't run on the source tree, only the test tree).

To debug this you are going to need to print the function produced in `expression._compile_constraints` to make sure it is what you are expecting.

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -341,21 +341,26 @@ def fit(self, msg_q=None,
...
                         # Parameters p is constrained on.
+                        # This does not work if the tying parameter is not in varying list, i.e.
+                        # if there are linked constraints
                         reference_params = [v for v in varying if str(v.name) in str(constraints[param_name])]
                         err_exp = str(constraints[param_name])
                         # Convert string entries into variable names within the code.
-                        for i, index in enumerate(reference_params):
-                            err_exp = err_exp.replace(reference_params[index].name, f"reference_params[{index}].value")
+                        for i, _ in enumerate(reference_params):
+                            err_exp = err_exp.replace(reference_params[i].name, f"reference_params[{i}].value")
```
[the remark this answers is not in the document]
gonzalezma: Agree. It seems the purpose was to handle the parameters missing the uncertainties, but I don't think this was ever functional. I did not dare to remove the code, but after your comment I think I will do it.

src/sas/sascalc/fit/expression.py
gonzalezma: I just checked this empirically, simply verifying the output of order_dependencies (and the result in SasView), and I see that the order of the dependencies is well determined before doing the reverse. For example, assuming that there is a single independent parameter (M1), but that a user will define constraints as M2=M1, M3=M2, and M4=M3, the function (without the reverse) returns the expected order [M2, M3, M4]. So I did not go yet in the detail of the logic of the function, but I'm sure that at least in my test case the reverse will make things fail.

pkienzle:
@gonzalezma you were right about `reverse()`. My test was wrong so the order was wrong. I fixed the existing test and added an explicit test for the a chained expression dependency. Be sure to pull before making more changes.

gonzalezma:
@pkienzle Thanks for checking and clarifying. I pulled your changes and will remove the now unnecessary code.

gonzalezma:
Last commit affects only the interface and should allow closing issue #2372. With this commit all parameters (either adjustable, fixed or constrained) will appear on the RHS of the Complex Constraint interface and can be used in the definition of a new constraint.

It also fixes the FittingWidget.paramHasConstraint method, which was working for polydisperse parameters but not for the standard one. With the changes in this commit, this method is no longer useful, but I decided to keep it as well as the commented command in ComplexConstraint.setupParamWidgets, in case we could adopt a different logic in future and decide that constrained parameters should not be used in the definition of a new constraint.

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -297,17 +297,14 @@ def fit(self, msg_q=None,
...
-                    param.value = uncertainties.ufloat(val, err)
-            else:
                 # Use the covariance matrix to calculate error in the parameter
                 fitted = uncertainties.correlated_values(values, cov)
                 for param, val in zip(varying, fitted):
                     param.value = val
+            except Exception:
+                # No convergence. Convert all varying parameters to uncertainties objects
+                for param, val, err in zip(varying, values, errs):
+                    param.value = uncertainties.ufloat(val, err)
```
pkienzle, line 307: There is a small concern that the Exception code is eating an error unrelated to numerical stability. If for instance the sizes of `varying`, `values` and `cov` do not align. Since the except block will raise an exception on misaligned `varying` and `values`, it may only be `cov` that is messed up.

I propose an additional line **before** the try block (untested):
```
assert len(values) == cov.shape[0] == cov.shape[1]
```
If you had it in the try block then a failure would be trapped by the except block and you would still silently ignore it.

A couple of flags:

Do we really want to rely on python's `a cond b cond c` is equivalent to `a cond b and b cond c`? If you use `a == b == c` in most other languages you may be surprised by the result (I suspect it is undefined behaviour in many cases, depending on whether it is interpreted as `(a == b) == c` or `a == (b == c)` or just rejected as bad syntax).

IIRC, `assert` is suppressed if python is started with the optimization flag so we should not be relying on it for correct code.

You could be boring and use the following, but I prefer the compact assert statement:
```
if len(values) != cov.shape[0] or len(values) != cov.shape[1]:
    raise ValueError("Covariance matrix has the wrong shape")
```

src/sas/sascalc/fit/BumpsFitting.py
```diff
@@ -341,21 +338,13 @@ def fit(self, msg_q=None,
...
                         # value.n returns value p
                         pvec.append(p.value.n)
-                        # value.n returns error in p
+                        # value.s returns error in p
                         stderr.append(p.value.s)
                     # p constrained based on another parameter
+                    # Warning: This will not work for parameters that are tied
+                    # to other parameter already constrained.
+                    # Is this still needed with the added _allComputedParamsUncertaintiesDefined test?
                     else:
```
pkienzle, line 347: Do you ever get into the `else` condition?

I think you can replace the entire if-else block with:
```
assert isinstance(p.value, (uncertainties.core.Variable, uncertainties.core.AffineScalarFunc))
# value.n returns value p
pvec.append(p.value.n)
# value.s returns error in p
stderr.append(p.value.s)
```

src/sas/sascalc/fit/BumpsFitting.py
gonzalezma: Fine. There was already an assert to check that values and errs existed, so I opted to add the simplest assert there as well.

src/sas/sascalc/fit/BumpsFitting.py
gonzalezma: Now that everything seems fixed, the execution flow is not entering the else condition any more in any of my tests. So I removed the else block entirely. But as a measure of precaution, I kept a try/except, so that if by any strange reason a parameter is still not an uncertainty object, at least the user will get the parameter value (+ some error message). But if this is judge completely unnecessary, I do not have any objection in removing the try/except.

pkienzle: My inclination would be to drop it, but I don't object to leaving it in.

In principle one should strive for 100% code coverage with tests. The except block is untestable (you can't trigger the except without extraordinary measures) . Let the higher up exception handler catch future broken code.

pkienzle:
I've only reviewed the code in sascalc not the GUI and I haven't tried running it, so I'll leave it to others to approve the PR.

wpotrzebowski (approved):
It works as expected if the weights are not modified. When I changed weights 1.5 form M1, and then change them all to 1.0 weight for M2 and M3 were not reset (see attached image, parameters and log). I think it is unrealeted issue and therefore I would vote for merging this one and addressing weighting separetly. 
Screenshot 2022-11-15 at 11 16 17

gonzalezma:
The weighting is correct (at least the fact that the values shown in the log are not the same as those given by the user in the interface is the expected behaviour, not a bug). I realize that this is confusing, but the idea is that the weights shown in the log are the effective weights of each data set that are passed to bumps in order to try that each data set contributes to the fit with a "force" proportional to the user weights given by the user in the interface (sorry, not sure that this is very clear either!).

When you give your 3 data sets a weight of 1, the effective weights shown in the log are automatically calculated to take into account the different number of points and error bars in each set, such that the 3 sets contribute more or less equally to the set. 

A clear documentation explaining how the weighting works is needed, but that is a separate issue.

src/sas/sascalc/fit/BumpsFitting.py
gonzalezma: Let's do that then. I just removed a few more lines :-)

wpotrzebowski:
What is bit puzzling with weighting is that one cannot come to defaults vaules i.e. if all values are set to 1.0 in parameter table the info in log shows: 
```
09:03:44 - INFO: ('FitPage1', 1.0)
09:03:44 - INFO: ('FitPage2', 1.0)
09:03:44 - INFO: ('FitPage3', 1.0)

```
Once the values are modified e.g. M1 weight is set to 1.5 and then you run fit the values changes (that's expected) but when you change them all to 1.0 again the "INFO" weights won't come back to one. I agree is beyond the scope of this PR and may not be desired behavior.

wpotrzebowski:
I figured one potential issue while doing the test on the latest code. This may be however more general issue with constrained fitting. When you set up restraint M2.radius = M1.radius and then in the M1 model panel deselect radius the fit still seems to start, however it returns the error: 

0```
9:14:06 - ERROR: Fitting failed: Traceback (most recent call last):
  File "sas/qtgui/Perspectives/Fitting/FitThread.py", line 79, in compute
  File "sas/qtgui/Perspectives/Fitting/FitThread.py", line 19, in map_apply
  File "sas/qtgui/Perspectives/Fitting/FitThread.py", line 16, in map_getattr
  File "sas/sascalc/fit/BumpsFitting.py", line 336, in fit
AssertionError  
```

I guess the proper treatment of this would be to remove constraint when certain parameter is disabled? Again, it may be beyond the scope of this PR

**Comment by gonzalezma** [comment]
Concerning the weights, the way to recover the default values is to deactivate the 'Modify weighting' option. But I agree that the difference between the values in the log info and the user interface is puzzling. So this is something to discuss and probably to improve in the future.

**Comment by gonzalezma** [comment]
Well spotted the error when a parameter is tied to a parameter that it is not a free parameter. The final cause for the error is that the fix parameter is not an uncertainty object, so the tied param will not be it either, and after the last changes discussed with Paul K this result in an assertion error. The question is if such a constraint definition makes sense and if we should allow it or not.

Note that fixed parameters are still allowed and correctly handled when used as part of a formula. E.g. defining 
M2.scale = M1.scale + 0.001 * M1.sld_core (where M1.scale is a free param and M1.sld_core is fixed) results in M2.scale having the expected value according to the formula and the same uncertainty as M1.scale.

So the quick fix here is to recover the try/except with a warning to the user that for a given parameter the uncertainty could not be computed, such that at least SasView does not stop.

Otherwise, we need to address the root problem as suggested by Paul K, meaning that we need to implement a method to identify such problematic definitions and then decide what to do, as I can see several possibilities:
1. Deactivate the constraint, as you propose.
2. Assume that the user intention was correct, so "unfree" the constrained parameter and fix it to the same value as the param on the RHS.
3. Inactivate the fit button and show a warning to the user explaining that this is an unclear case and asking him/her to intervene manually (either setting the RHS param as a free param or setting the LHS param as a fixed param with the value directly given by the user).

I can give a look to that, but I'm not sure that I can manage to find a useful solution in a reasonable time, so I would propose to go for the quick fix in this PR and then discuss in one of our calls if something better is needed.

**Comment by pkienzle** [comment]
The constraints should be independent of fitting. That is, I want to be able to set the two radii to be equal (so tied by an expression) even if neither are selected for fitting. The GUI should allow this without having to toggle fitting on to set up the constraint and off again to suppress fitting.

You can easily replace the assert with a condition, setting stderr to either NaN or zero for propagation of unconstrained values.  I suggest replacing the entire `if success-else` block with the following (untested):
```python
# Note: Derived values will not have uncertainty if the inputs are fixed, so no 'n' or 's'.
fitting_result.pvec = np.array([getattr(p.value, 'n', p.value) for p in pars])
fitting_result.stderr = np.array([getattr(p.value, 's', 0) for p in pars])
DOF = max(1, fitness.numpoints() - len(fitness.fitted_pars))
fitting_result.fitness = np.sum(fitting_result.residuals ** 2) / DOF
```

This produces the χ2 associated with the current set of parameter values, and the fitted parameters uncertainty estimates even if the fit quits early. Leave it to the GUI to decide how to inform the user on the basis of `fitting_result.success` rather than throwing away the information that we have.

For now you can reproduce current behaviour by adding:
```python
# TODO: Let the GUI decided how to handle success/failure.
if not fitting_result.success:
    fitting_result.stderr[:] = np.NaN
    fitting_result.fitness = np.NaN
```

There are other code simplifications that could be done but this is enough for now.

gonzalezma:
The problem is that we may have some ambiguous situations, where it is not completely clear the user intention. Some examples and the current behaviour of SasView are:

1. M1 = 1 (not selected for fitting), M2 = 2 (not selected for fitting), M2 = M1
    ==> Should M2 be set equal to 1 or 2?
    Currently, in SasView it is not possible to define the constraint M2=M1 from the menu (as only fittable parameters are shown in the RHS of the constraint menu), but it is possible to edit a line to add such constraint. In this case, the constraint will be applied and M2 = 1.

2. M1 = 1 (not selected for fitting), M2 = 2 (selected for fitting), M2 = M1
    ==> Should M2 be a free parameter (and therefore the constraint ignored)?
    ==> Should the constraint take preference and thus M2 be fixed to 1? This is what SasView does now.
    ==> Should the constraint take even more preference and assume that M1 should be fitted as well, so that M2 = M1 = fitted value?

In any case, the current behaviour does not seem unreasonable to me, so I would vote to keep it as it is. The errors that we have observed do not appear during the fitting, but in the post-processing of the uncertainties. I will try the new suggestion from Paul K. Otherwise, I will go back to a simple try/except block to take different actions depending if M1 is just a number (fixed parameter) or an uncertainty object (fitted param). This works also, and it is not very clear to me which information we would be throwing away.

pkienzle:
The existing constraint system is purely assignment. With a mathematical constraint such as `R' = (R^2 L)^{1/3}` you could imagine that any of pair of R', R, and L would be fittable and you could solve for the remaining parameter, but the code is not that clever. Instead you need to structure your constraint so that the fitted parameters are on the RHS, with the LHS computed from their values, and never fitted. In that context there is no ambiguity: M1 = M2 should be interpreted as M1 is assigned the value of M2, with M1 never fitted.

Given that constraints are only being used to tie parameters between models and almost(?) never to calculate arbitrary expressions you could scrap the existing GUI and replace it with something easier to control. This could build up the set of expressions needed to apply the constraints. That is outside the scope of this PR and better left as a discussion.

rozyczko:
1. Create two constraints, one of nonpolydisp. params and one of polydisp. params. E.g. `M2:radius.width = M1.radius.width` and `M2:radius = M1.radius`.
2. Go to M2 and right click -> Edit constraint on the `radius.width` row. This works.
3. Go to M2 and right click -> Edit constraint on the non-pd `radius` row.
```
11:42:16 - ERROR: Traceback (most recent call last):
  File "D:\projects\sasview_new\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1050, in editConstraint
    if not np.any([FittingUtilities.isParamPolydisperse(p, self.model_parameters, is2D=self.is2D) for p in params_list]):
  File "D:\projects\sasview_new\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1050, in <listcomp>
    if not np.any([FittingUtilities.isParamPolydisperse(p, self.model_parameters, is2D=self.is2D) for p in params_list]):
  File "D:\projects\sasview_new\src\sas\qtgui\Perspectives\Fitting\FittingUtilities.py", line 967, in isParamPolydisperse
    if '.width' in param_name:
TypeError: argument of type 'NoneType' is not iterable
```
You are probably trying to query all constraints based on the model of the first constraint rather than on both models?

gonzalezma:
@rozyczko Yes, this is something that I already noted on the code camp, but did not try to fix yet. I described the problem in issue 2355. All the work and testing in this PR was done using only the Const. & Simul. Fit tab. It's in my plans to try to solve that issue too, but I don't know when I'll have time. In the meantime, I would argue that this PR fixes the main problem of being able to constraint also polydisp. parameters and that I expect most users will use this functionality from the simultaneous tab, so issue 2355 is a minor problem and should not delay merging this PR. To discuss in the next call?

rozyczko:
> @rozyczko Yes, this is something that I already noted on the code camp, but did not try to fix yet. I described the problem in issue 2355. 

Ah! OK then, as long as we have this listed as an issue then I agree it's better to have the functionality in, before the codebase diverges and we have to spend time solving merge conflicts again, considering how big this PR is.

wpotrzebowski:
As discussed at the biweekly call we are merging this PR and saving related issues/discussion for the future.

PR #2357, opened 2022-10-31
Fixes in invariant prespective

### LINKED ISSUE #2334: Invariant failures
**Describe the bug**
A couple of major issues were found during code camp:
1. the q range is no longer settable. The range test boxes act as labels and the range sliders do not update the boxes either.
2. Using the example apoferritin.txt data, the perspective computes ok with the default values for background(0) and scale (1) and contrast (8e-6). It also works if you choose to extrapolate both low and high Q.  However, if one enters correct values (from the fitting of the data), namely 0.0023 background, 0.0003 scale and 4e04 SLD the calculator fails with:`Calculation failed: not enough values to unpack (expected 3, got 0)`. Enabling the extrapolations leads the same error being thrown for each of the extrapolations (so 3 errors).  E.g. the low Q extrapolation leads to `Low-q calculation failed: not enough values to unpack (expected 3,got 0)`

It is not clear why the correct values should fail unless it is due to the large error bars at high q which cannot be removed due to inability to constrain the q range used.

**To Reproduce**
1. Load the `apoferritin.txt` example data
2. send the data to the invariant perspective -- things should work
3. try to adjust the range (either with the sliders or manually in the panel) -- you cannot
4. Change the input parameters to reasonable values 
8. the computation now fails

**Expected behavior**
it should work

 - Version: 5.0.5

 - OS: [e.g. Windows 10]

wpotrzebowski:
## Description

As described in #2334 two issues related to invariant perspective were identified. The firts issue with read only fields for Qmin and Qmax should be fixed now. However the other problem is more deconvoluted as it is related to the fact that we calculate invariant based on `(scale * data) - background`, which requires correct setting for scale and background. As it turned out for apoferitin example the values coming from fitting may be over/underestimated and therefore Q* may become negative. Warning has been added to the code suggesting checking backround and scale value. In the future we may consider more inteligent guidance.

I've checked documentation. It seems to be consitent but maybe we should add note about scale/background issue? 

Fixes #2334 

## How Has This Been Tested?
1) Load data set 
2) Send to Invariant
3) Check if you can modify Qmin and Qmax
4) (to check other problem) Play with scale and background (e.g. small scale, high background) and check if you get proper warning. 

- [x] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Perspectives/Invariant/UI/TabbedInvariantUI.ui
+++ b/src/sas/qtgui/Perspectives/Invariant/UI/TabbedInvariantUI.ui
@@ -144,7 +144,7 @@
                   <bool>true</bool>
                  </property>
                  <property name="readOnly">
-                  <bool>true</bool>
+                  <bool>false</bool>
                  </property>
                 </widget>
                </item>
@@ -161,7 +161,7 @@
                   <bool>true</bool>
                  </property>
                  <property name="readOnly">
-                  <bool>true</bool>
+                  <bool>false</bool>
                  </property>
                 </widget>
                </item>
--- a/src/sas/qtgui/Perspectives/Invariant/UI/TabbedInvariantUI.ui
+++ b/src/sas/qtgui/Perspectives/Invariant/UI/TabbedInvariantUI.ui
@@ -144,7 +144,7 @@
                   <bool>true</bool>
                  </property>
                  <property name="readOnly">
-                  <bool>false</bool>
+                  <bool>true</bool>
                  </property>
                 </widget>
                </item>
@@ -161,7 +161,7 @@
                   <bool>true</bool>
                  </property>
                  <property name="readOnly">
-                  <bool>false</bool>
+                  <bool>true</bool>
                  </property>
                 </widget>
                </item>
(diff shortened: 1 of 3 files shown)


butlerpd (commented):
Code looks fine and addresses the very narrow scope of the PR in that the range boxes are now editable. However they seem to be doing nothing at all -- which may be why they were made not editable in the first place?  The range sliders are not visible unless extrapolations are set since the plot disappears unless one turns on extrapolation. However, moving them does not update the values in the range boxes nor does it seem to affect the invariant calculation either.

I would recommend finding out who turned off the editing and/or folks who've been using it like @phgilbert?  It may be that it should be merged and new tickets put in but I thought it was working well in 5.0.5 so wonder what has happened?

wpotrzebowski:
@butlerpd indeed... Good catch. I see branch of code where the data is set, so I can hopefully fix properly but as suggested I would investigate why the fields where disabled in first place.

wpotrzebowski:
https://github.com/SasView/sasview/commit/8a607a94fa237125345cb5e0fe1b14729e56f9f4

Issue #2316, opened 2022-10-28
Refactoring of preferences


lucas-wilkins:
The following things need to be done to the preferences panel

* Cancel button -> #2358
* #2313 -> #2358
* #2312 -> #2358
* #2317 -> #2381

This can mostly be achieved through:
1) Modifying the preferences dialog system to have a more formal class structure
2) Better integration with config

krzywon:
#2312: The `PreferencesWidget` class has a `restoreDefaults` attribute that houses a callable that is when the button is clicked. This just hasn't been implemented in any of the child classes yet. This should really change to a class method that child classes override. Update child classes to use this method. Question: should this reset the defaults in the current panel or in all panels?

#2313: Changes to make: Add a `Cancel` button, remove (most) connections from GUI elements, and update everything when the `OK` button is clicked. Would an `Apply` button be useful, too?

pbeaucage:
#2313 another, potentially easier solution would be to change the underlying config class to have some method set like `beginStagingChanges()`, `commitStagedChanges()` and `discardStagedChanges()`.  begin would be called on dialog open and basically `copy()` the structure to a backup place.  Then `commit...()` would replace main copy with the edited copy and `discard...()` would replace the stored version with the editable copy.

pbeaucage:
Not in general, depends on the config items.  The `Plotting` ones all take effect immediately.

lucas-wilkins:
Also, probably need to have a "your changes will not be applied untl you restart sasview" message for the display settings

krzywon:
Comments from #2306 that I am documenting here:

- [ ] `preferences.rst` word-smithing
- [ ] Potential performance improvements if one of `bumps.fitting` -or- `bumps.options` imports was removed from FittingPerspective.py (maybe needs it's own issue)
- [ ] Better handling of `name` passed when adding widgets
- [ ] Add `NotImplementedError` to `PreferencesWidget._addAllWidgets` and planned `PreferencesWidget.restoreDefaults`
- [ ] Better separation of GUI and logic for 'PreferencesWidget' children
- [ ] Move fitting options window back into fitting perspective to retain the perspective-as-plugin framework

krzywon:
> Also, probably need to have a "your changes will not be applied untl you restart sasview" message for the display settings

See #2317 - Tracking what is changed, we could better handle when a restart is required and only prompt the user when necessary.

lucas-wilkins:
Think this is done now.

PR #2358, opened 2022-10-31
2316: Refactoring preferences

### LINKED ISSUE #2312: Restore Defaults in the Preferences window does nothing
No changed option goes back to the default on clicking `Restore Defaults`.
Is it the GUI that needs refreshing or are the defaults not loaded at all?

### LINKED ISSUE #2313: Preferences immediately saves the values, even if OK isn't clicked
config file `config-5.json` gets written with currently entered values, without clicking the OK button.

krzywon:
## Description

This hopes to combine parts of what was done in #2306 and #2321, specifically related to formalizing the preferences panel and widget class structures.

As asked for in #2316, the following has been added to this PR:
- Settings are staged in the preferences panel until they are accepted and only then is the config updated (using the OK or Apply buttons)
- All staged changes are discarded if they are rejected (using the Close button or X)
- Restore defaults resets the active widget to the default values as read in from the configuration
- Added 'Cancel' and 'Apply' buttons
- Updated the help documentation to include expected behaviors and information on specific

Fixes #2312
Fixes #2313

## How Has This Been Tested?

- Unit tests that are in this PR
- From `run.py`:
  - Open preferences, modify settings, click Cancel, open preferences, and be sure previous changes are gone
  - Open preferences, modify settings, click X, open preferences, and be sure previous changes are gone
  - Open preferences, modify settings, click Ok, open preferences, and be sure previous changes are persistent
  - Open preferences, modify settings, click Apply, click Close and/or X, open preferences, and be sure previous changes are persistent
  - Open preferences, modify settings, click Apply, click Close and/or X, open preferences, click Restore Defaults, and be sure values reset to defaults

- [x] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/MainWindow/media/preferences_help.rst
+++ b/src/sas/qtgui/MainWindow/media/preferences_help.rst
@@ -1,11 +1,45 @@
-.. preferences_help.rst
-
+. preferences_help.rst
 .. J Krzywon wrote initial draft August 2022
-
+.. Last Updated: J Krzywon, October 2022
 .. _Preferences:
 
 Preferences
 ============
 
-SasView has a number of user-settable options available. Each heading will give more information about a
-particular group of settings. Any preferences not held within this window are planned to move here in a future release.
+SasView has a number of user-settable options available. For information on a specific preference, navigate to the
+appropriate page heading. Not all preferences are housed here, but will be eventually.
+
+A number of these preferences will only apply to the current SasView window and will reset when closing. Others,
+labelled *persistent* in this document, will be retained for the next time SasView is run.
+
+:ref:`Plot_Preferences`, :ref:`Display_Preferences`
+
+.. _Plot_Preferences:
+
+Plotting Preferences
+--------------------
+Plotting preferences only apply to new plots. Existing plots will retain existing settings.
+
+**Use full-width plot legends (most compatible)?**: With this option selected, plot legends will always be the full width
+of the plot it is on. The legend will also be partially transparent to better view the data. *persistent*
+
+**Use truncated legend entries?**: By selecting this option, legend labels are truncated to a single line of length
+*Legend entry line length* leaving only the beginning and end of the label. The central characters will be replaced with
+an ellipsis and whitespace. *persistent*
+
+**Legend entry line length**: This defines the maximum number of characters to display in a single line of a plot legend
+before wrapping to the next line. *persistent*
+
+.. _Display_Preferences:
+
+Display Preferences
+-------------------
+The display preferences modify underlying features of our GUI framework, Qt. For more information on each setting,
+please read more on the `Qt High DPI Settings <https://doc.qt.io/qt-5/highdpi.html#high-dpi-support-in-qt>`_.
+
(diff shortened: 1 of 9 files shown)


lucas-wilkins (commented):
You could use a QFormLayout to make things line up.

lucas-wilkins:
I've added type checking to config setting because currently the text fields are not validated (will add a config type checker next). This causes some pretty bad issues if you set an int to a string, for example (I tested it).

lucas-wilkins:
config will now raise a `TypeError` if you try and give it the wrong type. This would still need to be handled on the gui side (use `.validate(name, value)` to just test an entry

krzywon:
Thanks for the info. I'll add Qt validators for ints and floats and apply them to the appropriate inputs.

gonzalezma:
I think that the default option for Plotting Options/Use full-width ... should be that it is not checked. Otherwise one gets often really ugly plots with the legend appearing just in the middle of the screen. The alternative would be that for each plot the legend size is resized to the length of the longer label, but this will be harder to implement. With this option unchecked we recover the same behaviour as before, while given the user the possibility to modify that whenever the data labels are too large. Also after playing with the options and changing the legend length a couple of times I got an error:
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\MainWindow\GuiManager.py", line 1294, in showPlot
    self.filesWidget.displayData(plot, id)
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\MainWindow\DataExplorer.py", line 1087, in displayData
    self.plotData([(plot_item, plot_to_show)])
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\MainWindow\DataExplorer.py", line 1156, in plotData
    new_plot.plot(plot_set, transform=transform)
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\Plotting\Plotter.py", line 243, in plot
    if len(l)> config.FITTING_PLOT_LEGEND_MAX_LINE_LENGTH:
TypeError: '>' not supported between instances of 'int' and 'str'

So probably the plotting options need more testing until finding the best default params.

gonzalezma:
In the doc, I suppose this is not a % scaling. If so remove percent from
**QT Screen Scale Factor**: A percent scaling factor ...
And trying to change this factor, I got the following error:
Cannot set set variable 'QT_SCALE_FACTOR', improper type (Cannot coerce <class 'str'> to bool)
Could this be related to the type checking?

lucas-wilkins:
@gonzalezma 

> 'QT_SCALE_FACTOR', improper type (Cannot coerce <class 'str'> to bool) - Could this be related to the type checking?

Yes, that would be - though indirectly. Not sure why it would say bool, unless something is wired up wrong. It should be able to coerce a string to a float if it is valid (I think)

lucas-wilkins:
> @gonzalezma

Nope.

1) it wont convert strings to floats
2) the error message for this says bool, oops

krzywon:
I've added validators to the float and integer inputs and ensured the values are coerced before hitting the config

butlerpd:
moving legend location to be put into a new PR. Lucas to do proper review and hopefully approve

lucas-wilkins (approved):

PR #2368, opened 2022-11-03
Add masked data toggle to 2D plots

krzywon:
## Description

This adds a toggle to show/hide masked data points from 2D data plots. This is useful when visualizing (and slicing) multi-detector data that has masked overlap regions. A series of unit tests and user documentation were added for the new context menu action. Two previously broken unit tests for the 2D plotting were also fixed here.

## How Has This Been Tested?

In addition to the unit tests added here, I performed the following steps:
- Loaded data set with masked regions overlapping with unmasked regions
- Plot data to see masked areas
- Trigger toggle
- Only unmasked regions remain

- [x] Code has been reviewed


pkienzle:
If the masking is coming from one detector shadowing another, wouldn't it make more sense to split the data in two, one for foreground (unmasked) and another for shadow (masked)? This allows you to use sasview tools to view and manipulate them independently. You already handle files with multiple data sets so this doesn't add any complexity to the program.

The show/hide mask operation causes the foreground and shadow pixels to be interleaved in the plot. Instead you may want to regrid foreground and shadow to the same grid and add them. That way you are using the shadow to capture any intensity that leaked through the foreground. If you split them into separate data entries you can accomplish this by adding the entries with the data operation.

The only reason I see to keep foreground and shadow data in the same entry object is when you want to fit them with the same model. If they are separate then you will need to use simultaneous fitting simultaneous fitting for this, or support a "merge" data operation which interleaves the foreground and shadow rather than adding them on a common grid. I expect this use case is rare enough that this will be an acceptable workaround.

Different instruments may use masking in different ways. For example the pixels behind a beam stop could be masked, or there may be support structures that block the beam above some parts of the detector. Hot/dead pixels might also be masked if the reduction program knows about them. Again, separate datasets seems like a better solution.

butlerpd:
Not sure I follow exactly @pkienzle but as I understand it you are suggesting splitting every detector entry in two with one for "good pixels" and one for "bad pixels" (as far as I know almost all detectors have some region masked)? 

As I recall there was some discussion in the community years ago about whether reduction should remove "bad" pixels and the consensus was that should never be the case and that they should simply be flagged as masked.  Partly I think because one hates to "delete data" and partly because most masks are a bit arbitrary and it is deemed that the user may want to inspect the "bad" pixels and possibly even decide that masking was too aggressive and unmask some pixels. On the other hand they may decide the masking was not aggressive enough and choose to mask more.

Again, I'm not sure I totally understand what you are advocating here but not sure that it would be consistent with how the community seems to want to work with the data?

pkienzle:
Yes, I'm suggesting that it will make for a simpler and more flexible interface to split the masked and unmasked data on load. To handle mask adjustments you would need a tool to transfer data points between the two sets. Call it a mask editor 8-)

For the particular case of VSANS where some of the panels may be shadowed by other panels I'm suggesting the interpolating the data between the points is not the best algorithm. If there is enough intensity leaking through the front panel to matter then interpolating and adding makes more sense. (I hope this is not the case, or if it is, that it be handled by the reduction program; mostly I was suggesting what users might do with the additional flexibility of a separate data object.)

butlerpd:
Thanks @pkienzle. As usual this is still a bit beyond my skill level I think but I just wanted to make sure we were on the same page about what we were talking about and how it is used.  It seems we are and as long as it is transparent to poor users like me the details of the implementation shouldn't matter for usability? And coding considerations can then govern? ...  As long as it is not trying to force the rest of the world to change how they store their data 😄 That could also be done if important but would require a different group to get together and agree... But you said "internally" so I think we are good.

murphyryanp:
@krzywon, Sorry for the delay! The mask toggle works nicely. But, after playing around with things a bit more, I found a bug that prevents one from performing a circular average after masking the data.

Here are the steps to reproduce the error:
1. Load a 2D qxqy dataset
2. Create new plot (perform circular average works fine at this point)
3. Right click on graph and click "Toggle Masked Data" (all slicers work here and still ignore the mask)
4. Right click on graph and click "Perform Circular Average" (average plot is not generated and error is given below)

Here is the error info:
_13:45:07 - INFO:  --- SasView session started, version 5.0.5, 2022 ---
13:45:07 - INFO: Python: 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64 bit (AMD64)]
13:46:05 - ERROR: Traceback (most recent call last):
  File "sas\qtgui\Plotting\Plotter2D.py", line 338, in onCircularAverage
  File "sas\qtgui\Plotting\Plotter2D.py", line 307, in circularAverage
  File "sasdata\data_util\manipulations.py", line 608, in __call__
IndexError: boolean index did not match indexed array along dimension 0; dimension is 49152 but corresponding boolean dimension is 25153_

image

krzywon:
@murphyryanp - this now includes a fix for the bug you found. Take a look when you get a chance.

butlerpd:
Are we still waiting on @murphyryanp to approve?

murphyryanp (approved):

murphyryanp:
It works nicely!

PR #2381, opened 2022-11-09
2317: Prompt user to restart when preferences change

### LINKED ISSUE #2317: Changing preferences should prompt user to restart SasView
Changing the preferences and saving them by clicking OK should tell the user to restart the application.
Otherwise, it is unclear when/if the new values are loaded.

krzywon:
## Description

This PR prompts the user to restart SasView if changes are made that only take effect at the next startup of SasView.

**NOTE** This PR is currently chained off [2313-save-on-ok](https://github.com/SasView/sasview/tree/2313-save-on-ok). Once #2358 is merged, I will rebase this to `main`.

Fixes #2317

## How Has This Been Tested?

Run SasView, open the preferences window and change one of the display parameters that require a restart.

- If 'Cancel' is clicked, ensure no message dialog is generated
- If 'Ok' or 'Apply' is clicked, ensure a message is displayed with a description of changes that will only take effect on restart.

- [x] Code has been reviewed


src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
```diff
@@ -20,6 +20,7 @@
 BASE_PANELS = {"Plotting Settings": PlottingPreferencesWidget,
                "Display Settings": DisplayPreferencesWidget,
                }  # Type: Dict[str, Union[Type[PreferencesWidget], Callable[[],QWidget]]
+ConfigType = Union[str, bool, float, int, List[Union[str, float, int]]]
```
lucas-wilkins, line 23: Just a note, the full type of config elements can be defined as:

```
ConfigType = Union[List["ConfigType"], str, bool, float, int]
```

lucas-wilkins: ...sort of, because as its finite, its actually:

```
ConfigTypeTmp = Union[int, float, str, bool]
ConfigSubtypes = []
for i in range(11):
    ConfigSubtypes.append(ConfigTypeTmp)
    ConfigTypeTmp = List[ConfigTypeTmp]
ConfigType = Union.__getitem__(tuple(ConfigSubtypes))
```

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
```diff
@@ -94,6 +98,14 @@ def _saveStagedChanges(self):
         """ When OK or Apply are clicked, all staged changes should be applied to the config. """
         for k, v in self._staged_changes.items():
             setattr(config, k, v)
+        if any(self._staged_requiring_restart):
+            message = "SasView must restart for the following values to take effect:\n"
+            for val in self._staged_requiring_restart:
+                message += f"\t-{val}\n"
+            msgBox = QMessageBox(QMessageBox.Information, "", message, QMessageBox.Ok)
```
lucas-wilkins, line 105: I'm guessing adding an option for it to restart itself would be a massive pain to get working. I'm not even sure how other software does this.

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
lucas-wilkins: Doesn't really matter for your purposes though

lucas-wilkins (approved):

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
krzywon: Ah, yes, schema elements can be nested Lists. Theoretically, the nesting could be infinite, but practically speaking, never will be. Would it make more sense to put `ConfigType` into sas.system.config.config and import it from there?

lucas-wilkins: I put a hard limit on the nesting, because of `x = []`, `x.append(x)`. 

> Would it make more sense to put ConfigType into sas.system.config.config and import it from there?

Depends if you mean to represent the types that config can handle, or the types that the preferences can handle.

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
krzywon: I looked into this, and it's possible, but would add too much scope creep to this PR.

lucas-wilkins: cool

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
```diff
@@ -94,6 +100,16 @@ def _saveStagedChanges(self):
...
             setattr(config, k, v)
+        if any(self._staged_requiring_restart):
+            message = "SasView must restart for the following values to take effect. Do you wish to restart?:\n"
+            for val in self._staged_requiring_restart:
+                message += f"\t-{val}\n"
+            msgBox = QMessageBox(QMessageBox.Information, "", message, QMessageBox.Yes|QMessageBox.No)
+            msgBox.show()
+            if msgBox.exec() == QMessageBox.Yes:
+                self.parent.guiManager.quitApplication()
+                os.execl(sys.executable, os.path.abspath(__file__), *sys.argv)
```
rozyczko, line 111: Interestingly, this works fine in the installed version, despite `__file__` pointing to relative location of sasview.exe in pyinstaller generated setups. 
If you open the python shell in the installed version, the `__file__` attribute is not even defined.
I would probably use `sys.executable` instead of `os.path.abspath(__file__)`, since in the installed version it points to location of the exe file (without the need to expand the path with `abspath`) and in the dev build, it points to python.exe, with sys.argv containing `run.py` and other arguments.

But since this is working as is, I wouldn't put too much importance in the abve.

rozyczko (approved):
Works fine, except of one potential issue mentioned.
This is, however, just a mild nuisance for users and can be made into a separate issue.

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
```diff
@@ -80,10 +84,12 @@ def restoreDefaultPreferences(self):
         if hasattr(widget, 'restoreDefaults') and callable(widget.restoreDefaults):
             widget.restoreDefaults()
 
-    def stageSingleChange(self, key, value):
+    def stageSingleChange(self, key: str, value: ConfigType, config_restart_message: Optional[str] = ""):
```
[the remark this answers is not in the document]
lucas-wilkins: yeah, need to check if anything has actually changed at some point.

wpotrzebowski (approved):
Works as expected. Then only comment I have is that messange in the pop up window can be better aligned?
Screenshot 2022-11-15 at 11 29 41

lucas-wilkins:
> Works as expected. Then only comment I have is that messange in the pop up window can be better aligned? Screenshot 2022-11-15 at 11 29 41

Minor complaint. I don't think "factoring" is the right word here.

rozyczko:
One more issue - when the scaling value is empty, the parser throws:

```
15:48:40 - ERROR: Traceback (most recent call last):
  File "D:\projects\sasview_new\src\sas\qtgui\Utilities\Preferences\PreferencesWidget.py", line 21, in set_config_value
    value = dtype(value)
ValueError: could not convert string to float: ''
```
Hitting OK with empty line text doesn't trigger the warning and resets the value to default

krzywon:
> One more issue - when the scaling value is empty, the parser throws:

Ticketed in #2389

PR #2393, opened 2022-11-25
Replaced requirements list with install -r requirements

lucas-wilkins:
Just read requirements.txt instead of having a separate list of requirements in `ci.yml`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,12 +84,12 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install wheel setuptools
-
-          python -m pip install numpy scipy==1.7.3 pytest pytest-mock pytest_qt
-          python -m pip install tinycc h5py pyparsing html5lib reportlab==3.6.6 pybind11 appdirs
-          python -m pip install six numba mako ipython qtconsole xhtml2pdf pylint debugpy
-          python -m pip install qt5reactor periodictable uncertainties dominate importlib_resources
-          python -m pip install html2text superqt
+#          python -m pip install numpy scipy==1.7.3 pytest pytest-mock pytest_qt
+#          python -m pip install tinycc h5py pyparsing html5lib reportlab==3.6.6 pybind11 appdirs
+#          python -m pip install six numba mako ipython qtconsole xhtml2pdf pylint debugpy
+#          python -m pip install qt5reactor periodictable uncertainties dominate importlib_resources
+#          python -m pip install html2text superqt
+          python -m pip install -r build_tools/requirements.txt
 
       - name: Install PyQt (Windows + Linux)
         if: ${{ !startsWith(matrix.os, 'macos') }}


llimeht:
I agree with the spirit of the change. Having the list in two places is painful.

I think some changes are required prior to merging:
* the list of packages needs to be checked to make sure they are the same (they are not, see below)
* the overly complicated set of six separate installation steps can be cleaned up

Looking at those steps:
* the separate macos installation steps for matplotlib and lxml seem redundant now and can just go
* the pinning of PyQt only on macos needs to either be replicated in `requirements.txt` or not, and that step can go
* the windows-only installation of mako can go
* pytools were windows-only but are now installed everywhere? Is that desired? are they needed?
* the pyopencl installation can move into `requirements.txt` too (I think @pkienzle showed how that could be done in another discussion)
* `requirements.txt` is now installing `sphinx` everywhere rather than only when building docs

Looking at the packages list:
* the new method misses: 
  * debugpy (maybe it's a transitive dependency of something else?)
* the new method adds: 
  * bumps (this is definitely wrong since apparently we want the latest git version of bumps, not the latest release) 
  * docutils
  * jsonschema sphinx unittest-xml-reporting (my recollection is that it has an incomaptible version pin with setuptools, is unused on CI and mostly unused in the project)
* the new method adds packages everywhere that used to be only on windows: cffi pytools

I wonder if there needs to be either separate `requirements-dev.txt` (and maybe `requirements-doc.txt`) or functionality annotations within `requirements.txt` so that packages can be omitted.

We're also trusting that `pip` will remain able to install all the packages in one swoop. One of the reasons the list of packages got split to begin with was incompatible pinning from different dependencies (`a` required `b > 1`, while `c` required `b < 1` and a single `pip` run will just die with an error, but multiple `pip` runs will succeed because `pip` is not clever enough to enforce dependencies between invocations). Maybe some of the ecosystem has settled enough that we can indeed trust this to work...

lucas-wilkins:
I'm away for the next month or so, and I'm trying not to do any work on SasView in that time - not even got to the end of my first weekend!

I think your suggestions are good, but they do expand the scope of this PR quite a bit, and I'm not going to be doing anything major on this until I get back.

> the list of packages needs to be checked to make sure they are the same (they are not, see below)

My thoughts on this is that if any that were in the CI are missed from requirements.txt they should go in there, otherwise, it doesn't really matter that much - there's not much harm in having unused packages unless their installation cause an explicit issue. That said, a requirements `requirements-dev` isn't a bad idea, I would suggest the encompass both dev tools and docs tools.

> the overly complicated set of six separate installation steps can be cleaned up

Aforementioned scope creep. It sounds like you have something specific in mind though, you can just push to this branch if you like.

> Maybe some of the ecosystem has settled enough that we can indeed trust this to work...

I think so, I don't recall any issues with building from requirements.txt locally, but maybe needs to be checked.

> the pyopencl installation can move into requirements.txt too (I think @pkienzle showed how that could be done in another discussion)

I never figured out why it was separate, I assumed there was reasons. Sounds good if possible.

wpotrzebowski (commented):
I've made some further clean-up, which I think should work on all platforms (at least that's my experience from different local installation). We may need to revert non-binary installation of lxml on Mac for notarization step but shouldn't be concern for the moment.

wpotrzebowski:
Mac doesn't seem to like PyQt=5.15 (testing from the installer) 
Screenshot 2022-11-29 at 06 39 37
How does it look on Windows?

pkienzle:
The `requirements.txt` file can have conditionals. For example (untested):
```
appdirs
cffi
mako
    # for PyOpenCL build
docutils
    # for sasmodels
dominate
h5py
    # for sasdata
html2text
html5lib
jsonschema
importlib-resources
ipython
lxml; platform_system != "Darwin"
lxml --no-binary lxml; platform_system == "Darwin"
matplotlib~=3.5.2
numba
numpy
pyparsing
PyQt5; platform_system != "Darwin"
PyQt5==5.13; platform_system == "Darwin"
pyopencl \
    --only-binary=pyopencl \
    --find-links http://www.silx.org/pub/wheelhouse/ \
    --trusted-host www.silx.org \
    ; platform_system != "Windows"
    # TODO: move --trusted-host to pip.conf
    # https://github.com/pypa/pip/issues/2822#issuecomment-107078353
    # TODO: verify scope of --find-links
    # does it mean silx will be chosen over pypi for all packages that it hosts?
    # Maybe need a separate pip command for pyopencl :-(
pyopencl ; platform_system == "Windows"
pytools
qt5reactor
qtconsole
reportlab==3.6.6
scipy==1.7.3
six
uncertainties
xhtml2pdf

# Project related
# bumps, sasmodels and sasdata need to be built separately so
# that sphinx can borrow from their documentation.
bumps
sasmodels
sasdata
periodictable
tinycc
# nxunit does not yet have a separate package.
#nxunit
```

krzywon:
The installed version looks fine on Windows 10. @pkienzle gave a good example of using conditionals in a `requirements.txt` file in the previous comment, which is probably the way we need to go for now.

image

wpotrzebowski:
> The `requirements.txt` file can have conditionals. For example (untested):
> 
> ```
> appdirs
> cffi
> mako
>     # for PyOpenCL build
> docutils
>     # for sasmodels
> dominate
> h5py
>     # for sasdata
> html2text
> html5lib
> jsonschema
> importlib-resources
> ipython
> lxml; platform_system != "Darwin"
> lxml --no-binary lxml; platform_system == "Darwin"
> matplotlib~=3.5.2
> numba
> numpy
> pyparsing
> PyQt5; platform_system != "Darwin"
> PyQt5==5.13; platform_system == "Darwin"
> pyopencl \
>     --only-binary=pyopencl \
>     --find-links http://www.silx.org/pub/wheelhouse/ \
>     --trusted-host www.silx.org \
>     ; platform_system != "Windows"
>     # TODO: move --trusted-host to pip.conf
>     # https://github.com/pypa/pip/issues/2822#issuecomment-107078353
>     # TODO: verify scope of --find-links
>     # does it mean silx will be chosen over pypi for all packages that it hosts?
>     # Maybe need a separate pip command for pyopencl :-(
> pyopencl ; platform_system == "Windows"
> pytools
> qt5reactor
> qtconsole
> reportlab==3.6.6
> scipy==1.7.3
> six
> uncertainties
> xhtml2pdf
> 
> # Project related
> # bumps, sasmodels and sasdata need to be built separately so
> # that sphinx can borrow from their documentation.
> bumps
> sasmodels
> sasdata
> periodictable
> tinycc
> # nxunit does not yet have a separate package.
> #nxunit
> ```

The conditioning doesn't seem to work: https://github.com/SasView/sasview/actions/runs/3628155088/jobs/6118847855#step:6:23 (Ignoring PyQt5: markers 'platform_system != "Darwin"' don't match your environment). I also tried with sys_platform (which is also mentioned on stackoverflow as a possible solution).

wpotrzebowski:
Sorry. I am taking it back. I think it works but probably we need another PyQt version (other than 5.13).

butlerpd:
@wpotrzebowski is still trying to figure it out and may need to have more discussion with @lucas-wilkins when he gets back

wpotrzebowski:
This thing is really confusing. Currently the "best" results I get with the latest version 5.15.7 but it is not perfect. When I move the window the background/canvas of main window is not fully refreshed 
Screenshot 2022-12-09 at 06 11 30
And it is even more confusing when using 5.13.0, which we used for 5.0.5 the layout is completely messed up:
 
Screenshot 2022-12-09 at 06 13 47

I guess there must be some other dependencies that also need to be fixed to particular versions.

butlerpd:
can @wpotrzebowski have a chat with @lucas-wilkins now that both are back to sort out next step?

wpotrzebowski:
@lucas-wilkins I suggest we merge this one. This is an improvement over what we have now and PyQt on Mac is mess anyway, which we will need to deal with separately. Are you ok with it?

wpotrzebowski (approved):

lucas-wilkins:
@wpotrzebowski I'm good with it.

PR #2396, opened 2022-11-30
[Release 5.0.6] Cherry-picked 2178 - Breaking bug in P(r)

krzywon:
## Description

This is the first of what is to be many PRs for the release of 5.0.6. This is a cherry-pick of the single commit in #2178 to fix broken P(r) inversion functionality and unit tests.


butlerpd (approved):
This fixes is needed unless we drop back to an older numpy version.  It seems to fix the failing CI on github and has nothing that prevents it from being merged.  I think we should merge this  post hast so we can start pulling in other cherry picks

PR #2401, opened 2022-12-09
2385 incosistent tabbing order across the widgets

### LINKED ISSUE #2385: Incosistent tabbing order across the widgets
Some widgets in SasView are consistent wrt, tabbing order (Data Explorer) but most exhibit rampant randomness in focusing on elements when the Tab key is pressed.
The standard is to go right-then-down/left for all controls and widgets.

This is part 1 of the larger issue of improving keyboard driven access for SasView as discussed in https://github.com/SasView/sasview/discussions/2342

rozyczko:
## Description

This PR addresses one of the issues brought up in https://github.com/orgs/SasView/discussions/2342 and spun off into https://github.com/SasView/sasview/issues/2385.

Now, tab order for all widgets in SasView is consistent and goes in standard direction (left->right, up->down)

Fixes #2385

- [x] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Calculators/KiessigPanel.py
+++ b/src/sas/qtgui/Calculators/KiessigPanel.py
@@ -27,13 +27,9 @@ class KiessigPanel(QtWidgets.QDialog, Ui_KiessigPanel):
 
         # signals
         self.helpButton.clicked.connect(self.onHelp)
-        # self.computeButton.setVisible(False)
         self.closeButton.clicked.connect(self.onClose)
         self.deltaq_in.textChanged.connect(self.onCompute)
         self.deltaq_in.setText("0.05")
-
-        # Set focus away from Close
-        # self.computeButton.setFocus()
 
         # no reason to have this widget resizable
         self.setFixedSize(self.minimumSizeHint())
--- a/src/sas/qtgui/Calculators/KiessigPanel.py
+++ b/src/sas/qtgui/Calculators/KiessigPanel.py
@@ -27,13 +27,13 @@ class KiessigPanel(QtWidgets.QDialog, Ui_KiessigPanel):
 
         # signals
         self.helpButton.clicked.connect(self.onHelp)
-        self.computeButton.setVisible(False)
+        # self.computeButton.setVisible(False)
         self.closeButton.clicked.connect(self.onClose)
         self.deltaq_in.textChanged.connect(self.onCompute)
         self.deltaq_in.setText("0.05")
 
         # Set focus away from Close
-        self.computeButton.setFocus()
+        # self.computeButton.setFocus()
 
         # no reason to have this widget resizable
         self.setFixedSize(self.minimumSizeHint())
--- a/src/sas/qtgui/Utilities/UI/PluginDefinitionUI.ui
+++ b/src/sas/qtgui/Utilities/UI/PluginDefinitionUI.ui
@@ -135,8 +135,8 @@ value</string>
          <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
-&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'MS Shell Dlg 2'; font-size:7.8pt; font-weight:400; font-style:normal;&quot;&gt;
-&lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;br /&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'MS Shell Dlg 2'; font-size:6.6pt; font-weight:400; font-style:normal;&quot;&gt;
+&lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px; font-size:7.8pt;&quot;&gt;&lt;br /&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
         </property>
        </widget>
       </item>
@@ -145,6 +145,14 @@ p, li { white-space: pre-wrap; }
    </item>
   </layout>
  </widget>
+ <tabstops>
+  <tabstop>txtName</tabstop>
+  <tabstop>chkOverwrite</tabstop>
+  <tabstop>txtDescription</tabstop>
+  <tabstop>tblParams</tabstop>
+  <tabstop>tblParamsPD</tabstop>
+  <tabstop>txtFunction</tabstop>
+ </tabstops>
  <resources/>
  <connections/>
 </ui>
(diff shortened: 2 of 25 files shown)


krzywon:
I've tested the functionality but haven't reviewed the code. Here are a few thoughts, some of which might be out of scope for this PR.

The good:

- Tabbing is now left->right, top->bottom as expected in most places
- Tabbing skips disabled inputs but not uneditable inputs (useful for copying results)
- For the corfunc perspective, and SLD, Q-resolution, and Generic scattering calculator tools, tabbing seems reasonable

Needs work:

- Tabbing travels through uneditable boxes when they are blank but gives no indication the box has focus
- Data operations panel: tabbing takes you into the data windows. Is there a reason to ever give them focus?
- Density/Volume calculator: Tabbing gets stuck at 'Molar Volume' and 'Mass Density' Inputs
- Kiessig Calculator: Close is selected twice along the tabbing path, once when expected, but another time after the window tabs are highlighted
- Invariant Perspective: Data Explorer is given focus when switching to this perspective. Tabbing seems to get stuck at 'Porod Constant' text box
- Inversion Perspective: The Options section is still unpredictable where tabbing will take you
- Fitting Perspective: Selecting a model often causes the focus to shift back to the Fitting window but should probably stay at the model selector.

Suggestions:

 - Tabbing is only applied to the window with focus. There is no way (in Windows) for the data explorer to gain focus so there is no way to tab into that panel. Can we add a Ctrl/Command-Tab option to tab between panels, menu, and options bar?
 - Add some indication a text box, especially when an uneditable text box is blank, receives tab focus.

rozyczko:
> * Tabbing travels through uneditable boxes when they are blank but gives no indication the box has focus

Can you give me an example of a widget that has those? I can't find empty, read only boxes with focus stop...

> * Data operations panel: tabbing takes you into the data windows. Is there a reason to ever give them focus?

There is no reason as these windows only show data. Fixed.

> * Density/Volume calculator: Tabbing gets stuck at 'Molar Volume' and 'Mass Density' Inputs

This one is interesting - the user NEEDS to enter a value in this field, which is why the textbox holds on to the focus.
Once entered, you can keep tabbing.

> * Kiessig Calculator: Close is selected twice along the tabbing path, once when expected, but another time after the window tabs are highlighted

Very, very weird. Working on it.

> * Invariant Perspective: Data Explorer is given focus when switching to this perspective. Tabbing seems to get stuck at 'Porod Constant' text box

Fixed the ordering. One has to wait a short while after tabbing after Max Distance for the calculations to finish (Focus loss causes recalculation? Is this correct?).

> * Inversion Perspective: The Options section is still unpredictable where tabbing will take you

This again is an issue of recalculation being done on focus lost. This potentially needs fixing.

> * Fitting Perspective: Selecting a model often causes the focus to shift back to the Fitting window but should probably stay at the model selector.

I tried to address it with mixed success. Will look closer.

> * Tabbing is only applied to the window with focus. There is no way (in Windows) for the data explorer to gain focus so there is no way to tab into that panel. Can we add a Ctrl/Command-Tab option to tab between panels, menu, and options bar?

Hmm this could be a nice extension. Will address it.

> * Add some indication a text box, especially when an uneditable text box is blank, receives tab focus.

I don't think this can be easily done. It's better to remove uneditable boxes from receiving focus.

krzywon:
> > * Tabbing travels through uneditable boxes when they are blank but gives no indication the box has focus
> 
> Can you give me an example of a widget that has those? I can't find empty, read only boxes with focus stop...

The invariant perspective without data in the 'Invariant' tab - tabbing through takes 6 presses to go from max Q to the buttons, suggesting the volume fraction and other inputs are selected.

> > * Invariant Perspective: Data Explorer is given focus when switching to this perspective. Tabbing seems to get stuck at 'Porod Constant' text box
> 
> Fixed the ordering. One has to wait a short while after tabbing after Max Distance for the calculations to finish (Focus loss causes recalculation? Is this correct?).

Seems better

> > * Density/Volume calculator: Tabbing gets stuck at 'Molar Volume' and 'Mass Density' Inputs
> 
> This one is interesting - the user NEEDS to enter a value in this field, which is why the textbox holds on to the focus. Once entered, you can keep tabbing.
> > * Add some indication a text box, especially when an uneditable text box is blank, receives tab focus.
> 
> I don't think this can be easily done. It's better to remove uneditable boxes from receiving focus.

On a side note, these two are handled in #2258. In that PR, anything with focus gets a new background color and a red border is added to any input with an error. I won't get back to that PR for a while but some of these suggestions will likely be addressed for v6.0, but that work has a ways to go still.

src/sas/qtgui/Calculators/KiessigPanel.py
```diff
@@ -27,13 +27,13 @@ def __init__(self, parent=None):
 
         # signals
         self.helpButton.clicked.connect(self.onHelp)
-        self.computeButton.setVisible(False)
+        # self.computeButton.setVisible(False)
```
lucas-wilkins, line 30: Uncomment or delete

src/sas/qtgui/Calculators/KiessigPanel.py
```diff
@@ -27,13 +27,13 @@ def __init__(self, parent=None):
...
         self.closeButton.clicked.connect(self.onClose)
         self.deltaq_in.textChanged.connect(self.onCompute)
         self.deltaq_in.setText("0.05")
 
         # Set focus away from Close
-        self.computeButton.setFocus()
```
lucas-wilkins, line 36: same

src/sas/qtgui/Utilities/UI/PluginDefinitionUI.ui
```diff
@@ -135,8 +135,8 @@ value</string>
          <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
-&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'MS Shell Dlg 2'; font-size:7.8pt; font-weight:400; font-style:normal;&quot;&gt;
-&lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;br /&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'MS Shell Dlg 2'; font-size:6.6pt; font-weight:400; font-style:normal;&quot;&gt;
+&lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px; font-size:7.8pt;&quot;&gt;&lt;br /&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
lucas-wilkins, line 139: Should we be specifying font sizes absolutely?

lucas-wilkins (approved):

src/sas/qtgui/Calculators/KiessigPanel.py
rozyczko: done.

src/sas/qtgui/Calculators/KiessigPanel.py
rozyczko: done

src/sas/qtgui/Utilities/UI/PluginDefinitionUI.ui
rozyczko: This seems to be a standard Qt Designer thing. Quite a few .ui files have explicit font sizes for tooltips and html content displayed in forms. These values are never manually modified in the Designer so I assume they are defaults.
I could remove them by hand but then loading/saving the forms would write the font size directives again...

rozyczko:
> The invariant perspective without data in the 'Invariant' tab - tabbing through takes 6 presses to go from max Q to the buttons, suggesting the volume fraction and other inputs are selected.

Modified those fields to be focusable with clicks but not with tab.

PR #2420, opened 2023-01-16
2419 `setup.py` cleaning

lucas-wilkins:
Restructuring setup.py so that it doesn't require a maintained list of all the packages that need to be included.


llimeht:
I'll wildly speculate that the docs are failing to build because there doc coverage is now covering much more of the code due to the extra `__init__.py` files there.

.github/workflows/ci.yml
```diff
@@ -205,7 +205,7 @@ jobs:
         if: ${{ matrix.docs }}
         run: |
           cd docs/sphinx-docs/
-          python build_sphinx.py || true
+          python build_sphinx.py
```
llimeht, line 208: Yes please to this... but I'm not sure that it is routinely passing as a matter of course at this stage.

lucas-wilkins: we'll see what it does, but no, its not passing. Which is making figuring out the problem with the other changes harder.

llimeht: I suspect that all the `check_call`uses need to be tweaked so that if it fails, stdout/stderr end up in the log as part of the complaint, rather than disappearing.

llimeht: (yes, I know it's not *currently* passing, but I'm not sure it actually does on a good day... it's currently failing at apidoc but I think it actually normally returns errors from sphinx itself due to sketchy crossrefs etc)

llimeht: If apidoc is now hitting Qt code and used not to, it might be that it has to be run inside xvfb-run so that there's a display so that pyqt stuff can be imported.

lucas-wilkins: I'm not sure that its hitting very much new in this branch, the omission stuff was all out of date. But maybe it has been for a while.

docs/sphinx-docs/build_sphinx.py
```diff
@@ -176,25 +177,36 @@ def apidoc():
...
+            "-o", SASMODELS_API_TARGET, # Output dir.
+            "-d", "8", # Max depth of TOC.
+            "-H", "sasmodels", # Package header
+            SASMODELS_BUILD,
+            # omit the following documents from the API documentation
+            joinpath(SASMODELS_BUILD, "sasmodels", "models"),
+        ])
+
+    except CalledProcessError as e:
+        print(e.output)
```
llimeht, line 208: Any error is likely to be on stderr, which is not captured by `check_output()`. Lots of sharp edges to cut yourself on in the `subprocess` module.

The `check_output` could just become a `run` call (since that is now recommended anyway), ensuring that both stdout and stderr are captured. For `run`, either `check=True` will give the same exception on nonzero exit status, or check the `CompletedProcess` object rather than use exceptions.

https://docs.python.org/3/library/subprocess.html#subprocess.run

llimeht: I'm not sure whether sphinx considers that a supported interface — I know they're busy removing setuptools integration, for instance.

lucas-wilkins:
Gotcha!

lucas-wilkins:
This should all work now. Might be good to check the mac/linux installer @wpotrzebowski @llimeht

installers/sasview.spec
```diff
@@ -9,6 +9,12 @@ import sysconfig
 block_cipher = None
 PYTHON_PACKAGES = sysconfig.get_path('platlib')
 
+# Debugging stuff
+
+for filename in os.listdir("../docs/sphinx-docs"):
+    print(os.path.join("../docs/sphinx-docs", filename))
```
wpotrzebowski, line 15: Do we still need this print statement?

lucas-wilkins: Nope, forgot to remove it.

llimeht:
Looking at the contents of the installer, the python files seem to be the same which is good. I don't currently have a clean container or vm handy to test what the installer has produced in the absence of other OS components that I've installed.

However, all the documentation is missing from the installer. Looking at the build logs for linux:
* the call to genmodel.py is not creating the rst files for the sasmodels documentation (the `generating model/core_shell_bicelle_elliptical.rst` etc output is missing); perhaps something incorrect in the changes to absolute paths?
* the apidoc call is now recursing through all the qtgui's unit testing files such as `sas.qtgui.Calculators.UnitTesting` (which were previously omitted and probably should still be excluded as the tests aren't part of the API)
* `sas.sascalc.realspace` and `sas.sascalc.simulation` are apparently not supposed to be in the released version (see sphinx's conf.py) but now are and also in the documentation build
* the sphinx call fails - importing various modules will need a display because Qt is used. I have a recollection that one can either give sphinx an extensive list of mocks to use for PyQt, or it can be run inside a virtual display (as is done for the tests); this may be failing in main too... as the failing doc-build is suppressed with `||true` since the docs have never built cleanly and thus all failures are ignored there :(

Building sasview's documentation is horrible and fragile (partly because of the split between sasmodels and sasview as has been noted). While it's always nicest to merge a polished and finalised PR, perhaps the simplest path forward for the doc issue is to merge this PR and open two 'critical' tickets about (a) the docs don't currently build, and (b) CI doesn't fail when the docs don't build properly.

docs/sphinx-docs/build_sphinx.py
```diff
@@ -183,10 +193,7 @@ def apidoc():
         "-H", "SasView", # Package header
         SASVIEW_BUILD,
         # omit the following documents from the API documentation
-        joinpath(SASVIEW_BUILD, "sas", "qtgui", "GUITests.py"),
```
rozyczko, line 186: removal of this exclusion causes generation of unneeded documentation for UnitTesting, for all components.

rozyczko (commented):
long needed update of the scripts, especially the `spec` file for pyinstaller.
Minor issue with extra docs.
Installer works fine on Windows, with docs and everything.
As @llimeht noticed, the linux doc build needs looking at, though.

docs/sphinx-docs/build_sphinx.py
lucas-wilkins: whoops.

llimeht: that line on its own might not be the issue as that particular file doesn't exist any more. The extra `__init__.py` files might be causing it to recurse more though. I suspect a longish list of UnitTesting directories need to be added there instead.

lucas-wilkins: OK, I'll look into getting this right.

lucas-wilkins: I've added the paths, but I wonder if it might be better to get them all in the same test directory, in /test, not /src? @rozyczko - maybe you have some thoughts on this.

rozyczko: Indeed. `GUITests.py` has been deleted. I didn't see the extra `__init__.py` but their presence might explain the inclusion of the test dirs.

rozyczko: Originally, I put tests alongside modules, since the conversion from `wx` to `qt` was gradual and such a layout made more sense. We would convert one module, add tests, run them and go to the next module.
It might be beneficial to refactor the tests out to the top level `tests/` directory, which already contains unit tests for sascalc.

rozyczko: This obviously being a separate, pretty large project, so feel free to just fix the includes here, obviously.

lucas-wilkins:
@rozyczko Was that supposed to be a review? I'm going to make a separate issue for the docs.

PR #2422, opened 2023-01-18
1999: Clip fitting values set outside the fit range

### LINKED ISSUE #1999: 5.x: Parameter limits in the Fit Page are not respected.
This defect was originally reported by User AdrianR using 5.0.3 on 28 Dec 2020 but I cannot find a corresponding issue (@krzywon might...). However, User GregS has now reported the same issue using 5.0.4 which suggests no fix was implemented. So am logging the issue here.

Adrian said:
>I have been using SasView 5.03 under Windows 10 and noticed an anomaly when fitting and applying constraints (simply minimum or maximum values of parameters) that these are not always applied after some iterations of fitting.  The attached screen shot shows 'thickness1' exceeding the maximum but my impression is that this can happen with various models/parameters.  I am not certain from the various web pages whether this corresponds to other comments for which tickets have been generated or whether this may arise from unusual values.
ScreenShot_Parameters

Greg says:
>I was again trying to fit SESANS data, but the fit value of some SLDs went beyond my pre-defined limits. For models with fewer parameters, it seemed fine, but for this highly parameterised model, it didn’t restrict the fit to these limits.
Screenshot 2021-12-09 at 12 16 55

krzywon:
## Description

As described in [bumps/bumps#58](https://github.com/bumps/bumps/issues/58) and #1999, when the initial fit values are set outside the fitting bounds, the MPFit (L-M) and Quasi-Newton fitters fail to converge and may throw errors. This utilizes the bumps `FitDriver.clip()` feature to coerce the initial values within the bounds prior to fitting. A warning is logged when any values are coerced.

Fixes #1999

This fix should also be added to v5.0.6 once it's reviewed.

## How Has This Been Tested?

Run SasView, select the L-M (MPFit) optimizer, load a data set, send it to fitting, and select a model. From there, I tried 2 different things
1. Set the fit range of one of the parameters to not include the initial value. This was done by changing both the upper a lower limits.
2. Set the initial value of one of the parameters so it is outside the initial bounds of the fit.

Select the modified fit parameter to be fit and run the fit.

Previously, the fit would fail with an assertion error. Now the fit succeeds and a warning message is given.

I did not test the Quasi-Newton optimizer.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -2008,6 +2008,9 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             logger.error(msg)
             return
 
+        if results.mesg:
+            logger.warning(results.mesg)
+
         param_list = results.param_list # ['radius', 'radius.width']
         param_values = results.pvec     # array([ 0.36221662,  0.0146783 ])
         param_stderr = results.stderr   # array([ 1.71293015,  1.71294233])
--- a/src/sas/sascalc/fit/BumpsFitting.py
+++ b/src/sas/sascalc/fit/BumpsFitting.py
@@ -373,6 +373,7 @@ def run_bumps(problem, handler, curr_thread):
             return True
         return False
 
+    errors = []
     fitclass, options = get_fitter()
     steps = options.get('steps', 0)
     if steps == 0:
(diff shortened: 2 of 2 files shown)


smk78 (approved):
Functionality test on W10/x64 . Local build. Tested with the new MPfit L-M, N-M and Q-N optimisers.
This fixes the issue I reported in reopened #1999 . The error message no longer appears and Min, Max or Min & Max bounds are respected.
If the Min bound is outside the expected fitting range the message in BumpsFitting line 394 is displayed.
Approved.

wpotrzebowski (approved):

wpotrzebowski:
I reviewed the code and merged it. It can be now cherry picked for 5.0.6.

PR #2427, opened 2023-02-01
Release 5.0.6 releasenotes

butlerpd:
## Description

Adding release notes for SasView 5.0.6 in preparation for the first beta release of 5.0.6

## How Has This Been Tested?

N/A the text has been checke

### DIFF
--- a/docs/sphinx-docs/source/user/RELEASE.rst
+++ b/docs/sphinx-docs/source/user/RELEASE.rst
@@ -57,14 +57,14 @@ Pull Request Changes
 * Fix to allow multiple scattering script to run @wpotrzebowski `[#521] <https://github.com/SasView/sasmodels/pull/521/>`_
 * Fix error in core shell Ellipsoid documentation @pkienzle `[#512] <https://github.com/SasView/sasmodels/pull/512/>`_
 * Fix models with complex amplitudes not compiling on the fly @pkienzle `[#511] <https://github.com/SasView/sasmodels/pull/511/>`_
+* Correct paracrystal model error and rework documentation @butlerpd `[#545] <https://github.com/SasView/sasmodels/pull/545>`_
 
 Documentation Changes
 ^^^^^^^^^^^^^^^^^^^^^
 * Update optimizer help documentation @smk78 `[#2359] <https://github.com/SasView/sasview/pull/2359/>`_
 * Update contributor list @wpotrzebowski `[#2114] <https://github.com/SasView/sasview/pull/2114>`_
 * Update web links from http to https @smk78 `[#2087] <https://github.com/SasView/sasview/pull/2087/>`_ and `[#2265] <https://github.com/SasView/sasview/pull/2265/>`_
-* Update corfunc documentation @lucas-Wilkins `[#2047] <https://github.com/SasView/sasview/pull/2047/>`_
-* Correct paracrystal model error and rework documentation @butlerpd `[#545] <https://github.com/SasView/sasmodels/pull/545>`_
+* Update corfunc documentation @lucas-wilkins `[#2047] <https://github.com/SasView/sasview/pull/2047/>`_
 * Update gel_fit model documentation and fix formating @smk78 `[#541] <https://github.com/SasView/sasmodels/pull/541/>`_
 * Correct and update cylinder model documentation @butlerpd `[#539] <https://github.com/SasView/sasmodels/pull/539>`_
 * Restructuring and cross-linking of sasmodels docs @smk78 `[#534] <https://github.com/SasView/sasmodels/pull/534/>`_


docs/sphinx-docs/source/user/RELEASE.rst
```diff
@@ -13,6 +13,71 @@ Release Notes
...
+* Update optimizer help documentation @smk78 `[#2359] <https://github.com/SasView/sasview/pull/2359/>`_
+* Fix plot legend not updating on custom change @rozyczko `[#2362] <https://github.com/SasView/sasview/pull/2362/>`_
+* Cleanup (close) the bumps/DREAM results panel when associate data is removed from sasview @krzywon `[#2365] <https://github.com/SasView/sasview/pull/2365/>`_
+* Fix for failure to find custom_config preventing sasview from starting @krzywon `[#2407] <https://github.com/SasView/sasview/pull/2407/>`_
+* Fix error being thrown when the initial guess it outside the min/max range for the LM optimizer @krzywon `[#2422] <https://github.com/SasView/sasview/pull/2422/>`_
+* Fix problem with LM optimizer failing when GPUs are turned on by updating to the latest bumps version
+* Fix models with complex amplitudes not compiling on the fly @pkienzle `[#511] <https://github.com/SasView/sasmodels/pull/511/>`_
+* Fix error in core shell Ellipsoid documentation @pkienzle `[#512] <https://github.com/SasView/sasmodels/pull/512/>`_
+* Fix to allow multiple scattering script to run @wpetrzebowski `[#521] <https://github.com/SasView/sasmodels/pull/521/>`_
```
krzywon, line 63: `@wpotrzebowski`

docs/sphinx-docs/source/user/RELEASE.rst
```diff
@@ -13,6 +13,71 @@ Release Notes
...
+is outside the min/max range has been fixed, a problem with intermittent blanking
+of plots has been fixed, a number of defaults have been changed to be more reasonable
+and a number of issues in the documentation have been corrected and/or updated.
+
+This version of SasView is built with Sasmodels 1.0.7 and Bumps master.
+
+Pull Request Changes
+^^^^^^^^^^^^^^^^^^^^
+
+* Update corfunc documentation @lucas-Wilkins `[#2047] <https://github.com/SasView/sasview/pull/2047/>`_
```
krzywon, line 34: Random capitalization nitpick: `@lucas-wilkins`

docs/sphinx-docs/source/user/RELEASE.rst
```diff
@@ -13,6 +13,71 @@ Release Notes
 Features
 ========
 
+New in Version 5.0.6
+--------------------
+This is a small bug release which fixes some issues reported in earlier versions
```
krzywon, line 18: Historically, we have bulleted the highlighted changes so this should be something like:

```
 * The failure to start when installing on a new system due to issues finding the config file has been fixed
 * The startup speed has been improved.
 * The paracryatline models, which have been labelled as "under review" since 2018, have been corrected (bcc and fcc) and the documentation completely reworked (bcc, fcc, and sc).
 * Plots now properly support custom names in the legend, 
 * The LM optimizer failure on GPUs or when the starting value is outside the min/max range has been fixed
 * A problem with intermittent blanking of plots has been fixed
 * A number of default values have been changed to be more reasonable
 * A number of issues in the documentation have been corrected and/or updated.
```

krzywon (commented):
Historically, we have separated bug fixes from new features from documentation changes. That could wait for the final release but would be better to change that here so it's not forgotten later.

docs/sphinx-docs/source/user/RELEASE.rst
butlerpd: Oops .. good catch

docs/sphinx-docs/source/user/RELEASE.rst
butlerpd: Dang .. another one 😃

lucas-wilkins:
Not 100% sure about the fidelity of the attributions, but also I don't really care.

docs/sphinx-docs/source/user/RELEASE.rst
butlerpd: Yes, I am aware of that as well as the fact that last time we decided to break things out into documentation, features, bugs, and infrastructure. I considered this for a while as I agree it is very important to be consistent. However I chose to deviate slightly (while still maintaining the general style) as this is the very first time we are trying to follow strict semantic versioning and thus limiting ourselves to a pure bug fix -- so in principle we have argued that EVERYTHING is a bug ... though there is a lot of grey areas and one could even quibble that maybe some are beyond grey?  Still I decided that sticking with the spirit of the release all would be bugs so no need for a specific section.

The bulleted highlights last time was partly due to the shear number of them. This release is much more restricted and focused and since all the updates are fairly cleanly stated in a small bullet set already I thought it might be better to use a narrative approach. Otherwise we are just making two copies of a third or so of the full bulleted list?  I also considered **not** including **any** highlights on the grounds again that the list is relatively short already. In the end I chose to include some highlights because I felt there were definitely some significant difference in relevance to many of ours users and felt it might be useful to highlight that for them. Looking at past releases, this is not totally inconsistent as we have mixed narrative and bullets in various sections based on what seemed appropriate for the content of that particular release.
I am happy to do what the consensus is on this but since these were deliberate choices I thought I should lay out my reasons for discussion before making changes. I would propose that we merge once I fix the typos and release to get the beta out sooner than later for testing and sort out what we think the correct answer to these should be.

butlerpd:
LOL ... I see that @smk78 went ahead and separated things out while I was explaining the rationale behind the original choice.  No worries. I'm happy to go with the final consensus so agree it should be ready to merge into main then cherry pick back to release_5.0.6. 

Should I do this? if so when?

krzywon:
> Should I do this? if so when?

Once all CI runs succeed for this PR, I would suggest merging into `main`. You could start cherry-picking locally into a branch built off the `5.0.6 release` branch and push once this succeeds.

PR #2428, opened 2023-02-01
add beta version

butlerpd:
Beta version should be 5.0.6b1 not 5.0.6


wpotrzebowski:
Tested installer and checked docs.

PR #2434, opened 2023-02-06
[v5.0.6] Update installer.iss with correct version and copyright dates

krzywon:
## Description
This updates the version and copyright date in installer.iss for v5.0.6.

- [ ] Code has been reviewed

### DIFF
--- a/installers/installer.iss
+++ b/installers/installer.iss
@@ -2,7 +2,7 @@
 ; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
 
 #define MyAppName "SasView"
-#define MyAppVersion "5.0.6b1"
+#define MyAppVersion "5.0.6"
 #define MyAppPublisher "(c) 2009 - 2023, UTK, UMD, NIST, ORNL, ISIS, ESS, ILL, ANSTO, TU Delft and DLS"
 #define MyAppURL "http://www.sasview.org"
 #define MyAppExeName "sasview.exe"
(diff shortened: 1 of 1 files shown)


butlerpd:
I think it is too late for b1. The question however might be whether to change this to 5.0.6 (for all betas even) on the grounds that the final 5.0.6 will just be in the same directory?  When launching the version number is now prominently displayed in the top left corner of the SasView application window.  This would make releases a bit easier ... and less worry that the final release will default to 5.0.6bx instead of 5.0.6?

installers/installer.iss
```diff
@@ -2,8 +2,8 @@
 ; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
 
 #define MyAppName "SasView"
-#define MyAppVersion "5.0.5"
-#define MyAppPublisher "(c) 2009 - 2021, UTK, UMD, NIST, ORNL, ISIS, ESS, ILL, ANSTO, TU Delft and DLS"
+#define MyAppVersion "5.0.6b1"
```
butlerpd, line 5: For the installer we should just use 5.0.6 (for the betas and the final release).  Betas should be overwritten by the final release in any case.  Once that is changed this should be ready to merge.

butlerpd (changes requested):

butlerpd:
also .. why is it failing ubuntu unit tests?

krzywon:
> also .. why is it failing ubuntu unit tests?

The ubuntu failures were due to intermittent connection issues with azure.

PR #2435, opened 2023-02-08
Update entitlements.plist

wpotrzebowski:
## Description
Trying to resolve config file issue.

Fixes # (issue/issues)

## How Has This Been Tested?

- [ ] Code has been reviewed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -192,14 +192,6 @@ jobs:
         codesign --verify --options=runtime --entitlements ../../build_tools/entitlements.plist --timestamp --deep --verbose=4 --force --sign "Developer ID Application: European Spallation Source Eric (W2AG9MPZ43)" SasView5.app
         hdiutil create SasView5.dmg -srcfolder SasView5.app -ov -format UDZO
         codesign -s "Developer ID Application: European Spallation Source Eric (W2AG9MPZ43)" SasView5.dmg
-
-    - name: Publish installer package (OSX)
-      if: ${{ matrix.os == 'macos-latest' }}
-      uses: actions/upload-artifact@v2
-      with:
-        name: SasView-Installer-${{ matrix.os }}-${{ matrix.python-version }}-${{ github.sha }}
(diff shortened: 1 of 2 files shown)


krzywon:
We aren't sandboxing, but we are seeing a similar issue as in https://developer.apple.com/forums/thread/666634. If the key you added doesn't work, we might want to try this one next.
`<key>com.apple.security.temporary-exception.files.home-relative-path.read-write</key>`

butlerpd:
Would sandboxing bypass the dire warnings that notarization does not?  Also, is it possible that some macs are enforcing sandboxing while others are not? That said why not add both keys in this PR?

butlerpd:
@lucas-wilkins and @krzywon to check on whether to use temp or "permanent" then include and merge and cherry pick back to main.

lucas-wilkins:
I think it might need this one, if not now, but in the future:

com.apple.security.files.user-selected.read-write

Still looking for something home directory related

lucas-wilkins:
I think what's happening is that the entire concept of a home directory is being removed. Which is a pain for us!

We're also not going to be able to support sharing the same files across different apps, so, having multiple versions of sasview that have the same scripts will just be impossible on mac.

This is actually going to require some serious rethinking of things.

lucas-wilkins:
> <key>com.apple.security.temporary-exception.files.home-relative-path.read-write</key>

This is a funny key, I think it needs to be explicitly granted by apple, but only really makes a difference when things are distributed through the app store.

So, I don't think it will hurt to include it, but I'm also not sure it will make a difference. I'm not sure how much entitlements make a difference outside the app store either.

wpotrzebowski:
The program doesn't start with com.apple.security.temporary-exception.files.home-relative-path.read-write therefore I think we shouldn't include it. I've also changed the order of the last step in release.yml, so that artifact is published after "stapling" I don't think it makes any difference as it seems to behave exactly the same and my impression is that apple determines if app has been notarized by checking on-line. It may make difference when being off-line

PR #2443, opened 2023-02-25
Add +/- 1 and 3 sigma lines on residual plots

### LINKED ISSUE #1329: Meaningless error values on Theory and Normalised Residuals
There is a long standing issue, from at least as far back as Sasview 3

The error values associated with both Theory and Normalised Residual workspaces are all set to 1.0  (This is the dY column in "show info" on a workspace array.)

It happens that the statistical error on the Normalised residual for the default "Use dI" weighting scheme should actually be 1.0 (see below), but I suspect that all these errors are simply being set to 1.0 by default?

Discussion needed please by SAS experts as to what we should do here.

Regarding the Theory workspace - we could attempt to propagate error estimates from the errors on the fit parameters. We could simply not store and save any error value at all.

Regarding the Normalised Residuals - We minimise chi2=sum{ ( (I-C)/W )^2 }, the "normalised residual" is R = (I-C)/W where I is data, C is calculated, and W the weight for each data point.

Default is weight W = dI the error on data I, thus R = (I+-dI)/dI - C/dI, so error on normalised residual, dR = 1

If W=1 (the "none" option), dR = dI

If W=sqrt(I) then dR = dI /sqrt(I),  with the usual headache if I=0

If W =I then dR = dI/I,  with the usual headache if I=0

butlerpd:
## Description
Residual plots should now have 2 red lines at +/- 3 sigma and 2 dashed gray lines at +/-1 sigma.  There does not seem to be any documentation that needs updating.

The only change necessary for this was to use the `data1D_plot_role` already established. This allows the fix to require only a small 5 line block of code in the `PlotterWidget` in `plotter.py`.  Basically if `data.plot_role==data1D.ROLE_RESIDUAL` then add 4 horizontal lines (`ax.axhline()`). 

Fixes #1329

## How Has This Been Tested?
by running in developer environment

Load a data set (AOT microemulsion for example), send to fitting, choose a model (e.g. sphere/core_shell_sphere), hit compute/plot and/or fit (after choosing some parameters to fit).  The residual plots should have the extra 4 lines.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -227,6 +227,12 @@ class PlotterWidget(PlotterBase):
         if data.show_yzero:
             ax.axhline(color='black', linewidth=1)
 
+        # Display +/- 3 sigma and +/- 1 sigma lines for residual plots
+        if data.plot_role == data.ROLE_RESIDUAL:
+            ax.axhline(y=3, color='red', linestyle='-')
+            ax.axhline(y=-3, color='red', linestyle='-')
+            ax.axhline(y=1, color='gray', linestyle='--')
+            ax.axhline(y=-1, color='gray', linestyle='--')
         # Update the list of data sets (plots) in chart
         self.plot_dict[data.name] = data
 


krzywon:
This does what it says it does on the surface, but there is a wrinkle. The P(r) plot is tagged as a residual plot so the lines are drawn on those plots as well. Fixing another plot type goes beyond this PR so I would say we document the issue and not let it hold up this change.

wpotrzebowski (approved):

wpotrzebowski:
We agreed to merge it but we need to document issue with PR plot as pointed out by @krzywon

PR #2448, opened 2023-02-28
2447 plot roles

### LINKED ISSUE #2447: Remove +/-1 and 3 sigma lines from P(r) plots
In merging #2443 it was noted that the functionality review noticed the new sigma bounds _also_ appeared on P(r) plots. They should not. This needs to be addressed.

@lucas-wilkins says that the Corfunc plots should be unaffected.

krzywon:
## Description

Two new plotting roles are introduced here - stand-alone linear and stand-alone log. These are in addition to the residual role which has unique characteristics. These roles allow a plot to be displayed on its own without the parent data set. These new roles have been applied to the P(r) and polydispersity plots so the residual lines added in #2443 are not shown.

Fixes #2447

## How Has This Been Tested?

Run sasview. Run a fit with polydispersity. Residual plots have the +/- 1 and 3 lines, dispersity plots do not. Run a P(r) inversion. P(r) plots do not have the +/- lines.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -1044,8 +1044,11 @@ class DataExplorerWindow(DroppableDataLoadWidget):
                     or name in plot.name
                     or name == plot.filename):
                 # Residuals get their own plot
-                if plot.plot_role == DataRole.ROLE_RESIDUAL:
+                if plot.plot_role in [DataRole.ROLE_RESIDUAL, DataRole.ROLE_STAND_ALONE_LINEAR]:
                     plot.yscale = 'linear'
+                    self.plotData([(item, plot)])
+                elif plot.plot_role in [DataRole.ROLE_STAND_ALONE_LOG]:
+                    plot.yscale = 'log'
                     self.plotData([(item, plot)])
                 else:
                     new_plots.append((item, plot))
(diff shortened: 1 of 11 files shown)


butlerpd:
I am not convinced that the two new roles: `ROLE_STAND_ALONE_LINEAR` and `ROLE_STAND_ALONE_LOG` are the right answer. I rather think the  original commit of just a single `ROLE_STAND_ALONE` role would be appropriate.

Firstly, this is really log log and linear log (not linear linear as the name might imply). There can be log linear and linear linear. In fact, the polydispersity plots are linear linear for example. Furthermore, looking to the future, one probably wants to move the "change scale" from actually changing the data to x and y transforms. Finally, data1D objects already have a property that seems to do this `Data1D.ytransform` and `Data1D.xtransform` 2D data seems to have a similar thing but applies to the intensity I think so `Data2D.scale`? These are currently being used to plot the data on the appropriate scale (log or linear) so having a second way to do the same thing seems a bit ... dangerous? (spoken as someone who got bitten by competing ways to do the same thing before so that changing the default in the code did not actually change the default 😄 )

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -1090,13 +1091,14 @@ def displayData(self, data_list, id=None):
...
-            if (role == Data1D.ROLE_RESIDUAL and shown) or role == Data1D.ROLE_DELETABLE:
+            if (role in stand_alone_types and shown) or role == DataRole.ROLE_DELETABLE:
                 # Nothing to do if separate plot already shown or to be deleted
                 continue
-            elif role == Data1D.ROLE_RESIDUAL:
-                # Residual plots should always be separate
-                plot_to_show.yscale='linear'
+            elif role in stand_alone_types:
+                # Stand-alone linear and residual plots should always be separate
+                plot_to_show.yscale = 'linear'
```
butlerpd, line 1101: I believer line 1101 is either overwriting or being overwritten (that is part of the problem, even If I know there are two ways to achieve something I never know which one wins) by `FittingUtilities` function `residualsData1D` where line 550 has
```python
 residuals.ytransform = 'y'
```
actually in building that Data1D object, the axes labels are defined and the x axis is made linear for sesans data using the same transform on lines 551 and beyond
```python
    if reference_data.isSesans:
        residuals.xtransform = 'x'
        residuals.xaxis('\\rm{z} ', 'A')
    # For latter scale changes
    else:
        residuals.xaxis('\\rm{Q} ', 'A^{-1}')
    residuals.yaxis('\\rm{Residuals} ', 'normalized')
```
So it seems that the plottables Data1D object (I guess that was the idea of a plottable?) has all the hooks for defining how that data should be plotted while the plotting itself has functions to do the same thing? I'm not sure what the right answer should be but maybe that is monstrous can of worm?

If so then the easiest is to stick to using the Data1D hooks for everything in which case we just delete this line?

Also, shouldn't the comment be something like "identify plots that should be separate"?

krzywon: I tested this, and residual, polydispersity, and P(r) plots all continue to plot with a linear y scale after commenting out line 1101, so I removed it.

butlerpd (approved):
Code looks good now. I will defer to others on whether the long term choice should be to leave the scale choice to the plottable object .. though the more I think about it the more that seems right.
Functionality is tested on windows and the fix does what it says. I recommend we merge post haste.

rozyczko:
This looks good. the standalone role could be used as a starting point for refactoring of the "quick plots" which make the plotting subsystem more complex than it should be.

rozyczko (approved):
looks good

butlerpd:
do we need somebody to test on a mac or can we say that for this what has been done should be good enough and just merge?

wpotrzebowski:
Tested on Mac. Plots seem to work as expected. The only thing that I spotted is the lenged box on the plots is unnecessarily big. Not sure if this was introduced by this PR 

Screenshot 2023-03-14 at 08 18 44
Screenshot 2023-03-14 at 08 19 45
Screenshot 2023-03-14 at 08 26 29



We also need to remember to document new features

PR #2462, opened 2023-03-14
Fix wrong number of parameters on restore in slicer module

butlerpd:
## Description

Added the missing event parameter to all calls to restore. For some reason restore seems to be the ONLY function that where this was omitted. Also the RadiusInteractor.py was done correctly for some reason.  This should now fix all instances.

Fixes # (issue/issues)
2461

## How Has This Been Tested?

both the annular and box slicers were tested and the previous error was not generated

- [x] Code has been reviewed


butlerpd:
The question raised by @krzywon is whether this is the right answer of whether in fact the event parameter should be removed from the base class if it is never used. After reviewing and then discussion with @krzywon I believe, and @krzywon agreed that leaving the event is the better option. My thinking was that the original author's reason that all functions provided by the base class include the event is probably because, as an "interactor," it is triggered by an event and therefore the event should be provided to the functions in case someone wants to use it. While no instances of the various child interactors currently do, as we continue to refactor, add functionality etc, it may well happen that somebody wants to use the event. @krzywon points out that it also may be useful in debugging the event.

PR #2468, opened 2023-03-16
Get sasmodels from test branch

lucas-wilkins:
Investigating source of hanging CI


lucas-wilkins:
Think this is all resolved now, time will tell.

PR #2480, opened 2023-03-31
Add requests to requriements.txt

krzywon:
## Description

This adds the `requests` module to requirements.txt. This will be needed for sasview/sasdata#31 in order to load remote data by URL. The work I'm doing in https://github.com/SasView/sasdata/tree/31-load-data-from-url isn't finished yet, but thought I should get this in before I break the build here.

- [ ] Code has been reviewed

### DIFF
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -39,3 +39,4 @@ pywin32; platform_system == "Windows"
 superqt
 pyopengl
 pyopengl_accelerate
+requests


wpotrzebowski (commented):
I've noticed it is distributed under Apache-2.0 license, which we've just had a discussion on slack about. I think in this case we may need to be little be more careful and check if the code of the package is not distributed in the bundles. I guess we can take a look on how is packaged?

krzywon:
Good catch. I hadn't noticed the license. The bundling is what could cause us issues. If we released exclusively as a pypi package and listed it as a dependency, we should be fine, but our installers are a different story. I'll look into alternative modules.

butlerpd:
My understanding is that Apache2 is also a permissive licence like MIT and python licenses and unlike GPL. As such it should be ok to include?  Looking at the following writeup from FOSSA.com, there is a tiny caveat but I'm not sure it is relevant here nor is it clear that said caveat is "clear" to the authors of the piece.  I would thus argue it should be OK?

krzywon:
To eliminate any potential licensing issues, I've switched to the built-in `urllib.request` module instead of `requests`.

PR #2478, opened 2023-03-31
Pyside6 merge

rozyczko:
## Description

PyQt5 got replaced by Pyside6. This is needed for maintainability.

This is a big change from the point of view of codebase and requirement changes.
Please test and review carefully.

When testing the dev build please remember to **compile GUI elements** (.ui -> .py) by running e.g. `python setup.py build`

Fixes #2257

## How Has This Been Tested?

Local development version
Windows installer.

## Review Checklist (please remove items if they don't apply):

- [x] Code has been reviewed
- [x] Functionality has been tested
- [X] Windows installer (GH artifact) has been tested (installed and worked) 
- [ ] MacOSX installer (GH artifact) has been tested (installed and worked) 
- [ ] User documentation is available and complete (if required)
- [ ] Developers documentation is available and complete (if required)
- [ ] The introduced changes comply with SasView license (BSD 3-Clause)

PR #2483, opened 2023-04-11
Add argument to convertUI that forces full UI rebuild

krzywon:
## Description

This adds new arguments to convertUI.py and setup.py to force all UI elements to be rebuilt, regardless if they exist or not. This is partly in response to the cleanup needed to test #2478 but could also have utility for future dependency changes.

## How Has This Been Tested?

Ran `python convertUI.py -f`, `python setup.py rebuild_ui`, and `python setup.py clean`, and all 3 rebuilt the UI elements regardless if they existed or not.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/convertUI.py
+++ b/src/sas/qtgui/convertUI.py
@@ -5,7 +5,6 @@
 import os
 import sys
 
-
 def run(main, name, *args):
     saved_argv = sys.argv
     sys.argv = [name, *args]
@@ -20,7 +19,6 @@ def run(main, name, *args):
         pass
     return 0
 
-
 def pyrrc(in_file, out_file):
     """
     Run the qt resource compiler
@@ -28,14 +26,12 @@ def pyrrc(in_file, out_file):
     from PyQt5.pyrcc_main import main
     run(main, "pyrcc", in_file, "-o", out_file)
 
-
 def pyuic(in_file, out_file):
     """
     Run the qt UI compiler
     """
     from PyQt5.uic.pyuic import main
     run(main, "pyuic", "-o", out_file, in_file)
-
 
 def file_in_newer(file_in, file_out):
     """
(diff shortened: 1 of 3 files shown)


lucas-wilkins:
I think it should be part of "clean".

rozyczko:
> I think it should be part of "clean".

Should it?
I thought `clean` should delete all setup generated files. The proposed flag is to enforce the `.ui` -> `.py` rebuild, thus generating dependent python files.

src/sas/qtgui/convertUI.py
```diff
@@ -48,13 +51,17 @@ def file_in_newer(file_in, file_out):
...
+force_recreate = '-f' in args
+
 # look for .ui files
 for root, dirs, files in os.walk("."):
     for file in files:
         if file.endswith(".ui"):
             file_in = os.path.join(root, file)
             file_out = os.path.splitext(file_in)[0]+'.py'
-            if file_in_newer(file_in, file_out):
+            if force_recreate or file_in_newer(file_in, file_out):
```
rozyczko, line 64: this conditional is evaluated three times. It may be easier to pre-evaluate it.
```
redo = force_recreate or file_in_newer(file_in, file_out)
...
if redo:
...
```

rozyczko (approved):
The code looks good, minor stylistic points raised.

lucas-wilkins:
> Should it?
> I thought `clean` should delete all setup generated files. The proposed flag is to enforce the `.ui` -> `.py` rebuild, thus generating dependent python files.

I could have been clearer.

Removing existing automatic python files should be "clean", and that should be enough to trigger a rebuild when you run sasview.

However, it might also be good to have the actual making of the py files in build/rebuild

src/sas/qtgui/convertUI.py
krzywon: `file_in_newer` passes different files for `file_in` and `file_out` every time it is run, so this would create an invalid check. I could probably incorporate the later checks into this loop. Let me look into that.

krzywon:
I tried to clean up `convertUI.py` but the Ubuntu build failed so I reverted the changes and should probably not put any more time into it. I also incorporated the change suggested by @lucas-wilkins and the arguments `clean` and `rebuild_ui` now force the GUI to rebuild.

lucas-wilkins:
I've noticed that in main, automatic UI rebuilds don't seem to be happening.

lucas-wilkins:
I fixed that, hope the changes don't get in the way

PR #2485, opened 2023-04-14
2111 name changes in corfunc

lucas-wilkins:
* Outstanding renaming of entries in corfunc
* Tab ordering

The name "Hard block long period" is too long, so not doing.


wpotrzebowski (approved):
Collectively approved at the call

PR #2524, opened 2023-06-05
[5.0.6] PyPi requirements in setup.py

krzywon:
## Description

This updates setup.py to create a Windows package available for PyPi. This has not been tested on Mac or Ubuntu (yet).

Files: https://pypi.org/project/sasview/5.0.6b3/#files

Work left to be done:
- Test on Mac and various Linux distros
- Create specific Mac and Linux distros (as needed)
- Port to `main`: Will require (at minimum) PySide instead of PyQt
- Integrate into the build process (`.github/workflows`) for both the releases (PyPi bot account, GitHub secret for ssh, and add build and twine upload to yml files) and nightly builds (Append latest git hash to end of version number to ensure unique name)

I should probably document these issues more thoroughly in a separate issue.

## How Has This Been Tested?

On Windows:
```
> pip install sasview==5.0.6b3
> sasview
```
The SasView GUI launches!

- [ ] Code has been reviewed

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -281,7 +281,7 @@ packages.append("sas.qtgui")
 required = [
     'bumps>=0.7.5.9', 'periodictable>=1.5.0', 'pyparsing>=2.0.0',
     'lxml', 'h5py', 'sasmodels==1.0.7', 'qt5reactor', 'twisted', 'typing-extensions>=4.5.0',
-    'pyopenssl', 'cryptography<39.0'
+    'pyopenssl', 'cryptography<39.0', 'PyQt5', 'service-identity'
 ]
 
 if os.name == 'nt':
--- a/setup.py
+++ b/setup.py
@@ -280,7 +280,8 @@ packages.append("sas.qtgui")
 
 required = [
     'bumps>=0.7.5.9', 'periodictable>=1.5.0', 'pyparsing>=2.0.0',
-    'lxml', 'h5py',
+    'lxml', 'h5py', 'sasmodels==1.0.7', 'qt5reactor', 'twisted', 'typing-extensions>=4.5.0',
+    'pyopenssl', 'cryptography<39.0'
 ]
 
 if os.name == 'nt':


lucas-wilkins:
Agreed, I'm told there's plans to move away from CentOS7 to Rocky8 at ISIS in the near future. I'm waiting to hear more about what will happen and when.

andyfaff:
Do you use any of the pypa images to make the wheels? (Cibuildwheel is great for making a full matrix of wheels)

krzywon:
@andyfaff, this is in the infancy of creating PyPi distributions and currently am only doing so locally. The goal was something easy that works for v5.0.6 (which is set to release at any moment) and then build an infrastructure for future releases. `cibuildwheel` looks like a good option and will keep it in mind for the CI implementation.

lucas-wilkins:
Manually installing PyQt5 gives

image

wpotrzebowski:
Same on Mac - PyQt5 missing

llimeht:
@lucas-wilkins, I've only seen that error when Qt can't find a way to display, remembering that Qt has lots of different display methods available to it. On linux, that normally means it can't find some necessary X libraries. The [following line installs them](https://github.com/SasView/sasview/blob/main/.github/workflows/ci.yml#L75) for CI:

```sudo apt install libxkbcommon-x11-0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xinerama0 libxcb-xfixes0 x11-utils xvfb```

These are the Debian packages but might give the hints you need; some of these you clearly already have given you're working in a graphical session in the VM, but others you might not have. (xvfb is only for testing and so not needed for what you are doing)

There's also a hint there that you might be running under Wayland rather than X11; you might need to have some additional X11/wayland compat layer installed… I've not look at anything to do with Wayland yet.

BTW versions of pip newer than what you have there will not be happy with you running pip as root or installing into /usr/local; pip is getting very opinionated about using venvs with the desire to stop pip installs breaking distro packages. Given that is coming soon when your OS is upgraded or when you upgrade pip itself (pip v23 and python 3.11; PEP668), it might be worth starting with that workflow to save changing later.

rozyczko (commented):
Is this still a valid PR?
`qt5reactor` and `PyQt5` have been superseded by `PySide6`.

PR #2535, opened 2023-06-27
Rog and beta q

smalex-z:
## Description

Adjusted the Generic Scattering Calculator to automatically calculate and display the Radius of Gyration (only for non-element nuclear data). Furthermore, added Beta Q calculations into the Generic Scattering Calculation. A new box has been added to the QComboBox, giving 3 options now- "Fixed Orientation" (2D), "Debye Full Avg." (1D) and "Full Avg. w/ Beta(Q)" (1D). Selecting Full Average w/ Beta Q will perform the same actions as Debye Full Average, but will also plot a second graph of "Q vs Beta(Q)" next to the "Q vs form factor" graph in the main window.

periodictable import has been added as well.

## How Has This Been Tested?

- Import Nuclear Data (ex. PDB File), preferably smaller or else the form factor calculation will take a long time
- Radius of Gyration should be displayed correctly, rounded to .1 Angstrom (checked against other calculations)
- Select "Full Avg. w/ Beta(Q)" from the Dropdown list and hit compute
- After computation is complete, two graphs should show up in the main window- Q vs Form Factor and Q vs Beta Q. (Beta Q Calculation has been checked against other calculations)

NOTE: Have not tested against Magnetic Data or Element Data yet due to a lack of Magnetic or Element Data files. May need to test to ensure the program is not crashed by it. May have to disable Beta Q Calculations for Magnetic Data if that is the case.

- [ ] Code has been reviewed


src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -562,7 +567,9 @@ def change_is_avg(self):
         # update the averaging option fromthe button on the GUI
         # required as the button may have been previously hidden with
         # any value, and preserves this - we must update the variable to match the GUI
-        self.is_avg = (self.cbOptionsCalc.currentIndex() == 1)
+        self.is_avg = (self.cbOptionsCalc.currentIndex() in (1,2))
```
rozyczko, line 570: Since you only want to reject item 0, maybe this would look nicer?
`self.is_avg = self.cbOptionsCalc.currentIndex() != 0`

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -857,10 +864,70 @@ def update_gui(self):
...
+    def radius_of_gyration(self):
+        #Calculate Center of Mass(CoM) First
+        CoM = self.centerOfMass()
+
+        #Now Calculate RoG
+        RoGNumerator = RoGDenominator = 0.0
+
+        for i in range(len(self.nuc_sld_data.pos_x)):
+            coordinates = [0.0,0.0,0.0]
+            coordinates = [float(self.nuc_sld_data.pos_x[i]),float(self.nuc_sld_data.pos_y[i]),float(self.nuc_sld_data.pos_z[i])]
```
rozyczko, line 892: Since you're explicitly creating the list in the next line, line 891 is not needed.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -857,10 +864,70 @@ def update_gui(self):
...
+            cohB = periodictable.elements.symbol(self.nuc_sld_data.pix_symbol[i]).neutron.b_c
+
+            #Calculate the Magnitude of the Coordinate vector for the atom and the center of mass
+            MagnitudeOfCoM = numpy.sqrt(numpy.power(CoM[0]-coordinates[0],2) + numpy.power(CoM[1]-coordinates[1],2) + numpy.power(CoM[2]-coordinates[2],2))
+
+            #Calculate Rate of Gyration (Squared) with the formular
+            RoGNumerator += cohB * (numpy.power(MagnitudeOfCoM,2))
+            RoGDenominator += cohB
+
+        RoG = numpy.sqrt(RoGNumerator/RoGDenominator)
```
rozyczko, line 904: We get the denominator from an external source (`periodictable` module). This means we have to be cautious about plugging it without any checks. Please at least assure `RoGDenominator` is non-zero (and maybe non-NaN?)

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -857,10 +864,70 @@ def update_gui(self):
...
+    
+    def centerOfMass(self):
+        """Calculate Center of Mass(CoM) of provided atom"""
+        CoMnumerator= [0.0,0.0,0.0]
+        CoMdenominator = [0.0,0.0,0.0]
+        CoM = [0.0,0.0,0.0]
+
+        for i in range(len(self.nuc_sld_data.pos_x)):
+            coordinates = [0.0,0.0,0.0]
+            coordinates = [float(self.nuc_sld_data.pos_x[i]),float(self.nuc_sld_data.pos_y[i]),float(self.nuc_sld_data.pos_z[i])]
```
rozyczko, line 916: Similarly, line 915 can be deleted

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -857,10 +864,70 @@ def update_gui(self):
...
+            
+            #Coh b - Coherent Scattering Length(fm)
+            cohB = periodictable.elements.symbol(self.nuc_sld_data.pix_symbol[i]).neutron.b_c
+
+            for i in range(3): #sets CiN
+                CoMnumerator[i] += (coordinates[i]*cohB)
+                CoMdenominator[i] += cohB
```
rozyczko, line 923: A more "pythonic" way to fill arrays sequentially is with `.append`.
```
            for i in range(3): #sets CiN
                CoMnumerator.append(coordinates[i]*cohB)
                CoMdenominator.append(cohB)
```
This also means you can simplify definitions of CoMnumerator etc. to empty lists,
```
        CoMnumerator= []
        CoMdenominator = []
        CoM = []
```

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -857,10 +864,70 @@ def update_gui(self):
...
+
+        for i in range(3):
+            CoM[i] = CoMnumerator[i]/CoMdenominator[i] #center of mass
+
```
rozyczko, line 927: This loop can be replaced by `append` as well.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1435,6 +1563,12 @@ def plot_1_2d(self):
             data.yaxis(r'\rm{Intensity}', 'cm^{-1}')
 
             self.graph_num += 1
+            if self.is_beta or self.is_beta is None:
```
rozyczko, line 1566: you probably just mean
`if self.is_beta:`

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1357,12 +1427,70 @@ def complete(self, input, update=None):
...
+                fQ[a] +=  (cohB * (numpy.sin(currentQValue[a] * magnitudeRelativeCoordinate) / (currentQValue[a] * magnitudeRelativeCoordinate)))
+
+            #Beta Q Calculation
+            self.data_betaQ.append((fQ[a] **2)/(formFactor[a]))
+
+        #Scale Beta Q to 0-1
+        scalingFactor = self.data_betaQ[0]
+        for i in range (len(self.data_betaQ)):
+            self.data_betaQ[i] = (self.data_betaQ[i]/scalingFactor)
+
```
rozyczko, line 1492: Or just 
```
self.data_betaQ = [x/scalingFactor for x in self.data_betaQ]
```
without using the loop explicitly (list comprehension)

Alternatively, you could use numpy arrays for all your lists, making operations much easier.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1357,12 +1427,70 @@ def complete(self, input, update=None):
...
+        xmin = self.qmax_x * _Q1D_MIN
+        qstep = self.npts_x
+
+        fQ = [0 for j in range(self.npts_x)]
+        currentQValue = []
+        for a in range(self.npts_x):
+            if(self.npts_x == 1):
+                currentQValue.append(xmin)
+            else: 
+                currentQValue.append(xmin + (xmax - xmin)/(self.npts_x-1)*a)
```
rozyczko, line 1463: This can be simplified by pulling the first case out of the loop. You want to use as few conditionals inside loops as possible (helps with vectorization).

Maybe something similar to:
```
currentQValue[0] = xmin
for a in range(self.npts_x-1):
    currentQValue.append(xmin + (xmax - xmin)/(self.npts_x-1)*(a+1))
```

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1357,12 +1427,70 @@ def complete(self, input, update=None):
...
+        #Center Of Mass Calculation
+        CoM = self.centerOfMass()
+        self.data_betaQ = []
+
+        # Default values
+        xmax = self.qmax_x
+        xmin = self.qmax_x * _Q1D_MIN
+        qstep = self.npts_x
+
+        fQ = [0 for j in range(self.npts_x)]
```
rozyczko, line 1457: nice list comprehension. Could also use a bit simpler syntax
`f0 = [0.0]*len(self.npts_x)`

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1357,12 +1427,70 @@ def complete(self, input, update=None):
...
+                atomName = str(self.nuc_sld_data.pix_symbol[b])
+                #Coherent Scattering Length of Atom
+                cohB = periodictable.elements.symbol(atomName).neutron.b_c
+
+                coordinates = [float(self.nuc_sld_data.pos_x[b]),float(self.nuc_sld_data.pos_y[b]),float(self.nuc_sld_data.pos_z[b])]
+                
+                relativeCoordinate = [0.0,0.0,0.0]
+                relativeCoordinate[0] = coordinates[0] - CoM[0]
+                relativeCoordinate[1] = coordinates[1] - CoM[1]
+                relativeCoordinate[2] = coordinates[2] - CoM[2]
```
rozyczko, line 1479: you don't need coordinates to be lists. It will be clearer to just use simple variables here.
```
x = float(self.nuc_sld_data.pos_x[b])
y = float(self.nuc_sld_data.pos_y[b])
z = float(self.nuc_sld_data.pos_z[b])

r_x = x - CoM[0]
r_y = y - CoM[1]
r_z = z - CoM[2]
```

rozyczko (commented):
I let myself comment a bit on the general style.
Functionality seems to be OK and the code looks good.
No real issues, just some pointers on how to make the code look more "pythonic"

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
smalex-z: I considered it, but decided against it in the case that we would add more options/indexes in the future- does it make any difference functionally, or is it just a cosmetic fix?

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
smalex-z: I have changed it to leave it as NaN for now- pending Yun Liu's return to the NCNR, we will discuss how best to approach cases where the SLD is 0 or negative.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
smalex-z: Thanks- this is a remnant of my previous program prior to being integrated into Sasview. Sasview only allows # of Q-Bins >= 2, so self.npts_x cannot be equal to 1. I have removed the test and reintegrated the code.

rozyczko (approved):
lgtm

PR #2536, opened 2023-06-28
Post-v5.0.6 Release Update

krzywon:
## Description

This pulls in the handful of changes made in `release-5.0.6` branch and updates the version number to something sensible for the `main` branch.

- [ ] Code has been reviewed


rozyczko (commented):
Looks benign and sensible. Are these the only changes from 5.0.6 which should be merged?

(Merging `release-5.0.6` with `main` directly is probably not a good idea)

lucas-wilkins:
Is using an "a1" suffix supported everywhere?

wpotrzebowski:
At the developers call on July 4th, there was a discussion about next version number (6.0 vs 5.1).  The main concern was that we haven't reached feature parity with 4.x and for this 5.1 may be more appropriate for the next release. However from the user's perspective having the latest version with all the features is probably more important than reaching feature parity.

butlerpd:
In the interest of moving forward, I suggest we use 5.1.0 on the grounds that anything released with less than 6.0 would need to be another cherry picked version which would be down versioned? Unless we think that under proper semantic versioning rules main should be 5.x?

butlerpd:
Responding to @lucas-wilkins, in fact ax and bx are required when using strict versioning for alpha and beta versions which was the case a few years ago. We may have relaxed that but probably best to stick to the established convention rather than keep inventing new ones? In fact release 5.0.6 had b1 and b2 versions as I recall.

butlerpd:
If it runs from the installer then merge it.

PR #2543, opened 2023-07-03
Lowercase PySide6 executables for Linux compatability #2542

### LINKED ISSUE #2542: "PySide6-uic: command not found" when installing or running on linux
When attempting to install or run SasView on a Linux host, errors stating "PySide6-uic: command not found" and "PySide6-rcc: command not found" are produced, which prevent the process from completing.

This has been found to occur on both Arch Linux and Ubuntu while following the recommended dev environment setup instructions given in the wiki. 

These errors have been present since commit  `1509f69`.

The problem can be fixed by making the PySide6 executables (see run_line = ...) all lowercase in src/sas/qtgui/converUI.py before installing. These changes have been confirmed not to effect the Windows and MacOS installation processes by piotr and wpotrzebowski on the Slack channel.

ehewins:
## Description

This RP makes the PySide6 script names in converUI.py all lowercase, as their respective script names on Linux systems are all lowercase. Without this change, 'command not found' errors are produced when SasView is installed or run by following the dev environment setup instructions given in the wiki.

Fixes #2542 

## How Has This Been Tested?

The changes were made and found to get rid of the errors, allowing SasView to install and run correctly on both Arch Linux and Ubuntu. I have not personally tested the effect of these changes on Windows or MacOS, but wpotrzebowski and piotr have informed me via Slack that both upper and lower case work for both of these operating systems, so this change should not cause any problems.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/convertUI.py
+++ b/src/sas/qtgui/convertUI.py
@@ -24,7 +24,7 @@ def pyrrc(in_file, out_file):
     """
     Run the qt resource compiler
     """
-    run_line = f"PySide6-rcc {in_file} -o {out_file}"
+    run_line = f"pyside6-rcc {in_file} -o {out_file}"
     os.system(run_line)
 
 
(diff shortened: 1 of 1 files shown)


rozyczko (approved):
Works fine on Windows.
The fix targeted Linux, so it's reasonable to expect it works there.
According to Wojtek, this also works fine on MacOS so the branch should be merged.

wpotrzebowski:
Indeed, that should work on Mac, however, as I am not able to test the installer I am a bit hesitant to just merge it.

Issue #2541, opened 2023-07-03
Nightly Build artifact doesn't start on Mac


wpotrzebowski:
**Describe the bug**
While trying to run installer it immediately dies. 
Running from terminal gives error
```
Traceback (most recent call last):
  File "sasview.py", line 15, in <module>
  File "sas/cli.py", line 150, in main
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "sas/qtgui/MainWindow/MainWindow.py", line 7, in <module>
ImportError: dlopen(/Volumes/SasView5/SasView5.app/Contents/MacOS/PySide6/QtWidgets.abi3.so, 0x0002): Symbol not found: __ZNK11QHeaderView15initStyleOptionEP17QStyleOptionFrame
  Referenced from: <B4DC7158-7996-3C7E-AFED-A7B3957C1A66> /Volumes/SasView5/SasView5.app/Contents/MacOS/PySide6/QtWidgets.abi3.so
  Expected in:     <A1C89E5F-9C2D-394F-B0BA-D6F3B06D1094> /Volumes/SasView5/SasView5.app/Contents/MacOS/QtWidgets
[31891] Failed to execute script 'sasview' due to unhandled exception: dlopen(/Volumes/SasView5/SasView5.app/Contents/MacOS/PySide6/QtWidgets.abi3.so, 0x0002): Symbol not found: __ZNK11QHeaderView15initStyleOptionEP17QStyleOptionFrame
  Referenced from: <B4DC7158-7996-3C7E-AFED-A7B3957C1A66> /Volumes/SasView5/SasView5.app/Contents/MacOS/PySide6/QtWidgets.abi3.so
  Expected in:     <A1C89E5F-9C2D-394F-B0BA-D6F3B06D1094> /Volumes/SasView5/SasView5.app/Contents/MacOS/QtWidgets
[31891] Traceback:
Traceback (most recent call last):
  File "sasview.py", line 15, in <module>
  File "sas/cli.py", line 150, in main
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "sas/qtgui/MainWindow/MainWindow.py", line 7, in <module>
ImportError: dlopen(/Volumes/SasView5/SasView5.app/Contents/MacOS/PySide6/QtWidgets.abi3.so, 0x0002): Symbol not found: __ZNK11QHeaderView15initStyleOptionEP17QStyleOptionFrame
[...]

PR #2544, opened 2023-07-05
2541 nightly build artifact doesnt start on mac

wpotrzebowski:
## Description
Fixing OSX build from the main branch. It is pyside2 that is problematic and therefore it was removed from requirements (keeping only PySide6). 
It also updates the version of Python and the installer. These were probably not required, but it will be a useful update if it works on all platforms.

Fixes # (issue/issues)
#2541 

## How Has This Been Tested?
Installer downloaded and tested. 

- [x] Code has been reviewed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.9
 
       - uses: actions/checkout@v3
       - id: set-matrix
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -283,7 +283,7 @@ jobs:
     strategy:
       matrix:
         os: [ windows-latest, ubuntu-20.04 ]
-        python-version: [ 3.9 ]
+        python-version: [ 3.8 ]
       fail-fast: false
 
     name: Test installer
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
       - name: Install utilities to build installer
         if: ${{ matrix.installer }}
         run: |
-          python -m pip install pyinstaller
+          python -m pip install pyinstaller==5.7.0
 
       - name: Build sasview with pyinstaller
         if: ${{ matrix.installer }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -283,7 +283,7 @@ jobs:
     strategy:
       matrix:
         os: [ windows-latest, ubuntu-20.04 ]
-        python-version: [ 3.8 ]
+        python-version: [ 3.9 ]
       fail-fast: false
 
     name: Test installer
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
       - name: Install utilities to build installer
         if: ${{ matrix.installer }}
         run: |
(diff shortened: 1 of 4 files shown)


wpotrzebowski:
@rozyczko if you can please check if Windows installer works that would be great.

rozyczko (commented):
This build seems to run OK on Windows (quick, cursory test).
However, I would be careful with piling in two unrelated changes into this bugfix.
The bugfix itself consists of removal of `pyside2` from requirements.txt and replacing `pyside2` with `pyside6` in pytest.ini. Also, adding the missing GL linux library.

But,  unlocking the pyinstaller version and especially bumping Python from 3.8 to 3.9 should be probably done in separate, appropriately named branches and tested well first, I think.

rozyczko (approved):

wpotrzebowski:
@rozyczko good suggestion! I've modified this PR, so it only addresses the original issue. Can you take a quick look at the installer again (after the latest changes)? 

It would be also good if someone also tests ubuntu installer. @ehewins will you be able to do it?

rozyczko:
Yes, this works on Windows.

ehewins:
I can confirm that building and running SasView from this PR works perfectly on Linux/Ubuntu, but there is a problem with the .tar.gz archive packaged in the nightly-build release.
It's conventional to include an "INSTALL" for a "README" file with instructions on how to install the package (or a script to automate the process), but the archive only contains the sasview binary itself and all the dependencies. The sasview binary won't run unless the dependencies are installed properly somewhere, and at the moment the user has to work out for themselves how to do this in a way which doesn't cause any conflicts with their other programs.

llimeht:
> It's conventional to include an "INSTALL" or a "README" file with instructions on how to install the package (or a script to automate the process), but the archive only contains the sasview binary itself and all the dependencies. The sasview binary won't run unless the dependencies are installed properly somewhere, and at the moment the user has to work out for themselves how to do this in a way which doesn't cause any conflicts with their other programs.

Some sort of README would be good to explain what happens next in all of the downloads. For the Ubuntu tarball, there's not actually anything that is required — in particular, the user does _not_ have to figure out how to install it or deal with conflicts between these packages and system packages. There's a chance some extra X-related system packages need to be installed, as is done in the rather cursory CI test of the installer.

As a starting point, this is what is needed:
```
$ tar xf sasview5.tar.gz
$ ./sasview/sasview
```
Running it with an absolute path from any location also works:
```
$ mkdir ~/software
$ tar x -C ~/software -f sasview5.tar.gz 
$ /home/stuart/software/sasview/sasview
```

wpotrzebowski:
@ehewins thanks for testing. I will merge it now as it is critical for Mac users. It would be good to get documentation for ubuntu in place.

ehewins:
> Some sort of README would be good to explain what happens next in all of the downloads. For the Ubuntu tarball, there's not actually anything that is required — in particular, the user does _not_ have to figure out how to install it or deal with conflicts between these packages and system packages. There's a chance some extra X-related system packages need to be installed, as is done in the rather cursory CI test of the installer.
> 
> As a starting point, this is what is needed:
> 
> ```
> $ tar xf sasview5.tar.gz
> $ ./sasview/sasview
> ```
> 
> Running it with an absolute path from any location also works:
> 
> ```
> $ mkdir ~/software
> $ tar x -C ~/software -f sasview5.tar.gz 
> $ /home/stuart/software/sasview/sasview
> ```

Thanks for clarifying @llimeht . The sasview binary returns several "Failed to load module" errors when I run it in an Ubuntu VM, but the process goes off without a hitch when I try running it on my main Arch system. Perhaps it's an issue specific to my VM setup.

PR #2538, opened 2023-06-29
Gsc log spacing

smalex-z:
## Description

Fixes Issue #2537 where Log Spacing and Q-Minimum was unavailable to any users in the Generic Scattering Calculator. Now, an extra Input box and a checkbox has been added to let the user add their own Qx-Minimum values and choose whether they want to use Linear or Log Spacing. This feature is only available for 1D scattering: 2D scattering assumes the detector in the center, so Qx Minimum is set to negative QxMax, and 2D Scattering only uses Linear Spacing, not Log Spacing.

The python math class is now imported with this PR.

NOTE: Do not merge this PR with main, only approve it. Due to merging issues with the UI files, PR #2538, #2548, and an upcoming PR will be merged to main together when all 3 are approved.

Fixes #2537 

## How Has This Been Tested?

- Upload Nuclear Data into GSC
- Qx Min should be set to (QMax * -1) and Box should be disabled
- Log Spacing Box should be unchecked and disabled
- If Dropdown is changed to Debye Full Average:
- Qx Min should be set to (QMax * .001) and Box should be Enabled
- Log Spacing Box should be checked and Enabled

Upon calculation of BetaQ and Form Factor, both should follow either Linear or Log Spacing (based off User selection). 

## Review Checklist (please remove items if they don't apply):

- [ ] Code has been reviewed
- [ ] Functionality has been tested
- [ ] Windows installer (GH artifact) has been tested (installed and worked) 
- [ ] MacOSX installer (GH artifact) has been tested (installed and worked) 
- [ ] User documentation is available and complete (if required)
- [ ] Developers documentation is available and complete (if required)
- [ ] The introduced changes comply with SasView license (BSD 3-Clause)

PR #2545, opened 2023-07-05
Log explorer fix

smalex-z:
## Description

Fix found by @pkienzle. Fixes issue stemming from move to Pyside6 from pyqt5, where the Log Explorer stopped working after the switch. Previously, there were two instances of the function `emit()` that were inherited into the handler- one from logging.handler, and one from QObject.

Fixes #2538 

## How Has This Been Tested?

Run Sasview. Log messages should now show up in the Log Explorer.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -116,7 +116,7 @@ class GuiManager:
 
         # Fork off logging messages to the Log Window
         handler = setup_qt_logging()
-        handler.messageWritten.connect(self.appendLog)
+        handler.postman.messageWritten.connect(self.appendLog)
 
         # Log the start of the session
         logging.info(f" --- SasView session started, version {SASVIEW_VERSION}, {SASVIEW_RELEASE_DATE} ---")
--- a/src/sas/qtgui/Utilities/SasviewLogger.py
+++ b/src/sas/qtgui/Utilities/SasviewLogger.py
@@ -2,32 +2,32 @@ import os
 import sys
 import logging
 
-from PySide6.QtCore import *
+from PySide6.QtCore import QObject, Signal
 
-
-class QtHandler(QObject, logging.Handler):
-    """
-    Version of logging handler "emitting" the message to custom stdout()
-    """
+class QtPostman(QObject):
     messageWritten = Signal(str)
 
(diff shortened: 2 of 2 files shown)


rozyczko (commented):
Works fine but we now see DEBUG level logs, which were not printed by default before.

krzywon:
Note for clarity: the debug level logs are shown in **production**, not just from run.py. I'm leaning toward approving this PR and documenting the debug issues in a separate ticket unless someone smarter than me can find a simple solution.

pkienzle:
The default logger level is set to DEBUG in `sasview/src/sas/system/log.ini`:
https://github.com/SasView/sasview/blob/9f5eab2605d42e0c1a2ea285293a34eaf4103e2b/src/sas/system/log.ini#L46-L49
I suspect you will need a new handler entry to set the level to WARNING:
```ini
[handlers]
keys=console,log_file,qtconsole

…

[handler_qtconsole]
class=sas.qtgui.Utilities.SasviewLogger
formatter=simple
level=WARNING
args=tuple()
```

I don't know which takes precedence, (a) the simple formatter specified in the ini file:
https://github.com/SasView/sasview/blob/9f5eab2605d42e0c1a2ea285293a34eaf4103e2b/src/sas/system/log.ini#L11-L16
or (b) the formatter specified in the code:
https://github.com/SasView/sasview/blob/86c0879b4ab5d20bec74d1c2c2039ebf57f727b6/src/sas/qtgui/Utilities/SasviewLogger.py#L31-L34

I would expect (a) in the code to set the default and (b) in the config file to override. The latter contains an additional format item `%(name)s` so you should be able to tell which one is in use.

The docs for `sas.system.log.SetupLogger._find_config_file` claims that it will look for the .ini file in `~/.sasview` while in debug mode, but from the code that appears to not be the case:
https://github.com/SasView/sasview/blob/9f5eab2605d42e0c1a2ea285293a34eaf4103e2b/src/sas/system/log.py#L60-L75

It might be handy to allow the user to turn on DEBUG when they have problems and we want them to turn it on for more details (either on the command line, or in a `~/.sasview/log.ini` override file).

smalex-z:
I'm just getting a module object is not callable error. Would recommend merging unless someone else can fix it.

PR #2553, opened 2023-07-14
Polydisperse parameter check on model load

### LINKED ISSUE #2551: Polydispersity and models without disperse parameters
Currently, in the fitting perspective, the `Polydispersity` checkbox is always enabled and checkable. When a model is selected that has no polydisperse parameters, the polydispersity list is empty, which can confuse users. The checkbox should be disabled for models that have no polydisperse parameters. Noted by @Susyana.

rozyczko:
This adds a simple check for polydisperse parameters for the chosen model.
If there are such parameters, the polydisp checkbox and the polydisp tab are enabled.

Fixes #2551 

## How Has This Been Tested?

Local tests on windows

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1473,6 +1473,11 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         if self.page_parameters:
             self.updatePageWithParameters(self.page_parameters, warn_user=False)
 
+        # disable polydispersity if the model does not support it
+        has_poly = self._poly_model.rowCount() != 0
+        self.chkPolydispersity.setEnabled(has_poly)
+        self.tabFitting.setTabEnabled(TAB_POLY, has_poly)
+
         # set focus so it doesn't move up
         self.cbModel.setFocus()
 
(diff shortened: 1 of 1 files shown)


krzywon (approved):
The code looks good. This is an improvement over what already existed, so I am approving.

Potential improvements (but feel free to ignore any/all of these):
1. If PD is checked and the model is changed, the checkbox remains checked. The tab is no longer enabled if no PD parameters are in the model, so it's mostly a visual issue, but unchecking the box wouldn't hurt.
2. If no model is selected, the PD checkbox and tab are still enabled. Should this code be put into a class method that is run on tab creation, when the category changes, when the model changes, and (possibly...) when the structure factor changes? Adding a check if a model is selected to the front of `has_poly` should encompass all cases.

rozyczko:
> 1. If PD is checked and the model is changed, the checkbox remains checked. The tab is no longer enabled if no PD parameters are in the model, so it's mostly a visual issue, but unchecking the box wouldn't hurt.

Unchecking the disabled checkbox means modifying user chosen state. When the user switches back to the previous model with PD params, they expect the checkbox to remain checked. Seeing the checked status go away and come back is, I think, even more confusing (not to mention, a nightmare to track).

> 2. If no model is selected, the PD checkbox and tab are still enabled. Should this code be put into a class method that is run on tab creation, when the category changes, when the model changes, and (possibly...) when the structure factor changes? Adding a check if a model is selected to the front of `has_poly` should encompass all cases.

It is enough to make the check on the model change. I changed the initial state of this checkbox to `false` so the tab isn't active by default, just like magnetic.

krzywon:
> It is enough to make the check on the model change. I changed the initial state of this checkbox to `false` so the tab isn't active by default, just like magnetic.

There is one use case this breaks down that I can think of. A user selects a model with polydisperse parameters, then selects a new category. The model box goes blank, but the polydispersity checkbox and tab remain enabled. It's a small issue so I don't think it should hold up this PR.

PR #2557, opened 2023-07-19
Strip debug messages from production version of Qt console

pkienzle:
## Description

Suggested changes to logging controls.

Note that the Qt linkage is via the QtPostman signal. This is currently owned by QtHandler, but it could instead be owned by the console window and passed to the handler on creation. This would be consistent with the file handler which takes a log file name as input. With this change the set_qt_logger code would not have to check if there is an existing handler before creating its own.

Moving the QtHandler creation to log.ini causes a different problem: it requires the GUI in order to bind the QtPostman signal to the handler. An alternative is to create a QtPostman signal instance as a global so it is available for binding independently by logging handler and gui (e.g., as a class attribute of QtHandler).

As currently implemented the logging config is called even when we are using sasview to run a script file, so setting up QtHandler as part of logging config would force a Qt dependence on sascalc. Maybe it would be better if logging were only set up when the QtGui is about run?

## How Has This Been Tested?

A `--level` option was added to `cli.py` so that you can specify production or development when running sasview from the development branch. You can instead specify the message level that will be displayed. The `GuiManager.py` file includes some commented out logging lines at various severity levels. Or you can trace through the code to find a message at the right level and try to trigger it. For example, `matplotlib._afm._parse_header` will produce errors if you give it an almost usable header string.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -122,6 +122,9 @@ class GuiManager:
         logging.info(f" --- SasView session started, version {SASVIEW_VERSION}, {SASVIEW_RELEASE_DATE} ---")
         # Log the python version
         logging.info("Python: %s" % sys.version)
+        #logging.debug("Debug messages are shown.")
+        #logging.warn("Warnings are shown.")
+        #logging.error("Errors are shown.")
 
         # Set up the status bar
         self.statusBarSetup()
--- a/src/sas/system/log.py
+++ b/src/sas/system/log.py
@@ -16,7 +16,7 @@ Module that manages the global logging
 TRACED_PACKAGES = ('sas', 'sasmodels', 'sasdata', 'bumps', 'periodictable')
 IGNORED_PACKAGES = {
     'matplotlib': 'ERROR',
-    'numba': 'ERROR',
+    'numba': 'WARN',
     'h5py': 'ERROR',
     'ipykernel': 'CRITICAL',
 }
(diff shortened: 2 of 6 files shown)


src/sas/system/log.py
```diff
@@ -12,70 +12,59 @@
 Module that manages the global logging
 '''
 
+#BASE_LOGGER = 'sasview'
+TRACED_PACKAGES = ('sas', 'sasmodels', 'sasdata', 'bumps', 'periodictable')
```
krzywon, line 16: nitpick: Did you mean `TRACKED_PACKAGES`?

krzywon (approved):
Different cli options for logging seem to work (I tested production and development) and the installer does not display debug messages.

src/sas/qtgui/MainWindow/GuiManager.py
wpotrzebowski, line 1: It seems these are debugging comments and can be removed?

src/sas/system/log.py
pkienzle: I mean the packages for which the debug traces should be printed during development.

src/sas/qtgui/MainWindow/GuiManager.py
pkienzle: I left them in for manual testing of this PR.

These ones probably won't be needed again, but I find that when is code is complicated enough that I need print statements to trace it then it will probably fail in some future version and it is convenient to turn the traces back on. They also act as developer documentation, indicating which variables are important 8-)

PR #2520, opened 2023-05-16
Particle editor

lucas-wilkins:
I've made a new perspective which gives a simple interface for describing arbitrary particles. It's now at a stage where the basic bits work. Basically, just numerically integrate a particle defined by a function, with a bunch of GUI stuff to make it as painless as possible (e.g. visualisations, special syntax highlighting, automated conversion of coordinates). I'm looking for early feedback on the interface and workflow (i.e. high level comments, not a list things that don't work, I know what those are). 

It's currently not integrated into SasView properly, but it can be run by executing the script at `sas.qtgui.Perspectives.ParticleEditor.DesignWindow`

The plan is (assuming I fix the problem below) to add the following functionalities, which are only placeholders right now:
1) Magnetism
2) Parametric functions to allow fitting
3) More helper functions for building shapes
   i) function composition with rotations and translations
   ii) built-in shapes
4) More kinds of polydispersity (inc ensembles of random particles)
5) Structure factors
6) GPU acceleration

image

PR #2562, opened 2023-07-24
Syntax highlighting in Pyside6

rozyczko:
## Description

Missed converting the syntax highlighting module. General changes were made (headers, calls) but the actual highlighting was not working properly. It is now

Fixes # 2555

- [ ] Code has been reviewed


rozyczko:
> Oh, what were you using that for? I have recently written my own in #2520
> 
> I'm very much aware that the RE method is not the correct way of implementing these. Doing it properly with a (in)finite state machine was on my list of things to do.

This is used in the model editor - load/write a plugin and edit it. The python/C code is properly highlighted.
I am not sure why you think regex can't handle this? It seems to be doing good enough job here.
Of course it won't be as good as in professional IDEs but this hasn't been a goal...

PR #2548, opened 2023-07-07
Updated rG&rg

smalex-z:
## Description

Upon discussion with @butlerpd @pkienzle and @yunliu01, we decided it would be best to make the difference between Mass Based Radius of Gyration (Rg) and SLD Based Guinier Radius (RG) more obvious. I have separated the radius of gyration into two boxes, one for Mass-Based and one for SLD-Based.

I have also included code to change the SLD-Based RG based on the Solvent SLD, but that is commented out until a standard for determining protein volume can be agreed upon.

## How Has This Been Tested?

Open Generic Scattering Calculator

## Review Checklist (please remove items if they don't apply):

- [ ] Code has been reviewed
- [ ] Functionality has been tested
- [ ] Windows installer (GH artifact) has been tested (installed and worked) 
- [ ] MacOSX installer (GH artifact) has been tested (installed and worked) 
- [ ] User documentation is available and complete (if required)
- [ ] Developers documentation is available and complete (if required)
- [ ] The introduced changes comply with SasView license (BSD 3-Clause)

PR #2565, opened 2023-07-24
Custom fit models

smalex-z:
## Description

Adds a feature to automatically generate plugin models with Structure Factors from the Generic Scattering Calculator. Models are written in a singular combined C + Python, with 2 options for effective radius: equivalent sphere volume and Radius of Gyration. It also gives you the option to choose your own file name. There are 4 parameters- SLD, Solvent SLD, Swelling, and Protein Volume.

This PR combines 2 other PRs as well due to .UI merging issues: #2548 and #2538. 

## How Has This Been Tested?

Upload a file into nuclear data in the Generic Scattering Calculator- preferably a .pdb file. Choose one of the 1D Scattering options, check the Plugin Models checkbox, and change the desired filename if wanted.  Hit calculate and let it run.
After it is done calculating, the model file should automatically be created and put into the plugin_models folder of `.sasview`. You should be able to load it in the Fit Panel by going to Category - Plugin Models, then choosing your model name and choosing a structure factor as wanted.

- [ ] Code has been reviewed


butlerpd:
Jeff to check two other PR then close if ok.  Otherwise need to look at "does it do what it says on the tin" and overall code

krzywon:
I think three suggestions brought up by @pkienzle (appending to a numpy array, moving calculations into sascalc, and snake_case vs. CamelCase) should be fixed before merging this request.

#2538 and #2548 look fine at this point and can probably be closed in favor of this PR.

src/sas/sascalc/calculator/gsc_model.py
```diff
@@ -0,0 +1,156 @@
...
+    plugin_location = models.find_plugins_dir()
+    full_path = os.path.join(plugin_location, self.txtFileName.text())
+    if os.path.splitext(full_path)[1] != ".py":
+        full_path += ".py"
+
+    # generate the model representation as string
+    model_str = generateModel(self)
+    TabbedModelEditor.writeFile(full_path, model_str)
+
+    self.manager.communicate.customModelDirectoryChanged.emit()
```
krzywon, line 23: Importing qtgui from sascalc should never happen.

writeFile: Return the model string from here and use the TabbedModelEditor file writer from there -or- port the write method to here since it's pretty boilerplate.

self.manger doesn't exist in this context so this needs to move back regardless.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -577,11 +625,23 @@ def change_is_avg(self):
...
             self.txtMx.setEnabled(True)
             self.txtMy.setEnabled(True)
             self.txtMz.setEnabled(True)
+            self.txtQxMin.setEnabled(False)
+            self.txtQxMin.setText(str(float(self.txtQxMax.text())*-1))
+            self.checkboxLogSpace.setChecked(False)
+            self.checkboxLogSpace.setEnabled(False)
+            self.checkboxPluginModel.setEnabled(False)
+            self.checkboxPluginModel.setChecked(False)
+            self.txtFileName.setEnabled(False)
```
krzywon, line 644: Most of these checked/enabled values are opposite depending on `self.is_avg` and each pair could be combined into a single line outside the if/else statement.

```
self.txtMx.setEnabled(not self.is_avg)
...
self.txtQxMin.setEnabled(self.is_avg)
...
multiple = 0.001 if self.is_avg else -1
self.txtQxMin.setText(str(float(self.txtQxMax.text())*mulitple)
...
etc.
```

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -534,7 +560,26 @@ def change_data_type(self):
         self.update_gui()
         self.check_for_magnetic_controls()
 
+    def change_qValidator(self):
+            if float(self.txtQxMax.text()) == 0:
```
krzywon, line 564: I spoke to @yunliu01 about this and he seems to think the values you have here are acceptable, but this is a placeholder for future work using Rg to estimate the Q-range that should be used. I would suggest you add a  comment describing that future work otherwise this coercion of Q values looks suspect.

src/sas/sascalc/calculator/geni.py
```diff
@@ -514,4 +516,109 @@ def _spin_weights(in_spin, out_spin):
...
+    fQlist = np.empty(npts_x)
+    CoM = centerOfMass(nuc_sld_data)
+    r_x = np.subtract(nuc_sld_data.pos_x , CoM[0])
+    r_y = np.subtract(nuc_sld_data.pos_y , CoM[1])
+    r_z = np.subtract(nuc_sld_data.pos_z , CoM[2])
+    magnitudeRelativeCoordinate = np.sqrt(np.power(r_x, 2) + np.power(r_y, 2) + np.power(r_z, 2))
+    cohB = np.asarray([periodictable.elements.symbol(atom).neutron.b_c for atom in nuc_sld_data.pix_symbol])
+    
+    for i in range(npts_x):
+        fQ = np.sum(cohB*(np.sin(qX[i] * magnitudeRelativeCoordinate) / (qX[i] * magnitudeRelativeCoordinate)))
```
pkienzle, line 617: Fails if qX[i] = 0 or any r = center of mass.

You can use `np.sinc((qX[i]/np.pi) * dr)` to handle sin(0)/0.

This runs a bit slower than the incorrect sin(qr)/qr on my system. If it is a concern you can wrap it in numba.njit.

The following requires 50 μs on my machine vs 70 μs for Σ b sin(qΔr)/qΔr) and 85 μs for Σ b sinc(qΔr/π))
```python
@numba.njit
def f(q, b, dr):
     total = 0
     for bi, dri in zip(b,dr):
         qr = q*dri
         total += bi * np.sin(qr)/qr if qr != 0. else bi
     return total
```

wpotrzebowski:
I've tried it on several PDB models to see how it works on models with a large number of atoms. While it generally works fine I think we will need to put some limits on very large structures (e.g. virus capsids). For the following example https://github.com/Andre-lab/hbv_trSAXS/blob/master/HBVCP_empty_assembly/bayesian_models/capsid_T4.pdb it took about 11min to load the model and I interrupted curve computation after 30min.

yunliu01:
> I've tried it on several PDB models to see how it works on models with a large number of atoms. While it generally works fine I think we will need to put some limits on very large structures (e.g. virus capsids). For the following example https://github.com/Andre-lab/hbv_trSAXS/blob/master/HBVCP_empty_assembly/bayesian_models/capsid_T4.pdb it took about 11min to load the model and I interrupted curve computation after 30min.

I agree. We could set a limit, above which, the program can give a warning saying that it may be slow.
Because the calculation is used just once, sometimes, people may be patient enough to let it go through.
We could leave the option to the users to decide.

Also, I am wondering if there is a way to estimate the amount of time needed to load and compute the model. 
This information can allow the users to decide if they would like to wait or stop the computation.

Without knowing the details of how the PDB reader code, I might be wrong.
However, I would naively imagine that the loading time should be proportional to the number of atoms, N, in a PDB file.
And based on the equations used for the current P(Q) calculation in the generic scattering calculator, the needed computation time is probably proportional to N^2.  If we could estimate the computation time needed after reading the first 1000 atoms, I would imagine that it could give us enough information to estimate the time needed for the rest of calculation for loading the PDB file. And the calculation time for P(Q) might be estimated in a similar way.

wpotrzebowski:
Should we aim for 6.0.0 with this PR?

yunliu01:
It seems that the absolute intensity is not calculated correctly as the intensity is scaled by a factor about 1e4. This may be due to the unit conversation. The automatic generated plugin model code needs to have *f2 multiplied with 1e4 and *f1 multiplied with 1e2.

wpotrzebowski:
Additional comments in https://github.com/SasView/sasview/issues/2671

butlerpd:
Needs review of docs and fix abs scale.  The resolution issue probably something for a later issue.

wpotrzebowski:
Conflict also needs to be resolved before merging

butlerpd:
@butlerpd to do final functionality and docs check

wpotrzebowski:
MAC installer doesn't start. It seems to work fine from local dev env.

src/sas/sascalc/calculator/geni.py
lucas-wilkins: Has this been resolved?

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
lucas-wilkins: Has this been resolved?

src/sas/sascalc/calculator/gsc_model.py
lucas-wilkins: Has this been resolved?

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+    rsq = np.sum((CoM - coordinates)**2, axis=1)
+
+    rog_num = np.sum(masses * rsq)
+    rog_den = np.sum(masses)
+    guinier_num = np.sum(contrastSLDs * rsq)
+    guinier_den = np.sum(contrastSLDs)
+
+
+    if rog_den <= 0: #Should never happen as there are no zero or negative mass atoms
+        rog_mass = "NaN"
```
lucas-wilkins, line 545: Consider handling errors properly

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+        logging.warning("Atomic Mass is zero for all atoms in the system.")
+    else:
+        rGMass = np.sqrt(rog_num/rog_den)
```
lucas-wilkins, line 548: Inconsistent and opaque variable name

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+    elif (guinier_num/guinier_den) < 0:
+        guinier_value = (str(round(np.sqrt(-guinier_num/guinier_den),1)) + " Å")
+        logging.warning("Radius of Gyration Squared is negative. R(G)^2 is assumed to be |R(G)|* R(G).")
+    else:
+        guinier_value = (str(round(np.sqrt(guinier_num/guinier_den),1)) + " Å")
+
+    return [rog_mass,guinier_value,rGMass]          #[String, String, Float], float used for plugin model
+
+    
+def centerOfMass(nuc_sld_data):
```
lucas-wilkins, line 564: what is the type of this variable, looks like it is just some data without any hint about how one should expect it to be formatted. If its simply structures, just provide a type, otherwise make a dataclass

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+    """Calculate Center of Mass(CoM) of provided atom"""
+    CoMnumerator= [0.0,0.0,0.0]
```
lucas-wilkins, line 566: non-pythonic variable name

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+    if guinier_den == 0:
+        guinier_value = "NaN"
+        logging.warning("Effective Coherent Scattering Length is zero for all atoms in the system.")
+    elif (guinier_num/guinier_den) < 0:
+        guinier_value = (str(round(np.sqrt(-guinier_num/guinier_den),1)) + " Å")
+        logging.warning("Radius of Gyration Squared is negative. R(G)^2 is assumed to be |R(G)|* R(G).")
+    else:
+        guinier_value = (str(round(np.sqrt(guinier_num/guinier_den),1)) + " Å")
+
+    return [rog_mass,guinier_value,rGMass]          #[String, String, Float], float used for plugin model
```
lucas-wilkins, line 561: Don't rely on position to give meaning to variables, use a structured datatype, preferably a dataclass, or at a stretch, a named tuple

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+
+    #Avoid division by zero - May occur through contrast matching
+    if guinier_den == 0:
+        guinier_value = "NaN"
+        logging.warning("Effective Coherent Scattering Length is zero for all atoms in the system.")
+    elif (guinier_num/guinier_den) < 0:
+        guinier_value = (str(round(np.sqrt(-guinier_num/guinier_den),1)) + " Å")
+        logging.warning("Radius of Gyration Squared is negative. R(G)^2 is assumed to be |R(G)|* R(G).")
+    else:
+        guinier_value = (str(round(np.sqrt(guinier_num/guinier_den),1)) + " Å")
```
lucas-wilkins, line 559: unnecessary parenthesis

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
     return weight
 
+def radius_of_gyration(nuc_sld_data):
+    #Calculate Center of Mass(CoM) First
+    CoM = centerOfMass(nuc_sld_data)
+
+    #Now Calculate RoG
+    guinier_num = guinier_den = rog_num = rog_den = 0.0
+
+    dtype = np.dtype('double')
```
lucas-wilkins, line 522: is "double" a valid dtype specification. Also, why?

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+    x, y, z = nuc_sld_data.pos_x, nuc_sld_data.pos_y, nuc_sld_data.pos_z
+    pix_symbol = nuc_sld_data.pix_symbol
+    coordinates = np.array([x, y, z], dtype=dtype).T
+    slds, masses = np.empty(len(pix_symbol), dtype), np.empty(len(pix_symbol), dtype)
```
lucas-wilkins, line 526: multiple lines would be more readable

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
     )
     return weight
 
+def radius_of_gyration(nuc_sld_data):
```
lucas-wilkins, line 515: what is the type of nuc_sld_data?

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+    #Calculate Center of Mass(CoM) First
+    CoM = centerOfMass(nuc_sld_data)
+
+    #Now Calculate RoG
+    guinier_num = guinier_den = rog_num = rog_den = 0.0
+
+    dtype = np.dtype('double')
+    x, y, z = nuc_sld_data.pos_x, nuc_sld_data.pos_y, nuc_sld_data.pos_z
+    pix_symbol = nuc_sld_data.pix_symbol
+    coordinates = np.array([x, y, z], dtype=dtype).T
```
lucas-wilkins, line 525: vstack or hstack instead?

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
 
+def radius_of_gyration(nuc_sld_data):
+    #Calculate Center of Mass(CoM) First
+    CoM = centerOfMass(nuc_sld_data)
+
+    #Now Calculate RoG
+    guinier_num = guinier_den = rog_num = rog_den = 0.0
+
+    dtype = np.dtype('double')
+    x, y, z = nuc_sld_data.pos_x, nuc_sld_data.pos_y, nuc_sld_data.pos_z
```
lucas-wilkins, line 523: You wont run out of lines if you use more of them

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+        #Coh b - Coherent Scattering Length(fm)
+        cohB = periodictable.elements.symbol(nuc_sld_data.pix_symbol[i]).neutron.b_c
+
+        for j in range(3): #sets CiN
+            CoMnumerator[j] += (coordinates[j]*cohB)
+            CoMdenominator[j] += cohB
+
+    CoM = [] 
+    for i in range(3):   
+        CoM.append(CoMnumerator[i]/CoMdenominator[i] if CoMdenominator != 0 else 0) #center of mass, test for division by zero
```
lucas-wilkins, line 581: If you did all this in numpy it would be faster, also nump.divide has good ways of dealing with division by zero https://numpy.org/doc/stable/reference/generated/numpy.divide.html

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+    
+    return CoM
+    
+def create_betaPlot(qX, nuc_sld_data, npts_x, formFactor):
```
lucas-wilkins, line 585: types

src/sas/sascalc/calculator/geni.py
lucas-wilkins: No, also I agree

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+
+    
+    #Scale Beta Q to 0-1
+    scalingFactor = data_betaQ[0]
+    data_betaQ = data_betaQ / scalingFactor
+
+    return data_betaQ
+
+
+def FQ(qX, nuc_sld_data, npts_x):
```
lucas-wilkins, line 603: types

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
yunliu01: A comment has been added to this part of the code.

src/sas/sascalc/calculator/geni.py
yunliu01: np.sin(qr)/qr is changed to np.sinc(qr/np.pi).

src/sas/sascalc/calculator/geni.py
```diff
@@ -510,4 +512,109 @@ def _spin_weights(in_spin, out_spin):
...
+    slds, masses = np.empty(len(pix_symbol), dtype), np.empty(len(pix_symbol), dtype)
+    for i, sym in enumerate(pix_symbol):
+        atom = periodictable.elements.symbol(sym)
+        masses[i] = atom.mass
+        slds[i] = atom.neutron.b_c
+        #solvent_slds = atoms.volume() * 10**24 * float(self.txtSolventSLD.text()) * 10**5
+
+    #TODO: Implement a scientifically sound method for obtaining protein volume - Current value is a inprecise approximation. Until then Solvent SLD does not impact RG - SLD.
+    # contrastSLD = sld - solvent_sld         #femtometer
+    contrastSLDs = slds                       #femtometer
```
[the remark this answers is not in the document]
yunliu01: Agreed. Have changed constrastSLDs to contrastSLs. Also changed slds to coherentSLs, etc.

yunliu01:
Beta_Q should be one at Q=0. Therefore, calculated beta_Q needs to be normalized for the value at Q=0. This requires that both F_Q, and P_Q to be normalized properly at Q=0, i.e., F_Q(Q=0)=1 and P_Q(Q=0)=1. However, P_Q is currently not a normalized form factor. Therefore, the current code obtained the normalized Beta_Q by dividing it with Beta_Q[Q_min], the value of Beta_Q at lowest Q point. Note that when using the calculated form factor to fit the data, the code also needs to use the normalized form factor. The normalized form factor is also obtained by dividing the data with the value at the lowest Q point.

This approach could be problematic if a user forgets to choose the first Q point to be very small Q values. The documentation needs to be updated to let user know this whenever they use this function.

In principle, it is easy to calculate P_Q(Q=0) and F_Q(Q=0). In future, it is better to update the code to normalize the data with the exact values at Q=0 instead of using a value at a finite Q. With a quick look at the code, it seems that F_Q seems to be normalized correctly already. (This has not be fully tested yet.) However, the calculation of P_Q needs to be updated in future.

butlerpd:
So to be clear @yunliu01 -- you are saying that Beta_Q is normalized to Beta_qmin but that F(Q) is normalized to F(Q=0)?  Also, is not P(Q)=F(Q)^2 not then also normalized by P(Q=0) by definition? Finally if it is that simple should we not just normalize to Q=0 properly? Or is this considered too hard at this point?

butlerpd (approved):
Testing the install on windows (after merging the fast Debye calculator) works very nicely (and fast!). It looks to me that it is ready to merge.

Two points that should be addressed as soon as possible (and the first maybe as part of this release even?):
1. since plugin docs will now render in the installer version, it would be nice of some of the documentation currently buried in the help docs were moved (or even copied) to be part of the model docs as suggested in the new issue #2872. In particular an explanation of the new parameters of swelling and volume and how they should be used.
2. Because this model is not based on an analytical equation but on a calculated series of points, it is quite possible for the data to end up being outside the value that can be returned by the model.  If not, but the data has resolution, it is still possible that the fit needs data beyond what the model can return. It would be good to display a helpful message when the data+resolution is going to exceed the bounds of the calculated model telling the user that is the case and that they should recompute the plugin using a larger Q range. That may actually be a separate sasmodel issue

Finally I note that it is also possible that a user, particularly on a Mac, does not provide enough points in the Debye calculated plugin such that interpolation is bad in some areas (where there may be significant oscillations. Is there a way to flag that to the user? Is there a way to algorithmically assess it, maybe based on the smoothness of the curve?

wpotrzebowski:
It seems to work on Mac as well (with default GSC calculation in the backend).

lucas-wilkins (approved):

PR #2566, opened 2023-07-27
2550 wedge slicer

### LINKED ISSUE #2482: Problem with dragging slicer elements on plot
**Describe the bug**
When dragging items in the slicer interactor (e.g. move the right line in the y box sector), the item "jumps" to its final position on the plot only **after** releasing the mouse button. Depending on how fast one moves the mouse and how fast the computer is, this could be a substantial or a small effect

**To Reproduce**
1. Load some 2D data ( I used the example data shipped with sasview)
2. Sent the data to a new plot (`Create New` under `Plot`)
3. From plot context sensitive menu chose a slicer (`Box Averaging in Qx`, `Box Averaging in Qy`, `Annulus`, `Box Sum` and `Sector` have all been tested)
4. Using the mouse drag one of the lines. Then stop while still holding the mouse button down
5. Without moving the mouse, let go of the mouse key (dragging rapidly should make the jump more obvious, particularly on a slower computer).
6. The line will jump to its final position as the 1D plot is updated.

**Expected behavior**
The line should move with the mouse. Once the mouse no longer moves, the lines should not move anymore.  The problem as a user is that it is twofold: it is impossible to drag to a precise location and more importantly it is not clear which position is being used in creating the 1D plot of interes.

**Screenshots**
None

 - Version: main branch as of 4/9/2023. Should be in all 5.x branches and maybe in 4.x as well.

 - OS: tested on windows 10 but should be true on any OS

**Additional context**
The problem stems from the convoluted OO structure used, but fundamentally is because, after a drag event, the screen is updated **before** the underlying positional data for the lines is updated. 

This comes about because of the use of the base object (the self object of the object that instantiated the class making the call) and the fact that the class compostion/aggregation goes two deep. When to use `base.update` vs `base.base.update` can be confusing and was done backward in fact.  Suggest removing the call to base from the virtual class of `BaseInteractor` to remove confusion (and the multiple redundant redraws in one case) as well as the use of `base.base.update`

### LINKED ISSUE #2550: Wedge/Annulus-Sector Slicer

The issue will be the first step to enhance the slicer options as discussed in the post https://github.com/orgs/SasView/discussions/2198 . The new slicer shape defines a region of interest in 2D data with a four sided shape described by two radii and two concentric arcs (annular sector). As the geometric name "annular sector" might lead to confusion with the already existing sector and annulus slicer, it was proposed to call it "wedge" slicer. Further discussion of what to name it would be welcome.

The wedge is a combination of a sector with an annulus. The implementation would be similar to "Box Interactor X" and "Box Interactor Y", averaging over either Q or Phi within the selected region to produce a 1D plot of intensity against the other variable.
Much of the code involved will be taken from `AnnulusSlicer.py` and `SectorSlicer.py` and the classes they define. The proposed region of interest could be constructed using these classes.
Each is missing two parameters which the other specifies - AnnulusSlicer needs a start and end angle to produce a 1D plot against a finite angular region < 360 Degrees. The default being a full circle, when first plotted. SectorSlicer is missing a minimum and maximum radius to select a q range.

Alternatively, one can refactor the code and define a new "Wedge" slicer class. The two pre-existing slicers sector and annulus could be rewritten as simplified cases of this one. If I went for this second option, I'd take the opportunity to prune some repeated code from the classes in `SectorSlicer.py` (though this pruning will likely happen anyway sooner or later).

ehewins:
## Description

This PR addresses issue #2550, implementing the Wedge Slicer. The desired functionality is present, and the ideas behind `RadiusInteractor.py` could be applied to simplify `SectorSlicer.py` if the slicers are ever refactored. Also included are the cherry-picked commits from branch `2482-problem-with-dragging-slicer-elements-on-plot`, which served as a base for my work here.

Closes #2550 and also closes #2482

## How Has This Been Tested?

The slicer interactor elements were adjusted in various sequences, through both the GUI and the Edit Slicer Parameters window accessible via the right-click menu. It was found that these parameters behaved consistently, and each of the averaging directions produced convincing-looking 1D plots.
Whether the 1D plot results produced by the slicer are accurate depends on whether the classes `SectorQ` and `SectorPhi` from `sasdata/sasdata/data_util/manipulations.py` are working as intended. They should be sufficient for high resolution data, but there may be further work necessary to improve their reliability with lower resolution data.

- [ ] Code has been reviewed
- [X] User documentation is available and complete (if required)
- [X] Developers documentation is available and complete (if required)
- [X] The introduced changes comply with SasView license (BSD 3-Clause)


ehewins:
Following some further testing, I've discovered a bug in the `SectorPhi` class in `manipulations.py` which prevents certain phi and delta_phi angles from being set when attempting to perform "Wedge Averaging in Phi" . Considering this class has been unimplemented before now, I guess it's not surprising it has undiscovered bugs. Will update this post when I've found a fix.

src/sas/qtgui/Plotting/Slicers/ArcInteractor.py
```diff
@@ -0,0 +1,140 @@
...
+    Draw an arc on a data2D plot with a variable radius (centered at [0,0]).
+    User interaction adjusts the parameter r
+
+    param r: radius from (0,0) of the arc on a data2D plot
+    param theta: angle from x-axis of the central point on the arc
+    param phi: angle from the centre point on the arc to each of its edges
+    """
+    def __init__(self, base, axes, color='black', zorder=5, r=1.0,
+                 theta=np.pi / 3, phi=np.pi / 8):
+        BaseInteractor.__init__(self, base, axes, color=color)
```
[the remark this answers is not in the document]
ehewins: Yep, I can see the sense in that. All of the slicer interactors call `BaseInteractor.__init__()` however, would you advise making this change universally or just in WedgeInteractor, ArcInteractor and RadiusInteractor?

rozyczko:
I started looking at the functionality.

Loaded `P123_D2O_30_percent.dat` from the 2D examples and attempted to invoke the Phi wedge averaging.
The slicer shows up but throws:

```
  File "C:\projects\sasview\src\sas\qtgui\Plotting\Slicers\BaseInteractor.py", line 129, in onRelease
    self.moveend(ev)
  File "C:\projects\sasview\src\sas\qtgui\Plotting\Slicers\RadiusInteractor.py", line 132, in moveend
    self.base.moveend(ev)
  File "C:\projects\sasview\src\sas\qtgui\Plotting\Slicers\WedgeSlicer.py", line 253, in moveend
    self._post_data()
  File "C:\projects\sasview\src\sas\qtgui\Plotting\Slicers\WedgeSlicer.py", line 344, in _post_data
    super()._post_data(SectorPhi)
  File "C:\projects\sasview\src\sas\qtgui\Plotting\Slicers\WedgeSlicer.py", line 168, in _post_data
    sector = sect(self.data)
  File "C:\Users\piotrrozyczko\.conda\envs\sasview\lib\site-packages\sasdata\data_util\manipulations.py", line 1000, in __call__
    return self._agv(data2D, 'phi')
  File "C:\Users\piotrrozyczko\.conda\envs\sasview\lib\site-packages\sasdata\data_util\manipulations.py", line 848, in _agv
    err_data = data2D.err_data[np.isfinite(data2D.data)]
```

src/sas/qtgui/Plotting/Slicers/ArcInteractor.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        Clear this slicer and its markers
+        """
+        self.clear_markers()
+        try:
+            self.marker.remove()
+            self.arc.remove()
+        except:
+            # Old version of matplotlib
+            for item in range(len(self.axes.lines)):
```
rozyczko, line 66: should we be worrying about this "old version" path anymore? `remove()` is available in >3.0, which we have standardized on for quite some time now.

src/sas/qtgui/Plotting/Slicers/ArcInteractor.py
```diff
@@ -0,0 +1,135 @@
...
+        if r is not None:
+            self.radius = r
+        # Calculate the points on the arc, and draw them
+        for i in range(self.npts):
+            angleval = 2 * self.phi / (self.npts - 1) * i + (self.theta - self.phi)
+            xval = 1.0 * self.radius * np.cos(angleval)
+            yval = 1.0 * self.radius * np.sin(angleval)
+            x.append(xval)
+            y.append(yval)
+
```
rozyczko, line 91: this can be improved (performance) by taking out the factor and offset calculations out of the loop.
precalculate
```
angle_factor = 2 * self.phi / (self.npts - 1)
angle_offset = self.theta - self.phi
```
and just use those in the loop, e.g. with
```
x = [self.radius * np.cos(angle_factor * i + angle_offset) for i in range(self.npts)]
y = [self.radius * np.sin(angle_factor * i + angle_offset) for i in range(self.npts)]
```

src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
```diff
@@ -0,0 +1,345 @@
...
+    def validate(self, param_name, param_value):
+        """
+        Validate input from user
+        Values get checked at apply time.
+        """
+        MIN_Q_DIFFERENCE = self.dqmin
+        MIN_PHI_DIFFERENCE = 0.01
+        isValid = True
+
+        # Stitched together from other slicers. There could be a neater way.
```
rozyczko, line 221: You could probably invent something fancy here but maybe readability is better?


```def validate(self, param_name, param_value):
    """
    Validate input from user.
    Values get checked at apply time.
    """
    
    def check_radius_difference(param_name, other_radius_name, param_value):
        if np.fabs(param_value - self.getParams()[other_radius_name]) < self.dqmin:
            return f"Inner and outer radii too close. Please adjust."
        elif param_value > self.qmax:
            return f"{param_name.capitalize()} exceeds maximum range. Please adjust."
        return None

    def check_phi_difference(param_value):
        if np.fabs(param_value) < 0.01:
            return "Sector angles too close. Please adjust."
        return None

    def check_bins(param_value):
        if param_value < 1:
            return "Number of bins cannot be <= 0. Please adjust."
        return None

    validators = {
        'r_min': lambda value: check_radius_difference('r_min', 'r_max', value),
        'r_max': lambda value: check_radius_difference('r_max', 'r_min', value),
        'delta_phi [deg]': check_phi_difference,
        'nbins': check_bins
    }

    if param_name in validators:
        error_message = validators[param_name](param_value)
        if error_message:
            print(error_message)
            return False

    return True

```

rozyczko (approved):
Good code, just some minor comments not influencing the overall picture.

Great job of adding and correcting lots of comments in the code!

The only issue is the runtime exception when applying the slicer onto a residual plot, but this applies to all slicers, so needs to be made a separate issue.

PR #2568, opened 2023-07-31
unit conversion for gui is missing

### LINKED ISSUE #2561: Unit conversion for GUI is missing Å³
The unit conversion code here:

https://github.com/SasView/sasview/blob/37e7d1c95e9c7d56e78dc1b8bd4fd9016c4bb44f/src/sas/qtgui/Utilities/GuiUtils.py#L964-L981

should support the units listed in the rst_prolog here:

https://github.com/SasView/sasview/blob/37e7d1c95e9c7d56e78dc1b8bd4fd9016c4bb44f/docs/sphinx-docs/source/rst_prolog#L1-L24

(or the equivalent from sasmodels since that's where the units are being used in the parameter tables).

Within sasmodels I see:

* adsorbed layer uses `mg/m^2`,
* be polyelectrolyte uses `Ang^3`,
* broad peak uses `1/Ang` (prolog has `Ang^-1`),
* oriented particles use `degrees` (prolog has `deg`),
* core shell ellipsoid, hayter msa have `None` (should be empty?),
* gel fit has `cm^-1`,
* mono/poly gauss coil have `1/cm` (prolog has `cm^-1`), and
* polymer micelle has `1e15/cm^3` (prolog has `1e15cm^3)

Those missing from the prolog should be converted to the prolog equivalent or added to the prolog. For polymer micelle the `1e15cm^3` in the prolog should be corrected to `1e15/cm^3` for a density.

Sasview could use the rst prolog from sasmodels so the fixes don't need to be done in two places.

Maybe want a separate sasmodels ticket to update the units in the models.

rozyczko:
## Description

Unit coversion for GUI widgets is now based on sasmodels RST_PROLOG struct.
This struct is read in and parsed in GuiUtils to create a dictionary which is then used to convert ASCII units into HTML for display in the parameters table, report document and its serialized versions.

Several units, missing from rst_prolog were explicitly added and can/will be removed when sasmodels is updated.

Fixes #2561 

## How Has This Been Tested?

Local tests on win10.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Utilities/UnitTesting/GuiUtilsTest.py
+++ b/src/sas/qtgui/Utilities/UnitTesting/GuiUtilsTest.py
@@ -507,32 +507,6 @@ class GuiUtilsTest:
         s = "&#x212B; &#x221e;      &#177;"
         assert replaceHTMLwithASCII(s) == "Ang inf      +/-"
 
-    def testConvertUnitToUTF8(self):
-        ''' test unit string replacement'''
-        s = None
-        assert convertUnitToUTF8(s) is None
-
-        s = ""
-        assert convertUnitToUTF8(s) == s
-
-        s = "aaaa"
-        assert convertUnitToUTF8(s) == s
-
-        s = "1/A"
-        assert convertUnitToUTF8(s) == "Å<sup>-1</sup>"
-
-        s = "Ang"
-        assert convertUnitToUTF8(s) == "Å"
-
-        s = "1e-6/Ang^2"
-        assert convertUnitToUTF8(s) == "10<sup>-6</sup>/Å<sup>2</sup>"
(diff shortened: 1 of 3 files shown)


src/sas/qtgui/Utilities/UnitTesting/GuiUtilsTest.py
wpotrzebowski, line 1: Should we add test function for rst_to_html?

rozyczko: `testConvertUnitToHTML` covers some of the symbols but yes, we should probably test the method itself.

wpotrzebowski (approved):

PR #2572, opened 2023-08-01
Use a regex for version validity check rather than integer coercion

### LINKED ISSUE #2571: Malformed version bug - 6.0.0a1
**Describe the bug**
The commit https://github.com/SasView/sasview/commit/de6ed68880cff3650150ca575b8f4e9abf3c3fd7 is causing a MalformedFIle exception to be thrown when running from run.py. The first run of Sasview does not cause the error to be thrown, only subsequent runs. When sasview_version is read from the config.json file, the string is parsed and assumed to be three integers separated by `.`. The `0a1` as the last version throws a ValueError, causing SasView to not run. The config file loading is the only place this integer check is performed. The first time run.py is run, an old config file with the previous version is loaded and is overwritten by the program.

**To Reproduce**
1. Checkout `main`
2. Run `python run.py`
3. SasView starts!
4. Close SasView
5. Run `python run.py` again
6. Error thrown

**Expected behavior**
No error thrown when running run.py.

 - Version: main [6.0.0a1]

**Temporary Workaround**
Delete the config.json files in your `~/.sasview/` directory after sasview closes.

krzywon:
## Description

This fix replaces the int coercion with a regex when checking the version number loaded from the config.js file. This regex is used by `distutils.strictversion` which we use to check version numbers for our installer packages.

Fixes #2571 

## How Has This Been Tested?

Locally, I ran `python run.py` multiple times and verified SasView launched each time.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/system/config/config_meta.py
+++ b/src/sas/system/config/config_meta.py
@@ -1,3 +1,4 @@
+import re
 from typing import Dict, Any, List, Set
 import os
 import logging
@@ -150,8 +151,11 @@ class ConfigBase:
             raise MalformedFile("Malformed config file - no 'sasview_version' key")
 
         try:
-            parts = [int(s) for s in data["sasview_version"].split(".")]
-            if len(parts) != 3:
+            file_version = data["sasview_version"]
+            # Use the distutils strict version module regex to check if the version string is valid
+            #  ref: https://epydoc.sourceforge.net/stdlib/distutils.version.StrictVersion-class.html
+            matcher = re.compile(r'(?x)^(\d+)\.(\d+)(\.(\d+))?([ab](\d+))?$')
+            if not matcher.match(file_version):
                 raise Exception
 
         except Exception:
@@ -165,7 +169,6 @@ class ConfigBase:
 
 
         # Check major version
-        file_version = data["sasview_version"]
         file_major_version = file_version.split(".")[0]
         sasview_major_version = sas.system.version.__version__.split(".")[0]
 


llimeht:
+1 to supporting alpha/beta version strings

ehewins (approved):
Looks good! With these changes, SasView runs correctly once again.

krzywon:
> As main is broken, I'm going to merge now and leave the branch open in case @krzywon wants to implement my suggestion.

We use `distutils.version.strictversion` to check the version during our build process. Version strings that would pass in the `packaging` regex, e.g. `2.3.alpha4`, would cause our build process to fail. I used the `strictversion` regex to keep the versioning checks as consistent as possible among all our processes.

llimeht:
> We use `distutils.version.strictversion` to check the version during our build process. 

Do we still do that? I didn't find any direct uses with a `git grep` when I was looking yesterday. Perhaps it is buried in something from setuptools?

We need to migrate away from distutils in any case, so that's a limitation that solves itself when the build process is fixed.

From [PEP 632](https://peps.python.org/pep-0632/):
> Code that imports distutils will no longer work from Python 3.12.

We do have other uses of distutils in `docs/sphinx-docs/build_sphinx.py` and 
`docs/sphinx-docs/source/conf.py` that need to be fixed quite soon.

krzywon:
@llimeht, you are correct. We no longer use `strictversion` for version checks. This changed last October in a commit by you: https://github.com/SasView/sasview/commit/a4f91cf4c9fb821a452c6b0182fb064f061d38dd. I'll look into `packaging` to see if we can make use of it.

PR #2559, opened 2023-07-19
Avoid parenting mess by calling the widget directly

rozyczko:
## Description

When counting instances of plots, one needs to be careful with QtWidget parentage being updated.
The existing Manager->Plotter->PlotterWidget chain was working with Qt5 (and from what I remember was necessary to achieve correct MDI space location) but started failing with Qt6.
Getting rid of `Plotter` and instantiating `PlotterWidget` directly seems to be safer.

Fixes #2527

## How Has This Been Tested?

local testing on win10

## Review Checklist (please remove items if they don't apply):

- [ ] Code has been reviewed
- [ ] Functionality has been tested
- [ ] Windows installer (GH artifact) has been tested (installed and worked) 
- [ ] MacOSX installer (GH artifact) has been tested (installed and worked)

PR #2576, opened 2023-08-02
Local documentation generation

### LINKED ISSUE #706: make model help available for custom models (Trac #574)
Documentation for custom models does not currently appear in sasview help.

Using docutils, we can convert docstrings on the fly to html.

The basic mechanics are easy:

```#!python
from docutils import core
def convert(rst):
    # use replace_dollar from docs/_extensions/dollarmath.py
    # or configure mathjax to convert values between $$ to math
    rst = replace_dollar(rst)  
    parts = core.publish_parts(source=rst, writer_name='html')
    return parts['whole']
```

There will be some work required to get everything pretty, including proper style sheets and references to images.  Also, we won't want to produce the theory plots on the fly every time so they will need to be cached.

Migrated from http://trac.sasview.org/ticket/574
```json
{
    "status": "new",
    "changetime": "2018-09-09T19:06:41",
    "_ts": "2018-09-09 19:06:41.845014+00:00",
    "description": "Documentation for custom models does not currently appear in sasview help.\n\nUsing docutils, we can convert docstrings on the fly to html.\n\nThe basic mechanics are easy:\n\n{{{#!python\nfrom docutils import core\ndef convert(rst):\n    # use replace_dollar from docs/_extensions/dollarmath.py\n    # or configure mathjax to convert values between $$ to math\n    rst = replace_dollar(rst)  \n    parts = core.publish_parts(source=rst, writer_name='html')\n    return parts['whole']\n}}}\n\nThere will be some work required to get everything pretty, including proper style sheets and references to images.  Also, we won't want to produce the theory plots on the fly every time so they will need to be cached.\n",
    "reporter": "pkienzle",
    "cc": "",
    "resolution": "",
    "time": "2016-05-17T22:22:21",
    "component": "SasView",
    "summary": "make model help available for custom models",
    "priority": "minor",
    "keywords": "",
    "milestone": "SasView Next Release +1",
    "owner": "",
    "type": "enhancement"
}
```

tsole0:
## Description
**WORK IN PROGRESS**

Overhauls documentation viewing and editing system. When 'Help' is pressed anywhere in SasView, a window now appears that displays documentation using QT's WebViewEngine widget. This window also has buttons to edit the documentation, which will pull up the relevant RST (if user documentation) or Python file (if a model). When the 'Save' button is pressed, the relevant documentation is regenerated using a python to RST converter (if a model) and a somewhat skimmed-down version of Sphinx. A refresh of any open documentation windows is then triggered upon completion of the regeneration, displaying the user's changes. User-created documentation will also be generated and will display when 'Help' is pressed.

Sphinx will need to be included as a dependency for this update to work.

**TODO**
Fix various bugs, and, importantly, make sure the required files get ported over to the installed version and everything works on installed version.

Fixes #706

## How Has This Been Tested?

**Waiting for fixing of final bugs before describing**

## Review Checklist:

- [ ] Code has been reviewed
- [ ] User documentation is available and complete
- [ ] Developers documentation is available and complete - **NOTE:** developer documentation does not regenerate


src/sas/qtgui/Utilities/DocViewWidget.py
```diff
@@ -0,0 +1,216 @@
...
+            # Remove everything before /user/ (or \user\)
+            file = sub(r"^.*?(?=[\/\\]user)", "", path[0])
+
+            # index.html is the only rst file outside of /user/ folder-- set it manually
+            if "index.html" in file:
+                file = "/index.html"
+            self.editorWindow =  TabbedModelEditor(parent=self.parent,
+                                                   edit_only=True,
+                                                   load_file=file,
+                                                   model=False)
```
pkienzle, line 70: Separate the path manipulation logic from the editor startup. You can set `file=` in each branch of the condition, then use `load_file=file` after the condition to start the editor.

I don't think you need to escape the forward slashes in your regular expressions.

I'm not even sure you need regex. Once you strip the `file:///` prefix you can use `Path` to turn it into a path and test for "models" or "user" in path.parts. You can get the parts after user with:
```python
user_index = path.parts.index('user')
file = Path(*path.parts[user_index+1:])
```
If you do stick with regexs be sure to match `[\\/]user[\\/]` and `[\\/]models[\\/]` to reduce the chance collision with a filename such as `…/usermode.html`.

You use the same or similar path manipulations elsewhere, so create some helper functions to ease future maintenance.

src/sas/qtgui/Utilities/GridPanel.py
```diff
@@ -180,13 +181,17 @@ def addFitResults(self, results):
...
         """
         Open a local url in the default browser
         """
         url = "/user/qtgui/Perspectives/Fitting/fitting_help.html#batch-fit-mode"
-        GuiUtils.showHelp(url)
+        try:
+            self.helpWindow = DocViewWindow(parent=self.parent, source=url)
+            self.helpWindow.show()
+        except Exception as ex:
+            logging.warning("Cannot display help. %s" % ex)
```
pkienzle, line 194: Identical code appearing in multiple places. Can you do it once and share code?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -426,6 +464,38 @@ def updateFromEditor(self):
...
+            pass_in_file = os.path.basename(self.file_to_regenerate)
+        else:
+            pass_in_file = self.file_to_regenerate
+        command = [
+            sys.executable,
+            regen_docs_location,
+            pass_in_file,
+        ]
+        doc_regen_dir = os.path.dirname(regen_docs_location)
+        subprocess.run(command, cwd=doc_regen_dir) # cwd parameter tells subprocess to open from a specific directory
```
pkienzle, line 495: I didn't look in detail at the code but it seems familiar. Can you collect all your interactions with sphinx in one module?

rozyczko:
Just had a look at this feature in my (Windows 10) dev build.
Works fine, but a new help window is popping up every time I hit the Help button. This is especially concerning since the initial generation of HTML for display takes some non-negligible time and users have no visual cue that the process has started.
This means they will likely keep clicking Help and end up with a number of windows.

Please consider making a single instance of the help viewer and just update the content on new request.

docs/sphinx-docs/build_sphinx.py
```diff
@@ -260,7 +261,9 @@ def build():
 
     print("=== Copy HTML Docs to Build Directory ===")
     html = joinpath(SPHINX_BUILD, "html")
+    copy_tree(SPHINX_SOURCE, SASVIEW_DOC_TARGET)
```
krzywon, line 264: This is a duplicate of line 250. Is there a reason for doing this a 2nd time? If there is, documenting why would be helpful.

docs/sphinx-docs/doc-regen/makedocumentation.py
```diff
@@ -0,0 +1,132 @@
...
+import os
+import sys
+from os.path import join, abspath, dirname, basename
+import subprocess
+
+FILE_ABS_SOURCE = abspath(dirname(__file__))
+MAIN_PY_SRC = FILE_ABS_SOURCE + "/../source-temp/user/models/src/"
+if os.path.exists(MAIN_PY_SRC):
+    ABSOLUTE_TARGET_MAIN = abspath(join(dirname(__file__), MAIN_PY_SRC))
+    PLUGIN_PY_SRC = abspath(dirname(__file__)) + "/../../../../.sasview/plugin_models/"
```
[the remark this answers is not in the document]
krzywon: Many of the pathing issues could be overcome by copying all documentation files into the user directory on installation. Finding the user directory is trivial (see `find_plugin_models()` as an example).
```
Example structure:
~/.sasview/
    docs/
        original/  # The unedited files in case the user wants to blow away their changes (future work...)
        modified/ # Any modified files - load from here, if the file exists, otherwise grab the original.
```

docs/sphinx-docs/doc-regen/makedocumentation.py
```diff
@@ -0,0 +1,132 @@
...
+    Runs regenmodel.py/regentoc.py (specified in parameter regen_py) with all found PY_FILES
+    """
+    REGENMODEL = abspath(dirname(__file__)) + "/" + regen_py
+    # Initialize command to be executed
+    command = [
+        sys.executable,
+        REGENMODEL,
+    ]
+    # Append each filepath to command individually if passed in many files
+    if type(filepath) == list:
```
krzywon, line 53: `isinstance(filepath, list)` is usually preferred over type checking. https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance

docs/sphinx-docs/source/conf.py
```diff
@@ -37,7 +37,13 @@
               'sphinx.ext.mathjax',
               #'mathjax',  # replacement mathjax that allows a list of paths
               'dollarmath',
-              'sphinx.ext.viewcode']
+              #'sphinx.ext.viewcode',
+              ]
+
+import os
```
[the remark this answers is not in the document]
krzywon: This is also a duplicate import.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -352,7 +353,11 @@ def showHelp(self, url):
         """
         Open a local url in the default browser
         """
-        GuiUtils.showHelp(url)
+        try:
+            # Help window shows itself
+            self.helpWindow = DocViewWindow(parent=self, source=url)
```
krzywon, line 358: The window should be created in the GuiManager init so a new window is not generated every time the method is called. This method should only need to update the url and show the window if it's hidden. Closing the window should then hide it instead of destroying it.

src/sas/qtgui/Utilities/GuiUtils.py
```diff
@@ -530,18 +539,8 @@ def openLink(url):
...
-    #WP: Added to handle OSX bundle docs
-    if os.path.isdir(location) == False:
-        sas_path = os.path.abspath(os.path.dirname(sys.argv[0]))
-        location = sas_path+"/"+location
-    try:
-        webbrowser.open('file://' + os.path.realpath(location))
-    except webbrowser.Error as ex:
-        logging.warning("Cannot display help. %s" % ex)
+    pass
+# showHelp() IS TO BE DELETED IN NEXT COMMIT
```
krzywon, line 543: Removing an existing method can be dangerous in case it is used in places you haven't found. Adding a debugging message showing the method is deprecated and then a call to the new method would be appropriate for now.

docs/sphinx-docs/doc-regen/makedocumentation.py
```diff
@@ -0,0 +1,132 @@
...
+        PY_FILES = [join(directory, string) for string in files if not string.startswith("_") and string.endswith(".py")]
+        return PY_FILES
+
+def get_main_docs():
+    """
+Generates string of .py files to be passed into compiling functions
+Future reference: move to main() function?
+    """
+    # The order in which these are added is important. if ABSOLUTE_TARGET_PLUGINS goes first, then we're not compiling the .py file stored in .sasview/plugin_models
+    TARGETS = get_py(ABSOLUTE_TARGET_MAIN) + get_py(ABSOLUTE_TARGET_PLUGINS)
```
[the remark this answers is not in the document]
tsole0: I believe there are some deeper concerns here that would prevent the implementation of either of these solutions for the time being. The same collision resolution used by sasview should not be applied here because we're dealing with filepath names to pass to Sphinx and not the names of the models in SasView themselves. Filepaths have to be unique, so this shouldn't be a concern. The reason we need to have collision resolution here anyways is because `regenmodel.py` copies user created files into `source-temp`, the folder that stores the copies of models we use for editing documentation in this branch, because apparently Sphinx can't link to absolute paths. We search in both directories because the user may have added new models to the user model directory before `regenmodel.py` has run, and we can't run `regenmodel.py` first because it would take too long and we need to pass in only one file. A potential solution *could* be to use Jeff's suggested approach of putting `source-temp` inside the user model directory itself, but I think that's too big of a change for the time being and would involve some more research into whether or not Jeff's idea is compatible with Sphinx linking to source files.

docs/sphinx-docs/doc-regen/makedocumentation.py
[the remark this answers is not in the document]
tsole0: Implemented Paul's idea. I think Jeff's idea is a good idea for refactoring the code but would take a bit of work at the moment.

src/sas/qtgui/Utilities/GridPanel.py
butlerpd: `DocViewWIndow` is the shared code and replaces the old `GuiUtils.showHelp(url)`.  We have added a try except which could perhaps move to DocViewWindow? Otherwise I'm not sure I understand the question?

src/sas/qtgui/MainWindow/GuiManager.py
butlerpd: Would this not prevent one from having more than one help window open at once?

If so, that would be very bad IMO. I often open several (all browser windows currently) to different pages so I can easily go back and forth. It would be monumentally irritating if I had to close and open and close and open back and forth. The only thing we want to avoid is opening the **same** window multiple times.

If it does not prevent the above use case then this sounds like a reasonable idea.

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -1132,7 +1132,7 @@ def addDataPlot2D(self, plot_set, item):
         """
         Create a new 2D plot and add it to the workspace
         """
-        plot2D = Plotter2D(self)
+        plot2D = Plotter2DWidget(parent=self, manager=self)
```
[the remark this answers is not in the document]
butlerpd: I believe this is a result of some work that @rozyczko and @tsole0 both did independently to `DataExplorer.py` to bypass Plotter2D and call Plotter2DWidget directly to "avoid parenting mess" (closed PR #2559 ) .  I believe @tsole0 merged @rozyczko work in favor of his own. It is possible that he merged that branch into his rather than wait for it to get merged to main in order to keep his work moving? 

However, he made those changed himself originally because indeed they were required for his work so the short answer: it is in effectively in scope I believe (though this PR could have benefited from several smaller PRs )

Shorter answer .... it is no longer part of the PR changes now that main has been merged back in

docs/sphinx-docs/doc-regen/makedocumentation.py
```diff
@@ -0,0 +1,132 @@
...
+        SPHINX_SOURCE,
+        HTML_TARGET,
+        single_rst,
+    ]
+    try:
+        # Try removing empty arguments
+        command.remove("")
+    except:
+        pass
+    subprocess.check_call(command)
```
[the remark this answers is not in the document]
tsole0: `makedocumentation.py` should always be running in a subprocess in which the correct cwd is set, but I added appropriate code to ensure that this is always the case even if it somehow gets called and the cwd is not set correctly.

wpotrzebowski:
It has to be bundled with installer.

krzywon:
This needs to work in the installed version. To do this, the following changes are required:
1. Bundle sphinx and any other required dependencies in the release.
2. Save changes into a directory the user is guaranteed to have write access to.
3. Stretch goal: Determine a method for the user to contribute their changes to the documentation. This could be as simple as emailing a diff file to sasview help that we would have to handle, or as complicated as creating a pull request and/or issue directly from the App.

wpotrzebowski:
It seems that this PR is failing at the notarization step. I am looking into more details

wpotrzebowski:
It seems that some of Qt folders/modules are not signed. I am not sure but we may need to include more folder in fix_qt_folders script

wpotrzebowski:
Here is the log:
[notarization_log.txt](https://github.com/SasView/sasview/files/13943073/notarization_log.txt)

PR #2580, opened 2023-08-05
Adjusts scale and angular range in 1D plots from WedgeSlicer

butlerpd:
## Description

The 1D plots created by the new `WedgeSlicer` now returns:
* Properly converted radians to degrees for the x scale
* The default scale for both x (phi) and y (intensity) are linear.
* The Angular range is pi to pi (with 0 pointing along the positive x axis) as per the normal python and sasview convention

NOTE: this was tested and works well with uncommitted changes to manipulation.py in sasdata needed to fix the actual averaging. This can still be tested with sasdata/master I think as long as one does not allow the wedge to straddle the negative x axis. This should not affect merging since that problem exists in the current main of sasview.

NOTE2:  The caveat to the angular range is that when the wedge straddles the negative x axis the phi max will be greater than 2pi but the data will be in proper consecutive order. This is mostly coming from sasdata. While this may seem a bit odd it is the only rational way to display such data.

Fixes #2578

## How Has This Been Tested?

run on windows 10 developer and checked the wedge slicer (moving the different interactors around), and compared to the annular and sector interactors

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
@@ -138,6 +138,12 @@ class WedgeInteractor(BaseInteractor, SlicerModel):
 
         :param new_sector: slicer used for directional averaging in Q or Phi
         :param nbins: the number of point plotted when averaging
+        :TODO - Unlike other slicers, the two sector types are sufficiently
+        different that this method contains three instances of If (check class name) do x.
+        The point of post_data vs _post_data I think was to avoid this kind of thing and
+        suggests that in this case we may need a new method in the WedgeInteracgtorPhi
+        and WedgeInteracgtorQ to handle these specifics. Probably by creating the 1D plot
+        object in those top level classes along with the specifc attributes.
         """
         # Data to average
         data = self.data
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
```diff
@@ -175,7 +176,10 @@ def _post_data(self, new_sector=None, nbins=None):
             dxw = sector.dxw
         else:
             dxw = None
-        new_plot = Data1D(x=sector.x, y=sector.y, dy=sector.dy, dx=sector.dx)
+        # And here subtract pi when getting angular data back from sector averaging in
+        # manipulations to get back in the -pi,pi range. Also convert from radians to
+        # degrees for nicer display.
+        new_plot = Data1D(x=(sector.x - np.pi) * 180 / np.pi, y=sector.y, dy=sector.dy, dx=sector.dx)
```
ehewins, line 182: At the moment, this angular shift is happening even when plotting against Q. I think this line should go back to having `x=sector.x`, and then the following could be added after line 169:
```python
        if self.averager.__name__ == 'SectorPhi':
            sector.x = (sector.x - np.pi) * 180 / np.pi
```

butlerpd: Good catch @ehewins

butlerpd: In principle if class.__name__ ==name: is not a good construct I've been told. However there are 2 other uses of it in this one method so just going with it for this PR but left a "todo" comment. I suspect the right answer would be a create_plot method in post_data that handles those 3 specifics?

butlerpd:
As pointed out by @ehewins the slicers seem to be rife with the use of  `self.classname.__name__` which maybe needs to be looked at more carefully. In the meantime agreed to merge this.

Issue #2416, opened 2023-01-09
numpy 1.24 (in)compatibility


llimeht:
numpy 1.24 has now been released and so we will at some stage need compatibility with it. So far, I've only spotted 4 tiny things that need fixing, related to removal of deprecated functions. In particular, numpy 1.24 has finally removed the deprecated `np.float` definition. This should either be replaced by `float` if machine-variation is permitted (i.e. 32 bit or 64 bit depending on the architecture), or `np.float64` if a specific size is needed.

See: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

This is fortunately quite easy to fix in a backwards compatible way:

https://sources.debian.org/src/sasview/sid/debian/patches/numpy-1.24.patch/

The failures in `p_invertor.py` are caught by the test suite; I grepped the source looking for other instances which turned up the instance in GuiUtils.py.

**SasView version (please complete the following information):**
 - Version: 5.0.5 and main (and any other release)

**Operating system (please complete the following information):**
 - OS: all (tested on linux)

PR #2582, opened 2023-08-08
Added 3.11, removed 3.8

rozyczko:
## Description

Removed 3.8 and added 3.10, 3.11 to build/test matrix
Removed pinned versioning which caused issues with 3.11

Local win10 dev build and installer

- [ ] Code has been reviewed
- [X] Windows installer (GH artifact) has been tested (installed and worked) 


llimeht:
Would `np.float64` be better to be explicit about the precision? Probably makes no real difference unless someone is marshalling this through C somewhere.
https://sources.debian.org/src/sasview/5.0.5-5/debian/patches/numpy-1.24.patch/

This work should also close #2416.

rozyczko:
> Would `np.float64` be better to be explicit about the precision? Probably makes no real difference unless someone is marshalling this through C somewhere. https://sources.debian.org/src/sasview/5.0.5-5/debian/patches/numpy-1.24.patch/
> 
> This work should also close #2416.

Shouldn't make a difference, since the built-in float is dp anyway on non-32 bit arch.
It might be useful to have it spelled out explicitly, maybe, so let's do it then.

wpotrzebowski:
I was curious if OSX installer works on 3.11 and it seems to work fine!

butlerpd:
Not sure we need all 3 python versions. Maybe focus on 3.11 and try to make sure everything works there.  Needs lots of testing.  Aim to merge prior to code camp to ensure fulsome testing.

rozyczko:
> Not sure we need all 3 python versions. Maybe focus on 3.11 and try to make sure everything works there. Needs lots of testing. Aim to merge prior to code camp to ensure fulsome testing.

Good catch. We only need one python version for the release (3.11) but we want testing to be done on all supported versions (3.8, 3.9, 3.10, 3.11)

wpotrzebowski:
The installer starts on Mac, however, it crashes whenever I try to access any folder

`Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
Fatal Python error: PyGILState_Release: thread state 0x1026663a8 must be current when releasing
Python runtime state: initialized
`

wpotrzebowski (approved):

PR #2585, opened 2023-08-11
Remove Unused Dependency: h5py

gdrosos:
## Description

This pull request removes the unused dependency `h5py` from the `requirements.txt` and  `setup.py` configuration files. Moreover, the`log.py` is updated to ensure there are no references to `h5py`. The removal of this dependency is a finding from ongoing research aimed at identifying and eliminating code bloat within software projects.

## Rationale

Initially, the project incorporated the `h5py` library. However, over time, the codebase evolved, and currently the library is unused. Still, the package continues to be listed in the project's dependency files. Removing this unused dependency reduces the overall footprint of the application, mitigating potential security risks, and simplifying the dependency management process.

- [x] Code has been reviewed
- [ ] MacOSX installer (GH artifact) has been tested (installed) 

### DIFF
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -6,6 +6,7 @@ pytest_qt
 pytest-mock
 unittest-xml-reporting
 tinycc 
+h5py
 sphinx
 pyparsing 
 html5lib 
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -6,7 +6,6 @@ pytest_qt
 pytest-mock
 unittest-xml-reporting
 tinycc 
-h5py 
 sphinx
 pyparsing 
 html5lib 
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,7 @@ if build_qt:
 # Required packages
 required = [
     'bumps>=0.7.5.9', 'periodictable>=1.5.0', 'pyparsing>=2.0.0',
-    'lxml', 'h5py',
+    'lxml',
 ]
 
 if os.name == 'nt':
--- a/src/sas/system/log.py
+++ b/src/sas/system/log.py
@@ -17,7 +17,6 @@ TRACED_PACKAGES = ('sas', 'sasmodels', 'sasdata', 'bumps', 'periodictable')
 IGNORED_PACKAGES = {
     'matplotlib': 'ERROR',
     'numba': 'WARN',
-    'h5py': 'ERROR',
     'ipykernel': 'CRITICAL',
 }
 
--- a/src/sas/system/log.py
+++ b/src/sas/system/log.py
@@ -17,6 +17,7 @@ TRACED_PACKAGES = ('sas', 'sasmodels', 'sasdata', 'bumps', 'periodictable')
 IGNORED_PACKAGES = {
     'matplotlib': 'ERROR',
     'numba': 'WARN',
+    'h5py': 'ERROR',
     'ipykernel': 'CRITICAL',
 }
 


gdrosos:
Thanks @lucas-wilkins for pointing out the concerns with `sasdata` and the logging related to `h5py`.

Given your insights, I propose a phased approach. In this PR, we can proceed with removing `h5py` from `requirements.txt `and `setup.py` since its direct use in the main project is not present.
 However, to err on the side of caution, we can retain the logging references to `h5py` in `log.py.`

Once this PR is merged, we could open a separate issue to investigate the logging aspect further.

pkienzle:
Logging control belongs with the application rather than the library so leave those lines alone. For example, it would be confusing if you set the`numba` logging level to debug in `sasview` only to find that it is shut off again when `bumps` is imported.

gonzalezma:
What do you mean by unused dependency? I assume that h5py is needed to read/write NeXus files, so in this case it is not an unused dependency. Maybe that the conclusion is based on the analysis of the code base under sasview, while the requirements.txt file apply to the full SasView app (and therefore the analysis should include the code under sascalc and sasmodels)?
Sorry if I'm misunderstanding something.

gdrosos:
Thanks @gonzalezma for pointing that out. 
By "unused dependency," I was referring to any PyPI library not directly utilized in the `sasview` source code. I noted that `h5py` is indeed a dependency in [sasdata](https://github.com/SasView/sasdata/blob/b5a17f21b5122477554616e9ae886d7a6a90654f/requirements.txt#L2)  and assumed its presence there would negate the need for its inclusion in `sasview`. My apologies for any confusion.

wpotrzebowski:
It seems that tests require h5py. I guess they can be rewritten to use sasdata?

butlerpd:
Cannot merge this now - need to see if we can move tests to sadata and set up sasdata for proper ci if not already doing so.

krzywon:
Suggestion: Remove `h5py` and `lxml` from `setup.py` as both are only used by the `sasdata` package, but keep both in `requirements.txt`. With our current CI process, the Python environment is built using source from the `sasdata`, `sasmodels`, and `bumps` repositories and from the `requirements.txt` file, so we need that to be as complete as possible.

gdrosos:
Hello @krzywon, I reverted the changes in the `requirements.txt` accordingly, and now the PR proposes the removal of the `h5py` only from the `setup.py` file. Should I also remove `lxml` on the same manner? Because it seems that this library is directly invoked in the source code in this repo e.g. https://github.com/SasView/sasview/blob/55b1e9f6db58e33729f2a93b7dd1d8bf255b46f7/src/sas/sascalc/fit/pagestate.py#L26

krzywon (approved):
Looks good at this point

gdrosos:
Thank you @krzywon, please let me know if there is anything else that needs to be changed.

gdrosos:
I see that there is one failing check in macos, but given that it also occurs in #2597 qith th exact same error, It seems to be unrelated with the proposed changes on this PR.

krzywon:
> I see that there is one failing check in macos, but given that it also occurs in #2597 with the exact same error, It seems to be unrelated with the proposed changes on this PR.

This is a systemic issue related to the software signature we use. It should be resolved shortly.

As far as this PR goes, we often wait until our regular meetings before merging so we can discuss any outstanding issues that might not be obvious. Our next meeting is Tuesday of next week, so expect to either see this merged or new comments then.

wpotrzebowski:
I've been trying to resolve the signing issue. Unfortunately without success yet.

PR #2594, opened 2023-08-23
2593 wedge slicer fold negative q workaround

### LINKED ISSUE #2593: Neither fold=True nor fold=False is appropriate for WedgeSlicer
** Describe the issue **
The wedge slicer (specifically `WedgeInteractorQ`) is build on the same foundations as the sector slicer. With this comes "fold" functionality, which when enabled averages the data in the region of interest with the data on the opposite side of the origin. When fold is disabled, data on the opposite side of the origin to the main ROI is shown plotted against negative Q values. This is the current state of wedge slicer, but I had been unaware until recently that these negative Q values were present as the default graph scale is log-log, which obscures the negative Q values.
Neither of these fold behaviours are desired for the wedge slicer.

** Expected behaviour **
Only data enclosed within the main ROI should be plotted. 

** To reproduce **
1. In Sasview, Compute/Plot any 2D data from `example_data/2d_data`, using any model.
2. Right click the plot and select 'Wedge Averaging in Q'.
3. Change the scale of the output graph to linear, and observe that negative Q values are also plotted, corresponding to data outside the ROI.

** Additional context **
The proper fix (in my opinion) would be editing the `_Sector` class in `sasdata/sasdata/data_util/manipulations.py` to provide a third alternative to `self.fold = True` or `self.fold = False`, one which only processes the data within the specified region of interest. This could make the current version of `_Sector` even more confusing however, so I would rather work this into the `manipulations.py` rewrite and use a simpler fix in the meantime. This would involve adding some code to `WedgeSlicer.py` which looks for data associated with negative Q values in the `Data1D` object returned by `_Sector`, then discarding it before plotting.
If/When a method for handling this is added on the `manipulations.py` side, the commit fixing this issue can be kept or reverted with no ill effect.

ehewins:
## Description

Added an `else` clause to a pre-existing `if self.averager.__name__ == 'SectorPhi':` check, which removes any data points associated with negative Q values from the data returned by `SectorQ`.

Fixes #2593

## How Has This Been Tested?

The steps from the associated issue (detailing how to reproduce the bug) were followed, and the expected behaviour was observed with this fix present.

- [ ] Code has been reviewed
- [X] Developers documentation is available and complete (if required)

### DIFF
--- a/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
@@ -187,6 +187,18 @@ class WedgeInteractor(BaseInteractor, SlicerModel):
             # phi in manipulations to get back in the -pi,pi range. Also convert from
             # radians to degrees for nicer display.
             sector.x = (sector.x - np.pi) * 180 / np.pi
+        else:
+            # When using SectorQ, fold = False results in data with negative Q
+            # values also being plotted. This data is from the opposite side of
+            # the origin to the main ROI, and is unwanted. Here we discard it.
+            # If at some later point this is handled on the sasdata side, then
+            # this code can be deleted.
+            pos_q = [sector.x >= 0]
+            sector.x = sector.x[pos_q]
+            sector.y = sector.y[pos_q]
+            sector.dy = sector.dy[pos_q]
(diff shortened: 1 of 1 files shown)


wpotrzebowski:
As decided at the call on Sep 5th this part will be rewritten anyway and therefore can be closed

PR #2600, opened 2023-09-09
2577 orientation viewer doesnt work from nigthly build on mac

lucas-wilkins:
Fixed the orientation viewer parameters to align with that used by sasmodels.


wpotrzebowski (approved):
The code looks good and it works as expected (tested on Mac).

PR #2608, opened 2023-09-16
What's new dialog

lucas-wilkins:
What's New: Tip of the Day style dialog.

Currently a draft as depends strongly on config, and on files that need to be packaged correctly. There are some placeholder files included that should not be released.

Definitely need to check installers for this one!

Expected behaviour is that it will show messages for V5.1.7 and 7.0 (and not for 5.0.6).


src/sas/system/config/config.py
```diff
@@ -208,6 +208,9 @@ def __init__(self):
         # Default fitting optimizer
         self.FITTING_DEFAULT_OPTIMIZER = 'lm'
 
+        # What's New variables
```
wpotrzebowski, line 211: It probably doesn't matter but did you mean 6.0.0 here?

src/sas/qtgui/MainWindow/UnitTesting/MainWindowTest.py
```diff
@@ -13,8 +13,14 @@
 from sas.qtgui.Perspectives.Fitting import FittingPerspective
 from sas.qtgui.Utilities.HidableDialog import HidableDialog, ShowAgainResult
 
+from sas.system import config
 class MainWindowTest:
     """Test the Main Window GUI"""
+
+    def __init__(self):
+        config.override_with_defaults() # Disable saving of test file
+        config.LAST_WHATS_NEW_HIDDEN_VERSION = "999.999.999" # Give a very large version number
```
wpotrzebowski, line 22: Why do you need to hardcode this value?

wpotrzebowski (commented):
Code looks good. I will test functionality

wpotrzebowski:
I've tried to run it from the installer and it is failing with 
13:33:14 - sas.qtgui.MainWindow.MainWindow - ERROR - Application failed with: No module named 'sas.qtgui.Utilities.WhatsNew.messages'
13:33:14 - sas.qtgui.Perspectives.Fitting.FittingWidget - ERROR - Traceback (most recent call last):

rozyczko:
Please keep in mind that the additional (non-py) files need to be manually added as data to the pyinstaller script. This applies to the HTML files you created.

wpotrzebowski (approved):

PR #2646, opened 2023-09-22
Moving OSX signing to nigthly build (ver 2)

wpotrzebowski:
## Description

Some additional fixes to GH actions as the previous version was failing on nightly build

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -53,12 +53,12 @@ jobs:
           security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k DloaAcYP build.keychain
 
           cd installers/dist
-          python ../../build_tools/fix_qt_folder_names_for_codesign.py SasView5.app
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -40,7 +40,6 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Sign executable and create dmg (OSX)
-        if: ${{ matrix.installer && startsWith(matrix.os, 'macos') }}
```
rozyczko, line 43: won't removal of this line cause attempts of signing non-OSX binaries and therefore fail?

rozyczko (approved):
the code looks safe, one minor issue brought up but seems a non-issue after closer inspection. Good to merge

.github/workflows/nightly-build.yml
wpotrzebowski: This GH should run only on Mac. This is why it can be removed (I believe).

Issue #2438, opened 2023-02-15
Polydispersity settings not updating correctly


lucas-wilkins:
**Describe the bug**
Changing polydispersity parameters 

**To Reproduce**
Make a P*S model,
Switch between polydispersity settings and change values
Things do not update consistently
Not diagnosed fully yet, seems to be intermittent at this point

smk78:
Testing a recent 6.0.0a1 Github actions build on W10/x64.

Indeed, something bad has happened with Polydispersity!

![image](https://github.com/SasView/sasview/assets/10938679/33fd237a-4f66-470a-8ea9-8440c4f27fa4)

The function drop down is now appearing under Filename. The Filename box should only activate when an Array distribution is selected so that you can input your user-defined distribution.

Polydispersity still seems to function - and that is the only reason I'm not going to label this as a blocker for 6.0.0 - but I consider it highly undesireable that you can select a distribution other than a Gaussian and the GUI still says Gaussian.

This needs a prompt fix.

smk78:
Changing function the following are thrown:
```
ERROR: Traceback (most recent call last): File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3526, in File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3580, in onPolyComboIndexChange AttributeError: 'NoneType' object has no attribute 'blockSignals'
```
```
2024-03-27 15:20:22,929 : ERROR : sas.qtgui.Perspectives.Fitting.FittingWidget (FittingWidget.py:177) :: Traceback (most recent call last):
  File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3526, in <lambda>
  File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3580, in onPolyComboIndexChange
AttributeError: 'NoneType' object has no attribute 'blockSignals'
```

krzywon:
> The function drop down is now appearing under Filename. The Filename box should only activate when an Array distribution is selected so that you can input your user-defined distribution.
> 
> Polydispersity still seems to function - and that is the only reason I'm not going to label this as a blocker for 6.0.0 - but I consider it highly undesireable that you can select a distribution other than a Gaussian and

I can reliably recreate the issue seen by @smk78, not only in v6.0.0a, but also v5.0.6.

Steps to repeat:
1. Load a data set and send it to fitting.
2. Select a model, and PD, but do not select a structure factor.
[...]

Issue #1817, opened 2021-04-15
Name change causing mismatch between plots and data explorer


krzywon:
Changing the name of a data set can create a mismatch between the name displayed in the data explorer and the name shown in plots.

See [1772](https://github.com/SasView/sasview/issues/1772#issuecomment-768257509)

PR #2647, opened 2023-09-22
Various Multiplicity Model Fixes

### LINKED ISSUE #607: Multiplicity models do not appear in drop-down in Custom Model > Sum|Multi (Trac #473)
Reported by Saeed Mohan (saeed.mohan@ipleiria.pt):

I was attempting to add in my own custom model to Sasview, which is based on the !UnifiedPowerRg model. I thought the easiest thing to do would be to use the Sum Multi|(p1, p2) function under Fitting > Edit Custom Model, however I am not able to find this model in the options list.

Checked by SMK. He is correct.

What is more, the same bug exists in v3.0.0 and v2.2.1!!!

PDB checked and it appears that no multiplicity models appear in the dropdown (eg. RPA10, onion, etc).  This is probably working as designed since I suspect the interface was not obvious for a multiplicity model?  Interestingly there is also a very strange behavior that the shape independent -> Absolute_Power_Law seems to have been renamed in this dropdown to !PowerLawAbsModel ... perhaps due to the space in the name? 

Migrated from http://trac.sasview.org/ticket/473
```json
{
    "status": "new",
    "changetime": "2019-03-02T00:16:28",
    "_ts": "2019-03-02 00:16:28.434007+00:00",
    "description": "Reported by Saeed Mohan (saeed.mohan@ipleiria.pt):\n\nI was attempting to add in my own custom model to Sasview, which is based on the !UnifiedPowerRg model. I thought the easiest thing to do would be to use the Sum Multi|(p1, p2) function under Fitting > Edit Custom Model, however I am not able to find this model in the options list.\n\nChecked by SMK. He is correct.\n\nWhat is more, the same bug exists in v3.0.0 and v2.2.1!!!\n\nPDB checked and it appears that no multiplicity models appear in the dropdown (eg. RPA10, onion, etc).  This is probably working as designed since I suspect the interface was not obvious for a multiplicity model?  Interestingly there is also a very strange behavior that the shape independent -> Absolute_Power_Law seems to have been renamed in this dropdown to !PowerLawAbsModel ... perhaps due to the space in the name? ",
    "reporter": "smk78",
    "cc": "",
    "resolution": "",
    "workpackage": "SasModels Infrastructure",
    "time": "2015-11-10T12:00:02",
    "component": "SasView",
    "summary": "Multiplicity models do not appear in drop-down in Custom Model > Sum|Multi",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView 5.1.0",
    "owner": "",
    "type": "defect"
}
```

### LINKED ISSUE #1089: sum and product models won't work with multiplicity (Trac #1022)
See discussion in ticket #1075 and pull request [https://github.com/SasView/sasmodels/pull/53 #281].

This will also be a problem in mixture.py, if you are combining two models, one of which is a multiplicity model (combining two multiplicity models should raise an error). In that case, the control parameter will need to be tagged with the appropriate 'A_', 'B_', ... prefix before assigning, and the hidden() function, if present, will need to be wrapped so that the parameter names in the returned set are also prefixed.

Other parameters in model_info.ModelInfo should not be a problem. profile, profile_axes (for spherical shell models) are probably going to be drawn from the form_factor, and so don't need to be included. single and opencl are used internally to select a calculator, and can be undefined for this model (this model should present as a python model).

Migrated from http://trac.sasview.org/ticket/1022
```json
{
    "status": "new",
    "changetime": "2019-03-02T00:15:16",
    "_ts": "2019-03-02 00:15:16.703945+00:00",
    "description": "See discussion in ticket #1006 and pull request [https://github.com/SasView/sasmodels/pull/53 #53].\n\n\nThis will also be a problem in mixture.py, if you are combining two models, one of which is a multiplicity model (combining two multiplicity models should raise an error). In that case, the control parameter will need to be tagged with the appropriate 'A_', 'B_', ... prefix before assigning, and the hidden() function, if present, will need to be wrapped so that the parameter names in the returned set are also prefixed.\n\nOther parameters in model_info.ModelInfo should not be a problem. profile, profile_axes (for spherical shell models) are probably going to be drawn from the form_factor, and so don't need to be included. single and opencl are used internally to select a calculator, and can be undefined for this model (this model should present as a python model).\n",
    "reporter": "pkienzle",
    "cc": "",
    "resolution": "",
    "time": "2017-10-27T16:33:57",
    "component": "SasView",
    "summary": "sum and product models won't work with multiplicity",
    "priority": "major",
    "keywords": "",
    "milestone": "SasView 4.3.0",
    "owner": "",
    "type": "defect"
}
```

### LINKED ISSUE #1106: multiplicity not included in copy/paste parameters (Trac #1041)
When copying parameters for a multiplicity model such as spherical_sld, the multiplicity parameter is not include.  On paste, the multiplicity of the model in the fit page is not set.

Migrated from http://trac.sasview.org/ticket/1041
```json
{
    "status": "new",
    "changetime": "2017-12-01T13:02:18",
    "_ts": "2017-12-01 13:02:18.895037+00:00",
    "description": "When copying parameters for a multiplicity model such as spherical_sld, the multiplicity parameter is not include.  On paste, the multiplicity of the model in the fit page is not set.",
    "reporter": "pkienzle",
    "cc": "",
    "resolution": "",
    "time": "2017-12-01T13:02:18",
    "component": "SasView",
    "summary": "multiplicity not included in copy/paste parameters",
    "priority": "minor",
    "keywords": "",
    "milestone": "SasView 5.0.0",
    "owner": "",
    "type": "defect"
}
```

### LINKED ISSUE #1816: v5 errors when using S(Q) with a multiplicity model such as onion
after adding an S(Q) to say onion, when I change the number of shells I get:

- ERROR: Traceback (most recent call last):
  File "C:\sasview5\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 2625, in onMainParamsChange
    self.kernel_module.setParam(parameter_name, value)
  File "C:\sasview5\sasmodels\sasmodels\sasview_model.py", line 513, in setParam
    raise ValueError("Model does not contain parameter %s" % name)
ValueError: Model does not contain parameter n_shells

I suspect that once S(Q) is included the F(Q) is then also not being computed correctly, as the effective radius in S(Q) is wrong, only including the first shell, so suspect n_shells is being passed as zero.

[also can someone please add a new milestone, for 5.0.5 ? ]

### LINKED ISSUE #1897: Don't show SLD Profile button for models without a profile
Using Release 5.0.4, clicking the Show SLD Profile button in the unified_power_rg model returns
```
ERROR: SLD profile calculation failed.
```

It works in the onion model...

### LINKED ISSUE #2369: Copy Params to Format/Save Params to File throws error for Multiplicity models
**Describe the bug**
Whilst using the Unified_Power_rg model, User JamesD observed that trying to 'Copy params to Format' or 'Save Params to File' throws errors and does not work. The bug is easily to replicate.
```
15:23:15 - ERROR: Traceback (most recent call last):
  File "sas\qtgui\Perspectives\Fitting\FittingPerspective.py", line 120, in onExcelCopy
  File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3891, in onCopyToClipboard
  File "sas\qtgui\Perspectives\Fitting\FittingUtilities.py", line 806, in formatParametersExcel
IndexError: list index out of range
15:23:26 - ERROR: Traceback (most recent call last):
  File "sas\qtgui\Perspectives\Fitting\FittingPerspective.py", line 214, in onParamSave
  File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3896, in onCopyToClipboard
  File "sas\qtgui\Perspectives\Fitting\FittingUtilities.py", line 806, in formatParametersExcel
IndexError: list index out of range
```

**To Reproduce**
Run SasView and load up the Shape Independent > unified_power_Rg model
Compute the model
Attempt to Edit > Copy Params to Format/Save Params to File

**Expected behavior**
The model parameters should be exported. Though slightly different, this functionality works in 4.2.2

**Workaround**
Edit > Copy Params *does* work, but the format is not intended for the same use that Copy Params to Format/Save Params to File are

**SasView version**
Reproduced in 5.0.5, 5.0.4 & 5.0.3 so appears to be a general bug in 5.x

x64/W10

### LINKED ISSUE #2563: No polydispersity distributions for `core_multi_shell` model
**Describe the bug**
The Gaussian distribution is the only distribution available for `Distribution of thickness` in the `core_multi_shell` model. 

**To Reproduce**
1. In the fit panel, choose `sphere`, `core_multi_shell`,
2. Check `Option > Polydispersity`, then click on the `Polydispersity` tab,
3. See error: no drop-down menus for thicknesses.

**Expected behavior**
A drop-down menu in `Function` for each `thickness`.

**Screenshots**
Screenshot from 2023-07-24 13-02-04

**Note**
For the remaining drop-down menu (thickness2, not radius), a choice of distribution can lead to the following error:
```
12:58:47 - ERROR: Traceback (most recent call last):
  File "sasview/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py", line 3240, in <lambda>
    func.currentIndexChanged.connect(lambda: self.onPolyComboIndexChange(str(func.currentText()), i))
  File "sasview/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py", line 3316, in onPolyComboIndexChange
    self.kernel_module.set_dispersion(param.name, self.disp_model)
  File "sasmodels/sasmodels/sasview_model.py", line 816, in set_dispersion
    raise ValueError("%r is not a dispersity or orientation parameter"
ValueError: 'thickness[n]' is not a dispersity or orientation parameter
```

 - Version: 5.0.6

 - OS: Linux

krzywon:
## Description

This fixes a number of multiplicity issues including addition of models in the sum|multi custom combo boxes, copy/paste of parameters, copy to excel/latex files, linking to structure factors, and suppressing the `SLD profile` button for models without SLD parameters.

Fixes #1089
Fixes #1897 
Fixes #2369
Fixes #2563

## How Has This Been Tested?

Tested locally using the `core_multi_shell`, `onion`, and `unified_power_Rg` built-in models as well as the `core_multi_shell_cylinder` model from [sasmodels->newmodles0923](https://github.com/SasView/sasmodels/tree/newmodels0923). 

- [ ] Code has been reviewed


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -2161,8 +2161,6 @@ def updateModelFromList(self, param_dict):
         Update the model with new parameters, create the errors column
         """
         assert isinstance(param_dict, dict)
-        if not dict:
-            return
```
rozyczko, line 2164: :eyes:

rozyczko (commented):
Code looks good. Basic functionality checked. Multishell models serialize and deserialize properly.
Response to the shell combobox is as expected but we still lose the previously entered values on shell change. Should this be addressed? I believe it was mentioned as a bug in one of the issues.

RichardHeenan:
Built locally using 2369-multiplicity-copy and newmodels0923, I am happy so far, well done Jeff for fixing this.

core-multi-shell with one shell now gives same fitting result as core-shell-sphere, including polydisp on core radius and hard sphere S(Q)

core-multi-shell-cylinder with one shell now gives same fitting result as core-shell-cylinder, including with S(Q), and with reasonably close results with 3 or 4 polydispersity distributions (though understandably a bit slow to compute) 

[ Also - when I have the same data in more than one fit tab and use S(Q) the saving of P(Q) and S(Q) in the data explorer replaces previous one from another tab, which it ought not to.]

[Also there the new dropdown in "send to" for "replace" data in current fit tab is not working for me, the dropdown has only the one choice]

krzywon:
To answer the question at today's call by @butlerpd, this does **not** fix the issue related to the RPA model.

RichardHeenan:
I need to find some time to do some more testing, and also to merge my newmodels0923 branches with more recent changes.

One of my  new models, the lamellar_x5_ one, is prone to giving overflows & underflows, especially if using L-M fitting. When this happens the I(Q) in the data explorer comes back all as nan or inf. In my  local build it seems that (a) no error messages appear anywhere, (b) the model (or perhaps rather the comms with the gui) seems then to be corrupted, failing to work properly from that point onward, even with calculations that I know ought to work.  Testing simple forwards calculations of the I(Q) from the model using the "compare" py script suggests the model itself is behaving correctly. Need to to some more testing, perhaps make a temporary model that can be deliberately failed by a change of parameter value.

butlerpd:
My concern about the RPA not working is whether this is because the fix is not really correct, i.e. the special parameter is not being interpreted as defined.  If however it is just a case of extra code needed to fully implement the use of the special parameter I'm good with calling this an improvement and move forward approving.  @krzywon has agreed to look into this.

wpotrzebowski:
Regarding #1897 is the agreement to not show SLD profile?

wpotrzebowski:
@krzywon will take a look at #1089

wpotrzebowski (commented):
The issue pointed in https://github.com/SasView/sasview/issues/1089 doesn't seem to hold any more and I think we should be able to merge it now.

butlerpd:
> The issue pointed in #1089 doesn't seem to hold any more and I think we should be able to merge it now.

Does this constitute an approving review? 😄

wpotrzebowski:
This should be ready to merge

smk78:
Functionality test of https://github.com/SasView/sasview/actions/runs/7079265514 on W10/x64. Running side-by-side with 5.0.6.

All multiplicity models work and generate the same output & sld profiles as they did in 5.0.6

_core_multi_shell_ with 1 shell still gives same output as _core_shell_ (with same parameters); as it did in 5.0.6

#607 was already fixed in 5.0.6

#1089 was already fixed in 5.0.6, in the sense that PaulK's example in https://github.com/SasView/sasview/issues/1089#issuecomment-478231206 does work; **however the SLD profile calculation in the mixture model still fails**; but this appears to be a generic fail for any model which is not a pure multiplicity model so can we just remove the SLD Profile button in these instances?

**Combining two multiplicity models still does not work correctly.**

**Whilst we removed the _rpa_ model from the model choices on the Fit Panel, it still appears in the model choices in the Add/Multiply Model tool!** Plugins can be created with it!

Interestingly, if you _do_ create a plugin with the _rpa_ model (I did __rpa_+_line_) as a result of this PR you now get 5 additional (_rpa_) parameters that were missing in 5.0.6, in addition to an additional (erroneous?) multiplicity level parameter that shouldn't be in either base model!

#1106 also looks like it was fixed in 5.0.6, but there does seem to be a glitch, at least with the _spherical_sld_ model: I had a Fit Panel with this model with 2 layers in which I changed the thickness, interface & shape parameters for both layers away from the model default, I then copied the parameters, opened a new Fit Panel, selected the same model and pasted in the parameters. **The n_shells parameter erroneously incremented from 2 > 3**. Interestingly, this didn't happen in 5.0.6!

#1816 also looks like it was fixed in 5.0.6; no sign of the error @RichardHeenan reported

#1817 is fixed by this PR; the erroneous SLD profile button is no longer present on _unified_power_rg_ model

#2369 is fixed by this PR; parameters from multiplicity models (including the multiplicity parameter itself) can now be copied by Edit > Copy Params to... or saved to file

#2563 is fixed by this PR; all dispersity distributions are now available on all polydisperse parameters in multiplicity models

smk78 (changes requested):
Performed functionality review on W10/x64.

Mostly good, but a few oddities that it would be good to tidy up if possible.

gonzalezma:
Re-checked functionality. Basically same conclusions as Steve, but I found a remaining issue related to #2563. Now all the PD parameters have their correct menu, but when trying to change the distribution from gaussian to lognormal of thickness2 or thickness 3 (for a core_multi_shell model), the following errors were thrown:

`2024-01-21 21:20:11,199 : ERROR : sas.qtgui.Perspectives.Fitting.FittingWidget (FittingWidget.py:173) :: Traceback (most recent call last):
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3474, in <lambda>
    func.currentIndexChanged.connect(lambda: self.onPolyComboIndexChange(str(func.currentText()), all_items-1))
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3550, in onPolyComboIndexChange
    self.kernel_module.set_dispersion(param.name, self.disp_model)
  File "C:\Users\gonzalezm\Git\sasmodels\sasmodels\sasview_model.py", line 816, in set_dispersion
    raise ValueError("%r is not a dispersity or orientation parameter"
ValueError: 'thickness[n]' is not a dispersity or orientation parameter

2024-01-21 21:20:43,929 : ERROR : sas.qtgui.Perspectives.Fitting.FittingWidget (FittingWidget.py:173) :: Traceback (most recent call last):
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3474, in <lambda>
    func.currentIndexChanged.connect(lambda: self.onPolyComboIndexChange(str(func.currentText()), all_items-1))
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 3500, in onPolyComboIndexChange
    param = self.model_parameters.form_volume_parameters[row_index]
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
IndexError: list index out of range
`

butlerpd:
I think a number of the issues brought up by @smk78 would require a lot more work and are actually outside the scope of this PR. Multiplicity has a lot of issues in implementing in the GUI.

Unless @krzywon thinks some of these can be easily addressed, or are in scope (i.e. the new code is breaking something) I would recommend we merge this as a step forward?

butlerpd:
Sounds like actually still a few things to fix.

butlerpd:
When time to code freeze 6.0.0 into beta decide what to do with where we are.

krzywon:
> #607 was already fixed in 5.0.6

I'll close this issue and update my original comment.

> #1089 was already fixed in 5.0.6, in the sense that PaulK's example in [#1089 (comment)](https://github.com/SasView/sasview/issues/1089#issuecomment-478231206) does work; **however the SLD profile calculation in the mixture model still fails**; but this appears to be a generic fail for any model which is not a pure multiplicity model so can we just remove the SLD Profile button in these instances?

This is probably outside the scope of this PR. I'll create two new issues related to #1089. The first on the SLD profile calculation and the second on combining two multiplicity models. The second might be a sasmodels issue, but that is likely a question for @pkienzle.

> **Combining two multiplicity models still does not work correctly.**

I've added logic, when either of the selected models is a layered/multiplicity model, to remove all plugin and built-in multiplicity models from the other combo box so this can no longer happen.

> **Whilst we removed the _rpa_ model from the model choices on the Fit Panel, it still appears in the model choices in the Add/Multiply Model tool!** Plugins can be created with it!

The RPA model is no longer available in the list of models.

> #1106 also looks like it was fixed in 5.0.6, but there does seem to be a glitch, at least with the _spherical_sld_ model: I had a Fit Panel with this model with 2 layers in which I changed the thickness, interface & shape parameters for both layers away from the model default, I then copied the parameters, opened a new Fit Panel, selected the same model and pasted in the parameters. **The n_shells parameter erroneously incremented from 2 > 3**. Interestingly, this didn't happen in 5.0.6!

I'm still working on this, but I introduced this bug, so it needs to get fixed here before merging.

> #1816 also looks like it was fixed in 5.0.6; no sign of the error @RichardHeenan reported

I'll close this issue and update my original comment.

> #1817 is fixed by this PR; the erroneous SLD profile button is no longer present on _unified_power_rg_ model
> 
> #2369 is fixed by this PR; parameters from multiplicity models (including the multiplicity parameter itself) can now be copied by Edit > Copy Params to... or saved to file
> 
> #2563 is fixed by this PR; all dispersity distributions are now available on all polydisperse parameters in multiplicity models

Thanks for the report on these.

krzywon:
There are two outstanding issues still to be fixed before this is ready for another review.

> https://github.com/SasView/sasview/issues/1106 also looks like it was fixed in 5.0.6, but there does seem to be a glitch, at least with the spherical_sld model: I had a Fit Panel with this model with 2 layers in which I changed the thickness, interface & shape parameters for both layers away from the model default, I then copied the parameters, opened a new Fit Panel, selected the same model and pasted in the parameters. The n_shells parameter erroneously incremented from 2 > 3. Interestingly, this didn't happen in 5.0.6!

I'm still working on this, but I introduced this bug, so it needs to get fixed here before merging.

> Re-checked functionality. Basically same conclusions as Steve, but I found a remaining issue related to https://github.com/SasView/sasview/issues/2563. Now all the PD parameters have their correct menu, but when trying to change the distribution from gaussian to lognormal of thickness2 or thickness 3 (for a core_multi_shell model)

The name associated with the PD parameter is wrong. I'll need to fix it.

gonzalezma:
Tested functionality. Great work: All issues have been fixed. I just found two minor problems. If they can be solved, this would help to improve the user experience, but otherwise I don't think that they should prevent merging. 

One is that if one creates a mixture model where one of the models is a multiplicity model, the 'Show SLD Profile' button appears in the plugin model. However, when you click the button an error occurs:
'''
ERROR: SLD profile calculation failed.
'''

The second one is that changing the value of n resets all the "multiplicity" values to the default. For example, if I first do a fit with core_multi_shell and n=1 and then change to n=3, the previous values for the parameters sld_core, sld_solvent and radius are conserved, but sld1 and thickness1 not. This is more annoying in the polydispersity tab, if one has previously changed any of the distributions from gaussian to e.g. lognormal or schulz, as shown in the image:

multiplicity_issue

gonzalezma:
I don't know if this is unrelated to this branch, but I had to add an 'import subprocess' to the file convertUI.py to make it work.

krzywon:
> I don't know if this is unrelated to this branch, but I had to add an 'import subprocess' to the file convertUI.py to make it work.

The latest merge commit should fix this.

> One is that if one creates a mixture model where one of the models is a multiplicity model, the 'Show SLD Profile' button appears in the plugin model. However, when you click the button an error occurs:
'''
ERROR: SLD profile calculation failed.
'''

This is one of the items I was going to ticket from #1089. It's apparently been around for a while.

> The second one is that changing the value of n resets all the "multiplicity" values to the default. For example, if I first do a fit with core_multi_shell and n=1 and then change to n=3, the previous values for the parameters sld_core, sld_solvent and radius are conserved, but sld1 and thickness1 not. This is more annoying in the polydispersity tab, if one has previously changed any of the distributions from gaussian to e.g. lognormal or schulz, as shown in the image:

Very annoying. I'm looking into this now.

gonzalezma:
@smk78 just reported a similar problem in a comment to issue #2438, so perhaps this issue is not directly related to the changes here?

krzywon:
Final update:

> changing the value of n resets all the "multiplicity" values to the default.

This should now be fixed using a simple copy-paste of parameters.

> @smk78 just reported a similar problem in a comment to issue https://github.com/SasView/sasview/issues/2438, so perhaps this issue is not directly related to the changes here?

I can reliably recreate this in v5.0.6 (and maybe spent too much time trying to debug this). It appears to be related to simultaneous signals that cause conflicting behavior. I'll add more in the issue, but this has been around a while, is not related to this PR, and could be a complex fix.

krzywon:
For clarity during a review, here are the issues this PR fixes or improves:

- Copy params to latex/excel formats now works for multiplicity models.
- SLD profile button is suppressed for models with no SLD parameters.
- Changing the number of layers in a multiplicity model retains any modified values.
- Polydispersity distribution types are now available for layer parameters.
- P*S models no longer throw errors when changing multiplicity parameters.
- The RPA model is no longer available in the add/multi model editor window.
- Multiplicity models can no longer be combined using the easy add/multiply editor.
- Documentation changes
  - Differentiate `@` vs `*`
  - Suppression of layered models if one already selected
- Polydispersity tab no longer accessible with checkbox unchecked.
- Removed a number of `if not dict: pass` statements (they do nothing).

And issues this PR does not fix:

- PD distribution combobox sometimes ends up in the filename column - issue since v5.0.6 (at least)
- Changing the number of layers in a multiplicity model will reset PD params - #2438
- Combining two multiplicity models - issue since v4.0.0
- RPA (anything related to this) - issue since v5.0.0
- SLD Calculation failures - issue since ?

butlerpd (approved):
Everything that this PR claims to fix (which is not everything to do with multiplicity) seems to work as advertised save one thing which is still mostly fixed.  However this is a great improvement over 5.x so the issues noted below should not prevent merging as necessary.  That said here is what I found on windows 10:
* A sum model that contains a layered model with no SLD  does NOT suppress the SLD profile which will throw an error if pressed.  The suppression does seem to work in all other cases. 
* Loading a multi SLD (but not a multiplicity that does not contain SLD such as the `unified_power_Rg` model) throws a warning pop-up saying “Clipboard content is incompatible with the fit page.” This also happened after playing with the `core_mutli_shell` model and then adding a HS structure factor. As soon as the structure factor was chosen the pop-up occurred. It is almost as if choosing a model when it is an SLD profile model automatically tries to get something from the clipboard? 
* Finally, while the Copy parameters seems to work (as does Copy paste from a multiplicity model to another model), I notice that the table created includes extraneous parameters like the name of the model (and`multiplicity` in the case it is) This is clearly not part of this PR but is annoying I think?
* The other part of the Copy parameters which again is not likely to be part of the PR is that the TRUE FALSE flags for whether a parameter is being fit do NOT align. That seems to be true regardless of multiplicity or not.

krzywon:
> A sum model that contains a layered model with no SLD  does NOT suppress the SLD profile which will throw an error if pressed.  The suppression does seem to work in all other cases.

This is now fixed. I don't plan to fix the other issues noted here in this PR.

butlerpd:
This appears to fix both the first AND second points above. Not sure why? but I'll take it. 

However .... I do **not** understand why the onion model does not show the SLD. the reg expression should identify it and the model does contain the SLD function (`def profile`). I know I tested at least one of core_multi_shell and onion but maybe I forgot to test onion specifically (the `core_multi_shell` does in fact show the SLD profile button)?  Again I would not hold up merging for this.

butlerpd:
Right ... color me totally confused. This final fix now allows the onion model's SLD profile to be shown along with others and now that SLD profile is suppressed for non SLD multiplicity models even in sum/multiply, the `show profile` button still throws an error for the mixed models that do have an SLD profile.
```python
 ERROR: SLD profile calculation failed.
```

 I thought we checked that and it was working before? but looking at the code I'm not sure that is possible. It is the same error reported by @gonzalezma earlier and seems to come from the call to `onShowSLDProfile(self)` try except block aroun
```python
def onShowSLDProfile(self):
    """
    Show a quick plot of SLD profile
    """
    # get profile data
    try:
        x, y = self.kernel_module.getProfile()
    except TypeError:
        msg = "SLD profile calculation failed."
        logging.error(msg)
        return
```
Should we just suppress the profile for all mixture models?

On the other hand,  I think all the issues identified by @smk78 which led him to request changes have been addressed from my testing so that block on merging should be removed I think.

krzywon:
The SLD calculation error is one of the items I listed as 'not going to fix' in this PR. It's present in v5.0.6, so not something I introduced here.

butlerpd:
Fair enough - Pencils down you are telling me :-).  I was wondering about just disabling the button for all mixture models but then maybe that will just make us forget about the problem (out of sight out of mind?).  In which case we should be ready to merge IMO

PR #2652, opened 2023-09-25
PyInstaller pinned

rozyczko:
## Description

CI started to pull pyinstaller 6, which is messing up the Windows install generating a single file exe.
This started right when we finished the code camp...
The proper solution is to find out what changed in the default parameters and modify the SPEC file to not generate a single executable but the "normal" file tree.

## How Has This Been Tested?

GH windows build 

- [ ] Code has been reviewed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
       - name: Install utilities to build installer
         if: ${{ matrix.installer }}
         run: |
-          python -m pip install pyinstaller
+          python -m pip install pyinstaller==5.13.2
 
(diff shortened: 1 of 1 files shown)


wpotrzebowski (approved):

llimeht:
pyinstaller 6 seems to have lots of changes that are aimed at fixing problems with macos signing - are they needed, or is macos signing working OK at present? [https://pyinstaller.org/en/stable/CHANGES.html](https://pyinstaller.org/en/stable/CHANGES.html) 

It looks like there are *lots* of changes in pyinstaller 6 that will require a bit of work to catch up with and now is not the best time in the sasview release cycle to do that work.

PR #2664, opened 2023-10-20
Release 6.0.0 notes

wpotrzebowski:
## Description

This PR addresses two issues related to continuous integration for release_6.0: 1) It uses GH action that allows notarization with notarytool (compare to atool that won't be supported fron Nov 1st). 2) It automatically uploads release notes (so they don't disappear when upload action is pushed). We will need to make a decision whether this right approach for keeping GH release notes in the future. (One can still edit release page on GH and this file won't be aware of it). 

## How Has This Been Tested?
Installed sasview from installer file and checked if it runs (and if it is signed)

- [ ] Code has been reviewed

### DIFF
--- a/build_tools/release_notes/6.0.0_notes.txt
+++ b/build_tools/release_notes/6.0.0_notes.txt
@@ -1,0 +1,101 @@
+## New features
+- Orientation viewer
+- Corfunc refactored
+- Simultaneous fitting allows for weighting scheme
+- Preferences panel with display and plotting options (polydispersity and residuals plots can be hidden).
+- Improved label handling on plots
+- Residuals plots refactored
+- PDB reader refactored
+- Wedge slicer added
+- Sasdata package separated
+- Move to PySide6
+- Python 3.11 support
+- Required documentation (https://github.com/SasView/sasview/issues/2641)
+- Improved documentation
+- New Tutorials
+
+## Major bug fixes:
+- Handling of constraints for polydisperse parameters
+- Binning and FitPage plotting of SESANS data 
+- Fixed 1D slit-smearing function
+- Start-up speed improved
+- Magnetic SLD? 
+- Multiplicity model?
+
+## New models
+- New broad peak model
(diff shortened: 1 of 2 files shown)


build_tools/release_notes/6.0.0_notes.txt
```diff
@@ -0,0 +1,101 @@
...
+* Empty lines in data explorer by @rozyczko in https://github.com/SasView/sasview/pull/2643
+
+## New Contributors
+* @smalex-z made their first contribution in https://github.com/SasView/sasview/pull/2535
+* @ehewins made their first contribution in https://github.com/SasView/sasview/pull/2543
+* @lozanodorian made their first contribution in https://github.com/SasView/sasview/pull/2558
+* @gdrosos made their first contribution in https://github.com/SasView/sasview/pull/2585
+* @astellhorn made their first contribution in https://github.com/SasView/sasview/pull/2610
+
+**Full Changelog**: https://github.com/SasView/sasview/compare/nightly-build...v6.0.0-alpha
```
llimeht, line 101: Should this full changelog URL compare 5.0.6 to 6.0.0-alpha? Comparing to the nightly-build tag (which was tagged at some point during codecamp Oct 2022, from what I can see) doesn't seem right. 

I suspect there's also a stack of PRs that aren't listed in the What's Changed section, perhaps because they were merged prior to the `nightly-build` tag but after the branch point for `release_5.0.6`? (I've not looked in detail, but I think all the stuff I was doing in late 2022 wasn't destined for 5.0.6 but was held over for 6.0.0 and isn't listed there)

wpotrzebowski: Yes, the notes themself need to be worked on. This is more of the template. The main point of this is that release notes disappear every time someone pushes to the branch, so this is mostly to prevent empty records on this page: https://github.com/SasView/sasview/releases/tag/v6.0.0-alpha

Issue #2668, opened 2023-10-26
plot context menu is broken in 6.0.0a1


butlerpd:
**Describe the bug**
Using the latest build of 6.0.0a1, right clicking on most plots to get the context menu throw an error instead:
```
2023-10-26 11:09:38,405 : ERROR : sas.qtgui.Perspectives.Fitting.FittingWidget (FittingWidget.py:173) :: Traceback (most recent call last):
  File "sas\qtgui\Plotting\PlotterBase.py", line 272, in showContextMenu
  File "sas\qtgui\Plotting\Plotter.py", line 316, in createContextMenu
  File "sas\qtgui\Plotting\Plotter.py", line 397, in addPlotsToContextMenu
ValueError: operands could not be broadcast together with shapes (395,) (0,) 
```
Interestingly, the "extra plots" (i.e. residual and polydispersity) do not throw an error but pop up the context menu just fine. Could this be something to do with "role?"

**To Reproduce**
1. Load some data
2. click on `Create New` - alternatively click on `Send to Fitting` choose a model and hit calculate/plot
3. on the plot that pops up right click to get the context menu
4. Nothing happens! but if you look in the Log Explorer you will see an error like:
```
11:18:30 - ERROR: Traceback (most recent call last): File "sas\qtgui\Plotting\PlotterBase.py", line 272, in showContextMenu File "sas\qtgui\Plotting\Plotter.py", line 316, in createContextMenu File "sas\qtgui\Plotting\Plotter.py", line 397, in addPlotsToContextMenu ValueError: operands could not be broadcast together with shapes (66,) (0,)
```

**Expected behavior**
I expect the pop menu of course

**Screenshots**
NA

**SasView version (please complete the following information):**
 - Version: 6.0.0a1 uploaded to github at 5AM GMT on October 26

**Operating system (please complete the following information):**
 - OS: Windows 10

**Additional context**
I did not test plots from other perspectives

rozyczko:
switching to new numpy made checks like `!=[]` not just deprecated but removed. Fixing here.
No other occurrences of

PR #2670, opened 2023-10-27
fix for broken context menus in 1d plots

rozyczko:
## Description

switching to new numpy made checks like !=[] not just deprecated but removed. Fixing here and in related test.
I found no other numpy.array == [] or numpy.array != [] comparisons.

#2668

## How Has This Been Tested?

local Win10

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Calculators/SlitSizeCalculator.py
+++ b/src/sas/qtgui/Calculators/SlitSizeCalculator.py
@@ -116,7 +116,7 @@ class SlitSizeCalculator(QtWidgets.QDialog, Ui_SlitSizeCalculator):
         try:
             xdata = data.x
             ydata = data.y
-            if xdata == [] or xdata is None or ydata == [] or ydata is None:
+            if xdata.size == 0 or xdata is None or ydata.size == 0 or ydata is None:
                 msg = "The current data is empty please check x and y"
                 logging.error(msg)
                 return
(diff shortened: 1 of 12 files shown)


src/sas/qtgui/Calculators/SlitSizeCalculator.py
```diff
@@ -116,7 +116,7 @@ def calculateSlitSize(self, data=None):
         try:
             xdata = data.x
             ydata = data.y
-            if xdata == [] or xdata is None or ydata == [] or ydata is None:
```
krzywon, line 119: If `xdata` or `ydata` are `None`, their respective size checks will throw an exception.

krzywon (commented):
This should fix the issue, but there is one potential edge case that could cause issues noted in a comment.

I think this branch was based on `main` instead of `release_6.0.0`. Like in #2649, if we want those few commits in the 6.0.0, this isn't a problem.

rozyczko:
> I think this branch was based on `main` instead of `release_6.0.0`. Like in #2649, if we want those few commits in the 6.0.0, this isn't a problem.

Oops, yes, I branched it off of `main`. I think a better strategy would be to cleanly base it on `release_6.0.0` instead.
Unless, that is, the changes which crept in are what we would cherry pick from `main` anyway?

rozyczko:
Yes, this should be merged with `main` instead.
Cherrypicking the changes to `release_6.0.0` should be then performed or a separate PR with relevant changes.

butlerpd:
@krzywon it looks like all your questions have been answered? do you now approve? or is there more discussion here?

krzywon (approved):
Looks good now. Ready to merge IMO.

PR #2674, opened 2023-11-14
2658 Update Corfunc Docs

lucas-wilkins:
Documentation updated to reflect the changes made in the to Corfunc.

I've roughly followed the principles of diataxis, but it doesn't seem big enough to justify a distinction between a tutorial and a how-to.

This is currently a draft because I'm unsure about some of the details in the theory section. Namely, I have been told that corfunc provides an inverse of a 1D system, however, I cannot square this with the maths.

A 1D system will have a Q spectrum given by the cosine transform of the correlation function (under appropriate assumptions about small angles), however, corfunc performs the transform using q^2 cos(q x), which is appropriate for second derivative of the correlation function, not the correlation function itself.

For Gamma3, the recovered correlation is actually what is specified.

I need to be clear on exactly what is happening here before moving it from draft. @butlerpd and @smk78 said they would investigate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,7 @@ jobs:
       - name: Install X11 libraries (Linux)
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         run: |
+          sudo apt-get update
           sudo apt install libxkbcommon-x11-0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xinerama0 libxcb-xfixes0 x11-utils xvfb libegl-dev
 
       - name: Install Python dependencies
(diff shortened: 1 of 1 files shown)


smk78 (changes requested):
Overall this is a pretty thorough update to the corfunc docs. I approve it in principle, but there are a number of typos and a couple of other tweaks that need addressing before it is merged.

smk78:
@lucas-wilkins , I've made my suggested changes. A few things left:

- in corfunc-theory.rst, need something about Gamma3;

- in corfunc-technical.rst, the 2nd & 3rd equations under Large Q  just seem to duplicate what the 1st equation says?
  
image

- and I added some additional structural parameters(!), but I may not have gotten my rst/latex quite right...

- talking of which, the Stribek polydispersity needs defining, but you now have my book! :-)

butlerpd:
@lucas-wilkins and @smk78 I note this branch is still listed as a draft against main. Is the plan still to include this with the soon to be released 6.0.0 beta? If so we will need to also merge this into the 6.0.0_release branch.

PR #2700, opened 2023-12-22
Nightly builds fix

wpotrzebowski:
## Description
Fixing nightly build but not fully resolving an issue related to running workflows by developers from outside sasview organization. Trying to move signing to nightly_build.yml didn't work and therefore there is only notarization step done there. It seems that we may enable runs from outside developers by changing run trigger from pull_reguest to pull_reaquest_target. However, as far as I understand there is a security risk associated with it. I am not comfortable with making such a decision. 

This PR also changes names of linux and OSX packages to SasView6 (tar.gz, app and dmg). The question is whether we shouldn't drop this numbering for easier maintenance in the future. It may be however confusing for users that have multiple versions installed (at least on MAC). 
 

- [ ] Code has been reviewed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,12 +226,6 @@ jobs:
           iscc installers/installer.iss
           mv installers/Output/setupSasView.exe installers/dist
 
-      - name: Fix qt folders
-        if: ${{ matrix.installer && startsWith(matrix.os, 'macos') }}
-        run: |
-          cd installers/dist
-          python ../../build_tools/fix_qt_folder_names_for_codesign.py SasView5.app
-
       - name: Sign executable and create dmg (OSX)
         if: ${{ matrix.installer && startsWith(matrix.os, 'macos') }}
         env:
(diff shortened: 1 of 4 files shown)


wpotrzebowski:
@krzywon and @rozyczko can you take a look on this PR, please? It would be useful to merge before the camp next week, otherwise, main/nightly builds will keep failing.

.github/workflows/ci.yml
```diff
@@ -226,17 +226,31 @@ jobs:
...
+          codesign --verify --options=runtime --entitlements ../../build_tools/entitlements.plist --timestamp --deep --verbose=4 --force --sign "Developer ID Application: European Spallation Source Eric (W2AG9MPZ43)" SasView6.app
+          hdiutil create SasView6.dmg -srcfolder SasView6.app -ov -format UDZO
+          codesign -s "Developer ID Application: European Spallation Source Eric (W2AG9MPZ43)" SasView6.dmg
 
       - name: Build sasview installer tarball (Linux)
         if: ${{ matrix.installer && startsWith(matrix.os, 'ubuntu') }}
         run: |
           cd installers/dist
-          tar zcf sasview5.tar.gz sasview
+          tar zcf sasview6.tar.gz sasview
```
krzywon, line 253: Can we define this name somewhere in `matrix.py` using the current version? That would simplify a lot of the issues you seemed to face converting `sasview5` -> `sasview6` and will help future SasView developers when we move to v7 and beyond.

.github/workflows/ci.yml
```diff
@@ -226,17 +226,31 @@ jobs:
           iscc installers/installer.iss
           mv installers/Output/setupSasView.exe installers/dist
 
-      - name: Build sasview installer dmg file (OSX)
+      - name: Sign executable and create dmg (OSX)
```
krzywon, line 229: Do we need a conditional here to only sign when pushed to main? I only ask because this is diverging from our previous behavior.

krzywon (approved):
The Windows installer from the CI installs and opens without issue and the Mac CI passes. Seems to be good.

rozyczko (approved):
I concur with Jeff's comments about generalizing SasView[567..] name - this should all be part of a single config file. Other than that, the code looks safe and the windows installer seems to work as expected.

.github/workflows/ci.yml
wpotrzebowski: Yes, it will make sense. From the user's perspective, signing only makes sense when it is followed by notarization. (Gatekeeper complains when app signed but not notarized).

wpotrzebowski: And we do notarization from nightly-build on main branch

wpotrzebowski: If I understand correctly, as it is now:
```
 pull_request:
 push:
    branches: [main]
    tags: ['v*']
```
is done for pushes to main and all PRs, which is probably something we actually want?

wpotrzebowski:
Thanks for reviews @rozyczko and @krzywon! I will merge it once I know we can sign and notarize app. (As far as I can tell, ESS IT still needs to accept the updated agreement).

PR #2711, opened 2024-01-16
Constraint on parameter raises error

### LINKED ISSUE #2695: constraint on parameter raises error
**Describe the bug**
Trying to set a constraint on a parameter in 6.0.0 alpha throws a Traceback error
```
Traceback (most recent call last): File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1053, in editConstraint File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1053, in File "sas\qtgui\Perspectives\Fitting\FittingUtilities.py", line 968, in isParamPolydisperse TypeError: argument of type 'NoneType' is not iterable
```

**To Reproduce**
1. load a data set (I used apoferritin)
2. send to fitting
3. choose a model (I used core shell sphere)
4. Select a parameter (I chose thickness)
5. Right click on that parameter and choose `Constrain parameter to its current value` from the drop down
6. Right click on same parameter and now choose `Edit constraint` from the drop down
7. The error is thrown

**Expected behavior**
Instead of an error there should be a box to enter a constraint

 - Version: 6.0.0 alpha build from Dec 19, 2023

 - OS: Windows 10

**Additional context**
suspect this may be another pyside issue?

rozyczko:
## Description

Some of the logic was incorrect, when setting the single-parameter constraint and when querying the underlying model information.
Now, item query returns proper parameter names, and the context menu is properly constructed (i.e. no "Edit constraint" on simple value constraint added)

Fixes #2695 

## How Has This Been Tested?

Local dev build on Win10

- [x] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -780,7 +780,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         # but let's check the correctness.
         assert len(selected_rows) == 2
 
-        params_list = [s.data(role=QtCore.Qt.UserRole) for s in selected_rows]
+        params_list = [s.data() for s in selected_rows]
         # Create and display the widget for param1 and param2
         mc_widget = MultiConstraint(self, params=params_list)
         # Check if any of the parameters are polydisperse
@@ -1015,6 +1015,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             max_t = model.item(row, max_col).text()
             # Create a Constraint object
             constraint = Constraint(param=param, value=value, min=min_t, max=max_t)
+            constraint.active = False
             # Create a new item and add the Constraint object as a child
             item = QtGui.QStandardItem()
             item.setData(constraint)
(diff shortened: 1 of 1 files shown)


wpotrzebowski:
I've tested it on Mac and I don't seem to have access to "Edit Constrain" menu
Screenshot 2024-01-17 at 21 59 27
I've checked that it is avaialable in 5.0.6

gonzalezma:
The edit option should appear when you have a definition in the Simultaneous fit tab. When you simple constraint one parameter to its current value, the option is not present. If this is not what you tested, then the behaviour on Windows and Mac seems to be different.

wpotrzebowski:
I followed the steps described in https://github.com/SasView/sasview/issues/2695 and got different behavior with 5.0.6 and this build.

gonzalezma:
Yes, that's true, but I think this was intended, judging from Piotr's description of the PR. I agree with him that when constraining a parameter to its current value, the edit button makes not much sense.

wpotrzebowski:
OK, makes sense. I didn't read PR description properly.

wpotrzebowski:
Tested again with newly build Mac and works fine. Following the same logic, wouldn't make sense to also remove "Select and Deselect" items? If I understand correctly they shouldn't affect parameter if it is constrained? If so, probably separate PR and not for 6.0 (as we are fixing bugs now).

wpotrzebowski (approved):

danielballan:
Wow, what a great feature! Can't wait to see this merged!

danielballan:
I sure wish @butlerpd would post a review.

butlerpd (changes requested):
This is partly to test the request change but @krzywon says this is not how it worked in 5.0.6. Shouldn't we be keeping the functionality

krzywon:
> This is partly to test the request change but @krzywon says this is not how it worked in 5.0.6. Shouldn't we be keeping the functionality

The issue has been discussed. The input box in the constraint editor is partly broken and requires more work, and will be reinstated once it has been resolved. That work is beyond the scope of this PR, so we should merge this.

PR #2731, opened 2024-01-17
[old] Debye efficiency

klytje:
## Description

The previous Debye implementation for the generic scattering calculator has been replaced with external calls to the AUSAXS library, which is significantly faster. With the foundation already in place, future additions relying on this same library will be much easier to implement. 

The following dependencies are introduced:
* cpufeature to determine the CPU architecture and select an optimized version of AUSAXS,
* AUSAXS itself as shared libraries for all platforms. Currently these are fetched from a new empty sasview repository (https://github.com/SasView/ausaxs). I intend to modify the download path to instead point to the actual AUSAXS repository once it is opened to the public, hopefully within the next couple of months.

## How Has This Been Tested?
I have added an additional test case in `utest_sas_gen.py` where the previous calculation is compared with AUSAXS for a few different systems. 

## Issues
_Major_
- [x] Using the shared library approach on Windows results in a deadlock. Perhaps natively compiling my library with MSVC can help. This is a priority issue as it blocks future developments utilizing my library (_SAXS, water models, ..._)
_Solved: I refactored my library to a different threading model, which seems to have solved the issue_
- [x] The result will have substantial errors for highly ordered structures such as a crystal due to the binning approximation. These cases should be detected and forwarded to the old sasview calculator. 
_Solved: there was a minor bug in how weighted bins were used in my library_
- [x] MacOS is currently not supported since I cannot compile my library by myself for this system. @timsnow agreed to do the compilation once I fix some Clang-specific issues. 
_Solved: MacOS shared libraries are now availabe_
- [x] Ubuntu 20.04 currently fails. Based on a minor investigation by me & @rozyczko, it seems that this is due to the system being treated differently from Ubuntu 22.04 in the `ci.yml`. Perhaps @wpotrzebowski can help solve this? _Solved: the Linux library **must** be compiled with v20.04_

_Minor_
- [x] The user should be able to extend the q-range. 
_Solved: q-ranges within the default of my library will simply be interpolated. q-ranges outside this range will be somewhat slower but does seem to work_
- [ ] The user should be able to evaluate the profile for any system. Currently systems larger than 2000Å will result in my library crashing, though this should just result in the calculation being forwarded to the old sasview calculator. 
_Semi-solved: if my library crashes, the old sasview calculator is used. This is a terrible solution though, since the old calculator will never finish evaluating such huge systems. I'll try to find a more elegant solution later - for now, I think it is okay since such huge systems are rare_
- [x] Currently some chunking seems to be going on when using a large number of q-values, such that the calculation is split into multiple calls. This is counterproductive when using my library as the full q-range is calculated for every call. We should trace this down and change it. 
_Solved: the GUI chunks the calculation to add cancellation points and progress reports. Since I have manually added the latter and the former isn't possible with an external library call, this is not necessary and have been skipped with a simple if-check. In the future this should probably be refactored to a cleaner approach. I don't think the GUI itself should be responsible for chunking like this, but rather provide a listener object and allow the Iq models themselves to manage this_
- [x] A new process is spawned during the calculation, resulting in a big black terminal being created on top of the sasview GUI. 
_Solved: this doesn't happen when hooking into a shared library_

- [x] Update licensing to mention AUSAXS dependencies. 
_Tim & I found that this is not necessary since they are not part of the source code_
- [ ] Code has been reviewed
_Works on both Linux & Windows_
- [x] Windows installer (GH artifact) has been tested (installed and ships with AUSAXS) 
_Tested that the installer ships with AUSAXS libs and successfully uses them at runtime_
- [ ] MacOSX installer (GH artifact) has been tested (installed and ships with AUSAXS) 
- [x] Linux installer (GH artifact) has been tested (installed and ships with AUSAXS)
_Tested that the installer ships with AUSAXS libs and successfully uses them at runtime_

Note that should we decide to merge without Mac support, my new test should be updated to reflect this choice. Otherwise the `main` branch will consistently start failing this test when running it with Mac.


Susyana:
PDB upload works fine, calculation for 0.5 million atoms takes about 18min (on a not particularly fast windows PC). There is still a problem when the q-max is expanded past 0.5 (tested using klytje's executable), the calculated curve looks wrong at high q.

klytje:
> PDB upload works fine, calculation for 0.5 million atoms takes about 18min (on a not particularly fast windows PC). There is still a problem when the q-max is expanded past 0.5 (tested using klytje's executable), the calculated curve looks wrong at high q.

The new executable shipped with sasview itself has been corrected to evaluate points up to q=1Å^-1. Previously it was only evaluating up to 0.5Å^-1 and then doing a poor job of extrapolating the rest. 

Going beyond the q-range [1e-4, 1] is currently not supported. I'll introduce this functionality in a future update of my library.

klytje:
I have narrowed down the Windows dll issues to originating from a FreeLibrary system call, which for some reason causes a deadlock internally in my library. Reading through the ['dll best practice list'](https://learn.microsoft.com/en-us/windows/win32/Dlls/dynamic-link-library-best-practices) indicates this is a common issue in libraries utilizing statically initialized threads and/or thread-local memory. I will debug my own library for a solution to this sometime in the future; for now I'll leave it open as an issue.

wpotrzebowski (commented):
Even if we don't provide full functionality for MAC, we need to be able to build it. I vaguely remember that there was issue with tests failing on MAC. Is this is the case and the reason why Mac build is turned off?

klytje:
Since the new implementation defaults to the old should my library not be found or fail at runtime, I have to manually check for the presence of the library in the new test to ensure that it is actually running my library. This check fails on mac since there is no library. Of course, this is easy to change; we just have to decide if we want to ship with mac support or not first (I also added this info at the bottom of my PR). 

Regarding building, since my library is now publically available, I have forked it to the sasview account [here](https://github.com/SasView/AUSAXS) to keep a stable backup. I will recompile the Windows and Linux libraries from this fork later today. Compilation works just fine for both Linux & Windows with GCC, Clang, & MSVC, so I can't imagine it will be too difficult to adapt to mac as well. I just can't do it myself without the actual OS available.

klytje:
I'm a bit unsure about the licensing here. My project is licensed under the GPLv3 while sasview is BSD-3. Do I have to write an exception somewhere or are these already compatible considering sasview is only redistributing the compiled binaries under the BSD-3?

wpotrzebowski:
In my opinion things can become messy quite quickly if we start realeasing from different code bases. So my recommendation would be to still make a Mac installer working (even if it requires hacking CI). (@klytje I will try to help if you run into trouble). Releasing with GPL licenses is problematic. If this is distributed as a library, it can also have LGPL licenses, which I thought can be distributed with BSD but reading about this, I am not convinced that this the case. @klytje is there are particular reason (e.g. university policy) you release your code under GPL?

klytje:
I thought about just using CI to do the compilation as well, but with mac using x10 I suspect I'll run out of minutes long before I manage to get it working this month. Perhaps I could do it on the forked branch hosted by sasview? 

I'm no expert on licenses. We picked the GPL because it seems to do what we want: keep any future development of the project open-source. I see now that the LGPL does essentially the same thing, but allows linking to the compiled library under different licenses. Again I'm no expert, but to me it seems like if I just change the license to be LGPL it should work just fine for inclusion in sasview since we are not redistributing modified code under the BSD, only the compiled library itself.

klytje:
I realize now that the github runners are free for public repositories, so I don't have to worry about the minutes. 

I've managed to get the entire source to compile without errors on mac, but I'm having issues linking with the standard library. I've tried a few different approaches to solving the problems, but with only Google and the github runners to test on, it's proving both time-consuming and difficult to debug. @wpotrzebowski is this something you have experience with? I have the output available [here](https://github.com/AUSAXS/AUSAXS/actions/runs/8591008656/job/23539270382).

wpotrzebowski:
Good job @klytje! I have a little bit of experience and will take a look unless @timsnow already has a solution for us?

klytje:
Sounds good. I made a [mac branch](https://github.com/AUSAXS/AUSAXS/tree/mac) which requires clang v15 to compile since there was an annoying bug in v14 that I'd prefer not to work around. I'm not sure how useful it is, but the CI yaml is available [here](https://github.com/AUSAXS/AUSAXS/blob/mac/.github/workflows/build-and-test.yml). The only relevant CMakeLists is the one at the root folder of the project, very specifically [this](https://github.com/AUSAXS/AUSAXS/blob/e2c4f92abd519ff252683d76ed4afd4bf4f5aa7a/CMakeLists.txt#L118) section.

wpotrzebowski:
@klytje ok, I will try to build it locally. However my main question is how the fallback mechanism works?The way I understood it, is that if the library is not available it uses old code for GSC (this implies that we don't need to compile Mac to still have installer available). To rephrase it, my main concern is that in this PR Mac is not built at all https://github.com/SasView/sasview/blob/ee703982476c543ca0e9ba45adbb14821c9617fd/.github/workflows/matrix.py#L31

klytje:
I apologize - I think I was not being super clear in my previous reply. 

With this PR, the shared libraries are downloaded and shipped with the installers for Windows & Linux. 
At runtime, when the evaluation method is called, sasview will check for the presence of the library on all platforms. If it is not found in the expected location, the evaluation is forwarded to the old GSC. This is what will happen for mac after this PR. For the other platforms, my libary will be present and used instead. 

_Currently_, macos-latest is not being tested in this PR because my new test will fail due to these lines:
https://github.com/SasView/sasview/blob/ee703982476c543ca0e9ba45adbb14821c9617fd/test/sascalculator/utest_sas_gen.py#L229-L230
This is trivial to fix - just check if the test is being run on macos. I didn't do this yet because I dont't know if we want to include mac support in this PR or if we want to do it later.

wpotrzebowski:
Sounds good. I would suggest adding MacOS with current code base and fix failing tests. Full functionality for Mac can be then be added as a separate PR but at least we can make a progress on this one.

klytje:
Should be working now, but obviously I cannot test the mac installer.

wpotrzebowski:
The MAC installer is crashing but I think it is not related to your PR  (https://github.com/SasView/sasview/issues/2846)

klytje:
I think I solved the MacOS issue.

The architecture of the GitHub runner `macos-latest` is `arm64`, which is incompatible with the `x86-64` SIMD instructions used in my library. I considered this possibility already, and added slower alternatives for platforms not supporting these instructions, automatically toggled by definitions set at compile-time. However, I suspect the `arm64` Mac _does_ support these through its Rosetta translation layer, which translates `x86-64` executables into `arm64` the first time they are run, and thus enables these definitions. However Rosetta is not enabled by default, though it should boot up whenever a `x86-64` _executable_ is launched. But I suspect it does _not_ automatically start when a `x86-64` _library_ is loaded. 

So essentially Mac thinks it supports `x86-64` SIMD instructions and happily compiles my library using them, but then when it attempts to run it realizes it actually doesn't and segfaults. Perhaps the issue could be solved by enabling Rosetta before running sasview, but that doesn't seem like a good solution. Alternatively my library could be compiled simultaneously for both `x86-64` and `arm64` for MacOS, but I was unable to make this work. I might take a look at this again in the future; for now I've just disabled SIMD for Mac. I don't think this will have too much of an impact on the efficiency.

I'll do the rebase to 6.0.0 later.

andyfaff:
It should be quite easy to compile stand alone dylibs to  have both architectures in them. The harder part is to make universal wheels, although cibuildwheel should take care of a lot of the strife.

The proportion of x86_64 macs is going to be slowly decreasing over time, so it's worth adding arm support

wpotrzebowski:
Great to hear @klytje !

When starting SasView for the first time on the fresh machine the Rosetta windows pop-ups automatically. 
Screenshot 2024-04-02 at 22 02 25
And it seems that is required to run SasView. Therefore I don't think it should be problem with running x86-64 architecture.

@klytje once you rebase to 6.0.0 I test it

klytje:
> It should be quite easy to compile stand alone dylibs to have both architectures in them. The harder part is to make universal wheels, although cibuildwheel should take care of a lot of the strife.
> 
> The proportion of x86_64 macs is going to be slowly decreasing over time, so it's worth adding arm support

I've looked into these universal binaries, but was unable to compile my code into one. I'll give it another shot later (after 6.0.0) when I actually have access to a mac.

klytje:
> Great to hear @klytje !
> 
> When starting SasView for the first time on the fresh machine the Rosetta windows pop-ups automatically. And it seems that is required to run SasView. Therefore I don't think it should be problem with running x86-64 architecture.
> 
> @klytje once you rebase to 6.0.0 I test it

That's super weird - I thought for sure Rosetta not being enabled was the only explanation. Googling a bit further suggests Rosetta [may not support `AVX` instructions](https://medium.com/macoclock/m1-rosetta-2-limitation-illegal-hardware-instruction-a3b48fae02e). I don't understand why it would even allow code using them to be compiled in the first place.

klytje:
I'm not sure what went wrong, but rebasing onto the `release_6.0.0` branch resulted in a ton of conflicts and even differences - perhaps because of multiple merges with the `main` branch in the commit history? Anyhow, instead of dealing with these conflicts, I just made a [new PR](https://github.com/SasView/sasview/pull/2859) from a fresh fork of the branch; that seemed like a simpler solution.

butlerpd:
We have merged the ausaxs stuff now into 6.0 .... which will eventually get merged into main. Can we close this PR and delete this branch now @klytje ?

klytje:
Yep feel free to delete this one.

Issue #2390, opened 2022-11-23
Polydispersity parameters: Users don't know what's being reported. 


toqduj:
I've noticed that 44% of the SasView submissions to the Data Analysis Round Robin say that SasView reports volume-weighted size distribution parameters (means and widths in particular). 51% note that these are number-weighted, and some hedge their bets and don't say anything at all. 

This means it's unclear as to what the values mean that come out of SasView. That may be a problem with the user, but it could be resolved by being more explicit in the user interface. @timsnow noticed in the documentation that it should be number-weighted but might be volume-weighted for Schultz distributions? 

Anyway, PEBCAK, but given that even the more "advanced users" that, I suppose, contributed to the Round Robin can't figure this out, perhaps it's worth improving in the UI. These things shouldn't be hidden in the docs. 

If I understand correctly, @pkienzle uses volume-weighting internally for the integrations (as that makes more sense in terms of signal weighting anyway). Perhaps having a slider to switch between "number-weighting" and "volume-/mass-weighting" in the Polydispersity panel would solve both the awareness and the flexibility issues?

RichardHeenan:
But the "scale" parameters in sasview are a "volume fraction" if the sld's are correct, as we normalise the number distribution polydispersity to unit volume. We do this as the volume fraction is what an experimenter is most likely to know, as well as being helpful when thinking about S(Q). Ought to modify the polydispersity.html page linked above.

pkienzle:
Distributions use number density even though integrations use volume weighting. Documentation [here](https://www.sasview.org/docs/user/qtgui/Perspectives/Fitting/fitting_sq.html). Extended discussion with derivations [here](https://github.com/SasView/sasmodels/issues/205).

It would be quite a bit of work to use volume fraction distributions. You would need to specify it to the UI and send it through the fitting perspective into sasmodels. Within sasmodels you would need to send this flag to the kernel calculators so they remove the volume weighting for the individual integration points.

butlerpd:
I suspect that most of those reporting as a  volume weighted (or z average) distribution are in fact the more "advanced users"?  I've fount that most material scientists do not know about anything but number averages and therefore intrinsically assume number average without realizing it. This is exacerbated by the rise of microscopy techniques which intrinsically report number averages and lead to very confused users when the number is different from the volume weighted average obtained by scattering.

[...]

PR #2738, opened 2024-01-17
Define distributions on the UI

smk78:
## Description

This PR addresses a criticism aired in #2390 that Users seem confused by what type of average our polydispersity distributions represent, even though this is clearly stated in the docs and on our FAQ webpage. The specific request was that something be added to the UI.

The solution provided is shown here:
image

## How Has This Been Tested?
Tested as a developer build on W10/x64.

- [ ] Code has been reviewed


rozyczko (changes requested):
This is technically correct, however...
You are overwriting the existing title which is defined in `FittingWidgetUI.ui`.
It would probably be better to modify the title in the UI file instead.

gonzalezma:
I did the change proposed by Piotr. I think that this can be merged, but now someone else should review it.

rozyczko (approved):
looks good now!

PR #2742, opened 2024-01-18
815 more example data

smk78:
## Description

This PR adds the example data supplied in SasView/tutorials#13 into a folder example_data/magnetic_data.

This is, of course, where example_data _used_ to live... A separate PR (https://github.com/SasView/sasdata/pull/58) will add them to the sasdata repo.


src/sas/example_data/media/testdata_help.rst
```diff
@@ -229,26 +242,42 @@ Coordinate Test Data
...
+DLS Test Data
+................
+**Example data files are located in the \\dls_data sub-folder.**
 
-mag_cylinder
-  - SLD file that describes a single cylinder of radius 2nm and length of 4nm.
-    The cylinder has equal nuclear and magnetic SLD, magnetised along the length.
+dls_monodisperse / dls_polydisperse
+  - DLS data from a very dilute dispersion of 3 nm polymer latex nanoparticles in
+    h/d-water.
```
gonzalezma, line 279: Better H/D-water?

src/sas/example_data/media/testdata_help.rst
```diff
@@ -271,6 +300,38 @@ ISIS_98940
...
+**Example data files are located in the \\magnetic_data sub-folder.**
+
+S50
+  - 2D and 1D sector data from 10 nm nanospheres in d8-toluene (c_Fe2O3 = 50 mg/ml,
+    ~1 vol%) at an aligning magnetic field of 1.5 T in horizontal direction.
+  - See `10.1088/1367-2630/14/1/013025 <https://dx.doi.org/10.1088/1367-2630/14/1/013025>`_
+
+.. ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
+
+NR Data
```
gonzalezma, line 318: Should this be renamed to reflectometry data, as there NR and XR data sets?

src/sas/example_data/media/testdata_help.rst
```diff
@@ -271,6 +300,38 @@ ISIS_98940
...
+^^^^^^^
+Neutron (NR) or X-ray (XR) Reflectometry data sets primarily contain the
+interfacial reflectivity as a function of *Q*.
+
+.. note:: The Refl1D reflectivity model-fitting software uses the same fitting
+         engine (Bumps) as SasView.
+
+NR Test Data
```
gonzalezma, line 326: Why are needed both NR and NR Test Data? Can they not be put in the same folder?

src/sas/example_data/media/testdata_help.rst
```diff
@@ -297,39 +358,6 @@ spheres2micron_long
...
-NR Test Data
-................
-**Example data files are located in the \\nr_data sub-folder.**
-
-NR_Ni_down_state / NR_Ni_up_state
-  - Polarised (spin down/up) NR data from a Ni multilayer sample.
-
-.. ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
-
 Save States
```
gonzalezma, line 361: Should be saved states, as in the description below?

src/sas/example_data/media/testdata_help.rst
```diff
@@ -374,12 +392,25 @@ project_new_style
...
+
 .. ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
 
-Upcoming Formats
-^^^^^^^^^^^^^^^^
-Data in this folder are in formats that are not yet implemented in SasView but
-which might be in future versions of the program.
+Other Formats
+^^^^^^^^^^^^^
+Data in the \\other_formats folder are in alternative varaiations of formats used
```
gonzalezma, line 411: Typo: Remove in and correct variations

gonzalezma (approved):
I think this can be merged. Minor comments/questions concerning the help test for Steve to check.

src/sas/example_data/media/testdata_help.rst
smk78: Changed.

Folder name also changed.

src/sas/example_data/media/testdata_help.rst
smk78: It's not that they are in different folders, rather that the original author of this help doc chose to use a type of data heading and then a test data subheading.

src/sas/example_data/media/testdata_help.rst
smk78: Changed.

Folder named also changed.

src/sas/example_data/media/testdata_help.rst
smk78: Changed.

butlerpd:
Two questions:
1. Should all these example data be added here? or in sasdata?
2. Should this also be merged into 6.0.0_release? for the upcoming beta release?

smk78:
> Two questions:
> 
> 1. Should all these example data be added here? or in sasdata?
> 2. Should this also be merged into 6.0.0_release? for the upcoming beta release?

To take the last question first, yes, we will want these changes in 6.0.0. This PR is against main on the assumption it will be merged into the release_6.0.0 branch before release?

https://github.com/SasView/sasdata/pull/58 pushes these same changes (except for the coordinate & saved state examples) to sasdata. So in theory whether we build off main/release_6.0.0 or sasdata, the example data and testdata_help should be the same...

krzywon:
This has a lot of overlap with sasview/sasdata#58, but most of the overlap was meant for sasdata, not sasview. I think only the save states from here should stay.

krzywon (dismissed):
This has a lot of overlap with https://github.com/SasView/sasdata/pull/58, but most of the overlap was meant for sasdata, not sasview. I think only the save states from here should stay and then it will be ready.

butlerpd:
Indeed @krzywon is right - most of these changed files no longer exist so we should not be overwriting the work of removing them unless there is a good reason. It looks in fact like all the changes to removed files were changed appropriately in SasView/sasdata#58 in which case we just need to revert everything here except changing the name of the `save_state` folder to `save_states`?

or did I miss something?  will the scraping and bundling of docs occur properly?

smk78:
Ok, so I note https://github.com/SasView/sasdata/pull/58 got merged and I've looked at the example data in the release_0_9_0 branch and it looks good.

It is, however, missing the save_states folder and, more importantly, the coordinate_data because (https://github.com/SasView/sasdata/pull/58#issuecomment-1904244052) those "are only importable through SasView". Whilst it wouldn't be the end of the world if we didn't include the save_states in the release, we do need to keep the coordinate_data.

There is also the question of testdata_help.rst; is that now a Sasdata file or still a SasView one (I couldn't see a /media folder in Sasdata so I'm assuming it's still SasView)? Note that the testdata_help.rst in this PR has updates.

So... @butlerpd @krzywon  what's the best way to proceed here?

butlerpd:
This is now rebased against 6.0 release.  Coordinate data and save project examples are the two folders remaining in sasview's example_data folder.  The plan I believe is to move them over as well in a later release. In the meantime the \example_data from both repos are being merged when building the installer according to @krzywon so there should be no issue with losing the coordinate data.

Still needing addressing is the magnetic folder which looks to have been moved including these files so I guess that folder can just be removed from this branch and it will be ready to merge?

butlerpd (approved):
This is now ready to be merged according to all the reviews

Issue #2740, opened 2024-01-18
Validation of formulas for constraints needs to be improved


gonzalezma:
**Describe the bug**
The validation of the formula used in a constraint definition is buggy, refusing even simple valid formulas. In addition, the same (or very similar) code appears in two different places: sasview\src\sas\qtgui\Perspectives\Fitting\ComplexConstraint.py and MultiConstraint.py. In both files, the functions validateFormula and validateConstraint are defined.

**To Reproduce**
1. Create a project with 2 fitting pages and define a simple constraint, e.g. M2.scale = M1.scale
2. In FitPage2 go to the scale parameter and select Edit constraint (menu appearing at right mouse button click)
3. In the edit window, change M1.scale to M1.scale+0.1
4. See error (screenshot below)

Note that in 6.0.0a1 Edit constraint will not work for parameters other than background and scale, but this has been already reported (issues #2355 and @2695) and PR #2711 corrects this.

**Expected behavior**
The new constraint (as long as it is a valid expression) should be accepted.

**Screenshots**
![image](https://github.com/SasView/sasview/assets/10974838/8d984a44-4c37-42e2-a60b-1f01373983ab)

**SasView version (please complete the following information):**
 - Version: 6.0.0.a1

**Operating system (please complete the following information):**
 - OS: Windows 10

gonzalezma:
Yes. Probably it would help. Where is available?

Issue #2735, opened 2024-01-17
Constraints of polydisperse parameters are ignored when a project is loaded


gonzalezma:
**Describe the bug**
After opening a project, the constraints referring to PD parameters are lost. 

**To Reproduce**
1. Save a project where at least 1 constraint for a polydisperse parameter has been defined.
2. Open the project.
3. Check the constraints defined.

**Expected behavior**
All the constraints defined in the saved project, including PD ones, should be loaded.

**SasView version (please complete the following information):**
 - Version: 6.0.0a1

**Operating system (please complete the following information):**
 - OS: Windows 10

GrahamRobertsW:
I'm able to load constraints on polysdisperse parameters, however I cannot reload any constraint set by right clicking and "Constrain parameter to its current value". any constraints set via right click don't reload properly, and cannot have further constraints placed on them.

Issue #2356, opened 2022-10-31
Add All button in Complex Constraint window does not work for polydisperse parameters


gonzalezma:
**Describe the bug**
The Complex Constrain interface allows setting automatically all the possible constraints between 2 FitPages. However, the polydisperse parameters are ignored.

**To Reproduce**
1. Load 2 data sets and define a model that includes PD params.
2. Go to the Constrained or Simultaneous Fit.
3. In Add constraints, set M2.scale = M1.scale and then press Add All.
4. See error:
    4.1 All the parameters of the model page are added to the list of constraints, e.g.:
              M2.scale = M1.scale
              M2.radius = M1.radius
              M2.thickness = M1.thickness
          BUT also non-fittable parameters (e.g. sld) are added to the list. This should probably be reconsidered.
    4.2 The PD parameters (e.g. radius.width) are not added, so they have to be included one by one.

**Expected behavior**
Only adjustable parameters should be added to the list of constraints when using the Add All button.
The adjustable PD params should also be added.

**SasView version (please complete the following information):**
 - SasView 5.0.6? Tested in ESS_GUI_constrain_branch to be merged, as before it was not even possible to define constraints for PD parameters.

**Operating system (please complete the following information):**
 - Windows 10

gonzalezma:
Fixed by PR #2743

Issue #2355, opened 2022-10-31
Edit constraints from the Model tab does not work properly


gonzalezma:
**Describe the bug**
When a constraint has been defined (in the Constrained or Simultaneous Fit tab), it appears as blue in the Model or Polydispersity tab to indicate that it is no more a free parameter. By right-clicking in its corresponding Value box, a context menu is available allowing to select/deselect, remove or edit the constraint.

The select/deselect options work as expected, but not the other two. In addition, the behaviour is not the same for main and polydisperse (PD) parameters.

The remove constraint button effectively removes the constraint, but it removes also ALL the other constraints defined in the Model tab. BUT the remove button works as expected (deleting only the selected constraint) when it is applied to a PD parameter, or to scale or background. It is only when removing a model parameter, that all the other model parameters, including scale and background (but not the PD params) are removed.

The edit constraint button fails to launch the box when a main model parameter is selected, giving an error in the Log Explorer:
17:09:02 - ERROR: Traceback (most recent call last):
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1050, in editConstraint
    if not np.any([FittingUtilities.isParamPolydisperse(p, self.model_parameters, is2D=self.is2D) for p in params_list]):
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1050, in <listcomp>
    if not np.any([FittingUtilities.isParamPolydisperse(p, self.model_parameters, is2D=self.is2D) for p in params_list]):
  File "C:\Users\gonzalezm\Git\sasview\src\sas\qtgui\Perspectives\Fitting\FittingUtilities.py", line 963, in isParamPolydisperse
    if '.width' in param_name:
TypeError: argument of type 'NoneType' is not iterable

When applied to scale, background or a PD parameter, the box to define the constraint is correctly launched, but then it does not accept any new definition (the box background becomes red and only the cancel possibility is available)

**Expected behavior**
Remove should remove only the constraint of the selected parameter.
Define should update the constraint definition.

**Screenshots**
![image](https://user-images.githubusercontent.com/10974838/199057550-608267b4-52ed-451c-8c98-72b6e635d778.png)

**SasView version (please complete the following information):**
SasView 5.0.5 (ESS_GUI_constrain_poly branch to test the PD params)

**Operating system (please complete the following information):**
 Windows 10

[...]

PR #2743, opened 2024-01-18
2355 edit remove constraints

gonzalezma:
## Description

Fixes the problems with the remove/edit constraint options, including the constraints on polydisperse parameters.

Note that the validation of the formula has been disabled, as it is not working properly and this will require much more thought to fix. A separate issue (#2740) has been created for this.

Fixes # 2355

## How Has This Been Tested?

Create a project, define a few constraints (including polydisperse parameters) and play removing and editing them from the right mouse button menu.

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -961,6 +961,8 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         # First, get a list of constraints and symbols
         constraint_list = self.parent.perspective().getActiveConstraintList()
         symbol_dict = self.parent.perspective().getSymbolDictForConstraints()
+        if model_key == 'poly' and 'Distribution' in constraint.param:
+            constraint.param = self.polyNameToParam(constraint.param)
         constraint_list.append((self.modelName() + '.' + constraint.param,
                                 constraint.func))
         # Call the error checking function
@@ -1076,6 +1078,8 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         constraint.validate = mc_widget.validate
 
         # Which row is the constrained parameter in?
+        if model_key == 'poly' and 'Distribution' in constraint.param:
+            constraint.param = self.polyNameToParam(constraint.param)
         row = self.getRowFromName(constraint.param)
 
         # Create a new item and add the Constraint object as a child
@@ -1090,6 +1094,8 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         params = [s.data() for s in current_list.selectionModel().selectedRows()
                    if self.isCheckable(s.row(), model_key=model_key)]
         for param in params:
+            if model_key == 'poly':
+                param = self.polyNameToParam(param)
             self.deleteConstraintOnParameter(param=param, model_key=model_key)
 
     def deleteConstraintOnParameter(self, param=None, model_key="standard"):
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -780,7 +780,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         # but let's check the correctness.
         assert len(selected_rows) == 2
 
-        params_list = [s.data(role=QtCore.Qt.UserRole) for s in selected_rows]
+        params_list = [s.data() for s in selected_rows]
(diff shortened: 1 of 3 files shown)


wpotrzebowski:
I am not sure if I have encountered know issue but I am getting following error:
Screenshot 2024-01-18 at 22 06 25

wpotrzebowski:
Error appeared when I tried to constrain polydispersity parameters
`22:05:57 - ERROR: Traceback (most recent call last): File "sas/qtgui/Perspectives/Fitting/ConstraintWidget.py", line 1129, in onAcceptConstraint File "sas/qtgui/Perspectives/Fitting/FittingWidget.py", line 961, in addConstraintToRow TypeError: '`

gonzalezma:
You are right. I think that this could be because at some point I was working on another branch to address #2356. This could mean that this branch would not work with the PD parameters. I thought that I had tested for both types, though ... Perhaps I was on the wrong branch? I don't know which is the best way to proceed. If you want, I can create a new branch with the fixes for both #2355 and #2356 and create a single PR request from there, and we forget about this PR and the other one. What do you think?

wpotrzebowski:
@gonzalezma I think we can merge both at the same time. We just need to make sure what to test.

gonzalezma:
That sounds good. For info, a large part of the issues with the PD parameters is that they have a "coding name" (e.g. radius.width) and a "human name" (e.g. Distribution of radius), and depending on the workflow sometimes constraint.param has the coding name and others it has the human name, so it is needed to check and convert names before doing a comparison.
When I was testing this, I was using a saved project (and then adding the PD constraints manually, as those are not loaded when loading a project - issue #2735) and everything worked fine. But when I start from scratch I get the same error as you. I have some hopes that the fix for #2356 could solve this, so I would propose the following:
1. Revise this PR ignoring the PD constraints for now.
2. Revise the PR for #2356.
3. Once that both are accepted and merged with the release branch we can check if add/remove/edit constraints works also as expected for the PD ones.
Does this sound reasonable to you?

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1075,6 +1078,8 @@ def editConstraint(self):
         constraint.validate = mc_widget.validate
 
         # Which row is the constrained parameter in?
+        if model_key == 'poly' and 'Distribution' in constraint.param:
```
rozyczko, line 1081: Shouldn't "Distribution" be enough? All PD parameter names start with "Distribution" and I'm not sure whether any have "poly" in them

src/sas/qtgui/Perspectives/Fitting/MultiConstraint.py
```diff
@@ -102,6 +102,9 @@ def validateFormula(self):
         """
         Add visual cues when formula is incorrect
         """
+        # temporarily disable validation, as not yet fully operational
+        return
```
rozyczko, line 106: This should be made a separate issue then

rozyczko:
> 1. Revise this PR ignoring the PD constraints for now.
> 2. Revise the PR for [Add All button in Complex Constraint window does not work for polydisperse parameters #2356](https://github.com/SasView/sasview/issues/2356).
> 3. Once that both are accepted and merged with the release branch we can check if add/remove/edit constraints works also as expected for the PD ones.


Or maybe just merge changes from #2356 into this PR to make it a single commit/PR?
Would that be an option?

gonzalezma:
@rozyczko If you think it's preferable to merge changes from #2356 to this PR and delete the other one, that's ok for me. Shall I go ahead with this solution?

gonzalezma:
OK. I just merged the branch corresponding to issue #2356 and everything seems to work. So you should be able now to check:
- Remove or edit a constraint, including the PD ones.
- Adding all constraints at the same time (see #2356)
I will remove the PR that I had for the other issue.

wpotrzebowski:
I may be doing something wrong but I have a problem with setting polydispersity parameters. 
I have 2 fit pages both with polydispersity enabled and at least one polydispersity parameter selected, however I cannot add polydispersity constraints. 
 
(They are not visible from drop-down menu)
Screenshot 2024-01-20 at 07 37 46

(And "Adding all" doesn't work either)
Screenshot 2024-01-20 at 07 38 13

gonzalezma:
That's strange. Have you checked the polydispersity for M2 too and made at least 1 PD param on M2 fittable?

wpotrzebowski:
@gonzalezma ahh, sorry my bad I used push not pull request installer that didn't have changes from the earlier merged PR.

wpotrzebowski (approved):

wpotrzebowski:
It works fine and in my opinion is ready to be merged.

PR #60, opened 2017-04-04
#918: Windows build now includes all test data folders.

krzywon:

PR #2754, opened 2024-01-19
fixed xaxis label for updated SESANS nomenclature from z to delta

### LINKED ISSUE #2663: Update nomenclature for SESANS plots
I am the chair of the canSAS SESANS Reproducibility Working Group.

After much discussion in our group, we recently agreed recommendations for nomenclature for SESANS measurements. This is an update of previous nomenclature, but we feel that these should stick for now. Hopefully! There was at least agreement from the three facilities that run SESANS instruments...

The paragraph about it is on our wiki page.
https://wiki.cansas.org/index.php?title=SESANS_Reproducibility_Working_Group

There are two main changes that would need to be made in SasView. The variable for spin-echo length needs to updated from z to δ. The unit for both length scale and wavelength needs to be changed from A to Å.

I just checked and the Å (ångström) rather than A (amp) thing is also the case in the normal SAS window.

caitwolf:
## Description

Changes the x-axis label on the residuals SESANS plots from z to delta.

Fixes #2663 

## How Has This Been Tested?
Load SESANS example data, send to fitting, and plot residuals. This PR is tied to another in sasdata that fixes the x-axis label on the 1D data plot for SESANS (https://github.com/SasView/sasdata/pull/60).

- [ ] Code has been reviewed


butlerpd (changes requested):
Not sure that it was necessary to create a new shortcut for delta rather than just using the latex, but there is mixed usage already so guess it is OK

spin Echo length label is now generated for the plot .. but not the residual. Otherwise looks good

caitwolf:
The other plot should be updated when the changes in https://github.com/SasView/sasdata/pull/60 are implemented

caitwolf:
@butlerpd did you have a chance to try the new installer that includes the edits from the paired sasdata branch and PR?

butlerpd:
Also thinking this should be merged also into 6.0.0 for the beta release?

caitwolf:
I agree about putting this into 6.0.0, as well as the paired [PR #60 in sasdata](https://github.com/SasView/sasdata/pull/60). The last steps to be completed are:
- [ ] switch the target branch from main to 6.0.0
- [ ] test with the updated installer that includes changes on the sasdata side
- [ ] fix the ci.yml file back to the correct sasdata branch before merging

butlerpd (approved):
I understand that the choice of using the unicode or defining the symbol in rst_prolog was based on how other symbols were used in each file.  I still think this is confusing and it would be nice to refactor everything for consistency but that is certainly beyond the scope of this PR.

Otherwise the code looks good both here and the paired PR in sasdata and does what it should do on windows. 

SOME CAVEATS:
* Given this is about plotting greek symbols it may be wise to check that it behaves as expected on MacOS
* More importantly, changes to the ci.yml file should be reversed PRIOR to the merge
* The paired PR should be merged at the same time
* This should definitely go into 6.0.0.

butlerpd:
@wpotrzebowski will check the mac and deal with merging

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
```diff
@@ -550,7 +550,7 @@ def residualsData1D(reference_data, current_data, weights):
     residuals.ytransform = 'y'
     if reference_data.isSesans:
         residuals.xtransform = 'x'
-        residuals.xaxis('\\rm{z} ', 'A')
+        residuals.xaxis('\\delta ', 'A')
```
pkienzle, line 553: Shouldn't residuals properties inherit from reference data, at least for the xaxis and xtransform?

wpotrzebowski:
Looks fine on MAC

butlerpd:
Oops ... NOT ready to merge to 6.0.0 ... and will not be ready once PR #2808 is merged into main.

Basically `data_formats_help.rst` has moved from `sasview` to `sasdata`.  Thus the changes to `sasview/src/sas/qtgui/MainWindw/media/data_formats_help.rst` should be removed and added back to `sasdata/docs/source/user/data/data_formats.rst`. I would recommend reverting the ci.yml change also at this point to make the rebasing clean and easy.

Otherwise rebasing works and this will be able to be rebased onto release_6.0.0 without problem

src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
butlerpd: probably, but clearly beyond the simple scope of this PR. I guess that is part of the monster combo of plot refactoring and remaining work in sasdata project?

butlerpd (changes requested):
see most recent comment needing fixing before this can be merged

caitwolf:
@butlerpd I updated the docs after the rebasing step to correctly display the new delta character and also show the updated sesans fitting screenshots

butlerpd (changes requested):
This looks pretty good now and is running nicely under release_6.0.0. I do have a couple of requests:
* there is a missing \ symbol in from of a lambda in a math entry (simple fix)
* Thanks for updating the plots in the fitting documentation. That too often gets forgotten! However, they have a very low resolution and hard to read. It would be good if we could get the same resolution as the other figures in that file.

butlerpd:
Also .. why are we bothering with changing the `rst_prolog`? It no longer is getting used right? If not, we should revert that to keep things clean.

butlerpd:
> * there is a missing \ symbol in from of a lambda in a math entry (simple fix)

Right ... I forgot that is actually in the sasdata repo .... +1 to whoever said multiple repos are a pain 😄 - consider this a comment in sasdata instead please. Sorry about that.

caitwolf:
@butlerpd Updated the "This help document was last changed by..." statement and last two sesans fitting images in the fitting help doc. PR is ready for review once the checks have completed.

butlerpd (approved):
The graphics are now much better and match the rest of the document.

I did notice two small things that would be nice to fix still - don't know why I did not catch before. For 6.0 the `test` folder has been replaced with the `example_data` folder. @smk78 has changed most of those but probably did not realize (or forgot -- I see he is the original author of that line) there was a reference to it in this document in the sesans section?

While we are at it the line about 

> In the screenshot above, the radius of the sphere has been increased from its default value of 50 Å to 5000 Å in order to get the transform to show something sensible.

can probably just be removed. I'm not sure why @smk78 wrote that unless he had a different example graphic he used? I think the previous sentence is sufficient. Alternatively it could be replaced with a line saying something like:

> In the screenshot above for example, the radius of the sphere could be increased from its default value of 50 Å to 5000 Å in order to get the transform to show something more sensible.

PR #2772, opened 2024-01-21
Run.py: Prioritize sibling modules over installed modules

krzywon:
## Description

This PR prioritizes sibling packages over installed packages when launching run.py. This should better enable developers to modify sibling packages without the need for pip installation. This fix only applies to sasmodels, sasdata, bumps, and periodictable, but could be generalized to any package held in requirements.txt.

- [ ] Code has been reviewed

### DIFF
--- a/run.py
+++ b/run.py
@@ -54,14 +54,19 @@ def prepare():
     # Find the directories for the source and build
     root = abspath(dirname(realpath(__file__)))
 
-    # TODO: Do we prioritize the sibling repo or the installed package?
-    # TODO: Can we use sasview/run.py from a distributed sasview.exe?
     # Put supporting packages on the path if they are not already available.
     for sibling in ('periodictable', 'bumps', 'sasdata', 'sasmodels'):
-        try:
-            import_module(sibling)
-        except:
-            addpath(joinpath(root, '..', sibling))
+        # run.py is only used by developers. The sibling directory should be the priority over installed packages.
+        #   This allows developers to modify and use separate packages simultaneously.
+        devel_path = joinpath(root, '..', sibling)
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -54,14 +54,19 @@ def prepare():
...
+        #   This allows developers to modify and use separate packages simultaneously.
+        devel_path = joinpath(root, '..', sibling)
+        if os.path.exists(devel_path):
+            addpath(devel_path)
+        else:
+            try:
+                import_module(sibling)
+            except ImportError:
+                print(f"The {sibling} module is not available. Either pip install it in your environment or clone"
+                      f"the repository into a directory level with the sasview directory.")
```
rozyczko, line 69: This is a good thing. 
Normally, I would just plop in a <sibling>.pth file into the default python installation directory pointing to the locally git cloned package, but this allows for a more streamlined approach.

run.py
```diff
@@ -54,14 +54,19 @@ def prepare():
...
-            addpath(joinpath(root, '..', sibling))
+        # run.py is only used by developers. The sibling directory should be the priority over installed packages.
+        #   This allows developers to modify and use separate packages simultaneously.
+        devel_path = joinpath(root, '..', sibling)
+        if os.path.exists(devel_path):
+            addpath(devel_path)
+        else:
+            try:
+                import_module(sibling)
+            except ImportError:
```
rozyczko, line 67: Should we not exit here? What's the point of continuing if we can't import sasdata/sasmodels?

rozyczko (approved):
works as expected.

wpotrzebowski (approved):

PR #2778, opened 2024-01-21
Initial implementation of the undelete mechanism for fitting tabs.

rozyczko:
## Description

User request addressed - it is easy (too easy) to delete a fitting tab. It would be good to be able to recover it somehow.

This PR is a cruft before the proposed Undo stack is implemented, allowing for more streamlined action of recovering UI errors.

Here, we just store the deleted tab and if needed, populate the perspective with it.
The key shortcut to restore the deleted tab is `Ctrl-T`, akin to the web browser standard.

## How Has This Been Tested?

Local testing on win10

- [ ] Code has been reviewed


gonzalezma:
Tested functionality and works as intended, the last FitPage deleted is recovered. Some remarks:
1. The recovered FitPage gets a new FitPage number. It would be nice (if possible) to be able to recover the FitPage with the same number as it had (see also next point).
2. If one deletes a FitPage, then creates a new one by sending some data to fitting, and then do Ctrl+Tab, one gets two FitPages with the same number, which I guess will cause problems.
3. When the Constrained or Simultaneous Fit is open, the recovered FitPage gets the number 302.
4. The removed FitPage disappears from the Simultaneous Fit tab and the recovered page cannot be added. Even after closing and opening this tab, the recovered FitPage will not appear there.

butlerpd:
Given this is an intermediate quick fix, I'm guessing this should be included in 6.0.0 beta?

wpotrzebowski:
Maybe I am not testing it properly but it seems to work fine before I start fitting (I can undo delete). However, as soon as I run the fit I cannot do that for any page. BTW, on MAC is Cmd + T, which we will need to document.

gonzalezma:
Tested functionality (on Windows 10) and almost everything worked well, including the issue reported by @wpotrzebowski. I loaded several FitPages, made a fit and then deleted one and I could restore it using Ctr+T.
The only problem that I could reproduce is the one with the onion model. If I send a data set to Fitting, select the onion model and then remove the page before doing anything else I cannot restore it. However, as soon as there is at least one parameter selected to be fitted the page can be restored, so for practical purposes I would say this is working as expected and can be merged.

rozyczko:
> Tested functionality (on Windows 10) and almost everything worked well, including the issue reported by @wpotrzebowski. I loaded several FitPages, made a fit and then deleted one and I could restore it using Ctr+T. The only problem that I could reproduce is the one with the onion model. If I send a data set to Fitting, select the onion model and then remove the page before doing anything else I cannot restore it. However, as soon as there is at least one parameter selected to be fitted the page can be restored, so for practical purposes I would say this is working as expected and can be merged.

Good observation, which led me to a tab object lifetime issue. This is now solved and the onion model restores just fine.
(apart of the `n` value, but we knew about this parameter and its lack of correct serialization)

rozyczko:
Pinging for re-review, please

gonzalezma (approved):
Worked fine in all my tests, including the onion model so I think this can be merged. I think this is a useful intermediate solution to the problem of deleting unintentionally a FitPage (while waiting for a full Undo button), so I agree with @butlerpd suggestion of including this also in release 6.0.0.

lucas-wilkins:
Shall we merge this? @rozyczko @butlerpd - the only reason haven't clicked the button is because of the 6.0.0 question.

butlerpd:
Good point Lucas.  This has been sitting around for a while and somehow did not get pulled into the beta even though it was approved back in April? If we are going to do a beta 2 anyway I guess my vote would be to merge this into 6.0? That would be in the grey area of violating our rule of no new bug fixes (for prior bugs) but ...?  I guess the issue is that it would have to be rebased? ...or cherry picking the 3 commits might be easier?

What do others think/vote?

smk78:
I vote merge. But it is against ```main```.

rozyczko:
new version cherry picked to merge cleanly with main is here: https://github.com/SasView/sasview/pull/3140

krzywon:
I think we can close this PR now that the other is against main. @rozyczko?

rozyczko:
> I think we can close this PR now that the other is against main. @rozyczko?

We definitely can - I just wanted to make sure we can get to it from the other PR. But we can link to closed PR of course.

PR #2781, opened 2024-01-21
Close all plots when required

### LINKED ISSUE #2776: "Close all plots" doesn't close the Bumps results panel
**Describe the bug**
"Close All Plots" should close all plots related to the current fit page. This action leaves now the Bumps Results Page open.

**To Reproduce**
1. Run a fit
2. Click on "Window"/"Close All Plots"
3. See plots closed except of the offending "Results Page"

**Expected behavior**
All plots should be closed, as the caption says.

 - Version: 5.0.6/6.0.0

 - OS: Win10

rozyczko:
## Description

Set the visibility flag on result panel properly

Fixes #2776 

## How Has This Been Tested?

Win10 build

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -1894,7 +1894,9 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         """
         Close all currently displayed plots
         """
-
+        # results panel
+        self.parent.results_frame.setVisible(False)
+        # plots
         for plot_id in PlotHelper.currentPlotIds():
             try:
                 plotter = PlotHelper.plotById(plot_id)


gonzalezma (approved):
Works well. All plots are now removed, including the bumps panel.

butlerpd:
Was about to merge but have a question: does this really get rid of the plot? or just hide it? If the latter will that not cause a series of these plots to build up in memory just hidden from the viewer?

butlerpd (approved):
Actually this is coming from bumps I guess and not a plot managed in our usual way? It looks like running a fit on a new tab does NOT open a new results panel, but overwrites the previous results panel from a fit to a different tab. So basically hiding the plot is a reasonable approach?

I think in the future there is a desire to keep the results panel for each fit separately. However, I'm thinking that will require redoing completely how this plot gets generated so am going to approve this PR

gonzalezma:
As far as I understand, this is a separate window that has nothing to do with the plots (this is why it was not closed) and the setVisible(False) action is the same one that happens when the window is closed manually. But I let Piotr confirm.

wpotrzebowski:
It works on Mac as expected. I would vote for merging it if the behavior of the bumps panel is questionable (and can be addressed separately).

Issue #626, opened 2019-03-30
Data Operations should allow for reasonable numerical fluctuations in q (Trac #493)


butlerpd:
Data Operations allows a user to perform simple arithmetic on two data sets (add, subtract, multiply, divide).  At this time it does not however interpolate so that a check is made before performing a requested operation that the two data sets have exactly the same number of q points AND that each q point is '''EXACTLY''' the same.  This is generally appropriate but will cause the operation to fail even if the q values are essentially identical but differ by even one in the last decimal place.  

The attached 2D files provided  by Jun-Li Lin of the University of Illionois are an example.  These are large data sets presumably generated by a SAXS camera? Somehow the qx of 0.0755253 (either positive or negative) is so close to 0.07552525 that in the blank data set it evaluated to 0.0755252.  This was the ONLY discrepancy yet the subtract operation fails as expected.  this is roughly a difference of 1.5ppm and is clearly the same value.  However in order to fix we will have to decide on a cutoff: how close is close enough?  A more complicated but more satisfying answer would be to allow for an option that sets the maximum allowed jiggle.  

Another option that might be interesting is to allow interpolation. interpolation

Migrated from http://trac.sasview.org/ticket/493
```json
{
    "status": "closed",
    "changetime": "2016-10-07T15:28:42",
    "_ts": "2016-10-07 15:28:42.671742+00:00",
    "description": "Data Operations allows a user to perform simple arithmetic on two data sets (add, subtract, multiply, divide).  At this time it does not however interpolate so that a check is made before performing a requested operation that the two data sets have exactly the same number of q points AND that each q point is '''EXACTLY''' the same.  This is generally appropriate but will cause the operation to fail even if the q values are essentially identical but differ by even one in the last decimal place.  \n\nThe attached 2D files provided  by Jun-Li Lin of the University of Illionois are an example.  These are large data sets presumably generated by a SAXS camera? Somehow the qx of 0.0755253 (either positive or negative) is so close to 0.07552525 that in the blank data set it evaluated to 0.0755252.  This was the ONLY discrepancy yet the subtract operation fails as expected.  this is roughly a difference of 1.5ppm and is clearly the same value.  However in order to fix we will have to decide on a cutoff: how close is close enough?  A more complicated but more satisfying answer would be to allow for an option that sets the maximum allowed jiggle.  \n\nAnother option that might be interesting is to allow interpolation. interpolation",
    "reporter": "butler",
    "cc": "",
    "resolution": "fixed",
[...]

PR #2782, opened 2024-01-22
Interpolation for operations

caitwolf:
## Description

This PR adds interpolation to 1D data operations in SasView. Previously the q-values for each dataset had to match perfectly. Now the overlap range of the two datasets in q will be determined and the second dataset will be interpolated if points don't match within the specified tolerance.

Fixes # (issue/issues) - I could not find the relevant issue; please comment below when found.

## How Has This Been Tested?

Load 1D example data and try different operations. The points used for the operation, including the subset of dataset 1 and the interpolated points of dataset 2 are plotted with a new color when the data operation is computed.

- [ ] Code has been reviewed


gonzalezma:
I think this addresses issue #626. It was a closed issue, but probably this extends the functionality by allowing larger variations in q?

butlerpd:
Interesting .... Was that fixed in 4.x and not ported to 5.x? As far as I know it is NOT fixed in any way in 5.x? people have had difficulty doing simple math because the data were not exactly the same? Or am I wrong?

caitwolf:
> Interesting .... Was that fixed in 4.x and not ported to 5.x? As far as I know it is NOT fixed in any way in 5.x? people have had difficulty doing simple math because the data were not exactly the same? Or am I wrong?

After further investigation of #626, the fix only allows for slight variation between q points in the two datasets. You still needed q values of the same length and each pair of q values from dataset 1 and dataset 2 had to be within that tolerance value. 

The fix in this PR allows for datasets with different sizes of q values and will perform an interpolation if a q-value in dataset 1 doesn't find a close match within the specified tolerance in q values of dataset 2.

caitwolf:
ci.yml in .github/workflows has been updated to use the paired branch from the sasdata pull request

src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
```diff
@@ -126,7 +130,7 @@ def onClose(self):
 
 
     def onCompute(self):
-        """ perform calculation """
+        """ perform calculation - don't send to data explorer"""
```
rozyczko, line 133: You decoupled `compute` and `save` but gave no visual cue to the user that the calculation succeeded. Previously, getting a new item in the DE was a GUI hint.
Consider adding a message to the status bar.

src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
```diff
@@ -141,15 +145,31 @@ def onCompute(self):
 
         self.output = output
 
+        self.updatePlot(self.graphOutput, self.layoutOutput, self.output, color='#000000',
+                        operation_data=True, data_op=[self.data1, self.data2], color_op=["#44AA99", "#CC6677"],
```
rozyczko, line 149: we already define colours for BG_RED and BG_WHITE as module variables.
Please consider doing the same for the numerical RGB values - those don't "feel" nice.

src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
```diff
@@ -304,13 +324,14 @@ def onCheckChosenData(self):
...
-                self.cbData2.setStyleSheet(BG_RED)
-                return False
+            # handling data with different q values implemented by CMW 1-21-2024
+            # elif self.data1.__class__.__name__ == 'Data1D'\
+            #         and (len(self.data2.x) != len(self.data1.x) or
+            #                  not all(i == j for i, j in zip(self.data1.x, self.data2.x))):
+            #     logging.error('Cannot compute 1D data of different lengths')
+            #     self.cbData1.setStyleSheet(BG_RED)
+            #     self.cbData2.setStyleSheet(BG_RED)
+            #     return False
```
rozyczko, line 334: please delete the commented out code, if it isn't going to be used.

src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
```diff
@@ -432,16 +468,39 @@ def updatePlot(self, graph, layout, data):
...
+            op_data.scale = 'linear'
+            op_data.x = np.array([1e-5, 1])
+            op_data.y = np.array([data, data])
+            op_data.dy = np.zeros(2)
+            op_data.dx = np.zeros(2)
+            plotter.plot(data=op_data, hide_error=True, marker='-', color=color)
+
+            plotter.show()
+
+
```
rozyczko, line 503: too many empty lines (should just one between methods)

src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
```diff
@@ -51,6 +54,7 @@ def __init__(self, parent=None):
         # push buttons
         self.cmdClose.clicked.connect(self.onClose)
         self.cmdHelp.clicked.connect(self.onHelp)
+        self.cmdSave.clicked.connect(self.onSave)
```
rozyczko, line 57: Please add checks for empty data/charts in `onClose`.
Currently clicking on `Save` with empty data gives
```
ERROR: Traceback (most recent call last): File "C:\projects\sasview\src\sas\qtgui\Calculators\DataOperationUtilityPanel.py", line 172, in onSave self.onPrepareOutputData() File "C:\projects\sasview\src\sas\qtgui\Calculators\DataOperationUtilityPanel.py", line 184, in onPrepareOutputData self.output.name = name ^^^^^^^^^^^^^^^^ AttributeError: 'NoneType' object has no attribute 'name'
```

src/sas/qtgui/Plotting/PlotterData.py
```diff
@@ -105,50 +105,41 @@ def __str__(self):
...
+                if result.dx is not None and other._operation.dx is not None:
+                    result.dx[i] = math.sqrt((self._operation.dx[i]**2 + other._operation.dx[i]**2) / 2)
+                if result.dxl is not None and other._operation.dxl is not None:
+                    result.dxl[i] = math.sqrt((self._operation.dxl[i]**2 + other._operation.dxl[i]**2) / 2)
+                if result.dxw is not None and other._operation.dxw is not None:
+                    result.dxw[i] = math.sqrt((self._operation.dxw[i]**2 + other._operation.dxw[i]**2) / 2)
             else:
                 b = other
-            
+
```
rozyczko, line 137: please remove whitespace

rozyczko (changes requested):
Minor issues raised.
One functional issue found.

src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
caitwolf: This is a great idea! Warnings and info statements have been added to logging that let the user know whether interpolation was performed on Data2 and that the computation was successful.

src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
caitwolf: The plotting colors have been set as module variables.

src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
caitwolf: Checks for empty data and graphs were added throughout. I also noticed while implementing these that it was possible to save an output from a previous operation before you hit compute. I've added code that resets the output data to None if any of the operation conditions change so that unknown data isn't accidentally sent to the data explorer.

butlerpd:
Questions: should this go into 6.0.0 beta or be left for 6.1? Also ... this kinda smells like something that should be in sasdata? but that would clearly be another PR?

caitwolf:
> Questions: should this go into 6.0.0 beta or be left for 6.1? Also ... this kinda smells like something that should be in sasdata? but that would clearly be another PR?

I realized I never provided a link for the paired pull request in sasdata: https://github.com/SasView/sasdata/pull/62, but I did update the installer so that it was using the paired sasdata changes when testing this branch.

I could go either way on 6.0 vs. 6.1. I need make some changes as a couple tests are failing after responding to the last review. A more thorough update of the documentation should happen along with this feature, especially for the interpolation component. Depending on the timeline of 6.0 I may lean closer to 6.1

butlerpd (changes requested):
There was a formatting error -- the complete review is belwo

butlerpd (changes requested):
I have yet to review the code in this section but have done a functionality test using windows 10.  I have several observations, some of which may not be for this PR but others I'm sure are?

* Using the core contrast AOT test data and the apoferritin test data reveals a problem. The low Q data of **both** data sets is not considered. It is true that the apoferritin data extends beyond the AOT data both at low and high Q but there is still a significant region of overlap that is ignored. I suspect that in fact the data is not being interpolated.  At high Q the points are dense enough that I believe they are within tolerance - or many are. It is actually hard to visualize because I cannot zoom in due to the zoom error (#2668) which is still not fixed in main from which this was branched. Clearly this needs to be fixed?
  * NOTE: this does not seem to be a problem for sesans data
* Loading a 1D or sesans data sets throws the following error
```python
Traceback (most recent call last): File "sas\qtgui\Calculators\DataOperationUtilityPanel.py", line 239, in onSelectData1 File "sas\qtgui\Calculators\DataOperationUtilityPanel.py", line 528, in updatePlot File "sas\qtgui\Calculators\DataOperationUtilityPanel.py", line 442, in operationData1D AttributeError: 'NoneType' object has no attribute 'x'
```
* Trying to add sesans and sans1D data returns the error
```python
22:58:10 - ERROR: Unable to perform operation: x ranges of two datasets do not overlap.
``
That is correct in most cases but for SAXS data at high Q that might not always be true? What happens then? should Data on two different x scales (z vs 1/A) be able to be added?  1/nm vs 1/A should still be doable IF put on the same scale but sesans and not sesans?
* Help file needs to be updated. It could have used some work before but this change definitely needs to be added.
* I am unclear on the purpose of the `reset` button other than clearing the graphs in the data operations panel. Some quirks noted that are probably pre-this PR in which case they should be documented in issues:
  * Changing the operation (+ to * for example) resets the output data panel. However changing the input data (either 1 or 2) does not. This seems inconsistent behavior.
  * Once saved the new data shows up in data explorer and in the drop downs in the data operations panel. However loading in new data to the explorer does **NOT** show up in the drop down. Furthermore, clicking the `reset` button does not help either (so what is `reset` for?).
  * Finally, removing data from the data explorer with or without resetting the data operations panel does not remove the data. No error is thrown however when choosing those deleted files as they seem to still be somewhere?. Closing and re-opening the panel does however clear the now deleted files from the dropdowns.

wpotrzebowski:
It seems that https://github.com/SasView/sasdata/pull/62 needs to be merged first?

butlerpd:
@caitwolf and @DrPaulSharp @krzywon. Should this be closed as obsolete given work in sasdata? Or is it worth leaving in case someone has time to fix the issues?

PR #2788, opened 2024-01-22
Edited PR template

lucas-wilkins:
Tried to make the PR checklist more usable... now like this

**Installers**

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,13 +8,20 @@ Fixes # (issue/issues)
 
 Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.
 
-## Review Checklist (please remove items if they don't apply):
+## Review Checklist:
 
-- [ ] Code has been reviewed
-- [ ] Functionality has been tested
-- [ ] Windows installer (GH artifact) has been tested (installed and worked) 
-- [ ] MacOSX installer (GH artifact) has been tested (installed and worked) 
-- [ ] User documentation is available and complete (if required)
-- [ ] Developers documentation is available and complete (if required)
-- [ ] The introduced changes comply with SasView license (BSD 3-Clause)
+[if using the editor, use `[x]` in place of `[ ]` to check a box]
+
+**Documentation** (check at least one)
+- [ ] There is **nothing** that needs documenting
+- [ ] Documentation changes are **in this PR**
+- [ ] There is an **issue** open for the documentation (link?)
+
+**Installers**
+- [ ] There is a chance this will affect the **installers**, if so
+  - [ ] **Windows** installer (GH artifact) has been tested (installed and worked) 
+  - [ ] **MacOSX** installer (GH artifact) has been tested (installed and worked) 
+
+**Licencing** (untick if necessary)
+- [x] The introduced changes comply with SasView license (BSD 3-Clause)
 


wpotrzebowski (commented):
I like changes to documentation and licensing section but I am not sure about installers. I personally test all PR from installer as historically we had issues that it worked locally and Mac installer crashed (after some time). This was related to PyQt and I guess things are more stable now and we have better synchronization between GH actions and dev env. The other question is if one can actually judge if the change affects installer as in principle any change affects it. So maybe we need to reverse logic here? (Don't need to test installers if you are sure their not affected)

butlerpd:
as I recall it was and is way more than mac OS issues and QT. In fact it was a serious problem in 4.x which was Wx! More importantly I think it relates to the version of all the dependencies which are rarely, if ever, the same on a developer's local environment as it is in the installer? There may be other differences as well. After a number of fiascos, including a point release that had been mostly tested on build machines, there was a consensus that no change is fully tested unless it runs under the installer as that is what will be delivered to the end user.

So I guess I am agreeing with @wpotrzebowski ?

llimeht:
@lucas-wilkins I started work on two parts of that but didn't make it that far. 

-  In the `test-installer` job, current CI tries to start the application after running the installer, at least on Windows and Linux. 
     - If it doesn't even start and there's a fatal error, I *think* it will fail CI. In my testing, I think I pushed a deliberately broken installer at it and it failed, but it is worth checking that. (I was looking at this a couple of years ago when we were fighting with `setuptools`)
     - It is currently a pretty weak smoke-test and could be tightened up somewhat (run a batch fit for instance? export some plots?). 
     - It also needs to be extended to macos by someone who knows something about that OS.
- I also experimented with a separate job to simulate clicks on the UI, take screenshots and compare with some pre-approved ones. 
     - It was kind-of working under linux.
     - I never found a way of reliably making the screenshots on CI windows and so I just put this project aside for the time being. If others know ways of getting that to work on github CI, then let's look at it again.
     - Testing under macos might well be easier but it has the same blocker as `test-installer` before even getting to the screenshotting step, namely that I haven't got a starting point to even run the installer.
     -  I never got past some crude prototyping of how to the tools might work. We'd want to build an extensible and maintainable tool for this task - I think it would be easy to make something nice using pytest and syrupy and would probably want some CI bot integration to make PRs to update the screenshots when they changed and were accepted.

As a side-note, I suspect that running the UI-part of the test suite (`src/sas/qtgui/`) would be more benefit for the amount of effort required, even within the context of the installer being painful to currently test.

andyfaff:
For `refnx` I check that a pyinstaller app starts using this [script](https://github.com/refnx/refnx/blob/main/tools/app/check_app_starts.py). For a mac app one has to [navigate into the bundle](https://github.com/refnx/refnx/blob/main/.github/workflows/pythonpackage.yml#L156), but on WIndows you just use it on the executable.

wpotrzebowski:
The issues I was referring to, already happened after the transition to PyQt and if I remember correctly there was a problem with PyQt=5.15. It was a bit more tricky than just checking the installer because the program was failing after few minutes. But generally, I agree with the "smoke-test" approach as I think some will check installers regularly anyway.

butlerpd:
Looking at this again, I believe that what @lucas-wilkins is trying to do here make the template more appropriate for the PR submitter. 

In that vein, I think the discussion about testing with installers is moot a far as this PR is concerned since both `code review` and `functionality review`, including testing with installers rather than just locally, have been removed from the template (not something the author is allowed to do) 

The problem with testing installers is that, as a community we tend to have very short memories and could easily revert to testing on development platforms only before approving. I think the saying goes: "those who forget history are doomed to repeat it."

**suggestion**
* We create a wiki page with (short list) instructions on reviewing PR separated by coding vs functionality
* We add a line a the bottom of the template reminding reviewers of the link to the above (and to the wiki page on coding practices).
* We then approve this PR - I think overall everyone agrees?
* We create a priority issue about adding some extra testing as suggested above

Thoughts?

PR #2793, opened 2024-01-22
Quote file names in ConvertUI

### LINKED ISSUE #2791: Installation sasview developers_Convert UI does not create output if a space is present in thefolder name in the path
**Describe the bug**
During the installation of sasview developers enviroment, this kind of error was happaneng (see picture below):

**To Reproduce**

**Expected behavior**

**Screenshots**

 - Version: [e.g. 5.0.4]

 - OS: [e.g. Windows 10]

**Additional context**

MariannaGerina:
## Description

Please include a summary of the change and which issue is fixed. List any dependencies that are required for this change.
Fixed issue: convertUI does not create a outfiles when a space is present in a folder of the path. The issue was fixed properly quoting the files name in the convert UI.
Fixes #2791 

## How Has This Been Tested?
After this change,  run.py can start the application

- [ ] Code has been reviewed

### DIFF
--- a/src/sas/qtgui/convertUI.py
+++ b/src/sas/qtgui/convertUI.py
@@ -24,8 +24,8 @@ def pyrrc(in_file, out_file):
     """
     Run the qt resource compiler
     """
-    run_line = f'pyside6-rcc "{in_file}" -o "{out_file}"'
-    os.system(run_line)
+    run_line = ["pyside6-rcc", in_file, "-o" ,out_file]
+    subprocess.run(run_line)
 
 
 def pyuic(in_file, out_file):
@@ -34,8 +34,8 @@ def pyuic(in_file, out_file):
     """
     in_file2 = os.path.abspath(in_file)
     out_file2 = os.path.abspath(out_file)
-    run_line = f'pyside6-uic "{in_file2}" -o "{out_file2}"'
-    os.system(run_line)
+    run_line = ["pyside6-uic", in_file2 "-o", out_file2]
+    subprocess.run(run_line)
 
 def file_in_newer(file_in, file_out):
     """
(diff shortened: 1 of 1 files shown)


llimeht:
Good catch, and I suspect there are others like this elsewhere in the code.

Trying to robustly quote filenames will always be problematic (what happens if my directory name has a " in it?) and `os.system` passes the command through a subshell which can do odd things too (what if the directory name has a $ or \` in it?). `os.system` also look after the output (just gets dumped to the terminal, probably not an issue for these uses). The current code also doesn't check that the operation succeeded, which it really should.

Rather than trying to quote filenames to put them in a string and give them to `os.system` so that the shell can then try to split the string back up and then execute the command, how about putting the command into a `list` and giving it directly to `subprocess.run`. This has the benefit that the command doesn't get interpreted by a shell (with the default `shell=False`) so no quoting is needed. Additionally, output is easily captured, and it can raise an exception on failed operations.

[https://docs.python.org/3/library/subprocess.html#subprocess.run](https://docs.python.org/3/library/subprocess.html#subprocess.run)

butlerpd:
I think this should be against 6.0.0 not main. Particularly given the comments by @llimeht who points out that this is a tiny band-aid for a much larger problem than continuously bites us.  However his suggestion seems like it will require skills beyond mine or the author's and the question is whether anybody else will be able to address this "properly" in the near future? If not I would recommend we merge this into 6.0.0 and create an issue with @llimeht suggestions given it fixes a problem we had at the contributor camp?

butlerpd:
@lucas-wilkins suggests fixing the way @llimeht suggests but only for this issue before merging "should not be hard to do" - but says it is better than no change.

butlerpd:
@lucas-wilkins suggests this will be "easy" ... looking this up and reading a bit about it do you mean we should replace
```python
run_line = f"pyside6-rcc {in_file} -o {out_file}"
os.system(run_line)
```
with
```python
run_line = ["fpyside6-rcc", in_file, "-o" ,out_file]
subprocess.run(run_line)
```
??
It doesn't seem like any of the kwargs are necessary here? all defaults should be good?
This seems a bit scary to a white belt like me?

butlerpd:
OK thanks ... I'll give it a go then

Issue #2812, opened 2024-02-20
Ubuntu (latest) tests failing


klytje:
The CI system for Ubuntu (latest) currently fails when installing the X11 libraries, or more specifically the `libegl-dev` library ([case 1](https://github.com/SasView/sasview/actions/runs/7972097994/job/21765907856#step:5:64), [case 2](https://github.com/SasView/sasview/actions/runs/7934612039/job/21665970452)),  Going through the error report reveals that the cached download link (http://security.ubuntu.com/ubuntu/pool/main/m/mesa/libegl-mesa0_23.0.4-0ubuntu1%7e22.04.1_amd64.deb) is dead since the library has been updated. 

I've solved the issue in my own PR by adding a simple `sudo apt-get update` before attempting to download the libraries, but I don't know enough about the CI system to know if this is a good solution.

llimeht:
Interesting finds!

> I've solved the issue in my own PR by adding a simple `sudo apt-get update` before attempting to download the libraries, but I don't know enough about the CI system to know if this is a good solution.

Yes, that's a reasonable solution to handle security updates having been pushed out but the base images haven't been regenerated recently.

Perhaps a quick PR just for that change so that it can be merged and CI works for everyone would be worthwhile?

> Bonus info I discovered while checking previous CI runs: The Try running the installation (Linux) step hasn't been working for months due to a missing EGL library:

I assume that's a related problem that a package wasn't installed as desired or perhaps the library location changed between releases. Do you know what the right package is for the current Ubuntu images?

Yes, we should also do better in that testing... it's supposed to fail as you say. Could you please file a separate issue for that one so we don't forget?

klytje:
I implemented the changes in PR #2814, but it seems like the issue have solved itself. Should we merge it in anyway to prevent similar issues in the future?

llimeht:
Thanks @klytje! Yes, it's still worth fixing because it will happen again next time there is a delay between a security update coming out for packages used in CI and the images being regenerated.

PR #2814, opened 2024-02-29
Fix Ubuntu (latest) tests

klytje:
As mentioned in #2812, the CI system was previously failing for ubuntu (latest). Though the problem seems to have solved itself, this PR ensures it will not happen again.

### DIFF
--- a/.github/workflows/matrix.py
+++ b/.github/workflows/matrix.py
@@ -28,7 +28,7 @@ jobs = []
 os_release_list = [
 #    'ubuntu-20.04',
 #    'windows-latest',
-#    'macos-latest',
+#    'macos-latest', 
 ]
 
 # List of OS images to use for release tests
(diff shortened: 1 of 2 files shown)


.github/workflows/matrix.py
```diff
@@ -28,7 +28,7 @@
 os_release_list = [
     'ubuntu-20.04',
     'windows-latest',
-    'macos-latest',
+    'macos-latest',
```
llimeht, line 31: Would be nice to not load in a whitespace diff

llimeht (approved):
Thanks, looks good (other than the whitespace change)

.github/workflows/matrix.py
klytje: fixed

klytje:
Merging is blocked because `main` is protected; is this due to a lack of permissions from my side or because I should target a different base to merge into?

butlerpd:
Thanks @klytje! Nice work. The locked main branch is a hopefully brief holding pattern as we sort out what should be merged to `main` and what should be merged into `release_6.0.0` first and then pulled into `main` when `release_6.0.0` is merged back to `main`.  

I'm realizing as I write this the current  PR could be at risk since the `ci.yml` files have diverged and currently we have to resolve conflicts whenever merging `release_6.0.0` into `main`. I'm feeling this could lead to an accidental overwrite of this fix?

Ideally we should have a common `ci.yml`  which I think @krzywon has suggested?  Thoughts?

PR #2819, opened 2024-03-12
Revert "Quote file names in ConvertUI"

wpotrzebowski:
It seems that Mac Installer is failing because of this merge. Testing if this actually the case

### DIFF
--- a/src/sas/qtgui/convertUI.py
+++ b/src/sas/qtgui/convertUI.py
@@ -24,8 +24,8 @@ def pyrrc(in_file, out_file):
     """
     Run the qt resource compiler
     """
-    run_line = ["pyside6-rcc", in_file, "-o" ,out_file]
-    subprocess.run(run_line, check=True)
+    run_line = f"pyside6-rcc {in_file} -o {out_file}"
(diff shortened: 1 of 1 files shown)


krzywon:
I added a comma in line 37 so the line is now `run_line = ["pyside6-uic", in_file2 ", -o", out_file2]`. I have no preference either way.

butlerpd:
@wpotrzebowski is reverting to the original that was changed (twice) based on @llimeht suggestion and concurence of @lucas-wilkins. It should of course have had a comma so that was broken.

Of course if using the subprocess is causing the problem for mac then maybe we have to go back to that. But I suspect the issue was the missing comma which is required in the subproccess call. In which case I would think we should stick with @krzywon fix instead.

butlerpd:
As demonstrated by @caitwolf we also need to add `import subprocess` at the top. Duh!! Clearly I merged too fast and didn't even check the build. Very very bad form on my part. Apologies.

wpotrzebowski:
Closing this PR as #2821 resolved the underlying problem

Issue #2229, opened 2022-09-30
SESANS theories require data to initialise


lucas-wilkins:
When making theoretical models for SESANS, it is necessary to load a data file. This appears to be an issue with initialisation, but requires some investigation.

smk78:
It is probably unfair to call this a Defect. :-) The issue is that SasView grew up as an application for inverse-space (ie, Q) data analysis. SESANS data is in real-space. Thus one needs a way to 'tell' SasView that it is fitting real-space data. 

Addressing this issue will touch on matters raised in several other issues under the SESANS label.

wimbouwman:
I is indeed unfair to call it a defect. In Delft we use the theoretical models for SESANS frequently with students. It is no problem for them to load first a data-file.

butlerpd:
Priorities is another question altogether 😄

lucas-wilkins:
It's pretty hard to build a picture of what needs changing and what resources said change will require if there's not a catalogue of the behaviours that need to be changed.

smk78:
When SasView initialises it assumes that any data read, or any model to be calculated, is Q-space. Where it is reading data, in a suitable self-describing format, there is an obvious route to changing this behaviour (assuming the data is correctly described) and I think @krzywon has been going part way down this road in some of his branches. Note that at present the only thing telling SasView that some input data is SESANS is the file descriptor, not the data itself. But where you are just wanting to compute a model in real-space it is less clear how to proceed. There are probably several options. And those options mustn't make life a pain for those using SasView for 'traditional' Q-space fitting (ie, the core User community - no offence intended SESANS people!). Maybe there is a potential role here for the proposed Preferences dialog? But as @butlerpd & I have alluded too, if we're going to widen the range of data that SasView 'recognises' then there are other types too (eg, DLS & NSE with time bases instead of Q). Also, not all models will be applicable all of the time. So do we then add something to model descriptions to identify their validity? What then happens with backwards compatibility?

krzywon:
Currently, there is a flag in the `Data1D` and `Data2D` objects called `isSesans`. `sasmodels` checks this flag and other items in the data to determine the type of resolution and if any transforms are required. To do what we want here, and to keep it extensible for future analysis methods like DLS and NSE, adding a combo box with different analysis types (SAS and SESANS currently) would work. If no data is sent to the fit window, the user would be free to select the type of analysis desired. If data is present, the type should be auto-selected based on the data and locked.

[...]

PR #2839, opened 2024-04-01
SasData 66: Ensure realistic resolution values loaded from data

### LINKED ISSUE #2095: Enable resolution option from the start on theory models
Re ticket #1686, @rozyczko points out that the problem is that the theory curve must be calculated before resolution can be applied as resolution smearing requires a knowledge of the curve. However, from the user's perspective, that is totally unintuitive and can lead to frustration thinking that it is not allowed.

As soon as a new fit panel is instantiated, the resolution tab should have all resolution options available in the drop down. If the curve has not yet been computed, this will require computing the curve without smearing first and then applying smearing. This can however all be done behind the scenes from the user's point of view.

Nothing obvious comes to mind.

**Additional context**

krzywon:
## Description

This ensures the data resolution is not used when all resolution values loaded from a data set are 0.0. A bug was also fixed where the selected resolution was blank instead of None after selecting a model. In addition, custom pinhole and slit smeared resolutions can be applied to theory data once a model has been selected. This does NOT allow the Hankel transform to be applied to theory data due to differences inhow resolution functions are applied for SESANS data.

Fixes sasview/sasdata#66
Fixes #2095 (does not touch SESANS-specific #2229)
 
## Review Checklist:

  - NOTE: This is all inline documentation to ensure future changes do not change the behavior.

**Installers**


butlerpd:
Functionally this seems to fix the issue reported in https://github.com/SasView/sasdata/issues/66. However I don't think it really addresses the #2095 issue since it still requires one to click compute before being able to apply the smearing (as was always the case). However there may be more urgency to fix this based on new conversation in #2095?

caitwolf (approved):
Revisions look good and the resolution selection operates as described. I tested the Windows installer but could be good to have someone check the Mac installer?

caitwolf:
> Functionally this seems to fix the issue reported in [SasView/sasdata#66](https://github.com/SasView/sasdata/issues/66). However I don't think it really addresses the #2095 issue since it still requires one to click compute before being able to apply the smearing (as was always the case). However there may be more urgency to fix this based on new conversation in #2095?

I noted that the last change does now address #2095. Although I noticed that when smearing is applied to theoretical data, the data will be plotted immediately after the compute button is pressed. It will also briefly flash the theoretical data without smearing before showing the smeared data. When smearing is not applied to theoretical data, the compute button needs to be pressed followed by Compute/Plot in order to plot the data.

butlerpd:
Pursuant to @caitwolf review and some more testing with the latest installer on windows, there is indeed a discrepancy in behavior during the initial calculation of newly selected model between doing so with vs without resolution. I would however argue, and I think there is a ticket somewhere asking for this, that the correct behavior should be the one seen when resolution is requested.  Thus the change should be to the non-resolution functionality which is not part of this PR. So I thinks we should merge this.  On the other hand, I was unable to see the "flash" of the first non smeared calculation plotted as suggested by @caitwolf. That may be because my computer is too fast. Alternatively, if @caitwolf was testing from the development environment maybe it is not present in the windows installer?

That said, doing some more careful testing I note that there are a couple of small subtleties that I suspect should not be part of this PR?
* Entering a new value for one of the parameter and hitting return acts as if one has hit the `Calculate` or `Compute/Plot` button: it will only calculate but not plot the very first time (but the label of the button changes) and will calculate AND plot subsequent times. This is as expected in some sense though highlights the oddity of the initial `Calculate` only, when all subsequent actions show the new plot immediately.
* This equivalence is true also when resolution is selected so that in that case entering a value and hitting return (whether that is after entering the resolution value or a new value for a parameter) the calculation and plotting occur together even on the very first enter. Again I would argue this is as it should be.
  * HOWEVER, if the resolution value is entered without hitting return and the `calculate` button is pressed immediately after, nothing happens at all (the label on the button does not change either) and a second press of the button is required. This seems very odd!!

I also note that the polydispersity tab is active even though the polydispersity checkbox was not checked. However, I believe that was addressed in another PR?

krzywon:
> if the resolution value is entered without hitting return and the calculate button is pressed immediately after, nothing happens at all (the label on the button does not change either) and a second press of the button is required. This seems very odd!!

I'm seeing this in the installed Windows version, but not every time. A click is required to remove the focus from the dQ/Q box for some reason.

> On the other hand, I was unable to see the "flash" of the first non smeared calculation plotted as suggested by @caitwolf.

I see this in the installed version for times where Clicking the Calculate button the first time works.

> I also note that the polydispersity tab is active even though the polydispersity checkbox was not checked. However, I believe that was addressed in another PR?

Correct. This branch is a bit behind the release branch. Should I push a merge commit?

PR #2842, opened 2024-04-08
Add a button group to allow for proper logic of radio buttons

rozyczko:
## Description

When opencl is not present, the only option on the GPU Options dialog is "No OpenCL".
However, this "radio" button could be deselected and deselecting it caused issues when running tests.

This has been fixed by adding a QButtonGroup to add all the radio buttons into a layout.

Fixes #2841 

## How Has This Been Tested?

Local Win10 build

## Review Checklist:

[if using the editor, use `[x]` in place of `[ ]` to check a box]

**Documentation** (check at least one)
- [X ] There is **nothing** that needs documenting
- [ ] Documentation changes are **in this PR**
- [ ] There is an **issue** open for the documentation (link?)

**Installers**
- [ ] There is a chance this will affect the **installers**, if so
  - [ ] **Windows** installer (GH artifact) has been tested (installed and worked) 
  - [ ] **MacOSX** installer (GH artifact) has been tested (installed and worked) 

**Licencing** (untick if necessary)
- [x] The introduced changes comply with SasView license (BSD 3-Clause)

Issue #2846, opened 2024-04-09
Mac installer from main branch is not starting


wpotrzebowski:
**Describe the bug**
After starting Mac installer it crashes silently. Running it from the terminal gives an error

```
Traceback (most recent call last):
  File "sasview.py", line 15, in <module>
  File "sas/cli.py", line 135, in main
  File "sas/system/lib.py", line 17, in setup_sasmodels
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "sas/sascalc/fit/models.py", line 18, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "sasmodels/sasview_model.py", line 30, in <module>
    from . import weights
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "sasmodels/weights.py", line 11, in <module>
    from scipy.special import gammaln  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "scipy/special/__init__.py", line 777, in <module>
  File "scipy/special/_ufuncs.pyx", line 1, in init scipy.special._ufuncs
ModuleNotFoundError: No module named 'scipy.special._cdflib'
[41201] Failed to execute script 'sasview' due to unhandled exception: No module named 'scipy.special._cdflib'
[41201] Traceback:
Traceback (most recent call last):
  File "sasview.py", line 15, in <module>
  File "sas/cli.py", line 135, in main
  File "sas/system/lib.py", line 17, in setup_sasmodels
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "sas/sascalc/fit/models.py", line 18, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "sasmodels/sasview_model.py", line 30, in <module>
    from . import weights
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "sasmodels/weights.py", line 11, in <module>
    from scipy.special import gammaln  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "scipy/special/__init__.py", line 777, in <module>
[...]

PR #2848, opened 2024-04-09
Add a button group to allow for proper logic of radio buttons

rozyczko:
Cherry picked commit from https://github.com/SasView/sasview/pull/2842 rebased against `release_6.0.0`


krzywon (approved):
Assuming the build is successful, this should be okay to merge.

butlerpd:
Testing with windows installer and it fails to load with the same ` no module named 'scipy.special._cdflib'` error as recorded recently in #2846 and #2851.  Re my comment to PR#2851 about windows and release 6.0....?

PR #2851, opened 2024-04-11
Upgrading pyinstaller

wpotrzebowski:
## Description
It seems that we need to use earlier version scipy (1.12), to make MAC installer working. 
As always in such case, the fix should be removed once proper bug fix from scipy is provided. 
Addressing https://github.com/SasView/sasview/issues/2846

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,7 +202,7 @@ jobs:
       - name: Install utilities to build installer
         if: ${{ matrix.installer }}
         run: |
-          python -m pip install pyinstaller==5.13.2
+          python -m pip install pyinstaller
 
       - name: Build sasview with pyinstaller
         if: ${{ matrix.installer }}
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -34,4 +34,4 @@ superqt
 pyopengl
 pyopengl_accelerate
 sphinx
-scipy==1.11.4; platform_system == "Darwin"
+scipy==1.12.0; platform_system == "Darwin"
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -34,4 +34,4 @@ superqt
 pyopengl
 pyopengl_accelerate
 sphinx
-scipy=1.11.4; platform_system == "Darwin"
+scipy==1.11.4; platform_system == "Darwin"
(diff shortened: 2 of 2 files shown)


butlerpd:
Does this not affect windows as well? (See Julius' comment in #2846).  What about release 6.0? Does it not need t be downgraded as well?

butlerpd:
As suggested by @andyfaff in #2846, this is apparently as PyInstaller issue (see [stack overflow]( https://stackoverflow.com/questions/78287158/modulenotfounderror-no-module-named-scipy-special-cdflib-with-scipy-1-13-0) discussion started 8 days ago)  more than a scipy issue and should be fixed by upgrading to PyInstaller 6.6.0. As reported in the [PyInstaller 6.6.0 Changelog](https://pyinstaller.org/en/stable/CHANGES.html) published Friday: 

> ### Hooks
> * Update PySide6.Qt3DRender hook for compatibility with PySide6 6.7.0 (add hidden import for PySide6.QtOpenGL module). ([#8404](https://github.com/pyinstaller/pyinstaller/issues/8404))
> 
> * Update scipy.special._ufuncs hook for compatibility with SciPy 1.13.0 (add scipy.special._cdflib to hidden imports). ([#8394](https://github.com/pyinstaller/pyinstaller/issues/8394))
> 

Suggest we do **not** revert scipy but instead upgrade PyInstaller. This should be done Post Haste as it affects ALL builds (window and mac, release and main)

wpotrzebowski:
Ok, indeed it seems that pyinstaller version with the fix was released 3 days ago.  I noticed that we have quite discrepancy between main and release_6.0.0

For release 6.0.0

```
       - name: Install utilities to build installer
        if: ${{ matrix.installer && startsWith(matrix.os, 'windows') }}
        run: |
          python -m pip install pyinstaller==5.13.2
      - name: Install utilities to build installer
        if: ${{ matrix.installer && !startsWith(matrix.os, 'windows') }}
        run: |
          python -m pip install pyinstaller==6.3.0

```
while main installs `python -m pip install pyinstaller==5.13.2` for all platforms. However, when I tried to upgrade Windows in January, there were some issues and we decided to keep with the older version. 
So I agree we should upgrade pyinstaller but it may not be straight forward.

wpotrzebowski:
Closing as won't be able to resolve Windows at the minute

PR #2859, opened 2024-04-20
Debye efficiency

klytje:
## Description

The previous Debye implementation for the generic scattering calculator has been replaced with external calls to the AUSAXS library, which is significantly faster. With the foundation already in place, future additions relying on this same library will be much easier to implement. 

The following dependencies are introduced:
* cpufeature to determine the CPU architecture and select an optimized version of AUSAXS,
* AUSAXS itself as shared libraries for all platforms. Currently these are fetched from a new empty sasview repository (https://github.com/SasView/ausaxs). I intend to modify the download path to instead point to the actual AUSAXS repository once it is opened to the public, hopefully within the next couple of months.

## How Has This Been Tested?
I have added an additional test case in `utest_sas_gen.py` where the previous calculation is compared with AUSAXS for a few different systems. 

## Issues
_Major_
- [x] Using the shared library approach on Windows results in a deadlock. Perhaps natively compiling my library with MSVC can help. This is a priority issue as it blocks future developments utilizing my library (_SAXS, water models, ..._)
_Solved: I refactored my library to a different threading model, which seems to have solved the issue_
- [x] The result will have substantial errors for highly ordered structures such as a crystal due to the binning approximation. These cases should be detected and forwarded to the old sasview calculator. 
_Solved: there was a minor bug in how weighted bins were used in my library_
- [x] MacOS is currently not supported since I cannot compile my library by myself for this system. @timsnow agreed to do the compilation once I fix some Clang-specific issues. 
_Solved: MacOS shared libraries are now availabe_
- [x] Ubuntu 20.04 currently fails. Based on a minor investigation by me & @rozyczko, it seems that this is due to the system being treated differently from Ubuntu 22.04 in the `ci.yml`. Perhaps @wpotrzebowski can help solve this? _Solved: the Linux library **must** be compiled with v20.04_

_Minor_
- [x] The user should be able to extend the q-range. 
_Solved: q-ranges within the default of my library will simply be interpolated. q-ranges outside this range will be somewhat slower but does seem to work_
- [ ] The user should be able to evaluate the profile for any system. Currently systems larger than 2000Å will result in my library crashing, though this should just result in the calculation being forwarded to the old sasview calculator. 
_Semi-solved: if my library crashes, the old sasview calculator is used. This is a terrible solution though, since the old calculator will never finish evaluating such huge systems. I'll try to find a more elegant solution later - for now, I think it is okay since such huge systems are rare_
- [x] Currently some chunking seems to be going on when using a large number of q-values, such that the calculation is split into multiple calls. This is counterproductive when using my library as the full q-range is calculated for every call. We should trace this down and change it. 
_Solved: the GUI chunks the calculation to add cancellation points and progress reports. Since I have manually added the latter and the former isn't possible with an external library call, this is not necessary and have been skipped with a simple if-check. In the future this should probably be refactored to a cleaner approach. I don't think the GUI itself should be responsible for chunking like this, but rather provide a listener object and allow the Iq models themselves to manage this_
- [x] A new process is spawned during the calculation, resulting in a big black terminal being created on top of the sasview GUI. 
_Solved: this doesn't happen when hooking into a shared library_

- [x] Update licensing to mention AUSAXS dependencies. 
_Tim & I found that this is not necessary since they are not part of the source code_
- [ ] Code has been reviewed
_Works on both Linux & Windows_
- [x] Windows installer (GH artifact) has been tested (installed and ships with AUSAXS) 
_Tested that the installer ships with AUSAXS libs and successfully uses them at runtime_
- [ ] MacOSX installer (GH artifact) has been tested (installed and ships with AUSAXS) 
- [x] Linux installer (GH artifact) has been tested (installed and ships with AUSAXS)
_Tested that the installer ships with AUSAXS libs and successfully uses them at runtime_

Note that should we decide to merge without Mac support, my new test should be updated to reflect this choice. Otherwise the `main` branch will consistently start failing this test when running it with Mac.


wpotrzebowski:
@klytje I've tried Mac installer. It crashes silently as soon as I try to run Debeye calculation in GSC. I will try to debug it.

klytje:
> @klytje I've tried Mac installer. It crashes silently as soon as I try to run Debeye calculation in GSC. I will try to debug it.

I tried the installers again on Windows and Linux, since we're now on a different base. 

Windows works fine and is definitely using my library, though I noticed that the installer is still labelled `5.0.5`. 

Linux doesn't work at all - I'm getting "RuntimeError: Could not find the periodic table data files" on both Ubuntu 20.04 & 22.04. Did we change the installation procedure for Linux?

klytje:
@wpotrzebowski can you try with the newest installer?

wpotrzebowski:
Same thing with the new installer (also silently crashes), however running 
`from sas.sascalc.calculator.ausaxs.ausaxs_sans_debye import evaluate_sans_debye` from Python Shell inside sasview works fine.

wpotrzebowski:
However, this script fails on the last command (evaluate_sans_debye) also silently crashing whole sasview
```
from sas.sascalc.calculator import sas_gen
from sas.sascalc.calculator.ausaxs.ausaxs_sans_debye import evaluate_sans_debye
import numpy as np
pdb_file = '/Users/wojciechpotrzebowski/SASDJQ4.pdb'
pdbloader = sas_gen.PDBReader()
f = pdbloader.read(pdb_file)
coords = np.vstack([f.pos_x, f.pos_y, f.pos_z])
q = np.linspace(0.001, 1, 100)
w = np.random.rand(coords.shape[1])
I_out = evaluate_sans_debye(q, coords, w)
```

klytje:
Do you see a `AUSAXS: Started evaluating Debye equation.` message in the terminal?

At this point I think it must be due to some kind of incompatibility with my library. Can you try a `cc -dM -E -xc /dev/null | less` to see what instruction sets your mac supports?

Perhaps we will have to go back to the original plan & delay mac support until I can actually debug it myself.

wpotrzebowski:
I don't see `AUSAXS: Started evaluating Debye equation.`, which is weird because library seems to be in place. I've also tried swapping locally compiled lib with the one coming from the installer but it is the same (crashing).  The only thing that I am getting from the terminal is `zsh: floating point exception  ./sasview`.

wpotrzebowski:
And I am attaching the output from `cc -dM -E -xc /dev/null | less`
[instructions.txt](https://github.com/SasView/sasview/files/15058902/instructions.txt)

klytje:
I don't understand why your mac doesn't have SSE support - Rosetta 2 should support these. How old is your machine? When running the same command on the GitHub runners, they _do_ show up:
```
#define __SSE2__ 1
#define __SSE3__ 1
#define __SSE4_1__ 1
```
Perhaps this is another case of Clang compiling code it cannot actually run? Also, upon researching it further, it seems like we are bound to compile for the `x86-64` architecture since Rosetta does not support different architectures for the executable (sasview, x86-64) and libraries (mine). So it would be pointless to make a universal binary I think.

lucas-wilkins:
seems like stuff is generally not building...

klytje:
> seems like stuff is generally not building...

Sorry I'm in the middle of dropping MacOS support. 

Please try the artifacts from this build: https://github.com/SasView/sasview/actions/runs/8764060262

lucas-wilkins:
Please try the installer from this build: https://github.com/SasView/sasview/actions/runs/8764060262

I will do... I'm at home today, but will look at it at the next available opportunity.

klytje:
I've now disabled MacOS by postfixing `.disabled` to the downloaded shared library. During runtime sasview will then be unable to find the library and defaults to using the existing GSC implementation. This way advanced users will be able to just drop in a correctly named version of the library later when it is in a working state.

wpotrzebowski:
I've just tried the latest installer on MacOS (with the library disabled) but it still seems to be crashing silently (as before).

wpotrzebowski:
I also don't see any logs, so I don't know if the library is used or not.

src/sas/sascalc/calculator/ausaxs/sasview_sans_debye.py
```diff
@@ -0,0 +1,38 @@
+import numpy as np
+import logging
+
+def sasview_sans_debye(q, coords, weight, worksize=100000):
+    """
+    Compute I(q) for a set of points using the full Debye formula.
+    *q* is the q values for the calculation.
+    *coords* are the sample points.
+    *weight* is the weight associated with each point.
```
rozyczko, line 9: Since you are describing all arguments, please also add description for `worksize`

wpotrzebowski:
I did further investigation and it seems that `import cpufeature `line crashes SasView on Mac
@klytje any suggestion? Maybe some specific version?

src/sas/sascalc/calculator/ausaxs/ausaxs_sans_debye.py
```diff
@@ -0,0 +1,101 @@
...
+        attach_hooks()
+    return ausaxs_state is lib_state.READY
+
+def evaluate_sans_debye(q, coords, w):
+    """
+    Compute I(q) for a set of points using Debye sums.
+    This uses AUSAXS if available, otherwise it uses the default implementation.
+    *q* is the q values for the calculation.
+    *coords* are the sample points.
+    *w* is the weight associated with each point.
```
rozyczko, line 74: This is the main method for calculating the scattering. Maybe it could be made more "user friendly" with keyworded arguments?

rozyczko (approved):
Looking good!
Minor comments and questions raised but they aren't blocking the merge.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1400,30 +1400,36 @@ def complete(self, input, update=None):
...
-                outi = self.model.runXY(inputi)
-            out.append(outi)
-            if self.cancelCalculation:
-                update(time=t, percentage=100*(ind + chunk_size)/nq) # ensure final progress shown
-                self.data_to_plot = numpy.full(nq, numpy.nan)
-                self.data_to_plot[:ind + chunk_size] = numpy.hstack(out)
-                logging.info('Gen computation cancelled.')
-                break
+        # the 1D AUSAXS calculator cannot be chunked
+        if self.is_avg & len(input[1]) is 0:
```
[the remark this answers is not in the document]
klytje: I thought it was more clear what the if-statement is checking for with the explicit comparison, but I can remove it, sure.

src/sas/sascalc/calculator/ausaxs/ausaxs_sans_debye.py
```diff
@@ -0,0 +1,101 @@
...
+    READY = 2
+
+ausaxs_state = lib_state.UNINITIALIZED
+ausaxs = None
+
+def attach_hooks():
+    global ausaxs_state
+    global ausaxs
+
+    from sas.sascalc.calculator.ausaxs.architecture import OS, Arch, determine_os, determine_cpu_support
```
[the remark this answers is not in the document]
klytje: That's my C++ habits shining through - trying to reduce the scope of variables & imports as much as possible. I'll move it.

src/sas/sascalc/calculator/ausaxs/ausaxs_sans_debye.py
klytje: I'm not sure I follow what you mean by 'keyworded arguments' here

lucas-wilkins:
So, I ran the installer you mentioned (@wpotrzebowski - I take it mounting the file as a volume then running from there is the normal Mac workflow - I don't know, I new to this OS). It was *very* slow to start, like 5mins, but it did start. What checks should I be doing from there?

klytje:
> So, I ran the installer you mentioned (@wpotrzebowski - I take it mounting the file as a volume then running from there is the normal Mac workflow - I don't know, I new to this OS). It was _very_ slow to start, like 5mins, but it did start. What checks should I be doing from there?

What was slow to start? the installer, sasview itself, or a scattering calculation? If you didn't do it already, can you try running the GSC for a large-ish structure?

klytje:
> I did further investigation and it seems that `import cpufeature `line crashes SasView on Mac @klytje any suggestion? Maybe some specific version?

From the docs it seems that `cpufeature` only supports x86 & AMD64 architectures. Perhaps this is why it is failing on MacOS since it was never meant to run on that platform. I guess the solution is to just not use it for MacOS then.

src/sas/sascalc/calculator/ausaxs/ausaxs_sans_debye.py
rozyczko: As in 
```python
def evaluate_sans_debye(q=None, coords=None, w=None):
```
Meaning the method is called with e.g.
```python
iq = evaluate_sans_debye(q=my_q_set, coords=my_coords, w=my_weights)
```
so users don't have to remember which argument goes where and in what order.

wpotrzebowski:
> So, I ran the installer you mentioned (@wpotrzebowski - I take it mounting the file as a volume then running from there is the normal Mac workflow - I don't know, I new to this OS). It was _very_ slow to start, like 5mins, but it did start. What checks should I be doing from there?

When you start SasView on Mac from the new installer, Gatekeeper (or some other security system) does some checks and it takes a while. 5min sounds quite long though (in my case is usually 1-2 min). Next time you rerun SasView it should start immediately

lucas-wilkins:
Just trying windows version of this from Jeffs message: https://github.com/SasView/sasview/actions/runs/8848780399

rozyczko:
> Just trying windows version of this from Jeffs message: https://github.com/SasView/sasview/actions/runs/8848780399

this version worked for me on win10

lucas-wilkins:
Seems to work otherwise, unless there's a specific check I need to do.

rozyczko:
This is a standard windows message for unsigned binaries. We will sign the release installer, as usual

klytje:
Are we dropping support for Python 3.9 with version 6.0.0? Otherwise we have some `match` expressions we have to remove in the GUI interface

klytje:
`cpufeature` has now been removed as a dependency. Previously it was used to ensure that the host system supports the minimum requirements to run my library. 

Now we instead spawn a subprocess to run my library the first time it is called, which may end up crashing for any reason, such as `invalid operation` or `segfault`. However, since the crash is contained to a subprocess, it poses no danger to the main sasview thread which is then free to continue the calculation using the built-in GSC afterwards. If the subprocess does not crash, we assume it is safe to hook into the library with the main thread for subsequent evaluations, sparing us both the expensive subprocess spawning & library hooking. 

I have verified that this new approach works on both Windows & Linux when run as a script, and also with the installer on Windows. I've also verified that it succesfully captures and recovers from segfaults on Linux. We will have to test the installers again on Linux & Mac, since the documentation for using subprocesses in a frozen binary is not super clear if it will work or not on these platforms. This new approach should work for both `x64` and `arm64` MacOS assuming the Python instance is running as `x64`; with a pure `arm64` system it will fail to link to my library and default to using the built-in GSC. To support pure `arm64` I'll have to figure out how to compile my library as a universal binary for both architectures. 

The new test case is still disabled for MacOS since it seems the new workflow changed the Python instances from running as `x64` to `arm64`. 

@rozyczko perhaps you can re-review the updated `ausaxs_sans_debye.py` script?
@wpotrzebowski can you please test the new Mac installer? Please check in the terminal if it actually uses my library.

wpotrzebowski:
@klytje it seems to work fine. Just a small remark maybe we can turn it into Warning rather than error:
`12:22:39 - ERROR: AUSAXS calculator terminated unexpectedly. Using default Debye implementation instead.`

butlerpd:
I had the same problem with the slicer. Two points.
* It actually **is** in the log explorer which you can open (unhide) from the `view` menu.
* However currently only `ERROR` level loggings automatically unhide the logger view
* It is absolutely nothing to do with this PR.

My Question is whether `WARNING` level logs should also trigger the log window. It definitely should IMO (but not info and debug levels). Otherwise the poor user has no information to help them.  The other point of view I guess would be a warning is not an error which means things continue to work so why bother the user? Perhaps we should discuss this at the next call?

butlerpd:
> I guess the answer to that question depends on the severity of what people have generally logged as a `WARNING` previously. In this case I agree with @wpotrzebowski that failing to use my library should not be an `ERROR`, so we're stuck with a `WARNING` that some users could be interested in knowing was triggered.

Ok .. I recommend we make that a question for Tuesday's call to get broader input but it should not hold up this PR. We all agree we should not flag warnings as errors just to have them pop up automatically. So if there is a change to make it will not be in this PR anyway.

klytje:
This PR should be ready for merge; I'd just like someone to take a look at the rewritten `ausaxs_sans_debye.py` file, and the two updates to `run.py` first.

wpotrzebowski:
I've managed to run `ausaxs` on Mac from the local build. Dumping the same library to the app from the installer doesn't work but anyway, this means that we may have "a hack" for desperate Mac users.

src/sas/sascalc/calculator/ausaxs/ausaxs_sans_debye.py
```diff
@@ -0,0 +1,138 @@
...
+    READY = 2
+
+def _attach_hooks():
+    ausaxs = None
+    ausaxs_state = lib_state.UNINITIALIZED
+    from sas.sascalc.calculator.ausaxs.architecture import get_shared_lib_extension
+
+    # as_file extracts the dll if it is in a zip file and probably deletes it afterwards,
+    # so we have to do all operations on the dll inside the with statement
+    with resources.as_file(resources.files("sas.sascalc.calculator.ausaxs.lib")) as loc:
```
krzywon, line 22: This could be `os.walk('./ausaxs/')`  or `f for f in pathlib.Path().iterdir('./ausaxs/') if f.is_file()`which would eliminate the need to know the file extension. If multiple files are in the directory, each of them would be tried, so that might slow the calculation down, but nothing else should be there.

krzywon (commented):
Everything works locally (Windows). Code comments are for maintainability, style and PEP-8 and should not hold the PR up.

src/sas/sascalc/calculator/ausaxs/architecture.py
```diff
@@ -0,0 +1,34 @@
+from enum import Enum
+
+class OS(Enum):
```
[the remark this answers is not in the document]
klytje: Problem is that the `get_external_dependencies` script is run at build-time, before sasview is installed. So we cannot import modules from it yet. I thought about using relative imports, but decided against it since I wasn't sure if folders could potentially be shuffled around in the installed version.

test/sascalculator/utest_sas_gen.py
```diff
@@ -220,7 +220,59 @@ def test_calculator_1D(self):
...
+        from sas.sascalc.calculator.ausaxs import ausaxs_sans_debye
+
+        from sas.sascalc.calculator.ausaxs.architecture import get_os, OS
+        if get_os() is OS.MAC:
+            self.assertTrue(True, "AUSAXS library is currently not available for MacOS. Skipping test.")
+            return
+
+        if not ausaxs_sans_debye.ausaxs_available():
+            self.assertTrue(False, "AUSAXS library not found, test cannot be run.")
+            return
```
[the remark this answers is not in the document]
klytje: The entire point of this test is to verify that ausaxs gives the same results as the built-in GSC for a number of test files. If the ausaxs library is not available, it'll default to the built-in GSC, which will of course always be in agreement with itself. 

If we don't check that ausaxs is available, a potential bug preventing it from being loaded would never get caught by the tests.

src/sas/sascalc/calculator/ausaxs/ausaxs_sans_debye.py
klytje: Will this also work inside a zip file? I specifically chose this approach after a lot of discussions at the camp because it'll be able to handle that.

build_tools/get_external_dependencies.py
```diff
@@ -0,0 +1,52 @@
+import requests
+from enum import Enum
+
+class OS(Enum):
```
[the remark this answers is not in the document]
klytje: Have we completely left `Python 3.9` behind so it's safe to use `match` expressions?

test/sascalculator/utest_sas_gen.py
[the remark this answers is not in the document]
krzywon: Good point. Thanks for the clarification.

src/sas/sascalc/calculator/ausaxs/ausaxs_sans_debye.py
krzywon: I obviously didn't read your comment close enough. Ignore.

butlerpd (approved):
Tested on windows with apoferritin pdf file. Using 5.0.6 @yunliu01 says the computation for 30 points took 30 min on his windows computer. the same pdb using this PR version of the installer on my window took 8s for 100 points!!!! scaling the contrast and scale factor then comparing to the apoferritin.txt in our example data (which CLEARLY has some aggregation and also has the background subtracted but I suspect as always not perfectly, the result is quite impressive!!!

ApoferritinPDB_and_CoreShell

Way to go @klytje way cool!!

I think this can be merged

PR #2862, opened 2024-04-26
Pyinstaller part 2

### LINKED ISSUE #2813: Documentation window displays blank page on MacOS
**Describe the bug**
The new documentation window always displays a blank screen on MacOS instead of the expected HTML file. This usually only happens with the distributed app, but sometimes happens from a local build, depending on the local environment.

**To Reproduce**
1. Download the Mac app.
2. Run the Mac app.
3. Click any `Help` button.
4. A documentation window appears, but the page area shows a blank, white area.

**Expected behavior**
On step (4) of above, the HTML should be displayed.

 - Version: v6.0.0a1

 - OS: MacOS (possibly Ubuntu, but not clear)

### LINKED ISSUE #2886: the docs take forever to generate if the docs folder was deleted
**Describe the bug**
using the installer, the docs get copied to the user .sasview directory as a Docs folder.  Changes made to docs then get written to that location (which the user has write permissions to) and sasview regenerates the docs which takes time.  However, if the folder gets accidentally deleted, it appears as though the doc regeneration somehow still takes place rather than just copying from the installed directory?  If so this should be fixed. It may however be that it just takes that long to copy over the folder? this should be checked.

 - Version: 6.0.0a1

 - OS: e.g. Windows 10

**Additional context**
This should be addressed as part of the 6.0 release process

### LINKED ISSUE #2902: SasView 6.0.0-beta doesn't start on Linux
**Describe the bug**
I've been trying to run SasView on vanilla linux machine (Rocky Linux 9.3) but I am getting following error

Screenshot 2024-06-24 at 13 05 51

krzywon:
## Description

This is a tweak to how pyinstaller creates packages. Due to a couple of changes on the Mac build, this also adds some fixes to the documentation viewer to ensure Mac docs are copied into the user folder when required and are displayed properly.

Fixes #2886 
Fixes #2813 
Fixes #2902 

## Review Checklist:

**Installers**
- [x] This **will** affect the **installers**
  - [ ] **Ubuntu** installer (GH artifact) has been tested (installed and worked) 


klytje:
sasview now succesfully opens, but crashes the moment I click the 'open file' pane (both in the GSC and Image Viewer; so I assume it's all of them) with the error:
image

rozyczko:
On windows, this build seems to run fine.

krzywon:
> sasview now succesfully opens, but crashes the moment I click the 'open file' pane (both in the GSC and Image Viewer; so I assume it's all of them) with the error: image

@klytje - Thanks for the report. Are you seeing this error with anything other programs? It looks like either a permissions or environment variable issue. I'll see what I can do about this from our end.

https://askubuntu.com/questions/1351607/gtk-warning-could-not-load-a-pixbuf-from-icon-theme
https://bbs.archlinux.org/viewtopic.php?id=261143

klytje:
@krzywon I've never seen this error before. I am developing a GUI myself relying on GTK, so I'd be very surprised if that's the issue. Anyway, I've now tested it on all of my systems. 

The same issue is present on 2 separate Ubuntu 22.04 systems. On my Ubuntu 20.04 VM it works just fine. It would be relevant to try on other 22.04 systems; perhaps a VM if anyone has one of those running. 

I'm not sure if it is related to your PR though - this is the first time I've tried running the installer from the `release_6.0.0` branch.

klytje:
Using the `run.py` script which works just fine on all my platforms, I get the exact same errors _except_ the GuiManager one. Perhaps this is the one causing the crash?

llimeht:
If the SasView windows installer is used to do a system-wide installation (which it can do and I think that's even the default), then it's buggy to be using `{%USERPROFILE}` in the installer config (`.iss`). The problem in the context of a system-wide installation is which _single_ user will get the files in their profile, and what about other users? What happens with a central IT person insists on installing the software for the user, or SasView is installed on shared computers at a facility to be used by lots of different users?

There is an existing use of `{%USERPROFILE}` - but that's an empty directory that (I think) gets recreated at runtime if needed. (We should make sure that is true and then get rid of that entry in the `.iss`.)

To run sphinx at runtime in a user-writeable a location, some other creative solution is needed to correctly handle multi-user systems. Lazily copying the documentation across to the user profile on first start might work. Even better would be copying the documentation only once the user asks to edit it; rehoming the doc-base from the installer-provided files to the user-edited files only when needed would also address #2886.

Some work also needs to go into preventing data loss where the installer-provided documentation overwrites local edits to those files when upgrading versions. I think that would currently occur with the approach in this PR. (The above suggestions of lazily copying the files around could have similar issues depending on how they are implemented. Syncing is a hard problem.)

butlerpd (approved):
Testing functionality on Windows only, this does fix the problem identified in #2886.  My approval is solely for that part of this PR. Others should test the mac and linux issues.

tsole0 (approved):
Changes work on Mac and fixes #2813. My approval is specific to the Mac-dependent part of the PR.

PR #2877, opened 2024-05-08
2669: Default load directory persistence

### LINKED ISSUE #2669: Default directory is neither settable nor persistent (it used to be persistent)
**Describe the bug**
Using the latest build of 6.0.0a1, the application no longer remembers the last directory I used when loading data between sessions.  Moreover, the new settings panel does not provide any option for that either.  I'm not sure how far such behavior may go, but at one point we had implemented a persistent default directory between sessions which would seem like more desired behavior than the current.

**To Reproduce**
1. Start SasView
2. Click on `Load data` and navigate to your favorite data folder
3. Close out from SasView
4. Reload SasView
5. click on `Load data`
6. the pop up starts at the top of the 5.0.6 install directory again

**Expected behavior**
Ideally, there should be an option in the new settings menu to set the default directory or to make it default to the last used directory in a persistent fashion (between sessions).  In any case, defaulting to the top level SasView install directory seems silly. 

**Screenshots**

 - Version: 6.0.0a1 uploaded to github at 5AM GMT on October 26

 - OS: e.g. Windows 10

**Additional context**

krzywon:
## Description

This ensures the last directory any file was loaded from, whether it was from the file dialog or drag/drop, is used as the default directory the next time data is loaded. This value also persists between SasView instances.

Fixes #2669

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -153,6 +153,16 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         # Current view on model
         self.current_view = self.treeView
 
+    @property
+    def default_load_location(self) -> str:
+        return self._default_load_location
+
+    @default_load_location.setter
+    def default_load_location(self, value: str):
+        # Ensure the config entry is updated
+        self._default_load_location = value
+        config.DEFAULT_OPEN_FOLDER = value
+
     def createSendToMenu(self):
         self.actionReplace = QtGui.QAction(self)
         self.actionReplace.setObjectName(u"actionReplace")
@@ -218,6 +228,7 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         """
         Threaded file load
         """
+        self.default_load_location = os.path.dirname(url[0])
         load_thread = threads.deferToThread(self.readData, url)
         load_thread.addCallback(self.loadComplete)
         load_thread.addErrback(self.loadFailed)
@@ -252,7 +263,6 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         folder = str(folder)
         if not os.path.isdir(folder):
             return
-        self.default_load_location = folder
         # get content of dir into a list
         path_str = [os.path.join(os.path.abspath(folder), filename)
                     for filename in os.listdir(folder)]
@@ -1335,7 +1345,6 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         if not isinstance(paths, list):
             paths = [paths]
 
-        self.default_load_location = os.path.dirname(paths[0])
         return paths
 
     def readData(self, path):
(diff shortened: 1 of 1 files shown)


krzywon:
I downloaded the Mac installer and the persistence seems to work there. 

@butlerpd suggested adding an entry in the preferences window to give the user more power over this. I'll add that if I find the time

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -153,6 +153,16 @@ def __init__(self, parent=None, guimanager=None, manager=None):
         # Current view on model
         self.current_view = self.treeView
 
+    @property
+    def default_load_location(self) -> str:
+        return self._default_load_location
+
+    @default_load_location.setter
```
lucas-wilkins, line 160: Is there are reason why you've made a copy of the folder, and trying to synchronise two variables.

Seems to be against the general principle of storing things in only one place?

lucas-wilkins (changes requested):

src/sas/qtgui/MainWindow/DataExplorer.py
krzywon: Good catch. This could leave to divergence in the two variables if we aren't careful. Updated.

butlerpd (approved):
Works as advertised on windows and also uses the system file selection widget which is an added bonus.  I like it.

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -1326,16 +1335,14 @@ def chooseFiles(self):
         caption = 'Choose files'
         directory = self.default_load_location
         filter = wlist
-        options = QtWidgets.QFileDialog.DontUseNativeDialog | QtWidgets.QFileDialog.DontUseCustomDirectoryIcons
-        paths = QtWidgets.QFileDialog.getOpenFileNames(parent, caption, directory, filter, options=options)[0]
+        paths = QtWidgets.QFileDialog.getOpenFileNames(parent, caption, directory, filter)[0]
```
rozyczko, line 1339: The reason we used `DontUseCustomDirectoryIcons` was the occasional horrendous wait times when navigating to a directory with a number of subdirectories/files. This was a crippling bug on some windows machines.

This _might_ have been fixed in current version of PySide6 but please test.

krzywon: I've been having issues with the dialog since we moved to PySide6 where it won't display any files or folders except the current folder. Removing these options has fixed that on Windows, but I'll test on Mac.

As a side note - I realize these options are in a few other places and should be removed from all places if we go this route.

src/sas/qtgui/MainWindow/DataExplorer.py
lucas-wilkins: Yup

lucas-wilkins (approved):

krzywon:
TODO for @krzywon (yes, that's me): Check Mac build, particularly related to using the native file dialog and custom directory icons. If this works on Mac, apply this change to **all** instances of the file dialog that exist in SasView (Save/Load Project, GSC file loader, etc.). Long term, which would be even better, is to create a single file loader object that others call or inherit from to prevent duplicate code.

krzywon:
I accidentally merged `main` into this branch instead of `release_6.0.0` and then pushed. Revert commits were not working seemlessly, so I am closing this in favor of #2959.

PR #2879, opened 2024-05-09
2875 drop logger level required to force log explorer to pop up from error to warning

lucas-wilkins:
Log view shows on level 30 now instead of 40.

Removed unnecessary passes

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -803,7 +803,7 @@ class GuiManager:
         """
         """
         self.filesWidget.loadAnalysis()
-        pass
+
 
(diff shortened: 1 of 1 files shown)


krzywon (approved):
I tested this locally by trying to save an empty analysis. The log opened when the warning was thrown. Thanks for removing unneccessary passes.

Two suggestions:

- The inline comment above `record.levelno` says to show the log if error or worse. This should be updated to prevent confusion in the future.
- Remove 3 empty lines where `pass` used to be (lint).

lucas-wilkins:
> Remove 3 empty lines where pass used to be (lint).

We need to run black on the whole thing at some point soon. Think there is an issue for it.

PR #2882, opened 2024-05-10
Debye efficiency (macOS)

klytje:
Test branch for MacOS support of the external AUSAXS library.


lucas-wilkins:
Seems like its not finding it!

lucas-wilkins:
I'll be in the office for a little while longer if there's something you need doing.

klytje:
I think it's working - we didn't mark our new release as the latest, so on the first run the tests were still using the old library. I forgot we have two different actions running, one for `pull_request` and another for `push`, so I only requested a rerun of the first one with the new library.

butlerpd:
Question: is that possibly a chip issue? I think different macs can use very different chips?

Also out of curiosity: does it really crash? I see Mac saying it crashed and offering to kill the program but "ignore" is also an option and looking at the log explorer it looks like SasView caught the crash of ausaxs and fell back to the old method and is just taking a very long time (9% of that calculation at the time of the screenshot)? If one hits ignore does it continue?  If so and it is difficult to have it not crash on some chips/platforms it might be good to capture the failure better so that the user does not see the crash warning but just the log explorer popping up saying we could not use it falling back to slow method?

klytje:
> Question: is that possibly a chip issue?

Perhaps? Depends on the kind of optimizations Clang decided to use when @lucas-wilkins compiled it. I'll have to research how to specify what it can and cannot use; it's not obvious to me how to do this with universal binaries like this. The current implementation hides the actual crash message; I'll have to change this to debug further. 

> Also out of curiosity: does it really crash? 

A new process is spawned to 'test-run' my library, which indeed did fatally crash. Perhaps the OS is picking up on this and offering to kill the rest of the program as well? Such a popup is not ideal, to say the least. I'm not sure how to get rid of it. When I tested with Linux & Windows, neither of them seemed to care about the crashing process.

krzywon:
As another data point, I see the same error as @wpotrzebowski. I am running an M2 processor. Attached is the MacOS crash report which might help you debug the issue.
[sasview-2024-05-14-134817.txt](https://github.com/SasView/sasview/files/15351357/sasview-2024-05-14-134817.txt)

klytje:
@wpotrzebowski Can you test [this](https://github.com/AUSAXS/AUSAXS/actions/runs/9190285956) artifact with [this](https://github.com/AUSAXS/AUSAXS/blob/ci/test/test_shared_lib.py) script? Should be easier to debug through this than the sasview installers.

klytje:
@wpotrzebowski That's odd. I didn't change anything compared to when Lucas compiled it. What happens if you use this library with sasview? It should go in the `src/sas/sascalc/calculator/ausaxs/lib` folder.

wpotrzebowski:
Hmm, it doesn't work with just swapping lib files for the app from the installer. (Mac security-keeper is not happy about replacing files in the signed executable). I need to think if I can test it differently.  We probably can disable signing for this PR (it doesn't work for the moment anyway).

klytje:
I don't think you need an actual installer, you should be able to use it through the `run.py` script as long as the library is in the correct folder

wpotrzebowski:
> I don't think you need an actual installer, you should be able to use it through the `run.py` script as long as the library is in the correct folder

This is how I made it work before (with the library compiled on my machine), so I know this approach works but indeed it may be worth trying with external one.

wpotrzebowski:
> > I don't think you need an actual installer, you should be able to use it through the `run.py` script as long as the library is in the correct folder
> 
> This is how I made it work before (with the library compiled on my machine), so I know this approach works but indeed it may be worth trying with external one.

Yes, this works but I had to allow in security settings to run it. 
Screenshot 2024-05-22 at 14 39 29

Screenshot 2024-05-22 at 14 40 41

klytje:
@wpotrzebowski I wonder why you didn't get this warning when you were running my testscript? They hook into the library in the exact same way. Could this be related to why it doesn't work in the installer?

wpotrzebowski:
@klytje I don't think this is the reason. The signing script goes through the entire bundle and signs everything (I can probably check if your lib has been signed properly). I had to allow running your lib as it was downloaded from the Internet.

klytje:
@wpotrzebowski I updated the Releases so all new installer runs should include the new library version.

klytje:
Ah - the failing tests is my mistake. I recently changed the distance binning width from 0.1Å to 0.25Å, but this is apparently not good enough for the highly ordered diamond structure we're testing with. I'll fix this in the next version.

wpotrzebowski:
Without signing it returns the same error as for the signed package.

klytje:
> I´ve run test script on the library from the installer. Maybe this tells you something

Can you run `file libausaxs.dylib` for me?

wpotrzebowski:
> > I´ve run test script on the library from the installer. Maybe this tells you something
> 
> Can you run `file libausaxs.dylib` for me?

```
file /Volumes/SasView6/SasView6.app/Contents/Frameworks/sas/sascalc/calculator/ausaxs/lib/libausaxs.dylib
/Volumes/SasView6/SasView6.app/Contents/Frameworks/sas/sascalc/calculator/ausaxs/lib/libausaxs.dylib: Mach-O 64-bit dynamically linked shared library arm64
```

klytje:
Can you verify that the uncompressed library has both `x86-64` and `arm64`? This is what my GitHub runner tells me after compilation: 
```
 build/lib/libausaxs.dylib: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
 ```

**Comment by wpotrzebowski** [comment]
> Can you verify that the uncompressed library has both `x86_64` and `arm64`? This is what my GitHub runner tells me after compilation:
> 
> ```
>  build/lib/libausaxs.dylib: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
> ```

@klytje: What do you mean by uncompressed library?

**Comment by klytje** [comment]
@wpotrzebowski The library artifact you downloaded from my repo. 

I've read through the PyInstaller docs. Three observations:
1. PyInstaller has the ability to trim multi-arch binaries to only a single arch. I am sure this is what must be happening. I don't know why it is removing the wrong arch, though. 
2. PyInstaller has a `binary` section where you're supposed to put binaries. I moved the library there. Previously it was in the `data` section. Can you try using the latest installer with this change?
3. PyInstaller defaults to building for the architecture of the host machine unless otherwise specified. The `macos-latest` runners are all `arm64`, which explains why SasView is looking for that version of my library. But wouldn't this mean that the installer would not work with older Intel MacOS machines?

**Comment by wpotrzebowski** [comment]
For uncompressed
```
file libausaxs.dylib 
libausaxs.dylib: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
libausaxs.dylib (for architecture x86_64):	Mach-O 64-bit dynamically linked shared library x86_64
libausaxs.dylib (for architecture arm64):	Mach-O 64-bit dynamically linked shared library arm64
```

klytje:
Actually after checking your previous comments again, I am even more confused. The installer is definitely building for `arm64`: 
image
So PyInstaller seems to be doing it right, trimming away the `x86_64` version of my library (your earlier `file` output):
`libausaxs.dylib: Mach-O 64-bit dynamically linked shared library arm64`

But then at runtime, your system seems to be running SasView as `x86_64`? That's the library architecture it is looking for, anyway (from your earlier debugging output):
`OSError: dlopen [...] libausaxs.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64')`

I don't understand how this is possible? How can a `arm64` program be running in `x86_64` mode?

klytje:
Could this perhaps be due to the way an external Python process is spawned? 
image

The first time we try to hook into my library, we do it from a spawned child process to avoid propagating crashes. But this spawns a _fresh_ Python interpreter process. What would happen if you only have `x86_64` Python installed on your system, but the process is spawned from an `arm64` program?

wpotrzebowski:
The latest installer crashes silently.

klytje:
@wpotrzebowski can you try the latest installers? They should now be building with architecture-specific versions of my library

klytje:
@krzywon can you maybe give the latest installers a try since Wojciech is not available? Just try to run the generic calculator and see if it crashes

krzywon:
Thanks for the update, @klytje. I'll give it a try sometime this week.

krzywon:
I was able to test the latest build this morning. Unfortunately, I am still seeing the error pop up saying SasView stopped, but I also only see a single MacOS build. I am going to update this branch to the latest release to see if we can have parallel builds.

klytje:
> I was able to test the latest build this morning. Unfortunately, I am still seeing the error pop up saying SasView stopped, but I also only see a single MacOS build. I am going to update this branch to the latest release to see if we can have parallel builds.

It should already have been parallel builds? Anyway, if it's still crashing, I don't think we can get around waiting for the macbook to be shipped so I can do some proper debugging.

krzywon:
> It should already have been parallel builds? Anyway, if it's still crashing, I don't think we can get around waiting for the macbook to be shipped so I can do some proper debugging.

I reran the build so the parallel builds were actually created. I just tested the MacOS latest build and am still seeing the error when running the debye calculation. @rozyczko - any news on the MacBook?

klytje:
@krzywon or @wpotrzebowski Can you try again but running SasView from the terminal? I.e. navigating to the downloaded build and running `./SasView6.app/Contents/MacOS/sasview` ? I have a suspicion of what might be going wrong.

klytje:
I recompiled my own library to no longer automatically create a /temp folder on startup, and recompiled SasView to use this updated version. This seems to have fixed the crashing on my `intel` mac even when running the app directly. Can someone check if it also works on `arm`?

krzywon:
I was able to test this on my `arm` based Mac and the error message is no longer present when running any of the calculations.

klytje:
Latest build confirmed working on all three platforms. I think this is ready for review.

krzywon:
As noted in an email conversation, we are holding this for the next release.

wpotrzebowski:
Tested functionality on arm Mac and indeed it seems to work!

krzywon (approved):
After some thought, having this in the release would be useful for Mac users. Approved

PR #2890, opened 2024-05-30
6.0.0 release updates

### LINKED ISSUE #2887: year  needs to be changed in welcome window
**Describe the bug**
apparently the Welcome window (which contains the splash screen) has an independent entry for the (c) date. 2023 needs updating to 2024.

 - Version: 6.0.0a1

**Additional context**
Needs to be fixed as part of the 6.0 release process

krzywon:
## Description

This is an update to the release related information specific to 6.0.0.

Changes:
- Release notes
- Sister package release branches (sasmodels and bumps)
- DOI generated by Zenodo for v6.0.0
- Copyright year is now 2024
- Sasview version is now 6.0.0 (or 6.0.0b1 in some cases)
- Updates to `build_tools/release_automation.py`
  - Include sasdata for versioning and issues
  - Ability to read in contributor list from tsv file (if found)
  - Include help for the script
  - General code cleanup and lint

Still required:
- All versions should be 6.0.0 for the final release
- Update to contributors for the zenodo record

Fixes #2887

## Review Checklist:

**Installers**


build_tools/release_automation.py
```diff
@@ -88,10 +110,58 @@
...
+                    # TODO: Should we grab the latest affiliation information using the ORCID ID?
+                    record['orcid'] = values[5]
+                if values[2]:
+                    creators.append(record)
+                elif values[3]:
+                    record['type'] = 'Producer'
+                    contributors.append(record)
+                elif values[4]:
+                    record['type'] = 'Other'
+                    contributors.append(record)
```
butlerpd, line 159: I assume that this will mean that if both Producer and Other is checked only Other will get recorded? Probably reasonable but not sure what we agreed.. so far?

Also, could one future proof this by taking the record['type'] from the header line? not suggesting we need to make big changes now. Just asking the question.

build_tools/release_automation.py
```diff
@@ -136,15 +206,16 @@ def generate_zenodo(sasview_data, zenodo_api_key):
                                                             "and attach to Zenodo release record.\n- Publish Zenodo record")
     return newDOI
 
+
 version_template = \
-"""
-__version__ = "%s"
+"""__version__ = "%s"
 __release_date__ = "%i"
 __build__ = "GIT_COMMIT"
```
butlerpd, line 213: This line seems superfluous and outdated?  The original concept was that the installer would have a build number (as many software packages do) which would be the commit hash, but that only worked briefly. Does "GIT_COMMIT" have any utility? Of course, even if it is pointless, it may be something to remove after the release in which case we should ticket it.

docs/sphinx-docs/source/user/RELEASE.rst
```diff
@@ -13,6 +13,151 @@ Release Notes
...
+* Fixes problem with using deprecated node.js by @butlerpd in #2836
+* Run unit tests of Qt GUI and have CI gate on them by @llimeht in #2252
+* Remove ubuntu 18.04 from CI by @krzywon in #2439
+
+Python Dependencies
+___________________
+* Added 3.11, removed 3.8 by @rozyczko in #2582
+* Warnings removals and python 3.9 drop in tests by @wpotrzebowski in #2860
+* Pyside6 merge by @rozyczko in #2478
+* Pyinstaller 6 by @krzywon in #2854
```
butlerpd, line 148: This should say Change GUI package from PyQt5 to Pyside6. Also, This actually a big deal right? - no doubt some breaking changes and part of why this  is 6 not 5.1?

src/sas/system/version.py
```diff
@@ -1,4 +1,3 @@
-__version__ = "6.0.0a1"
-__release_date__ = "2023"
+__version__ = "6.0.0b1"
+__release_date__ = "2024"
 __build__ = "GIT_COMMIT"
-
```
butlerpd, line 4: Same comment as for `releasese_automation.py` - What is the point of `__build__` ?

docs/sphinx-docs/source/user/RELEASE.rst
```diff
@@ -13,6 +13,151 @@ Release Notes
...
+New Features
+^^^^^^^^^^^^
+* Improvements to Corfunc by @lucas-wilkins in #2450
+* Corfunc perspective needs export and report capabilities by @lucas-wilkins in #2065
+* Add a preferences window by @krzywon in #2167
+* Debye efficiency by @klytje in #2859
+* Rog and beta q by @smalex-z in #2535
+* Custom fit models by @smalex-z in #2565
+* 1763 implement the orientation viewer in 5x and GL Subsystem by @lucas-wilkins in #2394
+* Particle editor by @lucas-wilkins in #2520
```
butlerpd, line 32: again, lines beginning with numbers that I don't understand? Probably just need to remove them?

ALSO ... I don't think the particle editor is actually complete and available is it?

docs/sphinx-docs/source/user/RELEASE.rst
```diff
@@ -13,6 +13,151 @@ Release Notes
 Features
 ========
 
+New in Version 6.0.0
+--------------------
+This is a major release with a number of new features, enhancements, and bug fixes as described below. Many of the new
+features are now highlighted in our 'What's New' window displayed when SasView starts.
```
butlerpd, line 19: Going from v5 to v6 means more than feature additions. In principle breaking changes? In particular moving sasdata out of sasview and moving from PyQt to Pyside for the GUI package? Seems like it should be made clear somehow why this is not 5.1 (a feature release)?

Also - for a future PR probably but I would recommend highlighting at the top some of the big changes for users like a proper settings option and the combined orders of magnitude speed increases in both reading the PDB and then calculating the scattering from that PDB? Or maybe we just leave those now in "What's new" window?

butlerpd (commented):
I think there are at least a few items that should be addressed before merging. I still need to test the installer before finishing the review

build_tools/release_automation.py
krzywon: > I assume that this will mean that if both Producer and Other is checked only Other will get recorded? Probably reasonable but not sure what we agreed.. so far?

Correct. This also does not allow names to be in both the `creators` and `contributors` list.

> Also, could one future proof this by taking the record['type'] from the header line? not suggesting we need to make big changes now. Just asking the question.

I don't see why not, but I agree this can wait.

build_tools/release_automation.py
krzywon: Good question. Probably not necessary, but I agree we shouldn't hold up a beta release for this.

butlerpd:
The sphinx documentation does not have the correct version number. There is a separate place it gets set.

butlerpd (approved):
as mentioned the version of the sphinx docs needs changing. Otherwise it should be good to go. Approving assuming these things get fixed.

krzywon:
> The sphinx documentation does not have the correct version number. There is a separate place it gets set.

I'll take a look at this tomorrow.

krzywon:
> The sphinx documentation does not have the correct version number. There is a separate place it gets set.

Another place this is hard-coded that I missed. This build should have the correct version displayed in the documentation.

PR #2895, opened 2024-06-12
Cleaned unused item from config

lucas-wilkins:
Cleaned up config so that we don't have unused entries.

Fixes #2895


lucas-wilkins:
Resolves #2895

lucas-wilkins:
Fixes #2895

krzywon (approved):
Looks good.

PR #2901, opened 2024-06-21
Model Editor Bug Fixes

### LINKED ISSUE #1595: ESS_GUI: plugin editor not show model errors on save
The plugin editor says the plugin was saved and then proceeds to ignore it, changing the plugin model selected in the fit page dropdown to one that has no errors.

There is no error listed in the console, though a warning is appearing in sasview.log:
```
11:49:08 - WARNING: Failed to load plugin '.../.sasview/plugin_models/longcylinder.py'. See .../.sasview/plugin_models/plugins.log for details
```
and the error is properly listed in plugin.log

### LINKED ISSUE #1633: Investigate why the Model Editor is not applying the test on Apply
Created at request of @rozyczko :
The Model Editor window does not seem to have the test run implemented on Apply. Was this just missed, or was there a reason?

tsole0:
## Description

Fixes minor bugs in the model editor (`PluginDefinition.py` and its parent `TabbedModelEditor.py`) that impede usability or cause undesirable effects.

- Generates error message when user neglects to input a model name (84e4ebe)
  - Fixes behavior where a model syntax error (bad unicode) is thrown instead, which is unintuitive
- Allows user to create model even if default function text is unedited (b4c8f8c, bf3b57a)
  - Fixes bug in which user had to perform an edit action in the function box before model could be saved
- Changed behavior of model parameter tables in GUI (01b286c)
  - Previously would add new rows to tables even if no parameters were changed and would create unnecessary amounts of blank rows that could not be deleted
  - Previously new rows would be generated only after editing a parameter's value box and not the name of the parameter itself
- Fixes #1633 by ensuring both model and syntax checks are applied to user models regardless of whether they are defined using `PluginDefinition` or `ModelEditor` (558c7a1, 54a8aaf, 2b71022)
  - Note: Previously, only syntax was checked when using `ModelEditor` and only model tests were run when using `PluginDefinition`. The syntax checks were done using the `ast` library, which only checks for SyntaxErrors. I have kept this check as the first one that runs because it can often catch SyntaxErrors with more precision than using `GuiUtils.checkModel()`. In theory we don't need to run the `ast` check on models generated with the plugin editor, but it doesn't seem to add significant overhead and could even be useful if future changes break the model template.
- Fixes bug that prevented user from viewing/editng C model associated with a Python file if the C model contained errors when loading into the model editor. (eb3dee0, 14d1e0a)
  - New behavior: shows popup warning the user that the model loaded into the editor contains errors and must be fixed before using
- Fixes bug causing font to change in the function definition textbox if the user deleted all its text (788433b)
  - Note: bug was fixed by changing default text in box to Consolas, which is a Microsoft font. The fix still works on my Mac, however, probably because the system knows to substitute a similar monospace font. Even if the fix didn't work on Linux, this change should only make things better and I highly doubt it would create a worse result than before.

## How Has This Been Tested?

Verified correct behavior in SasView and checked wide variety of possible user inputs to search for additional bugs. Windows installer tested and verified.

## Review Checklist:

**Installers**


src/sas/qtgui/Utilities/PluginDefinition.py
```diff
@@ -152,9 +165,21 @@ def onParamsPDChanged(self, row, column):
...
+        # Check to see if the last two rows are empty. If so, remove the last row
+        remove_last_row = 0
+        for i in range(2):
+            for j in range(2):
+                cell_contents = self.tblParamsPD.item(self.tblParamsPD.rowCount() - (2 - i), j)
+                if cell_contents == None or cell_contents.text() == "":
+                    remove_last_row += 1
+        if remove_last_row == 4:
+            # If all four last cells are empty, remove the last row
+            self.tblParamsPD.removeRow(self.tblParamsPD.rowCount() - 1)
```
krzywon, line 181: This code block does not need to be here twice (also in 139-150). I will investigate. Likely a method that should be called instead of code repeat.

Note - they are separate calls for different tables. This can be abstracted.

krzywon (commented):
Functionally, this almost does what it says it does, but I'm not 100% certain this fixes #2901. 

If params are in the model text, but not defined in the param table, an error is displayed, but the model file is stil generated. This causes issues when correcting the model if the overwrite option is not selected. I'm not sure if this is the expected behavior or not. @rozyczko, what do you think?

krzywon:
Please note - I am planning to fix the issues I've found here.

krzywon (approved):
> Functionally, this almost does what it says it does, but I'm not 100% certain this fixes https://github.com/SasView/sasview/issues/1633.

This fixes the issue.

> If params are in the model text, but not defined in the param table, an error is displayed, but the model file is stil generated. This causes issues when correcting the model if the overwrite option is not selected. I'm not sure if this is the expected behavior or not.

This is a bigger fix than expected and should not hold up this PR. Many of the underlying checks require the file to exist, so I think this is ready.

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -371,57 +380,41 @@ def checkModel(self, model_str):
...
+                # last_lines = traceback.format_exc().split('\n')[-4:]
+                traceback_to_show = '\n'.join(last_lines)
+                child.setToolTip(traceback_to_show)
+
             # attempt to find the failing command line number, usually the last line with
             # `File ... line` syntax
-            for line in reversed(all_lines):
-                if 'File' in line and 'line' in line:
+            reversed_error_text = list(reversed(all_lines))
+            for line in reversed_error_text:
```
pkienzle, line 401: Better to use iterator form `for line in reversed(all_lines)` rather than turning it into a list that you are just going to throw away.

If you do need to assign an iterator to a variable, wrapping it in a tuple is a good defensive practice. Too often I've tried to reuse a sequence only to find that it empty the second time I traverse it. Tuples are immutable and slightly faster.

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -371,57 +380,41 @@ def checkModel(self, model_str):
...
+                if ('File' in line and 'line' in line):
```
pkienzle, line 402: Parsing compiler error messages feels brittle. Maybe a regexp with r"[Ff]ile.*[Ll]ine ([0-9]+)" will be a little more future proof?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -432,24 +425,47 @@ def updateFromEditor(self):
         if not w.is_python:
             base, _ = os.path.splitext(filename)
             filename = base + '.c'
-
         # make sure we have the file handle ready
-        assert(filename != "")
+        assert filename != ""
```
pkienzle, line 429: The python "optimizer" suppresses assert statements so technically we shouldn't rely on it for flow control. In practice we never run with the optimizer so not a problem. In this case we would fail on self.writeFile with a missing filename so even less of an issue.

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -432,24 +425,47 @@ def updateFromEditor(self):
...
         model_str = self.getModel()['text']
+        # Save the file
+        self.writeFile(filename, model_str)
+
+        # Get model filepath
+        plugin_location = models.find_plugins_dir()
+        full_path = os.path.join(plugin_location, filename)
+        if not w.is_python and self.is_python:
+            pass
+        elif os.path.splitext(full_path)[1] != ".py":
```
pkienzle, line 441: No idea where `filename` is coming from, but if it is `model.c` and the code is python would you want this saved as `model.c.py`? This will likely cause problems for the loader. 

You ought to check that filename without extension is a valid python symbol name [with something like `Path(filename).stem.isidentifier()`] and replace the extension with `.py` for python models.

Shouldn't this happen before `self.writeFile()` ?

Also, `self.writeFile()` must have similar code for path handling. Can you modify that function to return the full path or set it as an attribute?

pkienzle (approved):
Code looks okay. I didn't test functionality.

pkienzle:
Within "create new model" tab add a plugin name. The apply button does not become available until you switch to another field. Okay, a little confusing. Now go back into the plugin name field and change it. Clicking apply it is still using the old name, not the name that is typed into the field.

You may want to check whether the name already exists or does not pass the `isidentifier()` test as soon as you leave the field rather than waiting for the apply button to be pressed.

pkienzle:
I see that the box only accepts alphanumeric (and probably underscore but I didn't try). This still allows a name like 2cows.

I can't tell if 2cow is working because this branch is throwing the following error when I try to calculate a model (even a sphere model):
```
13:39:57 - sas.qtgui.Perspectives.Fitting.FittingWidget - ERROR - Traceback (most recent call last):
  File ".../sasview/src/sas/qtgui/MainWindow/GuiManager.py", line 1339, in deleteIntermediateTheoryPlotsByTabId
    self.filesWidget.deleteIntermediateTheoryPlotsByTabId(tab_id)
  File ".../sasview/src/sas/qtgui/MainWindow/DataExplorer.py", line 2085, in deleteIntermediateTheoryPlotsByTabId
    if item_model_id == model_id:
                        ^^^^^^^^
NameError: name 'model_id' is not defined
```

pkienzle:
Changing `model_id` to `tab_id` on that line lets me see that my `2cows` model works. Not sure if this is a good thing or a bad thing.

pkienzle:
* On mac when not in full screen the editor window pops up behind the application. Is it too difficult to put it within the MDI workspace?
* On mac plugin manager selecting Edit... for a model opens a new editor window even if there is already one open. Expected behaviour is to raise the existing window to the front.
* The traceback in the error log is hard to read with all the line breaks and indentation removed.
* Should check parameter table for name and initial value when creating model. If name is blank then it ignores all subsequent parameters. If value is blank it defaults to 1.
* Need to get current value from all fields on apply. Changing the name or value of the parameter then clicking apply uses the value from before the edit.
* Parameter table cannot input units, limits or description (future enhancement)

pkienzle (changes requested):
Make sure that the current value displayed in the text box, not the value before the text box was edited, is the one that is used in the new model file. The remaining suggestions are optional.

krzywon:
> Make sure that the current value displayed in the text box, not the value before the text box was edited, is the one that is used in the new model file. The remaining suggestions are optional.

I was able to reproduce this on MacOS, but not Windows. It seems focus is not removed from inputs when the buttons are clicked. I'll try to find a fix for this.

src/sas/qtgui/Utilities/PluginDefinition.py
```diff
@@ -5,12 +5,25 @@
...
-# tblParamsPD
-# txtFunction
+
+def remove_empty_table_rows(tbl: QtWidgets.QTableWidget):
+    """A helper function to remove empty rows in a PySide Table, if there are more than two empty rows at the end.
+    This function ensures there is always an empty row in the table.
+
+    :param tbl: A QTableWidget on i rows and j columns.
+    """
+    for i in range(0, tbl.rowCount()):
```
rozyczko, line 15: Would it be better to do the reversed row iteration? Otherwise you are risking skipping rows after removal.
e.g. `for I in range(tbl.rowCount() -1, -1 , -1)`

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -371,57 +390,41 @@ def checkModel(self, model_str):
...
             # GuiUtils.Communicate.statusBarUpdateSignal.emit("Model check failed")
             self.parent.communicate.statusBarUpdateSignal.emit("Model check failed")
-            # Put a thick, red border around the mini-editor
-            self.tabWidget.currentWidget().txtEditor.setStyleSheet("border: 5px solid red")
-            # last_lines = traceback.format_exc().split('\n')[-4:]
-            traceback_to_show = '\n'.join(last_lines)
-            self.tabWidget.currentWidget().txtEditor.setToolTip(traceback_to_show)
+
+            # Put a thick, red border around the editor.
+            from sas.qtgui.Utilities.CodeEditor import QCodeEditor
```
rozyczko, line 395: This is not a "heavy" import - please move it to the top of the file

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -347,18 +363,21 @@ def updateFromPlugin(self):
...
         try:
+            with open(full_path, 'r', encoding="utf-8") as plugin:
+                model_str = plugin.read()
             ast.parse(model_str)
+            GuiUtils.checkModel(full_path)
 
-        except SyntaxError as ex:
+        except Exception as ex:
             msg = "Error building model: " + str(ex)
             logging.error(msg)
```
rozyczko, line 382: The whole Exception handler here grew quite large - please consider refactoring it into a separate method, so the exception flow is clear.

src/sas/qtgui/Utilities/TabbedModelEditor.py
rozyczko: This is a valid complaint. We should not be using `assert` for type/value checks but instead perform explicit checks and notify the user with relevant message.
Yes - we already have quite a few asserts sprinkled all over the (non-test) code but I think we should be more "vocal" wrt. error checks.

backmari:
I tested the Model Editor on Linux. One small bug is that if I have an error in my model and then fix it, then the red border around the function text box goes away, but the tool tip is not reset:
Screenshot from Screencast from 02-20-2025 07:13:34 AM webm

A minor UX complaint is that I have to double click a cell to see it as editable (blinking cursor). After a while I discovered that the cell is editable also when selected with one click. This is different from e.g. the Fit Panel, where I get a blinking cursor when selecting a cell with one click. Not something that needs to be fixed right now though :slightly_smiling_face:

src/sas/qtgui/Utilities/PluginDefinition.py
summerhenson: At the very least, this is causing an extra blank row to be left if you delete the contents of the last row before the default empty one.

summerhenson:
Related to the bug Marie found, if you have an error in your model, click Apply, fix the error, and try to Apply again, you get an error message about the plugin name already being used.
Screenshot from 2025-02-20 12-12-19

My minor UX complaint is that you can replace a cell's contents after single clicking, but not delete - you have to double click for that. Double clicking for a cursor made sense to me since that's pretty standard for spreadsheets, but that comparison did mean that I spent about 5 minutes expecting it to work exactly like a spreadsheet and thinking it was impossible to delete a row since I could otherwise edit them just fine and it didn't occur to me to double click.

krzywon:
> * On mac when not in full screen the editor window pops up behind the application. Is it too difficult to put it within the MDI workspace?

This has been a problem for a number of windows. Let me see if I can fix it.

> * On mac plugin manager selecting Edit... for a model opens a new editor window even if there is already one open. Expected behaviour is to raise the existing window to the front.

This is expected behavior. Multiple editors open is by design.

> * The traceback in the error log is hard to read with all the line breaks and indentation removed.

This might be a next version type of thing.

> * Should check parameter table for name and initial value when creating model. If name is blank then it ignores all subsequent parameters. If value is blank it defaults to 1.

Again, by design - completely blank lines should be deleted (properly after my recent change) so any blank parameter names should stop file generation.

> * Need to get current value from all fields on apply. Changing the name or value of the parameter then clicking apply uses the value from before the edit.

This was fixed in a recent commit.

> * Parameter table cannot input units, limits or description (future enhancement)

See #3135 - Some work to enhance this editor is happening. If these suggestions aren't already there, it should be easy to implement them.

krzywon:
@summerhenson - I've noticed the same issue and it's too much for this PR (and 6.0.1).

pkienzle:
> > * On mac plugin manager selecting Edit... for a model opens a new editor window even if there is already one open. Expected behaviour is to raise the existing window to the front.
> 
> This is expected behavior. Multiple editors open is by design.

Two editor windows for the same model.

I suppose there is a use case: look at the previous implementation in one window while you update it in a second window, but I suspect this will mostly cause confusion.

PR #2917, opened 2024-06-28
MacOS universal binary

wpotrzebowski:
## Description

Testing if macos-13 build produces universal binary

### DIFF
--- a/.github/workflows/matrix.py
+++ b/.github/workflows/matrix.py
@@ -28,7 +28,7 @@ jobs = []
 os_release_list = [
     'ubuntu-20.04',
     'windows-latest',
-    'macos-latest',
+    'macos-13',
 ]
 
 # List of OS images to use for release tests


tsole0:
N.B. pinned packages cannot be moved to pinned-import section without additional changes. It seems that the installers need some packages to be installed before others (i.e. they must come before others in `requirements.txt`). Maybe we could move the pinned section to the top? But I have a feeling PySide needs to be one of the final packages, and this approach isn't sustainable if we need to pin an arbitrary package later.

wpotrzebowski:
So far I haven't seen any big differences in performance of "universal" installer and the one built on macOS-latest. What is consistent is that both perform much worse on GPU compare to "No OpenCL"

wpotrzebowski:
@krzywon sure! Sorry, I didn't see your comment earlier.

krzywon:
As discussed at the meeting today, I downloaded the runner built using Intel architecture onto my M2. The time it takes the program to load is significant, but only on the first launch. First launch took nearly two minutes, but subsequent launches were comparable to the launch time for the app built using silicon technology.

wpotrzebowski:
As discussed with @klytje and @krzywon I will try to create a truly universal library with both architectures included and not only intel one.

krzywon:
Based on the feedback by @timsnow in https://github.com/SasView/sasview/issues/2919#issuecomment-2301586680, we might want to pivot and build binaries for both. If you are unable to find a solution to create a truly universal binary, @wpotrzebowski, I can take that task on later today.

krzywon:
@wpotrzebowski - the dual installation generation seems to be successfull! I booted up my M2 to test both the macos-13 and macos-latest binaries. The Intel one takes a significant amount of time to load, but runs fine otherwise. The Silicon binary runs as expected. This might not be a draft much longer.

wpotrzebowski:
Thanks for testing @krzywon! Indeed, it should be ready to be reviewed now.

krzywon (approved):
I've tested the installers on an M2 Mac and both binaries launch. Someone will need to test an older Mac to be sure the macos-13 one launches appropriately.

wpotrzebowski:
I can ask Lionel (who initially reported the problem) but will need to expose it to the outside world. Is it ok if I hijack the beta2 asset (https://github.com/SasView/sasview/releases/download/v6.0.0-beta-2/SasView-6.0.0-alpha-.matrix.os.dmg) for it?

krzywon:
So long as you keep the original, and only create a second, intel-specific link there, I don't have a problem with that.

krzywon:
The Intel-based build was tested and shown to run properly on older Mac systems. I believe this is ready to merge, but I will hold off until our meeting tomorrow.

Issue #2913, opened 2024-06-27
corfunc-technical needs attention


smk78:
**Describe the bug**
In corfunc-technical.rst:

- the Stribeck polydispersity needs defining
- there are some incorrectly formatted equations at the foot of the page

lucas-wilkins:
Could have sworn I fixed this?

smk78:
![image](https://github.com/SasView/sasview/assets/10938679/9d0ebec0-155a-40e2-952f-f8a4fb21081a)

lucas-wilkins:
OK, so its always been like this. I can have a look.

PR #2923, opened 2024-07-01
Pin Scipy

lucas-wilkins:
Scipy latest release breaks things, pin until we can fix properly

See #2913


lucas-wilkins:
@krzywon Can you merge this asap? (assuming checks pass)

PR #2944, opened 2024-07-05
move new pinned dependencies to pinned section.

butlerpd:
Just moves the 2 new pinned versions (numpy and scipy) to the pinned version of `requirements.txt` and adds comments as to why they are pinned.


lucas-wilkins:
Fine, but why make two lists?

butlerpd:
> Fine, but why make two lists?

two lists? Not sure I follow. @krzywon alphabetized and organized pinned vs unpinned (and OS specific) into sections a while back, I assume because we have so many dependencies we were doubling up and also losing track of what is pinned etc?  We then pinned numpy and scipy (which were not pinned) without moving them to the pinned section.  If this is not useful then we should remove that section and reorder all the pinned back into the list? What about OS specific in that case? Or should that PR never have been merged in the first place?

Or is there something I did that I'm not noticing? Sorry for my confusion.

lucas-wilkins:
It doesn't really matter, just wondering why. It means you have to search in two parts of the file for a given package, not just one.

rozyczko:
I think the current solution is cleaner since we don't have a particularly long list of dependencies. They all neatly fit even on a small screen and placing them in sections for standard, platform-dependent and pinned makes sense.
Maybe the pinned version section could go to the top, though, since it is the most important (?)

rozyczko (approved):

Issue #2910, opened 2024-06-27
SasView does not restart after Preferences change


smk78:
**Describe the bug**
Having changed Preferences > Display Options to set automatic scaling the popup said that SasView needed to restart for the change to be applied. I confirmed the restart and the following appeared in an error popup window:

```
Unhandled exception in script
Failed to execute script 'sasview' due to unhandled exception: source code string cannot contain null bytes
Traceback (most recent call last):
  File "sasview.py", line 15, in <module>
  File "sas\cli.py", line 156, in main
  File "<frozen runpy>", line 290, in run_path
  File "<frozen runpy>", line 259, in _get_code_from_file
SyntaxError: source code string cannot contain null by
```

Issue #2388, opened 2022-11-22
Preference Panel High DPI Setting Issues


krzywon:
**Describe the bug**
The Display Settings in the preference panel has a few issues that could be resolved.

- Set a min and max to the settings that are reasonable (0.5 to 2.0)?
- Give a warning if the value changes by too much

**Additional context**
The checklist in [#2316](https://github.com/SasView/sasview/issues/2316#issuecomment-1295223654) has a bullet to show a preview window. Adding the preview window in tandem with this fix would be helpful.

PR #2950, opened 2024-07-11
2910 and 2446: Fixes to auto-restart after preferences change

### LINKED ISSUE #2446: `print` statements fail after preference restart
**Describe the bug**

`OSError(22)` is produced by print statements after an automatic restart is performed.

**To Reproduce**
Run `run.py` (I used PyCharm)
Change a preference that requires restart
Save, and run some code that has a print statement

 - Version: main

krzywon:
## Description

This fixes the issues related to restarting the installed version of SasView after making a preference change. This allows for different behaviors when running from python or from the executable.

Fixes #2446, #2910

## How Has This Been Tested?

I tested the installed version from https://github.com/SasView/sasview/actions/runs/9893932935 for both Windows and Mac. The auto-restart seems to work when changing the DPI and scaling values.

## Review Checklist:

**Installers**


wpotrzebowski (commented):
It seems that there are only two options i.e. python and main executable do we need "pass" 3 times?

krzywon:
> do we need "pass" 3 times?

Whoops, no. I had those in for debugging and just removed them.

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
```diff
@@ -30,6 +30,19 @@
...
 
 
+def restart_main():
+    # Restart the application...
+    executable = sys.executable
+    executable_filename = os.path.split(executable)[1]
+    if executable_filename.lower().startswith('python'):
+        # application is running within a python interpreter
+        python = executable
+        os.execv(python, [python, ] + sys.argv)
```
lucas-wilkins, line 40: This line looks weird, are we sure its right? Should one be in quotes, just checking

lucas-wilkins (approved):
Code looks mainly sensible, bit unsure about one line though. Going to approve, but please check

src/sas/qtgui/Utilities/Preferences/PreferencesPanel.py
krzywon: Both pythons are already strings, so no quotes needed, but the extra one is required for the restart when running from source.

butlerpd (approved):
tested windows installer and works as advertised on re-start.  That said, I still worry about those changes that require a restart. I used scale = 0.5 and while it restarted it was not only unusable I could not even get to the preferences menu to fix. I had to edit the config-6.json file.   We probably need another ticket for that. Something like a restart popup that is NOT scaled asking "keep changes or reject changes" maybe?

That however is not related to this PR so will also approve on functionality so ready to merge

lucas-wilkins:
> I could not even get to the preferences menu to fix

What do you mean by that?

krzywon:
> > I could not even get to the preferences menu to fix
> 
> What do you mean by that?

I'm guessing the scaling was so bad that he couldn't access the menu to open the preferences panel. This is a potential real issue and needs to be fixed - see #2388.

PR #2954, opened 2024-07-12
2912 moving sliders in corfunc throws errors

lucas-wilkins:
This should fix the error messages being shown when sliders in corfunc move.


lucas-wilkins:
Well, that doesn't work.

lucas-wilkins:
@smk78 - should be all good now

smk78 (approved):
Moving the sliders no longer throws errors to the Log Explorer. Fixes the problem it set out to fix.

PR #2966, opened 2024-07-18
2916 gsc plot windows can open hidden - also add GSC plot icon

### LINKED ISSUE #2916: GSC plot windows can open hidden
**Describe the bug**
When your Draw a loaded structure in the GSC, the plot window seems to be able to appear silently in the background behind other windows. Until you get used to this you sit there wondering how long the rendering is going to take!

lucas-wilkins:
Should fix #2916  - however, my computer is too fast to reproduce the issue with any of the example data. @smk78 - can you check?

### DIFF
--- a/src/sas/qtgui/Calculators/GenericScatteringCalculator.py
+++ b/src/sas/qtgui/Calculators/GenericScatteringCalculator.py
@@ -1,5 +1,8 @@
 import sys
 import os
+
+from PySide6.QtCore import QSize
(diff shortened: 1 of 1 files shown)


krzywon (approved):

butlerpd:
I think that was a problem I was having when checking GSC compute was now working except that what I saw was that the magnetic compute never showed the final 2D plot.  It does in this PR  so that seems fixed.

PR #2967, opened 2024-07-18
2904 graph is not getting the correct axis

### LINKED ISSUE #2904: Graph is not getting the correct axis
**Describe the bug**
When creating a graph of one of my data sets in the 6.0.0 beta, the y-axis range were not getting correctly scaled. This was before any manipulation of the graph range, zoom, or scale change. Simply creating a new graph/fitting would cause the axis scale to be incorrect. The y-axis is a lot more zoomed in than it should be.

Here is the file that isn't graphing properly:

**To Reproduce**
1. Go to Load data
2. Click on file named "hH_0D2O_20C" (file shared below, unable to upload as it is an ABS file)
https://drive.google.com/file/d/17I1IVp6_P6V05vKwOLX4F81sgXCMsQMq/view?usp=sharing
4. Click on "Create New" plot
5. See error
**The error in Fitting**
6. Send to Fitting
7. Put in any model
8. Press compute/plot or fit
9. See error

**Expected behavior**
A graph that correctly auto scales. 

**Screenshots**
**6.0.0b1** graph, where the issue is:
image
**5.0.6** graph, to give context to what this graph is supposed to look like.
image

 - Version: 6.0.0b1

 - OS: Windows 10

**Additional context**
I went through the files with Paul Kienzle, and we found that in Plotter.py line 284, the default_x_range  #and default_y_range are being set using self.ax.get_xlim() and self.ax.get_ylim(), these are not getting the actual default ranges given by matplotlib. We put a self.ax.relim() and self.ax.autoscale() above line 284-285 and found that the scale is then fixed.

lucas-wilkins:
Plots now get their default scaling from the data given to them. Hope this is the right solution and I've not missed anything.

Fixes #2904


krzywon:
We had something like this in the past, where the plots were scaled to the data, but this can create significant issues when a single outlying data point is loaded in. I don't think moving away from built-in MPL functionality is the right answer in this case. Instead, we need to investigate why the data set used by @XaelShan is being plotted the way it is. I have the file that triggered this and will add it to the original issue.

lucas-wilkins:
I think that the reason is that there is a minimum range, and the data is very flat, matplotlib matplotlib centres on the range, but then its logged, pushing it to the top.

 I did look at the data before doing this.

I don't see what the issue is with using this scaling method (I do remember people discussing before), I don't see how it would be problem with a single point, or with outliers - there are potential problems I can see with doing it naively, but scaling the range endpoints ensures that the range isn't zero, and that none of MPLs weird edge-of-plot cropping happens.

XaelShan:
I believe I have another dataset that also triggers this graphing problem. I will test it again and add it to the issue.

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -280,9 +280,27 @@ def plot(self, data=None, color=None, marker=None, hide_error=False, transform=T
             x_range = self.setRange.xrange()
             y_range = self.setRange.yrange()
         else:
-            # Use default ranges given by matplotlib
-            x_range = default_x_range = self.ax.get_xlim()
-            y_range = default_y_range = self.ax.get_ylim()
+            if isinstance(data, Data1D):
+                # Get default ranges from data
+                # factors of .99 and 1.01 provides a small gap so end points not shown right at edge
+                default_x_range = (0.99*np.min(x), 1.01*np.max(x))
```
rozyczko, line 286: Multiple occurences of these two parameters `0.99` and `1.09` would suggest replacing them with predefined (even local) constants.

rozyczko (approved):
Code looks good. Minor stylistic comment only.

PR #2977, opened 2024-07-21
2976 auxsas has nondeterminstic tests

### LINKED ISSUE #2976: AUXSAS has nondeterminstic tests
Make tests deterministic

lucas-wilkins:
## Description

AUXSAS has tests that rely on random numbers without setting a seed, this will (and has) cause random test failures.

Set seed

Fixes #2976

## How Has This Been Tested?

Tests passing will test it

### DIFF
--- a/test/sascalculator/utest_sas_gen.py
+++ b/test/sascalculator/utest_sas_gen.py
@@ -228,6 +228,8 @@ class sas_gen_test(unittest.TestCase):
         from sas.sascalc.calculator.ausaxs import sasview_sans_debye
         from sas.sascalc.calculator.ausaxs import ausaxs_sans_debye
 
+        rng = np.random.default_rng(1984)
+
         from sas.sascalc.calculator.ausaxs.architecture import get_os, OS
         if get_os() is OS.MAC:
             self.assertTrue(True, "AUSAXS library is currently not available for MacOS. Skipping test.")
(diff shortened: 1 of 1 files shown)


test/sascalculator/utest_sas_gen.py
```diff
@@ -241,6 +241,8 @@ def test_debye_impl(self):
         import glob
         pdb_files = glob.glob(os.path.join(os.path.dirname(__file__), 'data/debye_test_files', '*.pdb'))
 
+        np.random.seed(1984)
```
andyfaff, line 244:
```suggestion
        rng = np.random.default_rng(932810983)
```

andyfaff: Then on line 251, 269:

```
w = rng.random(size=(coords.shape[1],))
```

andyfaff: `Generators` are recommended for new code.

lucas-wilkins: They are, but its not new code. Change it if you like.

klytje: `w` is being passed to an external C function as a memory pointer. I don't think this will work as a generator?

klytje (approved):
Seems good to me. I never thought the weights could affect the result enough to push it beyond the 1% deviation limit. I strongly suspect the C60 structure to be responsible.

test/sascalculator/utest_sas_gen.py
lucas-wilkins: @klytje the generator just gets called the same-ish as np.random and makes an array.

lucas-wilkins:
> Seems good to me. I never thought the weights could affect the result enough to push it beyond the 1% deviation limit. I strongly suspect the C60 structure to be responsible.

There's often a finite chance of failure when you use random numbers, I've only seen it fail once since being merged, so it must be a very low chance. I'd say always use a seed in unit tests.

krzywon:
I just saw a failure of this in https://github.com/SasView/sasview/actions/runs/10059940403?pr=2917. I'm merging this now and will rerun the failing run to see if it corrects itself.

PR #2980, opened 2024-07-22
2979: Ensure multiplicity row is set when loading projects

### LINKED ISSUE #2979: Scale Parameter not Loading From Fit Projects
**Describe the bug**

The `scale` parameter is not being loaded from any project, not just constrained fit projects as was originally reported. All other parameters load properly, but the parameter table values for scale remain at their defaults.

Two representative projects that show the issue:
[testingsave.json](https://github.com/user-attachments/files/16338509/testingsave.json)
[constrained-fit-project.json](https://github.com/user-attachments/files/16338547/constrained-fit-project.json)

The projects load as expected in v5.0.6, so this is a v6.0 issue.

Reported by @XaelShan.

**To Reproduce**
1. Load either of the projects linked here.
2. Look at the scale parameter.
3. The scale should not be 1.0 and should be selected for fitting in most of the fit tabs.

**Expected behavior**
The proper values should be loaded.

 - Version: v6.0.0a1 and above

krzywon:
## Description

The multiplicity row is set to 0 by default, which is the `scale` parameter. When loading a project, this row is ignored because it is handled elsewhere. This sets the row to -1 if there is no multiplicity parameter.

Fixes #2979

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -4467,6 +4467,9 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             if self.kernel_module.is_multiplicity_model:
                 self.kernel_module.multiplicity=multip
                 self.updateMultiplicityCombo(multip)
+        else:
+            self._n_shells_row = -1
+            self._num_shell_params = -1
 
         if 'tab_name' in line_dict.keys() and self.kernel_module is not None:
             self.kernel_module.name = line_dict['tab_name'][0]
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -4467,6 +4467,9 @@ def updatePageWithParameters(self, line_dict, warn_user=True):
             if self.kernel_module.is_multiplicity_model:
                 self.kernel_module.multiplicity=multip
                 self.updateMultiplicityCombo(multip)
+        else:
+            self._n_shells_row = -1
+            self._num_shell_params = -1
```
rozyczko, line 4472: This is good, but maybe setting the default values for these two in lines 266/267 is even better?
(Seems to work as well)

rozyczko (approved):
Code looks good, potential modification requested.

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
krzywon: I think I found a conflict doing what you are suggesting during testing, but I can't remember what that was right now. I'll retest and add some inline documentation on why I chose this solution.

PR #2986, opened 2024-07-23
2982: Allow density to be blank in SLD Calculator

### LINKED ISSUE #2982: SLD calculator does not respect density in formula
**Describe the bug**
for mixtures, the SLD calculator supports adding the individual densities in line rather than having to calculate them manually. This works in the online version (https://www.ncnr.nist.gov/resources/activation/) assuming the density box is left empty. However leaving that box empty in SasView still assumes 1.0 (incorrect) and ignores the in line densities. eg entering `50%vol H2O@1 // D2O@1.1` will act as if the density of the mixture is 1 or whatever value the user puts into the density box.  This defeats the purpose.

**To Reproduce**

enter the above equation into the calculator, the SLD should be 2.887 but SasView will return 2.750 (what you would get assuming a density of 1 for the mixture) as given by the web calculator 

 - Version: 6.0.0b2 However it seems to also exist in 5.0.6

 - OS: Windows 10

**Additional context**
This seems to be a known bug and is mentioned in the documentation which tells the user to manually calculate the mixture density themselves and input that.

krzywon:
## Description

This allows the density value to be blank in the SLD calcuator tool for cases where the density is specified in the formula.

Fixes #2982

## How Has This Been Tested?

Tested locally that the result from the issue is the result when the formula and no density are entered.

## Review Checklist:

- [ ] There is an **issue** open for the documentation #3080 

**Installers**


wpotrzebowski:
I am still getting 2.75 from `50%vol H2O@1 // D2O@1.1 `

krzywon:
Check the updates I just added. This should be more robust in error catching before the information is sent for calculation.

rozyczko (approved):
the code looks good and the fix addressed the original issue.

wpotrzebowski:
Ok, I didn't read the instructions properly... It works. I am wondering however how users should know about the blanking density field? I guess one option is to leave it blank by default but then we won't be able to start from the example formula. We probably can add warning or at least tooltip?

krzywon:
> I am wondering however how users should know about the blanking density field?

I think a good solution for this is, if the density is specified in the formula, we could disable and gray out the density field and ignore any value in the box. This would give a visual user cue something has happened, while keeping the calculation true to the user intent.

butlerpd:
Actually, This issue will require some small changes to the documentation. I will open an issue around that.

butlerpd (changes requested):
The code changes look reasonable and already approved by @rozyczko. testing fairly thoroughly the parser seems to be doing a good job of robustly managing not only the new changes but the problem of re-computing while typing causing errors when one does not type too fast.

However, I do not believe the changes to the documentation are sufficient. Everything from `Specifying Materials or Mixtures in the Molecular Formula Field` down to `Biomolecules` should be read and edited appropriately.  This could be done in a separate PR as originally suggested or done in this PR given the docs have already started to be edited in this PR. Assuming the latter, this still needs changes. However if we decouple then this can be merged IMO

butlerpd:
@krzywon it may be more expedient to open another issue after all and let me do the documentation edits?  Up to you.

krzywon:
> it may be more expedient to open another issue after all and let me do the documentation edits? Up to you.

Working on it now in this branch

krzywon:
@butlerpd - take another look when you get a chance. I have updated the documentation and made another fix related to the formula entry to ensure all formulas in the documentation are valid.

butlerpd:
OK @krzywon - will work on that next then

butlerpd (commented):
went deep in a rabbit hole on this. It is much more complicated that I realized and I see why it took @katieweigandt so long to get the docs sorted! I now see why greying out the box may not always be the desired behavior. individual densities are required for all vol% entries in order to get the total scattering length. However non-ideal mixing means it may not always give you the right materials density.

I know there was some discussion about this earlier between @wpotrzebowski and @krzywon which led to replacing the "leave blank" with "grey out." I will admit that sounded good to me too this morning, but now I think that was the wrong answer. As @katieweigandt kept emphasizing in the docs: the best is to actually measure the density of your solution if you can.  It is just that this is often not easy to do. Thus I think the right answer would be to go back to the density box overriding but allowing it to be blank without it defaulting to 1.0. That is I think how it works now if we just remove the `.setEnabled(False)` line (but keep `density = None`). The error checking is much more robust now so that should work? It may require a bit more editing of the docs? There would be the earlier question by @wpotrzebowski about some better way of alerting the novice user I guess?

A couple of other points to note:
* Entering `@` , without any numbers following it, defaults to `@1.0` - I will argue that is reasonable and not worth trying to capture. If the user stops there I'm not sure what they would expect other than 1?
* Nested expressions may not grey out the density line even if the outer X components all have their densities reported. The density box always supersedes (see example in docs `20%vol (10%wt NaCl // H2O)@1.07 // D2O@1.11`) - This should probably be the default behavior anyway if we accept what I said above.

butlerpd (approved):
This looks good now. Looking at the code change I have not tried to do the full extensive testing again as only entries with multiple components will be affected (containing \\\\). 

This nicely addresses the problem of not allowing for non-ideal mixing and makes the SasView SLD app completely compatible with the online version I think. Moreover the color coded and tool tip on the density override is quite a nice aid for the user. All real errors, which will raise an exception from the `periodictables.py` all get caught and make the offending box background yellow so not worried about any combination that is not valid.

One confusing point about `periodictables.py` that maybe should be noted in the documentation is that `periodictables.py` knows the density of the elements and will assume them if a mixture does not specify its density.  But this is not related to this PR so I vote this can be merged and be called done for 6.0.0

PR #2991, opened 2024-07-25
2741: Fix PD parameter issue

### LINKED ISSUE #2741: Npts and Nsigma issue in polydisperse distributions
**Describe the bug**
Parameters npts and nsigma do not seem to be used correctly. Only the default values (npts=35, nsigma=3) produce the expected results. Overriding those values currently creates wrong distributions and in turn results in wrong fittings on data. 
See examples in the screenshots below. It seems like if the last parameter changed is PD[ratio] instead of npts or nsigma, the distribution will just be plotted with default values (like in Figure 5), but if the last parameter changed is npts or nsigma, a completely wrong distribution with a seemingly random number of points will be generated.

**To Reproduce**
1. Go to any shape-dependent model with polydispersity option checked 
2. Changing npts and nsigma will produce errors in the shape of the distribution and the number of points used to plot the distribution. 
3. Click 'Compute/plot'

**Expected behavior**
The distribution should appear correctly based on nsigma with the number of points specified by npts.

**Screenshots**
Figure 1:
Screenshot 2024-01-18 104911 
Figure 2:
Screenshot 2024-01-18 105124
Figure 3:
Screenshot 2024-01-18 105221
Figure 4:
Screenshot 2024-01-18 105649
Figure 5:
Screenshot 2024-01-18 105843

 - Version:6.0.0a1 

 - OS: Windows 11

**Additional context**
@murphyryanp discovered this issue initially. We think the problem could be in the sasview repo or the sasmodels repo. In sasview, the most recent updates that can affect this are in FittingWidget.py. In sasmodels, it seems like there's an update to the direct_model.py that changes how npts is calculated, but unclear if it's really any different from the older version.

krzywon:
## Description

A change to the PD params resulted in `width`, `npts`, and `nsigmas` to all change `width` of the underlying model. This corrects that behavior.

Fixes #2741

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1811,7 +1811,9 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             # PD[ratio] -> width, npts -> npts, nsigs -> nsigmas
             if model_column not in delegate.columnDict():
                 return
-            self.poly_params[parameter_name_w] = value
+            # Map the column to the poly param that was changed
+            associations = {1: "width", 4: "npts", 5: "nsigmas"}
+            self.poly_params[f"{parameter_name}.{associations.get(model_column, 1)}"] = value
             self.kernel_module.setParam(parameter_name_w, value)
 
             # Update plot
@@ -2218,6 +2220,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         """
         Update the model with new parameters, create the errors column
         """
+        print(param_dict)
         assert isinstance(param_dict, dict)
 
         def updateFittedValues(row):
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1811,7 +1811,9 @@ def onPolyModelChange(self, top, bottom):
             # PD[ratio] -> width, npts -> npts, nsigs -> nsigmas
             if model_column not in delegate.columnDict():
                 return
-            self.poly_params[parameter_name_w] = value
+            # Map the column to the poly param that was changed
+            associations = {1: "width", 4: "npts", 5: "nsigmas"}
+            self.poly_params[f"{parameter_name}.{associations.get(model_column, 1)}"] = value
```
rozyczko, line 1816: Shouldn't that be `.get(model_column)` ? I think we're guaranteed to get the proper column value here. Alternatively, maybe use "width" as the default, not `1`?

rozyczko (approved):
I was unable to reproduce the issue, so can't say this fixes the problem, but the code looks OK.

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
krzywon: Yes! I'll fix that.

PR #3000, opened 2024-07-30
2999 Fix slow startup time

### LINKED ISSUE #2999: Startup is slooow again
**Describe the bug**
Startup time increased significantly.

**To Reproduce**
1. Click on the sasview icon or
2. Run `python run.py`
3. Wait

**Expected behavior**
Wait time below 5s

 - Version: 6.0.0

 - OS: all

rozyczko:
## Description

GenericScatteringCalculator instantiation in GuiManager includes importing `sas.sascalc.calculator.geni`.
This module has been extensively "numbaised" and its import is very slow due to numerous JIT decorators.

The simplest/safest solution seems to be postponing the import until the GSC is actually needed.

Price paid: starting the GS calculator takes now a long time.

Fixes #2999 

## How Has This Been Tested?

Local testing on win10

## Review Checklist:

- [ X] There is **nothing** that needs documenting

**Installers**

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -206,6 +206,7 @@ class GuiManager:
         self.KIESSIGCalculator = KiessigPanel(self)
         self.SlitSizeCalculator = SlitSizeCalculator(self)
         self.ResolutionCalculator = ResolutionCalculatorPanel(self)
+        self.GENSASCalculator = None
         self.DataOperation = DataOperationUtilityPanel(self)
         self.FileConverter = FileConverterWidget(self)
         self.WhatsNew = WhatsNew(self)
@@ -1041,7 +1042,8 @@ class GuiManager:
         try:
             # delayed import due to numba instantiation in GSC
             from sas.qtgui.Calculators.GenericScatteringCalculator import GenericScatteringCalculator
-            self.GENSASCalculator = GenericScatteringCalculator(self)
+            if self.GENSASCalculator is None:
+                self.GENSASCalculator = GenericScatteringCalculator(self)
             self.GENSASCalculator.show()
         except Exception as ex:
             logging.error(str(ex))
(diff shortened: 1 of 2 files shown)


src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -206,7 +205,6 @@ def addWidgets(self):
         self.DVCalculator = DensityPanel(self)
         self.KIESSIGCalculator = KiessigPanel(self)
         self.SlitSizeCalculator = SlitSizeCalculator(self)
-        self.GENSASCalculator = GenericScatteringCalculator(self)
```
lucas-wilkins, line 209: `self.GENSASCalculator: "GenericScatteringCalculator" | None = None` instead? - keeps all variables declared in __init__ (or functions called from it at least)

lucas-wilkins (approved):
Seems reasonable. I think it would be good to make sure that we initialise all variables though.

src/sas/qtgui/MainWindow/GuiManager.py
rozyczko: This and also not reinstantiating the GSC but reusing its instance. Fixed

rozyczko: We have it already. `self.GENSASCalculator` lifetime is that of the application. Calling `show()` just makes the widget visible, while retaining the state.

lucas-wilkins: @rozyczko I don't think we do address what I was getting at (I should have said `closeEvent` btw, not `onClose`) - it needs to be set to not destroy the C++ side when the close signal is sent (apparently, this is dependent on Qts whims), as opposed to the issue with garbage collecting things on the python side. I've had bugs from this before, they we not fatal, but lead to things breaking in a subtly weird way - it's why the acknowledgements dialogue gets explicitly removed and re-instantiated.

Problem is python keeping alive something that Qt thinks should be dead, rather than python deleting something that Qt thinks should be dead.

lucas-wilkins:
Also PR 3000 !!!

src/sas/qtgui/MainWindow/GuiManager.py
rozyczko: You are right, `closeEvent` was reimplemented in other classes in this file, not the GSC class itself. Fixed.

PR #3007, opened 2024-08-06
2993 ascii file reader options

jamescrake-merani:
Created a prototype dialog menu for selecting various parameters for reading in an ASCII dataset file.


rozyczko:
Two minor suggestions:
- For Qt widget code, please place your files in the appropriate location in the source tree. This code should probably be placed somewhere in  `src/sas/qtgui/Calculators`.

- Our coding style for `QObject` derived classes (and logic around it) is the `PyQt` style, not PEP8:

```python
class MyWidget(QWidget):
    def loadItems(self):
        print 'loading items'
        
    def refreshResuls(self):
        print 'refreshing results'
```

jamescrake-merani:
Thank you for your suggestions Piotr.

>  For Qt widget code, please place your files in the appropriate location in the source tree. This code should probably be placed somewhere in src/sas/qtgui/Calculators.

Yes I see your point. I only had them in the current directory temporarily as I wasn't sure exactly where the code should have gone but I will move them to a permanent location later.

>  Our coding style for `QObject` derived classes (and logic around it) is the `PyQt` style, not PEP8:

Does this mean the methods in classes derived from `QObject` should use camel case names, and not snake case? Is this the same for attributes within the class?

rozyczko:
> Does this mean the methods in classes derived from `QObject` should use camel case names, and not snake case? Is this the same for attributes within the class?

Correct. Class names should be pascal case (`class BestClassEver`), method names should be camel case (`def bestMethodEver`) and attributes/variables should follow the snake case (`self.my_awesome_variable`)

(e.g. https://bitesofcode.blogspot.com/2011/10/pyqt-coding-style-guidelines.html )

src/ascii_dialog/dialog.py
```diff
@@ -0,0 +1,383 @@
...
+        """Returns all the columns that are required by the dataset type but
+        have not currently been selected.
+
+        """
+        dataset = self.current_dataset_type()
+        missing_columns = [col for col in dataset.required if col not in self.col_editor.col_names()]
+        return missing_columns
+
+    def duplicate_columns(self) -> set[str]:
+        """Returns all of the columns which have been sselected multiple times."""
```
pkienzle, line 368: typo: "sselected"

src/ascii_dialog/guess.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
```
pkienzle, line 1: No `main()`. Not a standalone executable.

src/ascii_dialog/guess.py
```diff
@@ -0,0 +1,32 @@
...
+
+from sasdata.dataset_types import DatasetType
```
pkienzle, line 3: Does this logic belong in `sasdata`? That is, could it be used by a script, a jupyter notebook or a web gui?

lucas-wilkins: It will be used by a file loader in sasdata too.

src/ascii_dialog/guess.py
jamescrake-merani: My editor likes to put this into new Python files but yes it doesn't really make sense for this to be here. I'll remove it.

jamescrake-merani:
@rozyczko 

> Correct. Class names should be pascal case (`class BestClassEver`), method names should be camel case (`def bestMethodEver`) and attributes/variables should follow the snake case (`self.my_awesome_variable`)

I've gone through, and corrected the method names to camel case. I just had one question: should properties use camel case, or snake case?

rozyczko:
@jamescrake-merani  
> I've gone through, and corrected the method names to camel case. I just had one question: should properties use camel case, or snake case?

Good question. I can see that we're pretty inconsistent here. These are method names, so we should use camel as described in the mentioned document.

src/ascii_dialog/dialog.py
```diff
@@ -0,0 +1,383 @@
...
+        # The table's width will always resize to fit the amount of space it has.
+        self.table.setSizeAdjustPolicy(QAbstractScrollArea.SizeAdjustPolicy.AdjustToContents)
+        # Add the context menu
+        self.table.setContextMenuPolicy(Qt.ContextMenuPolicy.CustomContextMenu)
+        self.table.customContextMenuRequested.connect(self.showContextMenu)
+
+        # Warning Label
+        self.warning_label = WarningLabel(self.requiredMissing(), self.duplicateColumns())
+
+        self.layout = QVBoxLayout(self)
```
rozyczko, line 104: In SasView we try to maintain consistency in terms of GUI design and have the UI code abstracted as much as possible, with all layouts defined in the .ui files (which then get converted into .py).
Using Qt Designer is very helpful in creating well-behaving dialogs and windows and with debugging layout issues without the endless modify code -> start sasview -> test functionality -> close sasview -> modify code.... loop.

However, this is a small widget so it's probably OK to create it explicitly in Python - we have several cases like this already (DataOperationUtilityPanel, some widgets in the resolution calc. etc)

src/ascii_dialog/dialog.py
```diff
@@ -0,0 +1,383 @@
+from PySide6.QtGui import QColor, QContextMenuEvent, QCursor, Qt
+from PySide6.QtWidgets import QAbstractScrollArea, QCheckBox, QComboBox, QFileDialog, QHBoxLayout, QHeaderView, QLabel, QMessageBox, QPushButton, QSpinBox, QTableWidget, QTableWidgetItem, QVBoxLayout, QWidget, QApplication
```
rozyczko, line 2: This line could be split into multiple lines.

src/ascii_dialog/dialog.py
```diff
@@ -0,0 +1,383 @@
...
+    def attemptGuesses(self) -> None:
+        """Attempt to guess various parameters of the data to provide some
+        default values. Uses the guess.py module
+
+        """
+        split_csv = [self.splitLine(line.strip()) for line in self.raw_csv]
+
+        self.initial_starting_pos = guess_starting_position(split_csv)
+
+        guessed_colcount = guess_column_count(split_csv,
```
rozyczko, line 160: Our standard is 120 chars per line so there is no need to split the lines so aggressively

jamescrake-merani: I see your point. I get my editor to split the lines automatically, and it looks like it was set to only 80 characters so I've bumped it up to 120.

src/ascii_dialog/dialog.py
jamescrake-merani: I see your point. I have QT Designer installed but I must admit that I haven't really used it before as most UI code I've written in the past has been done explicitly in code. But I will take a look at it more closely going forward.

jamescrake-merani:
I met with Lucas who suggested the following improvements to the UI.
- [ ] The label for the filename is a redundant because the tool also shows the current filename on the combo box.
- [ ] 'Load File' should become 'Select File'.
- [ ] Cancel button to quit the dialog without loading anything.
- [ ] Ignored columns should disable the unit combo box
- [ ] We don't really need the 'Included' label. Just have the checkbox, and centre it.
- [ ] The column selectors can be problematic if the window is too small, and there are too many of them. I need to look into this. Potentially by having a horizontal scroll menu but this may be a bit tricky to implement.
- [ ] Guesses still need improvements. Particularly, it gets the amount of columns on some files completely wrong. Lucas suggested looking at the first n lines (e.g. n = 100), and selecting the mode of columns.
- [ ] The wording of the warning labels can be confusing. Needs to be improved.
- [ ] Broken on dark mode (though this may be a general QT issue).
- [ ] The edit metadata button should be at the bottom.

jamescrake-merani:
Regarding

>  The column selectors can be problematic if the window is too small, and there are too many of them. I need to look into this. Potentially by having a horizontal scroll menu but this may be a bit tricky to implement.

I think this is less problematic when there are a normal amount of columns. Usually there is enough space, and it is easy for the user to add space if needed. I think this problem was worse when the guess of columns was wrong, and there were far too many of them. I'm therefore not sure if adding a horizontal scroll bar is needed.

krzywon (commented):
Comments and Suggestions:
- The column guesser seems pretty good in finding the numerical values.
- The data tool is pretty intuitive and works well (once it is connected to the data explorer)
- The metadata tool is confusing to use when files don't have metadata in the filename and might need to be simplified.
- Default units for dI and dQ don't make sense - they should match the units of I and Q.
- Intensity can have multiple unit types (1/cm, arbitrary, cts, Sr/m, etc) for SAS data. Is there a way to specify these other unit types?
- The package locations should probably be moved somewhere within the `qtgui` package at some point.

jamescrake-merani:
Thanks for the suggestions.

> The metadata tool is confusing to use when files don't have metadata in the filename and might need to be simplified.

I see your point. This is something I'll need to think about, and perhaps discuss with Lucas.

> Default units for dI and dQ don't make sense - they should match the units of I and Q.

This makes sense. I'll make it so the units always match what I, and Q's units are.

> Intensity can have multiple unit types (1/cm, arbitrary, cts, Sr/m, etc) for SAS data. Is there a way to specify these other unit types?

Not at present but I think perhaps I could make it so that each column could potentially have multiple unit types? And then I could specify all the unit types that intensity could have.

> The package locations should probably be moved somewhere within the qtgui package at some point.

Yes I agree its not in the right place. I suspect we'll probably end up moving it a bit later in the refactor branch as we may need to make some other decisions as to where new code goes.

src/ascii_dialog/col_editor.py
```diff
@@ -0,0 +1,87 @@
...
+    def onColumnUpdate(self):
+        column_changed = cast(ColumnUnit, self.sender())
+        pairing = pairings.get(column_changed.currentColumn)
+        if not pairing is None:
+            for col_unit in self.option_widgets:
+                # Second condition is important because otherwise, this event will keep being called, and the GUI will
+                # go into an infinite loop.
+                if col_unit.currentColumn == pairing and col_unit.currentUnit != column_changed.currentUnit:
+                    col_unit.currentUnit = column_changed.currentUnit
+        self.column_changed.emit()
```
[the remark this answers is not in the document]
jamescrake-merani: Yes I think you're right. I don't recall why I had this line here. Anyhow, I've deleted it.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -1404,4 +1405,12 @@ def particleEditor(self):
...
\ No newline at end of file
+        from ascii_dialog.dialog import AsciiDialog
+        dialog = AsciiDialog()
+        status = dialog.exec()
+        if status == 1:
+            loaded = load_data(dialog.params)
+            for datum in loaded:
+                logger.info(datum.summary())
+        else:
+            logger.error('ASCII Reader Closed')
```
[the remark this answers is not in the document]
jamescrake-merani: I only had this as a test for the status code. Eventually, the ASCII reader won't be in the dev menu but will be integrated into the new data explorer so this code won't be here anyway. I was thus thinking of keeping it for now but it will not stay there.

jamescrake-merani:
Taking a closer look, it might be a bit tricky for intensity to have more than one unit group because the unit selector dialog that appears when you click 'Select More' on the combo box only lets you select one unit group.

Instead, I was thinking we could perhaps have one unit group for intensity as a whole. Unit groups are currently defined in `units.py` in SasData. For example, the time unit group:

```python
time = UnitGroup(
  name = 'time', 
  units = [
    seconds,
    milliseconds,
    microseconds,
    nanoseconds,
    picoseconds,
    femtoseconds,
    attoseconds,
    minutes,
    hours,
    days,
    years,
])
```
So this `units` list could potentially just be a combination of two (or more) unit groups. Let me know what you think.

jamescrake-merani:
Turns out I had some code that was supposed to make the units on I, and dI etc match but there was a bug in it which I've fixed, and also both columns should have the same options for untis.

Issue #3030, opened 2024-08-15
Installer fails to install properly in existing SasView folder


butlerpd:
**Describe the bug**
When installing SasView from the installer on windows, the installer suggests a folder to install to.  Currently if that folder already exists the user is given zero warning (see #3029). However if the user chooses a different folder and that one exists, a warning is flagged asking if the user really wants to continue.

**PROBLEM STATEMENT**
In either case, if the exiting folder contains an older SasView version, the new install will not work.
Since 6.0.0b2 has a different icon it can be noticed that the saview.exe icon is the old one not the one for the new purportedly installed version.  It is not clear why this is happening but suspect that instead of deleting the content of the folder only certain files are replaced? or maybe none and only new files (with names that don't already exist) are installed?  Whatever the case it would appear that some mismatch of the old and the new are co-existing making the combination not operational?

**To Reproduce**
See #3029  for an example

**Expected behavior**
It should work in the folder it installed to. I would guess it should delete the contents of the folder (or the whole folder itself) and then install everything?

**SasView version (please complete the following information):**
 - Version: 6.0.0b2

**Operating system (please complete the following information):**
 - OS: Windows 10

**Additional context**
Discovered as part of the NIST CNR summer school. This was a universal problem for all windows students as they were specifically instructed to install both versions.

krzywon:
Do we want to uninstall existing versions if we install in the active directory? That is possible using pyinstaller.

butlerpd:
I would argue that the answer would be yes but only after warning the user this will happen, but it should never install into a directory that is not empty so if the user responds no the install should abort .. or ask for a new directory?

PR #3042, opened 2024-08-16
3029: Do not use previous App directory and warn user when installing in existing directory

### LINKED ISSUE #3029: 6.0.0b2 installer default folder is not 6.0.0 but whatever the last folder that SasView was installed to
**Describe the bug**
6.0.0b2 installer default folder is not 6.0.0 but whatever the last folder that SasView was installed to.  If the user accepts the defaults, as most do unless they feel very comfortable they know what they are doing, the previously installed will be overwritten unwittingly and the new 6.0 version will be in a confusingly named folder (most likely SasView-5.0.6 for example). There is **no** warning given to the user that the folder already exists and that it will be overwritten!

The worst of course is that currently this conflates with #3030 so that once it finishes, neither version is now functional!

**To Reproduce**
1. on a clean system, or in my test case a system devoid of 5.0.6 and 6.0.0b2 install version 5.0.6
2. See it start and works
3. Close SasView
4. Now install 6.0.0b2
5. The installer will default to the c:\SasView-5.0.6 folder
6. accept and note that no warning is thrown
7. Once completed, there is a quick flash, then another flash of the new splashscreen and then nothing
8. The log file shows no evidence that it was even started.
9. navigate the 5.0.6 folder to the sasview.exe
10. Note that the icon is wrong. Click on the icon
11. Note the same behavior as when finishing the install

**Expected behavior**
The default folder should be 6.0.0b2, or at the least 6.0.0. In either case, the user should be warned if the folder already exists.

**Screenshots**

 - Version: 6.0.0b2

 - OS: Windows

**Additional context**
Discovered as part of the NIST CNR summer school. This was a universal problem for all windows students as they were specifically instructed to install both versions.

krzywon:
## Description

This PR changes the INNO setup in order to always select a default directory (C:\Sasview-<version>), not the previous one SasView was installed in and warns the user if the directory already exists.

Fixes #3029 
Refs #3030

## How Has This Been Tested?

I downloaded the Windows installer generated on push and tested both of the above items work.

## Review Checklist:

**Installers**


butlerpd (approved):
code changes look reasonable (not that I know much about INNO setup). This is unfortunately a place where the version number has to be updated separately from what is in the `init` file.

Importantly installing 5.0.6 then installing this PR the installer defaults to suggesting the sasview-6.0.0 folder not the previously installed folder. Install works.  Installing again without removing 6.0.0 still defaults to the same folder but brings up a warning that the folder exists and asks if we want to proceed. Doing so works though that is because we are installing exactly the same files.

Unfortunately this does not address #3030. Installing on top of 5.0.6 for example does produce the warning (which it did before) but if the answer is "yes" the installation is completely broken and will not start (as before).

Ideally if the answer the user gives is "yes overwrite" we should remove the folder and recreate it. However the minimum should be to remove the option and just say we cannot install as the folder already exists.

Perhaps that should be another PR? I am approving this as fixing #3029 only as I'm guessing that was the intent?

butlerpd:
Never mind .. I see where indeed #3030 is being addressed separately. So this should be able to be merged.

PR #3074, opened 2024-08-20
3057: Change default settings

### LINKED ISSUE #3057: Default settings should be changed/updated
The new settings menu moves some of the existing menus such as Choice of Optimizer and GPU options, while providing new options for new flexibility in choosing default behaviors. However, in this case the defaults have been set to the behavior in 5.0.6

Given that some of these were added due to the large number of requests/complaints and based on the near universal choice at the NIST CNR summer school (among those who were aware or made aware of the new setting options), I strongly suggest that version 6.0.0 start by setting polydispersity plots to OFF and the toolbar bar menu on plots to ON. This is the state most people seem to want if they are actually trying to analyze the data from their experiment.

NOTE: as implied above this was identified during the NIST CNR summer school

krzywon:
## Description

This changes the default settings to show the mpl toolbar, and to not show polydispersity plots.

Fixes #3057

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/system/config/config.py
+++ b/src/sas/system/config/config.py
@@ -170,7 +170,6 @@ class Config(ConfigBase, metaclass=ConfigMeta):
         self.DEFAULT_OPEN_FOLDER = ""
         self.TOOLBAR_SHOW = True
         self.DEFAULT_PERSPECTIVE = "Fitting"
-        # self.DEFAULT_PERSPECTIVE = "Corfunc"
 
         # Default threading model
         self.USING_TWISTED = False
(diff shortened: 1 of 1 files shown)


juliuskarliczek (approved):
Everything looks good, started sasview from the branch and the automatic show of polydispersity plots is disabled as well as the matplotlib toolbar is enabled. 
Looked in the settings and the checkboxes are set correctly.
I will check the installers and write another comment

edit: the windows installer works with the changes but i do not have an osx machine to test the installer there.

PR #3082, opened 2024-08-21
Documentation Window Identify Loaded File

### LINKED ISSUE #2832: Documentation Viewer: could it identify the file
Whilst testing a Github Actions build for a PR review I was presented with the new Documentation Viewer when I went to check a help doc. Whilst I wasn't specifically testing the Viewer, I did notice that there isn't actually anything to say which documentation file you are actually viewing, like there is in a browser.

Could this be added, perhaps in the window name?

tsole0:
## Description
Documentation window now displays the path of the file currently loaded in its title bar, which changes when the user navigates to a different file.

**NOTE:** If it is decided to display only the title of the documentation as opposed to the filepath, I have already implemented this in branch [2832-documentation-viewer-identify-file](https://github.com/SasView/sasview/tree/2832-documentation-viewer-identify-file). 

Fixes #2832 

## How Has This Been Tested?

Tested on Mac but not on Windows--I may have overlooked something to do with windows pathing.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/DocViewWidget.py
+++ b/src/sas/qtgui/Utilities/DocViewWidget.py
@@ -224,28 +224,14 @@ class DocViewWindow(QtWidgets.QDialog, Ui_DocViewerWindow):
         try:
             current_path = re.search(r'//(.*?)\'\)', str(self.webEngineViewer.url()))
             current_path = current_path.group(1)
-            current_path = Path(current_path)
         except:
             self.setWindowTitle("Documentation")
             return
         
-        # Try to add the title of the HTML page to the window title
+        # Try to add the filepath to the window title
         try:
-            with open(current_path, 'r', encoding='utf-8') as file:
-                html_content = file.read()
-
-                # Find the first <h1> tag and extract its content
-                match = re.search(r'<h1[^>]*>(.*?)</h1>', html_content, re.IGNORECASE)
-                if match:
-                    h1_content = match.group(1)
-
-                    # Split the content by any HTML tags and keep only the text before any tags
-                    no_tags_title = re.split(r'<[^>]+>', h1_content)[0]
-                    # Set the window title with the extracted text
-                    self.setWindowTitle(f"Documentation—{no_tags_title.strip()}")
-                    return
-                else:
-                    pass
+            self.setWindowTitle(f"Documentation—{current_path.strip()}")
+            return
         except:
             self.setWindowTitle("Documentation")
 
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Utilities/DocViewWidget.py
```diff
@@ -207,8 +209,31 @@ def loadHtml(self):
...
+        Set the title of the window to include the name of the document,
+        found in the first <h1> tags.
+        """
+        # Convert QUrl to pathlib path
+        try:
+            current_path = re.search(r'//(.*?)\'\)', str(self.webEngineViewer.url()))
+            current_path = current_path.group(1)
+        except:
+            self.setWindowTitle("Documentation")
+            return
```
krzywon, line 229: This can be simplified to a single try/except with no return statements.

I'm leery about your regex. What are you trying to match here? I would suggest passing the url to `pathlib.Path` class and getting either the `name` for the filename, of `absolute()` for the pull path.

You should not use bare exceptions. `except (<A>, <TupleOf>, <ExceptionsIWillLikelyThrow>):` is ideal, but `except Exception:` should be the bare minimum.

```suggestion
        try:
            current_path = re.search(r'//(.*?)\'\)', str(self.webEngineViewer.url()))
            current_path = current_path.group(1)
            self.setWindowTitle(f"Documentation—{current_path.strip()}")
        except:
            self.setWindowTitle("Documentation")
```

tsole0:
Updated based on feedback. Was using regex to convert from QUrl formatting to usable filepath, but there is a built-in method `toLocalFile()` that is more robust.

src/sas/qtgui/Utilities/DocViewWidget.py
```diff
@@ -2,6 +2,7 @@
 import os
 import logging
 import time
+import re
```
krzywon, line 5: nitpick: import no longer used

src/sas/qtgui/Utilities/DocViewWidget.py
```diff
@@ -207,8 +209,24 @@ def loadHtml(self):
         settings.setAttribute(QtWebEngineCore.QWebEngineSettings.LocalContentCanAccessFileUrls, True)
         self.webEngineViewer.load(url)
 
+        # Set widget title to include the name of the document loaded
+        self.updateTitle()
+
```
krzywon, line 214: This might cause a 2nd running of `updateTitle()` . Line 210 should trigger the call as well if the urlChanged connection is working properly.

krzywon (approved):
Tested and it works.

tsole0:
Fixed 👍

PR #3083, opened 2024-08-21
remove/disable edit button

### LINKED ISSUE #3072: Remove `Edit` Button from Documentation Editor
After discussion with Jeff and Paul we've decided to **temporarily** remove the edit button from the 6.0.0 release.

### Rationale
- High latency in the regeneration of files
  - Sphinx regenerates all documentation files instead of just one (a harder problem to fix than you'd think)
- Unexpected results caused by starting two documentation processes simultaneously
- Regeneration process still somewhat opaque; creates confusion for users

Our currently proposed solution will be to generate a "preview" html file that should take a (much) shorter time to generate, with an option in the GuiManager menu bar to regenerate the entire documentation. Since our priority is finalizing 6.0.0, we will be implementing this at a later date and disabling the editing button to avoid confusion in the meantime.

#### Note
The underlying code for regenerating documentation will **NOT** be removed--only the button in the GUI. Model documentation will still be generated locally when it is loaded but doesn't already exist.

tsole0:
## Description

Remove the edit button in the documentation viewer window to simplify user experience for the 6.0.0 release. **Note** this PR does not remove the code for regenerating/editing documentation from the codebase--it only makes in inaccessible to the user. This will allow easy re-incorporation in a later patch/release.

Fixes #3072 

## How Has This Been Tested?

Tested on Mac--edit button is gone, making it impossible to initiate the documentation regeneration process without creating/loading a new user plugin (or plugin without documentation).

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/UI/DocViewWidgetUI.ui
+++ b/src/sas/qtgui/Utilities/UI/DocViewWidgetUI.ui
@@ -45,19 +45,6 @@
       </spacer>
      </item>
      <item>
-      <widget class="QPushButton" name="editButton">
-       <property name="minimumSize">
-        <size>
-         <width>100</width>
-         <height>0</height>
-        </size>
-       </property>
-       <property name="text">
-        <string>Edit</string>
-       </property>
-      </widget>
-     </item>
-     <item>
       <widget class="QPushButton" name="closeButton">
        <property name="minimumSize">
         <size>
--- a/src/sas/qtgui/Utilities/UI/DocViewWidgetUI.ui
+++ b/src/sas/qtgui/Utilities/UI/DocViewWidgetUI.ui
@@ -45,6 +45,19 @@
       </spacer>
      </item>
      <item>
+      <widget class="QPushButton" name="editButton">
+       <property name="minimumSize">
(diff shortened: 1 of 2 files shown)


wpotrzebowski:
Just a thought but maybe instead of hiding the button we can add the setting to Preferences that will control it?
Something like "Enable documentation editing", which will be off by default

src/sas/qtgui/Utilities/UI/DocViewWidgetUI.ui
```diff
@@ -44,19 +44,6 @@
        </property>
       </spacer>
      </item>
```
rozyczko, line 46: Rather than deleting this, you could just add
```python
self.editButton.setVisible(False)
```
 in `DocViewWidget.py`
It will be easier than restoring the deleted code later.

rozyczko (changes requested):
Consider setting the `visible` flag on the widget rather than deleting it.

tsole0:
Thanks for feedback! Suggested changes implemented.

rozyczko (approved):
looks good!

PR #3084, opened 2024-08-21
3016: Fix to PD changes

### LINKED ISSUE #3016: Changing polydispersity function also forces the width of that dispersity to be very large.
**Describe the bug**
NOTE: This may be fixed by PR #2991 as it sounds very related? That however needs to be verified before closing this ticket.

When turning on polydispersity (by checking the box in the fit panel) and then changing the polydisperse function to use (e.g. log normal), but touching nothing else (including leaving PD = 0 and NOT fittable) the curve immediately recalculates anyway as if the PD were something very large. Assuming data had already been fit without polydispersity turned on, hitting the fit button again will cause the polydisperse parameter (the whose dispersity function was changed, e.g. radius) to get much smaller as if a HUGE PD is being applied!!  Changing back to Gaussian with a PD of 0 still acts as if it now has a huge PD but now on the Gaussian distribution. However, turning on Polydispersity but doing NOTHING to the default distribution will fit normally,

This was tested with both the SDS, and apoferritin data in the 1D data examples and both the Schulz and Log Normal distributions were tried (besides the Gaussian).

**To Reproduce**
1. Load some data (e.g. apoferritin)
2. choose a model and fit (core shell for apoferritin)
3. Fit the data
4. Turn on polydispersity
5. Got to the polydisperse tab and choose a new distribution from the dropdown (e.g. the radius for apoferritin) - Gaussian is the default distribution
6. See the curve change (note it should NOT change until PD is something >0)
7. Click fit again. The radius will now look MUCH smaller as expected if the radius PD is very large
8. Changing back to a Gaussian distribution and try fitting again
9. The fit still has a huge PD

**Expected behavior**
PD should remain as set independently of choosing a different distribution. This means that visually nothing at all should happen when choosing a new distribution and the fits should not change between no polydispersity and a PD of 0 on the polydispersity. 

 - Version: 6.0.0b2

 - OS: Windows

**Additional context**
Identified during the NIST CNR summer school

krzywon:
## Description

This takes the fix applied in #2991 a step further. The original fix did not account for the error column, and continued to modified the `.width` param regardless of the value changed.

Fixes #3016

## How Has This Been Tested?

Tested locally - different function produce different shape distributions.

## Review Checklist:

**Installers**
  - [x] **MacOSX** installer (GH artifact) has been tested (installed and worked) 

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1814,9 +1814,10 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             if model_column not in delegate.columnDict():
                 return
             # Map the column to the poly param that was changed
-            associations = {1: "width", 4: "npts", 5: "nsigmas"}
-            self.poly_params[f"{parameter_name}.{associations.get(model_column, 'width')}"] = value
-            self.kernel_module.setParam(parameter_name_w, value)
+            associations = {1: "width", delegate.poly_npts: "npts", delegate.poly_nsigs: "nsigmas"}
+            p_name = f"{parameter_name}.{associations.get(model_column, 'width')}"
+            self.poly_params[p_name] = value
+            self.kernel_module.setParam(p_name, value)
 
             # Update plot
             self.updateData()


butlerpd (approved):
Looks reasonable and behaves as expected now.

I don't think it is new to this PR? But I noticed that the first change of polydisperse function on a given parameter triggers a run of the calculation but not for any future changes of function. I would think it is either a calculation whenever the function changes or never?

Issue #2972, opened 2024-07-19
Documentation Editor Does not updated changes to built-in model docs


butlerpd:
**Describe the bug**
When editing built-in models, the documentation editor is replaced by the model editor which in principle would allow editing of the c file as well as the python math besides the documentation.  However, changes to the documentation do not trigger a document regeneration. I strongly suspect that the math changes do not get updated either.  Loading a regular document page and editing it will trigger a regeneration when hitting `save`, but that regeneration does not appear to affect the built in model documentation.

**Expected behavior**
There are several approaches that maybe require some discussion.
* First we could just remove the edit button on the builtin models docs
* Second, we could add a button to the model editor that generates the docs separate from saving. This would allow the user to check their documentation from the model editor (it would be truly annoying for the model docs to get generated every time  I change a line of code while debugging my new model).
* We make the builtin model docs `Edit` button link to the standard doc editor window. In this case we probably don't want to suck in the whole file into the editor but just the rst portion (i.e. the top of the file between the two sets of triple quotes)
* Something else?

I'm currently thinking the second option is probably the best both for the user and the code structure?

**SasView version (please complete the following information):**
 - Version: 6.0.0b1

**Operating system (please complete the following information):**
 - OS: Windows 10

PR #3088, opened 2024-08-22
Fix model documentation generation

### LINKED ISSUE #3077: Can't generate documentation for user models
At some point over the past few months it seems that the documentation editing window has lost the ability to successfully display (or maybe generate?) newly created user models.

To replicate:

1. Create a new model in the model editor
2. Save the model and select it in the fitting window
3. Press `Help`. Documentation generation will run
4. `Your file cannot be accessed` message

tsole0:
## Description
Fixes bugs in the generation of model documentation preventing it from being generated/displayed correctly. Additionally fixes bug preventing Sphinx logging information appearing in DocRegenInProgress window.

Fixes #3077, #2972 

## How Has This Been Tested?

Not yet implemented fixes to toctree generation.

## Review Checklist:

**Installers**


src/sas/qtgui/Utilities/DocViewWidget.py
```diff
@@ -43,19 +45,41 @@ def __init__(self,
...
-                return
+                logging.info('Documentation regeneration is already in process. '
+                            'Please wait for it to complete before attempting again.')
         except KeyboardInterrupt as msg:
+            self.close()
             logging.log(0, msg)
 
+    def close(self):
+        # Ensure the runner and locks are fully released when closing the main application
+        if self.runner:
```
pkienzle, line 79: I wonder what happens if the `closeSignal` from line 51 is triggered while in the polling loop of of line 63? It looks like  `self.documentation_lock.release(); self.runner.kill` will be called on line 67-68 will be called. Multiple `release` should be harmless. Not sure about multiple `kill`. Depending on the exact timing of `closeSignal` and another call to `compute`, the new runner might itself be killed.

Separate question: do you need some sort of `wait` on the killed thread?

Third question: is `closeSignal` triggered from the same thread as `compute` is running? Does `close` even have access to the runner process handles?

I'm not familiar enough with Qt signals and threads, so maybe the code is fine.

It may be easier to work with a (semi-)permanent documentation thread, and a queue of documentation build requests. You still may want a separate process within that thread for cancelling in-progress builds and capturing failed builds.  It'll be be easier to reason about with a clear separation from Qt signals.

Alternatively, use only Qt multiprocessing and signaling for the doc builds, assuming they have tools for managing  worker queues within their environment.

PR #3090, opened 2024-08-23
Make the what's new page appear in front of the main window on Linux

jamescrake-merani:
## Description

Previously, on my Ubuntu machine, the what's new page would appear behind the main window. This is problematic because the user will not immediately see the what's new page, and also the main window will not be responsive until the what's new window has been closed. This caused me a lot of confusion initially.

The reason for this is because the what's new parent is being set to the GuiManager not the MainWindow, and also this parent does not get passed to the QDialog instance thus QT doesn't internally know about its parent.

One thing that perhaps should be discussed is whether this fix should be implemented for any other windows/widgets in the application. I noticed other widgets are using GuiManager as their parent but wasn't sure if this was intentional, and I only wanted to apply the fix specifically for the what's new window in case something else broke.

## How Has This Been Tested?

Executing the script in the `run.py` file. After the fix, the what's new page appears on top of the main window as expected.

## Review Checklist:

**Installers**

**Licencing** 

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -209,7 +209,7 @@ class GuiManager:
         self.GENSASCalculator = None
         self.DataOperation = DataOperationUtilityPanel(self)
         self.FileConverter = FileConverterWidget(self)
-        self.WhatsNew = WhatsNew(self)
+        self.WhatsNew = WhatsNew(self._parent)
         self.regenProgress = DocRegenProgress(self)
 
     def loadAllPerspectives(self):
@@ -661,7 +661,7 @@ class GuiManager:
         self.welcomePanel.show()
 
     def actionWhatsNew(self):
-        self.WhatsNew = WhatsNew(strictly_newer=False)
+        self.WhatsNew = WhatsNew(self._parent, strictly_newer=False)
         self.WhatsNew.show()
 
     def showWelcomeMessage(self):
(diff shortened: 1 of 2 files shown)


rozyczko (commented):
I don't really see how these changes would fix the issue. Maybe they do, but it seems parentage should not influence the visibility.

How about calling `raise` on the WhatsNew widget instead?

```python
        self.show()
        self.raise_()
        self.focusWidget = QtWidgets.QApplication.focusWidget()
        self.setFocus()
```

or maybe even make it modal, so users have to click on OK to make it go away?

rozyczko (commented):
Please have a look at `def plotSelectedPlot` in GuiManager.py (line 299).

This method is bringing up front a selected plot. Comments in there might be useful if the proposed `.rase_` call doesn't work.

jamescrake-merani:
> How about calling raise on the WhatsNew widget instead?

I've tried doing this instead of setting the parent but it does not appear to work on my machine. Setting the parent is the only thing that seems to work. Interestingly, the `show` method doesn't block input on the main window unlike before.

> but it seems parentage should not influence the visibility

My guess is that Qt is using the parent internally to decide where the windows should show up. This seems to be a Linux specific issue though as I'm not aware of it happening anywhere else.

> or maybe even make it modal, so users have to click on OK to make it go away?

I've tried to explicitly set it as a modal box with `self.setModal(True)`. But If I'm not mistaken, I believe the box already was modal as it blocks input on the main window until it has been dismissed.

> Please have a look at `def plotSelectedPlot` in GuiManager.py (line 299).

I tried the `showNormal` method instead of `show` but the outcome appears to be the same.

rozyczko (approved):
Platform-specific solutions are always sub-optimal but in this case, reparenting the widget doesn't sound too bad, especially if it doesn't change the functionality on other platforms.

Really annoying that `raise_` and `showNormal` are not giving us what they are supposed to do...

PR #3099, opened 2024-08-28
Add Reparameterization Editor

tsole0:
## Description

*Based off 6.0.0 release but not intended to be included in 6.0.0 release*
Create a user interface for creating custom reparameterized models. The new window lives in Fitting > Reparameterize Model, and provides a more graphical experience for reparameterizing a model. For more about reparameterization, refer to the [SasView documentation on reparameterization](https://www.sasview.org/docs/user/qtgui/Perspectives/Fitting/plugin.html#reparameterized-models). Feature list:

- Fields for specifying model name and redefinition (= mapping the old parameters to the new parameters)
- An intuitive user interface for selecting a model to be reparameterized
- Dropdown menus for easily navigating parameters and parameter attributes
- A specialized popup allowing the user to customize all aspects of a new parameter (could even be incorporated into the model editor in the future)
- Detailed parameter checking/warning system that autodetects parameter-based errors *before* model compilation
- Model error checking on 'Apply'--syntax errors appear in log explorer

## Known limitations
Cannot use arbitrary numpy function imports--imports to the reparameterization file are not seen by sasmodels, and sasmodels will not accept anything in the "translation" string besides `var = expr` type syntax, precluding import statements.

## How Has This Been Tested?
Community feedback and testing encouraged!

## Review Checklist:

- [x] There is **nothing** that needs documenting (although the reparameterization docs are a bit old)

**Installers**


tsole0:
This PR is based on the 6.0.0 branch but is not intended for the 6.0.0 release.

pkienzle:
The execution context is C code in the polydispersity loop. You have access to any functions defined for the base kernel.

I thought I had provisions for putting in arbitrary C functions using `reparameterize(..., source=source)`, but that was only for adding additional libraries from `sasmodels/models/lib`. I don't think you can add C files relative to the reparameterization definition file since it is not on the source search path but I haven't tested.

You can put in arbitrary C code, but you have to do so by hacking the returned `model_info` object [untested]:

```python
parameters = [
    # name, units, default, [min, max], type, description
    ["volume", "Ang^3", 1e5, [0, inf], "volume", "ellipsoid volume"],
    ["eccentricity", "", 1, [0, inf], "volume", "polar:equatorial radius"],
]

source = [] # list of additional libraries needed for the code

translation = """
radius_polar = eccentricity*Re(volume, eccentricity)
radius_equatorial = Re(volume, eccentricity)
"""

c_code = f"""
#line 1 "{__file__}"
static double Re(double volume, double eccentricity) {
    return cbrt(volume/eccentricity/M_4PI_3);
}
"""

model_info = reparameterize("ellipsoid", parameters, translations, __file__, source=source)
model_info.c_code = f"{model_info.c_code}\n{c_code}" if model_info.c_code else c_code
```

Meanwhile, add a ticket to sasmodels asking for `reparameterize(..., c_code="...")`, and another to check if helper code can be saved as `model_reparam.c` in the same directory as the `model_reparam.py` file.

pkienzle:
Rather than reparameterization we could build the variants into the model.

For example, for ellipsoid provide all four parameters (volume, eccentricity, Re, Rp) but require that the user choose two.

This could be done with an enumeration (6 variants: V-ε, V-Re, V-Rp, ε-Re, ε-Rp, Re-Rp), or by setting the unused parameters to NaN.

If we are likely to go this route, then it would be better to skip the reparameterization infrastructure that we will need to support indefinitely for backward compatibility.

pkienzle:
Regardless of whether we use `reparameterize()` or model variants, it would be nice to show derived internal parameter values to the user.

This is challenging because the derived parameter will have a number density distribution arising from the polydispersity.

Even more challenging, there will be uncertainty in that distribution through the interaction with uncertainty on the fitted parameter values and polydispersities.

Maybe put this as a ticket on the sasmodels wish list.

krzywon:
This should be superseded by https://github.com/SasView/sasview/pull/3136. Keeping it open for now, but likely will be closed next week.

lucas-wilkins:
Close as rebase PR exists (#3136)

PR #3100, opened 2024-08-30
3014: Beta approximation does not run

### LINKED ISSUE #3014: Beta approximation not working correctly
**Describe the bug**
The beta approximation no longer produces the beta or S_eff subplots. This suggests beta may not in fact be being applied. A few quick tests support that hypothesis, though better test data would be required to prove it. The priority depends on which is the case. I will set the priority as critical under the assumption that it is in fact not actually calculating beta even though the GUI suggests to the user it is doing so. If it is just the plots not being added to the explorer I would rate the severity as moderate.

**To Reproduce**
1. Load some data (2% hSDS for example) and send to fitting
2. choose sphere and HMSA structure factor and set parameters as given in tutorial (https://www.sasview.org/downloads/basic_1d_fitting_in_sasview_v5.pdf)
3. hit fit then check the subplots under the data set in the explorer
4. There will only be S, P and residuals. S_eff and beta will be absent

**Expected behavior**
In 5.0.6 once fit, the data file will have 5 subplots rather than just two. The two missing from 6.0.0 are the beta and S_eff plots.

 - Version: 6.0.0b2

 - OS: both windows and mac

**Additional context**
Reported as part of the NIST CNR summer school bugs

### LINKED ISSUE #3053: Phi and theta stuck at zero in 2D theory calculations
**Describe the bug**
When plotting oriented 2D scattering patterns from theory models, setting either theta or phi to 0 at any point, disables that parameter. Any additional changes to those angles are not updated in the simulated data images. This makes it difficult to manually fit. It is not clear (and hard to test) if the same would happen if the fitting algorithm sets one of those values to zero which would be extremely problematic.

**To Reproduce**
1. create a theory model for any anisotropic shape (e.g. cylinder) and click the 2D box before plotting
2. change angles theta and phi (and psi for prism models)
3. Now set one to zero
4. Try to move off of zero
5. Pattern does not change from that at 0

**Expected behavior**
the plot (and underlying theory data) should recalculate with the new angel and not use zero as being unchangeble.

 - Version: 6.0.0b2

 - OS: Windows 10

**Additional context**
Identified during the NIST CNR summer school and thus labelled 6.0.0 release as agreed - However, whether this is a longer standing issue has yet to checked

### LINKED ISSUE #3108: Bug with radius_effective_mode
**Describe the bug**
Looking into https://github.com/SasView/sasview/issues/3101 in more detail there appears to be a bug with radius_effective_mode **which has been introduced since release 5.0.6**.

Whilst the steps to reproduce might seem very niche, they are a set of steps that reliably reproduce the bug. There may be other sequences of steps that cause the same thing.

**To Reproduce**
1. Run SasView
2. Load some data (eg, 2% SDS); send to fitting
3. Create an ellipsoid @ hayter_msa model
4. Set radius_equitorial = 40
5. Tick background, radius_equitorial, volume_fraction & charge
6. Fit
7. Set radius_effective_mode = unconstrained
8. Fit
9. Set radius_effective_mode back to something other than unconstrained
10. Fit.

The radius_effective drops to zero!

And nothing you can do will recover it to a computed value based on the setting on radius_effective_mode, and the S(Q) remains frozen in subsequent fits.

**Expected behavior**
radius_effective should always take the value determined by the setting of radius_effective_mode. 

 - 6.0.0b2

 - W10/x64

krzywon:
## Description

The changes related to multiplicity models (#2647) took a step too far and broke parameter updates for `structure_factor_mode` and other non-layer-parameter combo boxes. This removes an extra if statement that should not be reachable for the layer parameter. This might have a collateral effect on multiplicity models, though I couldn't find any.

Fixes #3014
Fixes #3053 (possibly)
Fixes #3108 (possibly)

## How Has This Been Tested?

Selected a P*S model and selected the beta approximation. The calculation generated the expected plots. Tested various features of multiplicity models to see if they still worked and did not notice anything off. Also tested the pinning of param values to 0.0 and that is, tangentially fixed.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -2940,8 +2940,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
             # don't try to update multiplicity counters if they aren't there.
             # Note that this will fail for proper bad update where the model
             # doesn't contain multiplicity parameter
-            if self.kernel_module.params.get(parameter_name, None):
-                self.kernel_module.setParam(parameter_name, value)
+            self.kernel_module.setParam(parameter_name, value)
         elif model_column == min_column:
             # min/max to be changed in self.kernel_module.details[parameter_name] = ['Ang', 0.0, inf]
             self.kernel_module.details[parameter_name][1] = value


smk78:
Tried having a look at this but experienced something which I think is a wider problem (see https://github.com/SasView/sasview/issues/3101). So bit difficult to pass judgement on this PR at present.

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -2940,8 +2940,7 @@ def onMainParamsChange(self, top, bottom):
             # don't try to update multiplicity counters if they aren't there.
             # Note that this will fail for proper bad update where the model
             # doesn't contain multiplicity parameter
-            if self.kernel_module.params.get(parameter_name, None):
```
pkienzle, line 2943: Suspicious: `bool(0.0)` is false, so if the current value of the parameter is 0 then this code wasn't updating it.

I didn't look at the context, but this change may fix other bugs.

krzywon: There are a number of issues when params become 0. We should also test to see if this solves https://github.com/SasView/sasview/issues/3053

krzywon: Out of curiosity, I decided to test and, unless I am mistaking the issue, this should now fix #3053.

krzywon:
@smk78 - do you have time to test this? Getting this into 6.0 is a priority.

smk78 (approved):
I have re-reviewed using https://github.com/SasView/sasview/actions/runs/10638002500 on W10/x64, testing for each of the issues reported in #3014 , #3053 , #3101 and #3108 . As far as my testing could discern, this PR fixes the anomoulous behaviour in each of those. On that basis I think it is good to go.
I have not tested for #3109 , mentioned in this PR, because I could not tell if the commits in that had been merged with this branch.

krzywon:
Thanks for the review, @smk78. This does not fix #3109, so there's no need to review that.

PR #3104, opened 2024-09-04
3086 weird plot auto scaling

### LINKED ISSUE #3086: Weird plot auto scaling
**Describe the bug**
Not sure when this happened but during PR testing today (always from installers) I discovered that the auto scaling of plots behavior has changed in very strange ways.

Using my favorite apoferritin data set:
* Sending to fitting and plotting the window scales to have the theory curve fit from the top left to bottom right... exactly. This becomes obvious if you make scale tiny so that the theory curve is entirely well below the data (e.g a scale of 1e-6). Hitting home on the graph will however reset so that both data and theory are all contained in the window
* Sending the data to a plot and scaling to Kratky then bring up a linear fit and shrink the range for fitting then fit. The plot now blows up to again show the fit line not the data (in my case the bottom left to the top right of the liner fit line. So the data out of the fit range is not visible. Exiting the modal fit panel and clicking on the graph home in this case does nothing to the graph. The only way to get the whole contents of the window back into the viewport is to manually change the x and y axis ranges.
* Sending the data to a plot and scaling to Porod then bring up a linear fit shrink the range for fitting then fit. The plot now blows up to again show the fit line not the data. So the data out of the fit range is not visible. Exiting the modal fit panel and clicking on the graph home in this case does nothing to the graph itself (as before) but makes the window get smaller and smaller every time you hit home/reset
* Sending the data to a plot and scaling to Guinier then bring up a linear fit. Immediately the data is shoved into the left hand side as the default x axis range is for the values in Q though the data are now in Q^2. Shrink the range for fitting then fit. No zoom takes place and other than the fact that the default zoom was wrong it behaves normally. If you zoom in prior to opening the linear fit panel, the data will stay zoomed till actually doing the fit then the window will revert to as above (with all the data on the left hand side of the plot) and continue "normally." Exit the modal window and click the home/reset button. Now the graph zooms to have the fit line go from the top left corner to the bottom right corner (so "weird" but in the opposite order from the fitpage "weird"). And why the different linear fits all behave very differently in their auto-zooming is another mystery.

**To Reproduce**
See above

**Expected behavior**
clicking fit in any case should never also perform a zoom. And to the extent a zoom has happened hitting home/reset should recover the window

 - Version: 6.0.0b2 Several PR installers from 8/21/2024

 - OS: Windows 10

lucas-wilkins:
## Description

Plot now picks range based on all plots, not just the last one

Fixes #3086

## How Has This Been Tested?

Fit some data, look at the plots

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -331,15 +331,13 @@ class PlotterWidget(PlotterBase):
 
         return False
 
-    def _plot_bounds(self, offset=0.01) -> tuple[tuple[float, float], tuple[float, float]]:
+    def _plot_bounds(self, offset=0.05) -> tuple[tuple[float, float], tuple[float, float]]:
         """ Get the appropriate bounds for the plots
 
         :param offset: add a small fraction of the absolute value of each end to each end
         :returns:
         """
 
-        min_scale = 1-offset
-        max_scale = 1+offset
 
         x_min, x_max = np.inf, -np.inf
         y_min, y_max = np.inf, -np.inf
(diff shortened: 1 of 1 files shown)


krzywon (approved):
Looks fine and fixes the issue

Issue #2975, opened 2024-07-19
Copy and pasting parameters lead to results tuple being empty


XaelShan:
**Describe the bug**
After copy and pasting parameters (AND loading project) from one fit page to another, returns the error that 

> ERROR: Traceback (most recent call last): File "C:\Users\tns14\sasview\src\sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 2109, in fitComplete res_list = result[0][0] ~~~~~~~~~^^^ IndexError: string index out of range

Specifically only with GPU option checked (even after getting all tests passed and pip install pyopencl)
*future edit note: after copy and pasting parameters, opencl option no longer passes all tests. To clarify, before copy/paste, all 77 opencl tests passed. After copy/paste, 60 tests _failed_*

**To Reproduce**
1. Load datasets
2. Load into separate fitting pages
3. Choose the same model for both pages
4. Click on copy on one page, paste in the other page
5. Click on compute for both, one will not form results
6. See error

**Expected behavior**
Should return a fit with copied parameters. Instead, returns error and fit does not complete; compute/plot also returns a graph with no fit/line (only data).

**Screenshots**
Before copy/paste parameters:
![image](https://github.com/user-attachments/assets/77fa5381-594b-4599-90c5-a706bcc6753a)

after copy/paste parameters (same effect as opening a .json saved project)
![image](https://github.com/user-attachments/assets/3c94754a-0256-44e4-81d4-437d76904088)

compute/plot results in this:
![image](https://github.com/user-attachments/assets/ac27a58b-a071-4078-9965-0fe131afea8f)

and fit itself does not return anything, instead just returning an ERROR

**SasView version (please complete the following information):**
 - Version: 6.0.0b1

**Operating system (please complete the following information):**
 - OS: Windows 10

**Additional context**
models that I know have this error are: core-shell-bicelle, core-shell-sphere.
Also note that any fits AFTER switching to gpu/opencl _but_ BEFORE copy/paste work. It is only AFTER copy/paste that the opencl tests no longer pass and fits no longer work.

butlerpd:
Also identified in NIST CNR summer school

rozyczko:
I am trying to reproduce this issue. 
However, I don't understand the connection of GPU tests failing and compute action failing as related.

I did:

1. Loaded two datasets (cyl_400_20 and cyl_400_40)
2. Selected `core_shell_bicelle` for both of them
3. Ran `Compute/Plot` for the first dataset
[...]

PR #3112, opened 2024-09-25
Various 6.0.0 fixes

### LINKED ISSUE #3055: Full Debye calculator hangs
**Describe the bug**
Probably related to #3054, depending on the PDB file loaded into the GSC, the full Debye computation can either be very fast, or hang (no errors thrown).  Two levels of hanging have been observed: In one case it hangs after it seems to have started in the sense that the compute button changes to an apparently active cancel button, but pressing cancel has no effect. In the other case, the failure happens before the cancel button appears and the window is locked with the compute button grayed out. Note that the hanging has only been seen in files that generated warnings PLUS a traceback error.

**To Reproduce**
Using the data provided in #3054, use the data that throws a single traceback to observe the first behavior. The file that produces two traceback errors will lead to the second behavior. 

**Expected behavior**
If the calculator is fed a malformed file an exception should be raised and clean exit from the code executed. The user should then be provided with the exception error.

 - Version: 6.0.0b2

 - OS: Windows 10

**Additional context**
Identified during the NIST CNR summer school and thus tagged for 6.0.0 release as agreed. However whether this is a pre-existing problem has not yet been verified.

krzywon:
## Description
This fixes two wrinkles found in a couple of issues. The GSC could load invalid files, which then failed to run the computation on those files. The GPU Options window would not change the selected option when the tests were run. Both of these are fixed here.

~~NOTE: If I find a solution to https://github.com/SasView/sasview/issues/2975#issuecomment-2375220706, I will try to get it in here~~ (long-standing issue that won't be fixed in 6.0.0)

(maybe) Fixes #3055 
refs #2975

## Review Checklist:

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/GPUOptions.py
+++ b/src/sas/qtgui/Perspectives/Fitting/GPUOptions.py
@@ -140,13 +140,15 @@ class GPUOptions(PreferencesWidget, Ui_GPUOptions):
         """
         Run sasmodels check from here and report results from
         """
+        # Set the SAS_OPENCL value prior to running to ensure proper value is used
+        no_opencl_msg = self.set_sas_open_cl()
+        self._unStageChange('SAS_OPENCL')
         self.model_tests = sasmodels.model_test.make_suite('opencl', ['all'])
         number_of_tests = len(self.model_tests._tests)
         self.progressBar.setMinimum(0)
         self.progressBar.setMaximum(number_of_tests)
         self.progressBar.setVisible(True)
         self.testButton.setEnabled(False)
-        no_opencl_msg = self.set_sas_open_cl()
 
         test_thread = threads.deferToThread(self.testThread, no_opencl_msg)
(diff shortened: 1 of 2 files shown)


butlerpd (approved):
Code for the second fix looks good.  Also the second seems reasonable but I'm don't know the full flow of that part of the code.

Tested the windows installer and both fixes behave as advertized. 

The second problem is a solid fix. It turns out that running the Debye calculator on the file read with errors not only hung, but even if reset, and a proper PDB file was loaded, the calculation would still hang, meaning once you tried to calculate a poorly formed PDB one had to exit SasView and restart. That is now fixed!

The first problem now fixes the first of the two reported errors, namely that if the OpenCL radio button is checked and the test then run, the gpu tests are run. However, the fix raises the question of whether the `Apply` button is still relevant? I don't see that it is.

Also the second problem reported that the OpenCl tests never complete and cause the application to become unstable if canceled persists

PR #3113, opened 2024-09-27
Update Version for 6.0.0RC1

krzywon:
## Description

This updates the version number for the release candidate to 6.0.0rc1.


build_tools/release_notes/6.0.0_notes.txt
wpotrzebowski, line 1: I believe we don't need this file anymore. It may be good opportunity to delete it?

krzywon: Good catch. Removed.

pkienzle (approved):

PR #3128, opened 2024-10-17
3127 adding custom text to a graph causes it to go blank

### LINKED ISSUE #3127: Adding custom text to a graph causes it to go blank
**Describe the bug**
I've tested this both on Mac, and Ubuntu though I don't have a Windows machine on hand to test with. The error seems more reproducible on Mac as it doesn't seem to always occur on Ubuntu.

**To Reproduce**
1. Load in some data, and do a fit
2. Right click on one of the graphs, and add custom text
3. On Mac this is enough to reproduce the error in my testing but on Ubuntu you might need to resize the graph before it goes blank.

**Expected behavior**
Custom text should get added onto the graph.

**Screenshots**
image

**SasView version:**
 - Version: SasView RC2

**Operating system:**
 - OS: Ubuntu 24.04, and Mac OS Sonoma 14.6.1 

**Additional context**
This is the error that gets printed to the console:
```Traceback (most recent call last):
  File "matplotlib/backends/backend_qt.py", line 498, in _draw_idle
  File "matplotlib/backends/backend_agg.py", line 387, in draw
  File "matplotlib/artist.py", line 95, in draw_wrapper
  File "matplotlib/artist.py", line 72, in draw_wrapper
  File "matplotlib/figure.py", line 3156, in draw
  File "matplotlib/layout_engine.py", line 183, in execute
  File "matplotlib/_tight_layout.py", line 266, in get_tight_layout_figure
  File "matplotlib/_tight_layout.py", line 82, in _auto_adjust_subplotpars
  File "matplotlib/artist.py", line 1412, in _get_tightbbox_for_layout_only
  File "matplotlib/_api/deprecation.py", line 457, in wrapper
  File "matplotlib/axes/_base.py", line 4499, in get_tightbbox
  File "matplotlib/artist.py", line 365, in get_tightbbox
  File "matplotlib/text.py", line 960, in get_window_extent
  File "matplotlib/text.py", line 897, in get_unitless_position
TypeError: float() argument must be a string or a real number, not 'NoneType'
```

I'll take a look at the error, and see if I can fix it.

jamescrake-merani:
## Description

Fixes #3127 

## How Has This Been Tested?

Tested on Ubuntu. Haven't had a chance to test it on Mac but will do later.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -531,10 +531,7 @@ class PlotterWidget(PlotterBase):
         # MPL style names
         styles = ['normal', 'italic', 'oblique']
         # QFont::Style maps directly onto the above
-        try:
-            mpl_font.set_style(styles[extra_font.style()])
-        except:
-            pass
+        mpl_font.set_style(styles[extra_font.style().value])
 
         if len(extra_text) > 0:
             new_text = self.ax.text(pos_x,
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -531,7 +531,9 @@ class PlotterWidget(PlotterBase):
         # MPL style names
         styles = ['normal', 'italic', 'oblique']
         # QFont::Style maps directly onto the above
-        mpl_font.set_style(styles[extra_font.style().value])
+        # Get an index of the font style which we can then lookup in the styles list.
+        font_index = extra_font.style().value
+        mpl_font.set_style(styles[font_index])
 
         if len(extra_text) > 0:
             new_text = self.ax.text(pos_x,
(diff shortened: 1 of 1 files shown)


krzywon (approved):
It's not clear to me why this fixes the issue based on the changes, but Mac plots were blanking when adding text in RC2, and are displaying properly here.

Would it be possible to add some in-line comments? That might give others a better understanding of why these changes were made.

jamescrake-merani:
Good point! I will add comments. After which, if they look good to you then I think this PR is ready to merge.

The bug seems to have been the culprit of two issues:

- The style of the font was being set but the wrong value was being passed into the `set_style` method. This should have caused an exception but it looks like there was some exception handling that just ignored the exception, and continued. This left the plot in a broken state, and I think whenever Matplotlib tried to draw the graph again, it wouldn't be able to properly. This might also explain why the behaviour differs between operating system; it might have something to do with how the graph is being drawn.
- There was also a problem with how the position of the text was being retrieved. Normally, when someone left clicks on the graph, it will update the `xclick`, and `yclick` positions. However, this *only* works on a left click; not a right click. You need to right click to access the context menu in order to add text but if you never left clicked, there would be no position to add the text to. My fix simply updates these coordinates on *every click* so that, after accessing the context menu, there should be a position to add the text to now.

PR #2918, opened 2024-06-28
Update model editor

tsole0:
## Description

- Added GUI box for direct form volume function editing to `PluginEditor.py` (dd7819a, 98c0a31, 6dc2f84)
  - Auto-insert first polydisperse parameter into GUI box (b35ee1c)
  - Add tooltip and sample function to GUI box (3ea3929)
- Changed text of GUI for clarity (3ea3929)
- Updated Python model template
  - Add flags and explanatory text (b3bc16f, 5ccb501)
  - Adds Effective Radius and Volume Ratio functions to non-polydisperse model template (d28cba7)
- Added capability to generate and save C models (bb37054, 84c7bdb, 2cc5296, 8bf8960, 3c26555, 8f7f7f2)
  - Added checkboxes to plugin editor giving user option to generate Python and/or C model templates (1251c38)
  - Created template for C models containing all kernel-defined functions and instructions for use (84c7bdb, f16633f, 5ccb501)
- Linked Python and C files together by inserting `source = ['filename.c']` into Python model template if both Python and C are generated (27ac6d6)
- Run model checks on C code and report errors (NOT a C syntax check, but a MODEL check via `sasmodels`) (0c61d51, a1d04b3)
- Added error popups for intuitive use:
  - Error if user does not specify model type (54b1159, bb61fb7)
  - Override dialog if user tries to save a model that contains errors asking if they would like to save or continue editing (20212f6, 7790233)
  - Info popup if user generates C model with no python model informing them that model checks will not be run (faefb31)
  - Warning popup if user loads a model with errors into model editor (a1d04b3)
- Parameter tables in plugin widget GUI expand to fit available space (068cdb1)

**Note: Merged #2917 into branch to fix tests**
**Note: Shares two commits with [model-editor-bug-fixes](https://github.com/SasView/sasview/tree/model-editor-bug-fixes) branch (ba4cbc7, 26e6b59) because ba4cbc7 originated on this branch and 26e6b59 was necessary for error-checking C models**

Linting:
de25165, 3ea3929

## How Has This Been Tested?

**TODO: windows installer, mac installer**

## Review Checklist:

[if using the editor, use `[x]` in place of `[ ]` to check a box]

**Documentation** (check at least one)
- [ ] There is **nothing** that needs documenting
- [ ] Documentation changes are **in this PR**
- [ ] There is an **issue** open for the documentation (link?)

**TODO: Need minor documentation changes!!**

**Installers**
- [ ] There is a chance this will affect the **installers**, if so
[...]

PR #3135, opened 2024-10-25
Update model editor (rebase)

krzywon:
## Description

This is a rebase of the branch in #2918, from `release_6.0.0` to `main`, using interactive mode to eliminate the extra merge commit from a separate branch. This should be a parallel PR at this point, but will require some checking. The changes and checklists can all be seen in the original PR.


wpotrzebowski:
Started testing functionality on Mac and getting error when clicking on Apply 

`21:33:47 - ERROR: Traceback (most recent call last): File "sas/qtgui/Utilities/TabbedModelEditor.py", line 269, in onApply File "sas/qtgui/Utilities/TabbedModelEditor.py", line 366, in updateFromPlugin File "sas/qtgui/Utilities/TabbedModelEditor.py", line 668, in canWriteModel AttributeError: 'TabbedModelEditor' object has no attribute 'builtinmodels'`

src/sas/qtgui/Utilities/PluginDefinition.py
```diff
@@ -178,8 +259,20 @@ def onFunctionChanged(self):
...
+            self.model['func_text'] = new_text
+            self.modelModified.emit()
+
+    def onFormVolumeFunctionChanged(self):
+        """
+        Respond to changes in form volume function body
+        """
+        # keep in mind that this is called every time the text changes.
+        # mind the performance!
+        #self.addTooltip()
```
jamescrake-merani, line 272: If this line has been commented out, can it just be removed?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -149,10 +160,8 @@ def onLoad(self, at_launch=False):
 
             if self.model is True:
                 # Find location of model .py files and load from that location
-                if os.path.isfile(user_model_name):
-                    filename = user_model_name
-                else:
-                    filename = MAIN_DOC_SRC / "user" / "models" / "src" / (self.load_file + ".py")
+                filename = user_model_name if os.path.isfile(user_model_name) \
+                    else MAIN_DOC_SRC / "user" / "models" / "src" / (self.load_file + ".py")
```
jamescrake-merani, line 164: Is this compatible with the changes made in the platformdirs branch? I can't remember if we're still using the `MAIN_DOC_SRC` constant.

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -246,10 +255,8 @@ def onCancel(self):
         """
         Accept if document not modified, confirm intent otherwise.
         """
-        if self.is_modified:
-            saveCancelled = self.saveClose()
-            if saveCancelled:
-                return
+        if self.is_modified and self.saveClose():
+            return
```
jamescrake-merani, line 259: Could this be simplified by being:
```python
if not (self.is_modified and self.saveClose():
    self.reject()
```

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -389,60 +456,55 @@ def findFirstError(self, full_path):
...
-            if re.match(r"[Ff]ile.*[Ll]ine ([0-9]+)", line):
-                # If model check fails (not syntax) then 'line' and 'File' will be in adjacent lines
-                error_line = re.split('line ', line)[1]
-                try:
-                    error_line = int(error_line)
-                    break
-                except ValueError:
-                    # Sometimes the line number is followed by more text
+            # Put a thick, red border around the editor.
+            from sas.qtgui.Utilities.CodeEditor import QCodeEditor
```
jamescrake-merani, line 473: Is there a reason this import is deferred? I know we sometimes do it when an import can take a while but is that true in this situation?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -389,60 +456,55 @@ def findFirstError(self, full_path):
...
+
+            # Find all QTextBrowser and QCodeEditor children
+            text_browsers = self.tabWidget.currentWidget().findChildren(QtWidgets.QTextBrowser)
+            code_editors = self.tabWidget.currentWidget().findChildren(QCodeEditor)
+
+            # Combine the lists and apply the stylesheet
+            for child in text_browsers + code_editors:
+                child.setStyleSheet("border: 5px solid red")
+                # last_lines = traceback.format_exc().split('\n')[-4:]
```
jamescrake-merani, line 482: Again, this line appears to be commented out. Can it be removed?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -507,47 +566,114 @@ def regenerateDocumentation(self):
...
+            self.parent.helpWindow.regenerateHtml(self.filename_py)
+
+    def noModelCheckWarning(self):
+        """
+        Throw popup informing the user that no model checks will be run on a pure C model.
+        Ask user to acknowledge and give option to not display again.
+        """
+        msgBox = QtWidgets.QMessageBox(self)
+        msgBox.setIcon(QtWidgets.QMessageBox.Information)
+        msgBox.setText("No model checks will be run on your C file until a python file of the same name is created in your plugin directory.")
```
jamescrake-merani, line 578: I know I'm being *very* pedantic here so feel free to ignore but Python should have a capital P.

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -507,47 +566,114 @@ def regenerateDocumentation(self):
...
+        msgBox.setWindowTitle("No Python File Detected")
+        buttonContinue = msgBox.addButton("OK", QtWidgets.QMessageBox.AcceptRole)
+        doNotShowAgainCheckbox = QtWidgets.QCheckBox("Do not show again")
+        msgBox.setCheckBox(doNotShowAgainCheckbox)
+
+        msgBox.exec_()
```
jamescrake-merani, line 584: Why is the `exec` method being called with an underscore?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -564,51 +690,170 @@ def getModel(self):
...
         return self.tabWidget.currentWidget().getModel()
 
+    def addTab(self, filetype: str, name: str):
+        """
+        Add a tab to the tab widget
+        :param filetype: filetype of tab to add: "python" or "c"
+        :param name: name to display on tab
+        """
+        if filetype in ["python", "py", ".py"]:
+            #display_name = Path(self.filename_py).name
```
jamescrake-merani, line 700: Lines commented out here.

jamescrake-merani (changes requested):
The code is very clean. Just a few minor comments.

src/sas/qtgui/Utilities/TabbedModelEditor.py
krzywon: Both are valid in PySide6 and act the same, but `exec_()` is an older syntax that may go away. [Relevant Stackoverflow discussion](https://stackoverflow.com/questions/22610720/should-i-use-app-exec-or-app-exec-in-my-pyqt-application). I will convert to `exec()` to future-proof this.

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -455,45 +512,42 @@ def updateFromEditor(self):
...
-        try:
-            self.c_editor_widget.txtEditor.setStyleSheet("")
-            self.c_editor_widget.txtEditor.setToolTip("")
-        except AttributeError:
-            pass
-        self.editor_widget.txtEditor.setStyleSheet("")
-        self.editor_widget.txtEditor.setToolTip("")
+        if self.is_python and self.findFirstError(full_path.with_suffix(".py")) >= 0:
+            # select bad line
+            cursor = QtGui.QTextCursor(w.txtEditor.document().findBlockByLineNumber(error_line-1))
```
rozyczko, line 519: Here, `error_line` is not defined. I'm getting errors related to this when testing.

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -507,47 +561,114 @@ def regenerateDocumentation(self):
...
+        msgBox.setDefaultButton(buttonContinue)
+
+        # Execute the message box and wait for the user's response
+        userChoice = msgBox.exec()
+
+        # Check which button was clicked and execute the corresponding code
+        if msgBox.clickedButton() == buttonContinue:
+            return False
+        elif msgBox.clickedButton() == buttonSave:
+            # Save files anyways
```
rozyczko, line 610: a side effect of this check is that if I hit `Escape`, the whole dialog disappears without saving or without giving users a chance to do anything.

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -507,47 +561,114 @@ def regenerateDocumentation(self):
...
+            c_tab_open = self.isWidgetInTab(self.tabWidget, self.c_editor_widget)
+
+            # Check to see if we have a certain model type open, and if so, write models
+            if py_tab_open and c_tab_open:
+                    self.writeFile(py_file, self.editor_widget.getModel()['text'])
+                    self.writeFile(c_file, self.c_editor_widget.getModel()['text'])
+            elif py_tab_open:
+                    self.writeFile(py_file, self.editor_widget.getModel()['text'])
+            elif c_tab_open:
+                    self.writeFile(c_file, self.c_editor_widget.getModel()['text'])
```
rozyczko, line 623: or just
```python
if py_tab_open:
    self.writeFile(py_file, self.editor_widget.getModel()['text'])
if c_tab_open:
    self.writeFile(c_file, self.c_editor_widget.getModel()['text'])

```

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -564,51 +685,168 @@ def getModel(self):
...
+            self.tabWidget.addTab(self.editor_widget, name)
+            self.editor_widget.modelModified.connect(self.editorModelModified)
+        elif filetype in ["c", ".c"]:
+            self.c_editor_widget = ModelEditor(self, is_python=False)
+            self.tabWidget.addTab(self.c_editor_widget, name)
+            self.c_editor_widget.modelModified.connect(self.editorModelModified)
+
+    def removeTab(self, filetype: str):
+        """
+        Remove a tab from the tab widget.
```
rozyczko, line 705: The two cases where `removeTab` is called (lines 185 and 92) use integers as arguments (`1` and `0`).
How is this going to work with this signature?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -564,51 +685,168 @@ def getModel(self):
...
         """
+
+        def formatPythonFlags() -> str:
+            """Get python flags for model and format into text"""
+            checkbox_values = {
+                'chkSingle': getattr(self.plugin_widget, 'chkSingle').isChecked(),
+                'chkOpenCL': getattr(self.plugin_widget, 'chkOpenCL').isChecked(),
+                'chkStructure': getattr(self.plugin_widget, 'chkStructure').isChecked(),
+                'chkFQ': getattr(self.plugin_widget, 'chkFQ').isChecked(),
+            }
```
rozyczko, line 806: why not just `self.plugin_widget.chkSingle.isChecked()`?

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -743,9 +1038,146 @@ def VR(%(args)s):
...
+    // WARNING: The use of Iqxy is generally discouraged; Use Iqabc instead
+    //    for its better orientational averaging and documentation for details.
+    // IMPORTANT: Only define ONE calculation for I(Q): either Iq, Iqac, Iqabc, or Iqxy;
+    //    remove others.
+    
+    return 1.0;
+}}
+"""
+
 if __name__ == '__main__':
```
rozyczko, line 1178: Not sure this is really needed. We no longer test those widgets as separate windows, but rather inside SasView. Please consider removing the whole conditional

src/sas/qtgui/Utilities/TabbedModelEditor.py
```diff
@@ -743,9 +1038,146 @@ def VR(%(args)s):
...
     app = QtWidgets.QApplication(sys.argv)
     sheet = TabbedModelEditor()
     sheet.show()
-    app.exec_()
+    app.exec()
```
rozyczko, line 1182: The file fails the ruff check with 8 errors. Consider fixing those.

rozyczko (changes requested):
Some actual failures encountered, so please have a closer look

src/sas/qtgui/Utilities/TabbedModelEditor.py
krzywon: Those instances are removing tabs from tabWidget, not calling this method. This isnt called anywhere, so I will likely remove it.

src/sas/qtgui/Utilities/PluginDefinition.py
```diff
@@ -167,19 +219,58 @@ def onParamsPDChanged(self, row, column):
...
+        for row in range(self.tblParamsPD.rowCount()):
+            for column in range(self.tblParamsPD.columnCount()):
+                if self._checkCell(row, column):
+                    break
+            else:
+                # If column loop finishes, continue to next row
+                continue
+            # If column loop was broken, break out again
+            break
+
```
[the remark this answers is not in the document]
krzywon: This is cleaner and I'll use it. My only issue with your suggestion is you must iterate through the entire column before checking if any cells in that row are empty.

jamescrake-merani (approved):
Happy with stuff on my side. I see there's one more outstanding issue that Piotr has highlighted before we merge but otherwise I think this is ready.

krzywon:
Items still outstanding for this PR that can be finished post feature-freeze:

- User documentation required for new features
- Capture keyboard interrupts gracefully in TabbedModelEditor (was this available previously?)
- Better documentation in the C model generated by the easy model editor

src/sas/qtgui/Utilities/TabbedModelEditor.py
krzywon: I agree, but for the sake of the feature freeze, I've added this as something that needs fixing during the bug fix phase of the release.

butlerpd:
I've done a functionality review though not as thorough as I would like given the time. There is one serious bug (easy to fix?) and a number of lesser issues:
* The big issue is an error in the code that adds an extra `parameters = [ ` line in front of the comment line explaining the meaning of each column (and is never closed) when generating the python file. This happens every time apply is pressed making it impossible to actually get a working model ready .
* ER should be commented out by default. With it on everything can be multiplied by S(Q). This will cause lots of confusion. However a comment above the commented out ER function explaining that uncommenting will allow it to be used with S(Q) but beware that the return value needs to be a useful value.
* It would be much better if validation was used in the parameter box. I note that if I use a string for the initial value the code defaults to making the initial guess = 1 but …
* Also there should be the new version of ER that is required for use with C models which should also be all commented out by default with explanation.
* Return should act as tab in parameter table. Once you get into the idea of hitting tab it works beautifully but return is also a common behavior and the fact that causes an extra line to be generated makes one think the "cursor" has moved to a new box somewhere but typing will erase my carefully chosen variable name - bad form...
* Default for the volume editor box should not be vol = 0 (1st polydisperse parameter * 0.0) probably times 1.0 would be good since I(Q) will be divided by that value at some point... and the default anyway is volume = 1.  A note in the comment however should mention that I(Q) will be divided by this value so the IQ equation needs to be written understanding that. It will be different than the equation they write if volume =1.
* if a C file is generated an extra line should be added to the python file with the source= line. That line should however be commented out till they have a working C file.  This last could be viewed as an extra feature as much as failure of expectation when generating  a C file?


That said, I think this is will be more helpful and less prone to users making incorrect polydisperse models as pointed out by @RichardHeenan a while back.

butlerpd (approved):
Given we have agreed to merge with bugs which will then be addressed in the bug fixing phase of the release, I am approving. Note that the change requested by Piotr seems to have been addressed? so this should be ready to merge

Issue #3021, opened 2024-08-15
Plot transforms should be owned by Plotter objects not Data objects


krzywon:
**Describe the bug**
The x and y transforms are currently owned by the Data objects that are held by in the Data Explorer. This is leading to issues when replotting data that was previously plotted and rescaled. Ideally, the transforms should be owned by each plot, not each data set. The only transform the data object should have is a suggested transform which is immutable.

This should be for 6.0.1, not 6.0.0, due to the number of places this would need to change.

**To Reproduce**
Follow the steps outlined in #1783

**Expected behavior**
Each new plot should have the default scaling.

**SasView version (please complete the following information):**
 - Version: 5.0.0+

PR #3148, opened 2024-11-14
Set proper plot scales for ancillary plots

krzywon:
## Description

This sets the default axis scales for plots based on their plot role. All plots are set to log/log in v6.0.0, including residuals, polydispersity, and P(r) plots. This also removes a matplotlib version check that should no longer be relevant.

## Review Checklist:

- [X] Documentation changes are **in this PR**

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -186,12 +186,8 @@ class PlotterWidget(PlotterBase):
             self.yscale = 'linear'
 
         # Include scaling (log vs. linear)
-        if version.parse(mpl.__version__) < version.parse("3.3"):
-            ax.set_xscale(self.xscale, nonposx='clip') if self.xscale != 'linear' else self.ax.set_xscale(self.xscale)
-            ax.set_yscale(self.yscale, nonposy='clip') if self.yscale != 'linear' else self.ax.set_yscale(self.yscale)
-        else:
-            ax.set_xscale(self.xscale, nonpositive='clip') if self.xscale != 'linear' else self.ax.set_xscale(self.xscale)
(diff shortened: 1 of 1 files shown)


krzywon:
I am making more changes to this in order to resolve #3021. I am removing `xtransform` and `ytransform` from the PlotterData object and solely using the DataRole. This will then act as a hint for new plots, but existing plots will not be transformed when new data is appended.

butlerpd (changes requested):
The problems seem to be getting worse not better? I'm wondering if we need to pull back on the changes and wait for the full plot refactor (that would be mean reverting whatever changes were made in 6.0.0 that caused the problems in the first place)? The old plotting infrastructure is clearly in need of that complete overhaul.

So far I tested only 3 things
* load 1D file and create plot. 
  * now try to do a linear fit (the data is not linear but no matter it should draw a line). The data disappears from the plot with a straight line added
  * Try changing scale .. say to Gunier - it still plots as log log but presumably on transformed data? Same with Porod
  * Now try changing to linear/linear -- we get the log log plot back instead
  * one could go on. This part worked mostly in 6.0 except that the ranges were chosen in very strange ways (so still a problem)
* Load a 2D data set
  * try an slicer. No 1D data is generated. Instead the following error is thrown:
 ```python
18:08:12 - ERROR: Traceback (most recent call last): File "sas\qtgui\MainWindow\DataExplorer.py", line 1127, in displayData AttributeError: 'DataRole' object has no attribute 'stand_alone'
````
  * In 6.0.0 it works but plots everything on log log so the annulus slicer (which should be linear on the x axis for phi) gives a very "interesting" plot
* Load a sesans data set
  * it now plots in log log and shows no data when using the test data `sphere_isis.ses`
  * in 6.0.0 just plotting the data produces a linear/linear plot as expected

At this point I've not done any fitting or played with any other plotting functions but not sure how useful an exercise that can be at the moment?

PR #3156, opened 2024-12-04
Testing periodictable v2.0 

krzywon:
## Description

This is a test to see if sasview is compatible with the changes in  `periodictable` v2.0.


krzywon:
The SLD calculator runs as expected. A few other tests might be needed, but we are nominally forward compatable with upcoming release of periodictable.

PR #3160, opened 2024-12-17
Delete intermediate theory plots by tab id not model id.

### LINKED ISSUE #3159: Error when plotting models
**Describe the bug**
This one seems to depend on some particular paths in the code but is reproducible. When activated, changing a parameter in the model and hitting "return" (NOT not just click away!!) the entire model parameter window goes blank
NoParams

**To Reproduce**
1. Start SasView
2. Add a model to the first fit page (e.g. sphere)
3. Turn on polydispersity
4. Set polydispersity (e.g. to .1)
5. calculate, then plot
6. Create a second fit page
7. add the exact same model (in this case "sphere")
8. Calculate, then plot
9. Return to first fit page
10. change the value of a parameter (I tried scale and radius so far) and **HIT RETURN** this step is important. Apparently just clicking away avoids the error path
11. the parameter screen goes blank
12. choose some other action (e.g. hit "Compute/Plot" or even just try to quit the program
13. Monster error pops up.
14. System is now in weird state. hitting "Compute/Plot" now throws the following error
```python
16:19:27 - ERROR: Traceback (most recent call last): File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 2461, in onPlot File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 2500, in showTheoryPlot RuntimeError: Internal C++ object (HashableStandardItem) already deleted.
```
15.  I recommend exiting the program at this point (click quit a second time if the error was generated on the first click)

**deduced so far**
* The two fitpages **must** use the same model
* Only the first fitpage needs to have polydispersity set.
* But -- the first model must be run with polydispersity set (not just turned on .. so some non zero value) PRIOR to starting the second fitpage

 - Version: 6.0.0

 - OS: tested on Windows 10

jamescrake-merani:
## Description

I've tried to fix #3159 by changing the `deleteIntermediateTheoryPlotsByModelID` method so that it deletes by tab id, and not model id. This was causing the bug as we had 2 tabs with the same model, and it was trying to delete theory items from the other tab.

While this fixes the problem, I'm not entirely sure how suitable this fix is as I'm not aware of why this method was needed in the first place, and thus it could potentially break some other functionality. It would probably be useful to discuss this fix at the call. 

## How Has This Been Tested?

Followed the reproduction instructions Paul provided in #3159. I am no longer able to reproduce the bug in this pull request.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -2069,7 +2069,7 @@ class DataExplorerWindow(DroppableDataLoadWidget):
         self.theory_model.appendRow(model_item)
         return model_item
 
-    def deleteIntermediateTheoryPlotsByModelID(self, model_id):
+    def deleteIntermediateTheoryPlotsByTabId(self, tab_id):
         """Given a model's ID, deletes all items in the theory item model which reference the same ID. Useful in the
         case of intermediate results disappearing when changing calculations (in which case you don't want them to be
         retained in the list)."""
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -1331,12 +1331,12 @@ class GuiManager:
             return
         per.currentTab.setTheoryItem(item)
 
-    def deleteIntermediateTheoryPlotsByModelID(self, model_id):
+    def deleteIntermediateTheoryPlotsByModelID(self, tab_id):
         """
         Catch the signal to delete items in the Theory item model which correspond to a model ID.
         Send the request to the DataExplorer for updating the theory model.
         """
-        self.filesWidget.deleteIntermediateTheoryPlotsByModelID(model_id)
+        self.filesWidget.deleteIntermediateTheoryPlotsByModelID(tab_id)
 
     def updateModelFromDataOperationPanel(self, new_item, new_datalist_item):
         """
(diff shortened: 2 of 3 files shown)


rozyczko:
I think that this looks sensible. DataItem is connected to a fitting tab, with a unique ID, so all dependent plots should be checked against that unique ID.

src/sas/qtgui/MainWindow/DataExplorer.py
```diff
@@ -2077,8 +2077,8 @@ def deleteIntermediateTheoryPlotsByModelID(self, model_id):
                 return
             match = GuiUtils.theory_plot_ID_pattern.match(data.id)
             if match:
-                item_model_id = match.groups()[-1]
-                if item_model_id == model_id:
+                item_tab_id = match.groups()[0]
```
rozyczko, line 2080: why do you compare the first, not last match?

rozyczko (approved):

butlerpd:
I think this should be against release_6.0.1 rather than main. Interestingly, checking 5.0.6 hitting return at the appropriate juncture immediately throws the traceback error but the parameter panel never goes blank?

jamescrake-merani:
> I think this should be against release_6.0.1

Yes I think you're right. I had it on main as that was the default when Github creates a PR from an issue. I'll try to change it. 

> Interestingly, checking 5.0.6 hitting return at the appropriate juncture immediately throws the traceback error but the parameter panel never goes blank?

Hmm interesting. I was going to try this for myself on IDAaaS but there seems to be some downtime at the moment but I'll have a look when thats back.

jamescrake-merani:
Looks like the Mac build failed but by the looks of it, it is just a CI issue. The error is:

```hdiutil: create failed - Resource busy```

jamescrake-merani:
> Interestingly, checking 5.0.6 hitting return at the appropriate juncture immediately throws the traceback error but the parameter panel never goes blank?

This is actually the behaviour I get on Linux, and I just tested 5.0.6, and got the same result (albeit on Linux again). My guess is that it depends when the parameters are redrawn, and at one point the theory item is accessed.

PR #3166, opened 2025-01-03
Use platformdirs package for all user-related file paths

### LINKED ISSUE #3142: Linux filepaths do not comply with the XDG Base Directory Specification
**Describe the bug**
This was raised during the review of the Flatpak. Right now, we are storing all of the configuration files, and compiled models that SasView generates in a directory that lives directly on the home directory. This does not comply with the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/)  (a kind of standard for Linux desktop applications).

Configuration files should be placed under `~/.config/sasview`, and user specific data should be stored in `~/.local/share/sasview`. The latter should also included the `sasview.log` file which should not be on the home directory. The idea of this is to not pollute the user's home directory.

This is a low priority issue as it should hopefully not block the Flatpak from being published. In fact, many popular applications break this rule as well. But it is probably a change we'll want to make in the future to promote better practices.

If a fix were to be implemented, there is an issue of older installations using the old directories, and then these being missed by newer installations. To get around this, I would suggest having a precedence system where we check one location first before checking another directory. This is a common approach taken by many other Linux applications.

**SasView version:**
 - Version: 6.0.0

**Operating system:**
 - OS: Any Linux distribution

krzywon:
## Description

This moves our user file handling to the `platformdirs` package for os preferred pathing. This is used to store log files, configurations, and app data. This was suggested by @llimeht in the wheels discussion in #3161. This automatically creates subfolders for each version so docs no longer require versioning in their paths.

### What has changed:
- No longer storing data in `~/.sasview`
  - The files are not all be in the same place depending on the OS
- Using more OS prefered storage locations for application files
- Updated calls to the directories to ensure new locations are used for docs, plugin models, the log file, categories, and config files.
- Added user documentation to let users know where these files live (Waiting on Flatpak info...)

### Good ideas, but outside the scope of this PR
- Add a method to open explorer windows directly to each file directory for to give the user easy access to the directories - OS-agnostic methods to do this don't exist.
- Copy example data from the as-installed location to the user directory - Working on in separate branch built off this one: https://github.com/SasView/sasview/tree/platfromdirs-example-data-move

### Other work to be done:
- Any site on sasview.org that references the user directory will need to be updated - https://github.com/SasView/sasview.github.io/issues/28

Fixes #3142

## How Has This Been Tested?

CI is able to launch both the Windows and Ubuntu 22 binaries.

## Review Checklist:

- [ ] There is an **issue** open for the documentation

**Installers**
- [X] This **will** affect the **installers**


src/sas/system/user.py
```diff
@@ -1,18 +1,120 @@
...
+        files = [f for f in os.listdir(old_plugins) if os.path.isfile(os.path.join(old_plugins, f))]
+        for file in files:
+            if not Path(new_plugins, file).exists():
+                shutil.copy2(Path(old_plugins, file), Path(new_plugins, file))
+    # Copy config file over
+    new_config_dir = Path(get_config_dir())
+    config_name = f'config-{sasview_version.split(".")[0]}'
+    old_config = old_sasview_usr_dir / config_name
+    new_config = new_config_dir / config_name
+    print(old_config.exists())
```
pkienzle, line 118: remove debug statements

rozyczko:
Shouldn't this be now against `main`?

src/sas/system/user.py
```diff
@@ -1,18 +1,122 @@
...
-def get_user_dir(create_if_nonexistent=True):
-    """
-    The directory where the per-user configuration is stored.
 
-    Returns ~/.sasview, creating it if it does not already exist.
+def get_dir_and_create_if_needed(path: PATH_TYPE, create_if_nonexistent: bool = True):
+    path = Path(path)
+    if create_if_nonexistent and not os.path.exists(path):
+        path.mkdir(parents=True, exist_ok=True)
+        path.chmod(764)
```
jamescrake-merani, line 26: Why did you choose this specific mode for `chmod`? When I comment it out (and delete the directories created before so it creates them from scratch), I don't get the permission error anymore. But if this permission is needed for something else then it might break something down the line. In any case, we'll probably have to change it.

krzywon: No specific reason. I'll remove this line and see if that fixes things.

jamescrake-merani (approved):
It now works on my machine, and code LGTM.

jamescrake-merani:
I've just noticed this is still marked as a draft. Is there any work left do on this?

krzywon:
> I've just noticed this is still marked as a draft. Is there any work left do on this?

Two items left:

* Documenting new user file locations (Required)
* Adding methods and GUI hooks to open user file locations (I strongly suggest this be added now that files are in different places)

pkienzle:
This is making pathing os specific rather than os agnostic. The previous system didn't care which operating system it was using.

The plugins were copied to `~/Library/Application Support/SasView/plugin_models`, including weights.

Logs are appended to ~/Library/Logs/SasView/sasview.log

This file already had logs from existing versions of python.

**Bug**: The new log entries are missing the date portion of the timestamp.

**Feature request**: Show the path to the file on disk in the model editor window.

**Feature request**: Plugin polydispersity distributions are not appearing in the plugin manager.

**Bug**: On my intel mac I'm getting the following when I click "Load File" on the model editor window:
```python
12:25:42 : ERROR : sas.qtgui.Perspectives.Fitting.FittingWidget (FittingWidget.py:158) :: Traceback (most recent call last):
  File ".../sasview/src/sas/qtgui/Utilities/TabbedModelEditor.py", line 162, in onLoad
    filename = QtWidgets.QFileDialog.getOpenFileName(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../site-packages/qtpy/_utils.py", line 159, in _from_kwarg_name_to_kwarg_name_
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
TypeError: PySide6.QtWidgets.QFileDialog.getOpenFileName((<sas.qtgui.Utilities.TabbedModelEditor.TabbedModelEditor(0x600000510780, name="TabbedModelEditor") at 0x14f4acc80>, 'Open Plugin', PosixPath('~/Library/Application Support/SasView/plugin_models'), 'SasView Plugin Model (*.py)', None)) is wrong (missing signature)
```

**Feature request**: Store base doc tree in the application install directory. Only store patches in the user data.

The current Doc tree is 170 MB. If you end up with a new tree for each point release that's a lot of data.

In the unlikely case the user has edited the docs, will the edits show up when a new version is installed? The only way I see that happening is if the edits are represented as a patch file.

krzywon:
> This is making pathing os specific rather than os agnostic. The previous system didn't care which operating system it was using.

I updated the PR description.

> The plugins were copied to `~/Library/Application Support/SasView/plugin_models`, including weights. [edit: weights were correctly copied to `~/Library/Application Support/SasView/weights`]

Thanks, I updated the documentation with this.

> Logs are appended to ~/Library/Logs/SasView/sasview.log
>
> This file already had logs from existing versions of python.

The log file gets copied to the new directory on app launch, if no log exists in the new location.

> **Bug**: The new log entries are missing the date portion of the timestamp.

Fixed.

> **Feature request**: Show the path to the file on disk in the model editor window.
> 
> **Feature request**: Plugin polydispersity distributions are not appearing in the plugin manager.

Beyond the PR scope at this point.

> **Bug**: On my intel mac I'm getting the following when I click "Load File" on the model editor window:

Fixed.

> **Feature request**: Store base doc tree in the application install directory. Only store patches in the user data.

Another beyond the PR scope, though, noted.

src/sas/qtgui/MainWindow/media/sasview_files.rst
```diff
@@ -0,0 +1,93 @@
...
+and a categories file that allows the end user to change how their models are organized in the fitting perspective. The
+general configuration file, config-<v>.json is a json file that stores a mapping of config variables to their updated values.
+This file is only generated and stored locally if something in the :ref:`Preferences` panel has been modified. The categories
+file, categories.json, stores the user-preferred model organization, allowing a user to add model categories, move models
+into new categories, and not show certain models. More information on this feature is availabe in :ref:`Category_Manager`.
+
+OS-specific file locations for configuration files:
+ - Windows: ~/AppData/Local/sasview/SasView/
+ - MacOS: ~/Library/Application Support/SasView/
+ - Linux: <TODO>
```
jamescrake-merani, line 35: ~/.config/SasView/

src/sas/qtgui/MainWindow/media/sasview_files.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Log Files
+---------
+SasView creates a time-stamped log file that includes all messages entered into the log explorer, as well as a number of
+other messages related to application startup and shutdown.
+
+OS-specific file locations for log files:
+ - Windows: ~/AppData/Local/sasview/SasView/Logs/
+ - MacOS: ~/Library/Logs/SasView/
+ - Linux: <TODO>
```
jamescrake-merani, line 47: ~/.local/state/SasView

src/sas/qtgui/MainWindow/media/sasview_files.rst
```diff
@@ -0,0 +1,93 @@
...
+
+OS-specific file locations for plugin files:
+ - Windows:
+    - Plugin Models: ~/AppData/Local/sasview/SasView/plugin_models/
+    - Compiled Models: ~/AppData/Local/sasview/SasView/compiled_models/
+ - MacOS:
+    - Plugin Models: ~/Library/Application Support/SasView/plugin_models/
+    - Compiled Models: ~/Library/Application Support/SasView/compiled_models/
+ - Linux:
+    - Plugin Models: <TODO>
```
jamescrake-merani, line 66: Plugin models: ~/.local/share/SasView/plugin_models
Compiled models: ~/.local/share/SasView/compiled_models

src/sas/qtgui/MainWindow/media/sasview_files.rst
```diff
@@ -0,0 +1,93 @@
...
+Documentation
+-------------
+SasView moves its documentation into a user location when launching the app, if the documentation does not already exist.
+Each version of SasView has its own documentation that may be different between versions, so a separate directory is used
+for each version of the doc files.
+
+OS-specific file locations for documentation:
+ - Windows: ~/AppData/Local/sasview/SasView/<sasview.version>/doc/
+ - MacOS: ~/Library/Application Support/SasView/<sasview.version>/doc/
+ - Linux: <TODO>
```
jamescrake-merani, line 80: ~/.local/share/SasView/6.0.1/doc/

src/sas/qtgui/MainWindow/media/sasview_files.rst
```diff
@@ -0,0 +1,101 @@
...
+.. _UserFiles:
+
+SasView User Files
+==================
+
+During the installation process and while running the app, SasView creates a number of files that are stored outside
+the install location on the end users computer. These files include configurations, plugin models, compiled plugins,
+documentation, and example files. Please refer to the specific section to find where each type of file is stored.
+
+File paths that start with the '~' character refer to the user directory, which is typically `C:\\\\Users\\<username>` in
```
jamescrake-merani, line 15: The amount of backslashes on this directory doesn't look right to me.

src/sas/qtgui/MainWindow/media/sasview_files.rst
```diff
@@ -0,0 +1,101 @@
...
+.. _Plugin_Files:
+
+Plugin Files
+------------
+Plugin models, whether they be written by the user (:ref:`Writing_a_Plugin`), downloaded from the model marketplace
+(`Model Marketplace <https://marketplace.sasview.org/>`_), or sent from another SasView user, should be stored in a
+specific location to allow SasView to find them. During normal fitting operations, models are compiled and the compiled
+files are stored in a parallel location.
+
+OS-specific file locations for plugin files:
```
jamescrake-merani, line 66: I'm not sure whether it would be clear to someone reading this what the distinction is between the 'Plugin Models', and 'Compiled Models' directories. Furthermore, I think saying 'OS-speciific file locations for **plugin** files' is a bit confusing because it might imply that you need to put stuff into the `compiled_models` directory to add a plugin model.

jamescrake-merani (changes requested):
Code LGTM. Just a few comments regarding the documentation.

src/sas/qtgui/MainWindow/media/sasview_files.rst
krzywon: A single backslash acts as an escape character, so two consecutive backslashes result in a single backslash in the built documentation.

https://docutils.sourceforge.io/docs/user/rst/quickref.html#escaping

jamescrake-merani: Ah I see. That makes sense.

jamescrake-merani (approved):

PR #3181, opened 2025-01-13
Removing macOS-13 from the regular build

wpotrzebowski:
## Description

As decided at the contributor camp in Lund, we will only build macos-latest in the regular builds and the macOS-13 only for the release. 

Fixes # (issue/issues)

## How Has This Been Tested?

## Review Checklist:

**Installers**
  - [x] **MacOSX** installer (GH artifact) has been tested (installed and worked) 

### DIFF
--- a/.github/workflows/matrix.py
+++ b/.github/workflows/matrix.py
@@ -29,7 +29,6 @@ os_release_list = [
     'ubuntu-20.04',
     'ubuntu-22.04',
     'windows-latest',
-    'macos-13',
     'macos-latest',
 ]
 


krzywon (approved):
I would argue this doesn't fully fix the issue, since the `macos-13` build is never trigged for releases, so the issue should remain open for now. This does, however, solve the more immediate issue of the notarization clashes, so I think this can be merged.

butlerpd:
So we should unlink the issue from this PR so that it doesn't close automatically? Otherwise it looks like we can merge this now?

PR #3184, opened 2025-01-14
3157: Fix v6.0.0 plot scaling issues

### LINKED ISSUE #3157: SESANS models not being displayed in plot (Sasview 6.0.0)
When plotting SESANS data, the experimental data is properly displayed but not the model curves (in particular negative values). It is necessary  to right click on the plot > change scale > linear plot to solve the issue (even though the SESANS plot is linear on both axes by default). This is possibly because the model thinks it is a log log plot and avoid plotting negative values?

sasview bug

SasView 6.0.0
Windows 11

krzywon:
## Description

This is a lesser implementation of the work already in #3148. This fixes the plot scaling issues introduced in v6.0.0, but doesn't make as clean a separation of the data and plots as I originally hoped to make for the release.

Refs #3021 (A start for this, but doesn't fully remove the x/y transforms from PlotterData classes and still uses them in a couple of places)
Fixes #3157

## How Has This Been Tested?

Tested from source on Win11. Verified the following:
- SAS data and fits are plotted together on log/log scale
- SAS residuals are on log/lin scale
- SESANS data and fits are plotted together on lin/lin scale
- SESANS residuals are on lin/lin scale
- All residual plots (SAS and SESANS) have the +/- 3 lines
- Polydispersity plots are on lin/in scale
- No change to 2D plot scaling

## Review Checklist:

**Installers**


krzywon:
MacOS builds failing due to notarization issue that is fixed in #3183

butlerpd (changes requested):
I'm still not convinced converting to roles at this stage was/is a good idea. It is time consuming to test even because there are so many nooks and crannies but here is my best. 

This is certainly behaving better than the last PR However there remain several issues which still need fixing:
* Residuals
  * The plot now uses a linear scale for the y axis; good. However the data still seems to think it is on a log scale I think (even though it is not) and thus no negative residuals get plotted (as if the mirror images was covered over 😄) .
* Sectors
  * Always plot log log even if the x axis is in degrees. So annual and wedge phi are incorrect. They are done correctly  in 5.0.6
* Linear plots
  * Axis default (home button) are the untransformed limits. In other words, the x and y range are based on the data Q and I(Q) not the transformed (e.g. Q^2 or Ln(I) etc). The result is that, if one is lucky, the data is squished into a tiny corner of the plot, if not, and the plot goes negative such as in the cross-sectional Guinier, the data disappear altogether!!
* Interesting error that seems to be introduced by this PR but not sure how? but it is not present in 6.0.0.
  *  When using a fitpage without any data to create a theory plot (but NOT if there is data associated with the fitpage) the following error gets thrown when hitting compute/plot (and yes, it gets thrown twice)
```python
19:37:38 - ERROR: Traceback (most recent call last): File "sas\qtgui\MainWindow\GuiManager.py", line 1339, in deleteIntermediateTheoryPlotsByTabId File "sas\qtgui\MainWindow\DataExplorer.py", line 2085, in deleteIntermediateTheoryPlotsByTabId NameError: name 'model_id' is not defined
19:37:38 - ERROR: Traceback (most recent call last): File "sas\qtgui\MainWindow\GuiManager.py", line 1339, in deleteIntermediateTheoryPlotsByTabId File 
```
Things tested that do work now:
* Polydispersity seems to work fine now
* Show SLD seems fine
* Other perspectives seem fine

krzywon:
> * Residuals
>   
>   * The plot now uses a linear scale for the y axis; good. However the data still seems to think it is on a log scale I think (even though it is not) and thus no negative residuals get plotted (as if the mirror images was covered over 😄) .
> * Sectors
>   
>   * Always plot log log even if the x axis is in degrees. So annual and wedge phi are incorrect. They are done correctly  in 5.0.6

Both of these should be fixed now. Please check again when you are able.

> * Linear plots
>   
>   * Axis default (home button) are the untransformed limits. In other words, the x and y range are based on the data Q and I(Q) not the transformed (e.g. Q^2 or Ln(I) etc). The result is that, if one is lucky, the data is squished into a tiny corner of the plot, if not, and the plot goes negative such as in the cross-sectional Guinier, the data disappear altogether!!

I'm still looking into this. We might want to create a separate issue, since this is also happening in v6.0.0.

> * Interesting error that seems to be introduced by this PR but not sure how? but it is not present in 6.0.0.
>   
>   * When using a fitpage without any data to create a theory plot (but NOT if there is data associated with the fitpage) the following error gets thrown when hitting compute/plot (and yes, it gets thrown twice)
> 
> ```python
> 19:37:38 - ERROR: Traceback (most recent call last): File "sas\qtgui\MainWindow\GuiManager.py", line 1339, in deleteIntermediateTheoryPlotsByTabId File "sas\qtgui\MainWindow\DataExplorer.py", line 2085, in deleteIntermediateTheoryPlotsByTabId NameError: name 'model_id' is not defined
> 19:37:38 - ERROR: Traceback (most recent call last): File "sas\qtgui\MainWindow\GuiManager.py", line 1339, in deleteIntermediateTheoryPlotsByTabId File 
> ```

This error was fixed by PR #3190. I've integrated that changed into my branch so that doesn't seem to happen anymore.

krzywon:
> Axis default (home button) are the untransformed limits. In other words, the x and y range are based on the data Q and I(Q) not the transformed (e.g. Q^2 or Ln(I) etc). The result is that, if one is lucky, the data is squished into a tiny corner of the plot, if not, and the plot goes negative such as in the cross-sectional Guinier, the data disappear altogether!!

I think I have this figured out now. The plot axis range was based off the original data, not the data view, which is why it was always the same value. My latest commit should hopefully fix this now.

butlerpd (commented):
A thorough functionality review on all plots indicates that we are almost there as well as revealing a few other issues to be reported elsewhere.

The only real issue remaining is that the wedge slicer in Q defaults to linear linear. This may be because the implementation seems to be setting the `fold=false` creating the mirror data in negative Q space. This was never caught because the default plot was correctly log log till now. This needs to be log log by default again as the wedge by definition should be `fold=true` as there is only one wedge on one side of Q space. Also the wedge must be using the same default scaling as the sector average since after plotting a wedge in Q slice, even after clearing the slicer, the sector slicer will plot in linear linear. These should be decoupled.

Note: it should be a trivial one liner to fix the fold issue in which case it would be good to do so as part of this PR. However, this is a longer standing issue and can have a "good first issue" made for it.

One other thing that would be nice to fix but was apparently an issue in 5.0.6 would be to apply the same fix that was just done for linear plots of data to linear plots of "theory."  Indeed, linear plots of theory data had apparently always had the plot limits determined by the non transformed x and y.  Moreover, this caries over EVEN when doing a freeze theory and then plotting that theory data, which is now supposed to be actual data (but clearly is not behaving 100% as such).  Again if the fix is simple based on having already done it for the regular data that would be nice but not necessary as this is a longer standing issue? 

there are is one other issue to do with sesans fitting but clearly not relevant here. All the plotting looks good.

krzywon:
> The only real issue remaining is that the wedge slicer in Q defaults to linear linear. This may be because the implementation seems to be setting the `fold=false` creating the mirror data in negative Q space. This was never caught because the default plot was correctly log log till now. This needs to be log log by default again as the wedge by definition should be `fold=true` as there is only one wedge on one side of Q space. Also the wedge must be using the same default scaling as the sector average since after plotting a wedge in Q slice, even after clearing the slicer, the sector slicer will plot in linear linear. These should be decoupled.

I fixed this here. The other suggestions should probably become separate issues for future releases, at this point.

src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
```diff
@@ -200,14 +200,14 @@ def _post_data(self, new_sector=None, nbins=None):
...
-        new_plot.plot_role = DataRole.ROLE_ANGULAR_SLICE
         # If the data file does not tell us what the axes are, just assume...
         if self.averager.__name__ == 'SectorPhi':
             # angular plots usually require a linear x scale and better with
             # a linear y scale as well.
             new_plot.xaxis("\\rm{\phi}", "degrees")
             new_plot.xtransform = 'x'
             new_plot.ytransform = 'y'
+            new_plot.plot_role = DataRole.ROLE_ANGULAR_SLICE
         else:
```
butlerpd, line 211: I would think that the two lines above the added line (`new_plot.ytransform = 'y'`  and `new_plot.xtransform = 'x'`) should at least commented out if not deleted?

butlerpd (approved):
This does fix the introduced issue. I will add issues for the remaining issues as suggested. Modulo the question on the code this should be ready to merge

PR #3187, opened 2025-01-16
3020: Linear fit no longer causes graph rescaling

### LINKED ISSUE #3019: Linearized plot fit window is modal
**Describe the bug**
Because the linearized Fit panel is modal, the user has to close the panel, losing all their work, if they want to zoom into the plot differently... or use the sliders to visually refine the data range etc.

**To Reproduce**
1. Load a data set and send to a plot
2. change scale to Guinier
3. right click on the plot and chose the data->linearized plot
4. the modal window appears and nothing can be done on the plot until it is closed.

**Expected behavior**
The Fit panel should be a regular panel and not modal. Modal windows normally are there to require the user to respond to some important issue not for regular interaction. It should act more like 

 - Version: 6.0.0b2

 - OS: Windows 10

**Additional context**
Identified as part of the NIST CNR summer school

### LINKED ISSUE #3020: Linearized fits rescale to the full data set whenever a fit is performed
**Describe the bug**
When performing linearized fits on a data set, hitting fit will reset the graph range to the full range of the data. Worse it seems to be applying the range from the scale of the original data?  Either way, since most Linearized plots apply to a small region it becomes very difficult to deal with, particularly since the fit panel is modal (see #3019), requiring the user to close the fit, losing all their work, resetting the zoom factor on the graph then starting the fit panel again.  The instructor's comment to me was "At this point users just go use another software when they need to do a linear fit."

**Expected behavior**
Fitting should have no affect on the graph ranges.

 - Version: 6.0.0b2

 - OS: Windows 10

**Additional context**
Reported during the NIST CNR summer school

krzywon:
## Description

This fixes the issue with the linear fits causing the plot scaling to reset every time the fit is performed. ~~As a side bonus, this also removes the linear fit window modality, so you can interact with the plot with the linear fit window open.~~ After some discussion, I have restored the modality of the Linear Fit window. The linear fit system changes based on the plot scale so more work is needed to link those, and that is too much for 6.0.1.

~~Note: The Q-range sliders are having issues in this current implementation, but I thought this was such an improvement over what it is in 6.0, this should be included in the bug fix release, and maybe I can fix the slider issue in another PR before the release.~~

~~Fixes # 3019~~ Modality restored on Jan. 31, 2025
Fixes #3020

## Review Checklist:

**Installers**


src/sas/qtgui/Plotting/LinearFit.py
```diff
@@ -79,19 +78,22 @@ def __init__(self, parent=None,
         self.txtBerr.setText("0")
         self.txtChi2.setText("0")
 
-
         # Initial ranges
-        self.txtRangeMin.setText(str(max_range[0]))
-        self.txtRangeMax.setText(str(max_range[1]))
+        tr_min = GuiUtils.formatNumber(max_range[0])
+        tr_max = GuiUtils.formatNumber(max_range[1])
+        self.txtRangeMin.setText(str(tr_min))
```
rozyczko, line 84: `GuiUitils.formatNumber` will only return a string. There is no need to cast it to str again

src/sas/qtgui/Plotting/LinearFit.py
```diff
@@ -304,32 +308,36 @@ def floatInvTransform(self, x):
...
-        elif self.xLabel == "x^(4)":
-            return numpy.sqrt(numpy.sqrt(x))
-        elif self.xLabel == "log10(x)":
-            return numpy.power(10.0, x)
-        elif self.xLabel == "ln(x)":
-            return numpy.exp(x)
-        elif self.xLabel == "log10(x^(4))":
-            return numpy.sqrt(numpy.sqrt(numpy.power(10.0, x)))
-        return x
+        from sas.qtgui.Utilities.GuiUtils import xyTransform
```
rozyczko, line 311: is this import so expensive that you had to defer it? If not, please consider moving it to the top.

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -671,20 +671,18 @@ def onLinearFit(self, id):
                     xlabel=self.xLogLabel,
                     ylabel=self.yLogLabel)
         fit_dialog.updatePlot.connect(self.onFitDisplay)
-        if fit_dialog.exec_() == QtWidgets.QDialog.Accepted:
-            return
```
rozyczko, line 675: There was a good reason why we wanted to have this dialog modal. Too bad I don't remember that reason, but I _think_ it was Paul's suggestion and also something present in 4.x

But since nobody remembers anything, let's just make it a regular, non-modal dialog.

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -827,17 +825,15 @@ def xyTransform(self, xLabel="", yLabel=""):
...
-        tempx = fit_data[0]
-        tempy = fit_data[1]
         self.fit_result.x = []
         self.fit_result.y = []
-        self.fit_result.x = tempx
-        self.fit_result.y = tempy
+        self.fit_result.x = temp_x
+        self.fit_result.y = temp_y
         self.fit_result.dx = None
         self.fit_result.dy = None
```
rozyczko, line 839: You could use the chance to add type hints to `LinearFit.py` and lint it properly with `ruff` :)

rozyczko (commented):
The code looks OK, so apart of recent Paul's observations with the runtime issues this should be fine.

src/sas/qtgui/Plotting/Plotter.py
jamescrake-merani: I suspect this is to stop the user from modifying the plot while they have the fit dialog open.

To give an example, I managed to break it by opening the fit dialog, then going back to the plot, and changing the scale. Going back to the dialog, pressing the 'Fit' button gives me the following error:

```
12:13:03 - sas.qtgui.Perspectives.Fitting.FittingWidget - ERROR - Traceback (most recent call last):
  File "/var/home/jamescrake-merani/repos/sasview-repos/sasview/src/sas/qtgui/Plotting/LinearFit.py", line 252, in fit
    self.drawSliders()
  File "/var/home/jamescrake-merani/repos/sasview-repos/sasview/src/sas/qtgui/Plotting/LinearFit.py", line 330, in drawSliders
    self.clearSliders()
  File "/var/home/jamescrake-merani/repos/sasview-repos/sasview/src/sas/qtgui/Plotting/LinearFit.py", line 345, in clearSliders
    self.q_sliders.clear()
  File "/var/home/jamescrake-merani/repos/sasview-repos/sasview/src/sas/qtgui/Plotting/QRangeSlider.py", line 52, in clear
    self.clear_markers()
  File "/var/home/jamescrake-merani/repos/sasview-repos/sasview/src/sas/qtgui/Plotting/QRangeSlider.py", line 105, in clear_markers
    self.line_min.clear()
  File "/var/home/jamescrake-merani/repos/sasview-repos/sasview/src/sas/qtgui/Plotting/QRangeSlider.py", line 202, in clear
    self.remove()
  File "/var/home/jamescrake-merani/repos/sasview-repos/sasview/src/sas/qtgui/Plotting/QRangeSlider.py", line 208, in remove
    self.inner_marker.remove()
  File "/home/jamescrake-merani/repos/sasview-repos/sasview/.venv/lib/python3.12/site-packages/matplotlib/artist.py", line 256, in remove
    raise NotImplementedError('cannot remove artist')
NotImplementedError: cannot remove artist
```
So for this reason, I'm leaning towards the view that we should keep this as a modal dialog. Unless there's a way to stop this from happening.

jamescrake-merani (commented):
I agree with what Piotr has already said regarding the code quality. I have tested this branch on my machine, and it fixes both issues. However changing the linear fit dialog so that it is not modal seems to cause issues if you change the plot scale. I guess the reason we don't want it to be modal is so the user can still interact with the plot while looking at the data in the linear fit dialog. Is there any trivial way we can lock the scale of the plot while this window is open perhaps?

krzywon:
I think this PR could benefit from the work I did in #3184, but for now, I'll look into the suggestions made to see if I can improve this.

src/sas/qtgui/Plotting/Plotter.py
krzywon: I was able to create a different error. The LinearFit window changes based on the plot scales to know whether a Guinier or basic linear fit is needed. I'll add a comment about this to #3019, but the modality will likely need to be reverted for 6.0.1. 

To reproduce the error I'm seeing:
- Open a plot on log/log scale
- Open the linear fit window
- Change the plot scale to Guinier with the linear fit window open
- Run a linear fit
- Error!

PR #3190, opened 2025-01-20
Model ID Not Defined

### LINKED ISSUE #3159: Error when plotting models

jamescrake-merani:
## Description

For some reason, a commit seems to have been left out of #3160 which leads to an error whenever a plot is created. I think this may have been a mistake I made when cherry picking commits so I could rebase the branch onto release 6.0.1. I may have inadvertently left out the first commit.

Fixes #3159 

## How Has This Been Tested?

I have followed the reproduction steps mentioned in the original issue, and I am not able to reproduce the bug anymore.

## Review Checklist:

**Documentation**

**Installers**

### DIFF
--- a/src/sas/qtgui/MainWindow/DataExplorer.py
+++ b/src/sas/qtgui/MainWindow/DataExplorer.py
@@ -2081,8 +2081,8 @@ class DataExplorerWindow(DroppableDataLoadWidget):
                 return
             match = GuiUtils.theory_plot_ID_pattern.match(data.id)
             if match:
-                item_model_id = match.groups()[-1]
-                if item_model_id == model_id:
+                item_tab_id = match.groups()[0]
+                if item_tab_id == tab_id:
                     # Only delete those identified as an intermediate plot
                     if match.groups()[2] not in (None, ""):
                         items_to_delete.append(item)


krzywon (approved):
Fixes the issue and code looks fine.

Issue #3209, opened 2025-02-11
New version available box


lucas-wilkins:
Make a popup when a new version is available, it should have

* a button to open the browser to the latest version
* a "don't show this again" checkbox
* an entry in the config for the last version that the "don't show me again" was selected for (so it will still show again if there is an even newer version)

krzywon:
As discussed today, the initial implementation will be included in v6.0.1. To close this issue, the following items will need to be changed:

- The check for new versions menu option should open the new pop-up window
- The new version is available message should be removed from the status bar
- All code related to displaying the status bar message should be removed

lucas-wilkins:
I'm not sure that's right @krzywon, checking for new versions from sasview should run regardless of whether you have dismissed the box, and it should compare it the version you're running. As I said in the call, the new box behaves differently to the other stuff regarding the checks it makes - it compares to the latest version you have installed, and/or last one you dismissed the dialog for.

Additionally, you wouldn't want to use the same popup - it would need to be changed for the new context. 
The way it pops up is different too.

We could get rid of it from the status bar, but is there any reason to?

Basically, everything is different in this code, because it does different things. The existing stuff isn't bad.

krzywon:
> We could get rid of it from the status bar, but is there any reason to?

Yeah, you're right. I agree there is no need to get rid of the status bar updates. I'm saying we don't need two different code bases that perform similar functions. The existing logic should probably be moved into the new system.

PR #3212, opened 2025-02-12
3211: Fix for sasmodel doc build

krzywon:
## Description

This disables OpenCL during the sasmodels doc generation to allow the model figures to be generated under Windows.

Fixes #3211

## How Has This Been Tested?

CI

## Review Checklist:

**Installers**
- [X] There is a chance this will affect the **installers**, if so
  - [ ] **Windows** installer (GH artifact) has been tested (installed and worked) 
  - [ ] **MacOSX** installer (GH artifact) has been tested (installed and worked) 

**Licencing** (untick if necessary)
- [x] The introduced changes comply with SasView license (BSD 3-Clause)

PR #3216, opened 2025-02-14
3209 new version available box

lucas-wilkins:
## Description

Adds a popup that appears at startup when there is a new version. Fulfils the requirements specified in #3209 

Fixes # (issue/issues)

## How Has This Been Tested?

Tested as best I can without live testing from the web. Looks OK, possibly worth doing a live test though.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
+++ b/src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
@@ -116,9 +116,8 @@ def get_current_release_version() -> Optional[tuple[str, str, tuple[int, int, in
 
 def a_newer_than_b(version_a: tuple[int, int, int], version_b: tuple[int, int, int]) -> bool:
     """ Check if version_a is strictly newer than version_b"""
-    return version_a[0] > version_b[0] \
-        or version_a[1] > version_b[1] \
-        or version_a[2] > version_b[2]
+    
+    return version_a > version_b
 
 def maybe_prompt_new_version_download() -> Optional[QDialog]:
     """ If a new version is available, and Show a dialog prompting the user to download """
--- a/src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
+++ b/src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
@@ -1,0 +1,189 @@
+import webbrowser
+from copy import copy
+from typing import Optional
+
+import json
+import logging
+import re
+
+from PySide6 import QtWidgets
+from PySide6.QtCore import QSize
+from PySide6.QtGui import QIcon
+from PySide6.QtWidgets import QDialog, QVBoxLayout, QHBoxLayout, QLabel, QWidget, QCheckBox, QPushButton, QSpacerItem
+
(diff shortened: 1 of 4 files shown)


krzywon:
Note - the builds here are failing because of the issue fixed in #3212.

krzywon:
This works as expected (tested by changing the version in `sas.sasytem.version` to 5.0.7) where a window pops up on program startup and announces a new version is available. Looking at it, this might a new feature... We can discuss this at one of our upcoming meetings.

jamescrake-merani:
> Looking at it, this might a new feature...

The idea was that we wanted to get this change in as soon as we could because we can't apply it to previous versions. So once this change is implemented in 6.0.1, a 6.0.1 user will get notification that 6.1.0 has been released when that happens. While I appreciate 6.0.1 is meant to be mostly a bug fix release, in my opinion this PR should be an exception to that. But yes we could discuss further in future meetings.

src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
```diff
@@ -0,0 +1,182 @@
...
+        return version_string, url, parse_version(version_string)
+
+
+    except Exception as ex:
+        logging.info("Failed to get version number", ex)
+
+
+def a_newer_than_b(version_a: tuple[int, int, int], version_b: tuple[int, int, int]) -> bool:
+    """ Check if version_a is strictly newer than version_b"""
+    return version_a[0] > version_b[0] \
```
jamescrake-merani, line 119: Perhaps there is a slightly more concise way to write this? I guess its not a big deal with it only being three lists though.

jamescrake-merani (approved):
Code LGTM. Lets discuss @krzywon 's point on Friday before merging.

src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
pkienzle: Test with [3, 1, 0] and [4, 0, 0]. With "or" logic then it skips the major version and sees that 1 > 0 and so returns true. Simply using `version_a > version_b` should do the right thing.

pkienzle: Note: simple comparison reports (4, 0, 0, 'a1') as newer than (4, 0, 0).

butlerpd:
How is this different from the current logic that pops up a notice below the log box indicating a new version? I appreciate that it was not always obvious to the user and that as annoying as popups are this might be the place to use one, that is a only a GUI change right?

Which, to @krzywon point, I think this is basically a bug fix since we are saying the long standing implementation does not work well enough?

src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
lucas-wilkins: yeah, will fix

krzywon (approved):
I tested the latest change and the issue noted by @pkienzle is now fixed. Approved.

PR #3220, opened 2025-02-20
Include Debian-specific pathing in documentation system

krzywon:
## Description

Doc location change based on the Debian system (unclear if this works for Ubuntu yet...)

## How Has This Been Tested?

:dizzy_face: Hopefully people with Linux-based systems can test this...

## Review Checklist:

**Installers**
  - [ ] **Ubuntu** installer (GH artifact) has been tested (installed and worked) 
  - [ ] **Debian** installer (GH artifact) has been tested (installed and worked)

### DIFF
--- a/src/sas/sascalc/doc_regen/makedocumentation.py
+++ b/src/sas/sascalc/doc_regen/makedocumentation.py
@@ -3,6 +3,7 @@ Creates documentation from .py files
 """
 import logging
 import os
+import platform
 import sys
 import subprocess
 import shutil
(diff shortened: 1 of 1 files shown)


src/sas/sascalc/doc_regen/makedocumentation.py
```diff
@@ -27,30 +28,39 @@
...
 RECOMPILE_DOC_LOCATION = HELP_DIRECTORY_LOCATION
 IMAGES_DIRECTORY_LOCATION = HELP_DIRECTORY_LOCATION / "_images"
 SAS_DIR = Path(sys.argv[0]).parent
 
 # Find the original documentation location, depending on where the files originate from
-if os.path.exists(SAS_DIR / "doc"):
+if platform.system() == 'Linux':
+    BASE_DIR = Path("/usr/share/doc/sasview/")
+    ORIGINAL_DOCS_SRC = BASE_DIR / "doc-source"
+    ORIGINAL_DOC_BUILD = BASE_DIR / "html"
```
pkienzle, line 48: This will be wrong if you are running conda-forge version or a locally compiled version on a system which has sasview installed as a package. Similarly for other tests against 'Linux'.

jamescrake-merani (changes requested):
I'm not able to get documentation to work using the `run.py` script. I tried to test the PyInstaller bundle but I wasn't able to get it to run without crashing due to an unrelated error.

I think, as Paul alluded to, the issue is that the paths here make assumptions about where SasView is installed which may not be accurate on the user's machine. I haven't yet tested on the Flatpak as I need to create a new build.

I will need to discuss this further at the technical meeting because there's a few things about the documentation system I have questions about.

llimeht:
I can confirm that the patch there is not suitable for inclusion in sasview.git - it's fine in the Debian package because the Debian package controls where everything is installed and it's always in the same location -- in the FHS-defined locations. It's possible to then skip all calculations of where resources might be found because we know where the resources actually are. For installations that don't use FHS locations (such as flatpaks, conda, pip, ...), that's not going to work because they can get installed in lots of different spots, or (and this is a significant thing to fix) not installed at all.

There is only one reliable file location that can be used from within the sasview module and that is where sasview module is located on disk, and therefore where data files within that module are to be found. Using `__file__` is ok(ish) but will fail in perverse ways with modules in zip files, pyinstaller etc; using `importlib.resources` is the portable approach. Any other code that tries absolute paths (`/usr/share/doc/...`) or calculate some relative paths to locations outside the module (`../doc`) is going to be fragile, messy, and probably wrong.

If the docs get installed inside the module, then all except `run.py` and `pip install -e` are then solved in one hit and it becomes simpler to imagine how we might deal with those last two. This is a longer-term approach that I am working on as agreed in our contributor camp discussions.

krzywon:
Based on feedback, this won't meet our needs so I am closing this.

PR #3247, opened 2025-04-01
Poresize

achicks15:
## Description

Please include a summary of the change and which issue is fixed. List any dependencies that are required for this change.

Fixes # (issue/issues)

## How Has This Been Tested?

## Review Checklist:

**Installers**


src/sas/sascalc/poresize/SizeDistribution.py
```diff
@@ -0,0 +1,163 @@
...
+        self.set_weights()
+        
+        ## Return Values after the MaxEnt? 
+        self.BinMagnitude_maxEnt = np.zeros_like(self.bins)
+        self.chiSq_maxEnt = np.inf
+        self.Iq_maxEnt = np.zeros_like(self.Iq)
+
+
+
+    def set_qmin(self, value:float):
```
jamescrake-merani, line 48: Having setters like this is not very Pythonic in my eyes. We could use Python properties (and in particular the setter attribute):
https://docs.python.org/3/library/functions.html#property

But I'm also not sure we even need to use properties in this instance.

test/poresize/MaxEntTest.py
```diff
@@ -0,0 +1,89 @@
+import pytest
+from maxEnt_method import sizeDistribution
+import unittest
+import csv
+import numpy as np
+from sasmodels import resolution as rst
+from sasdata.dataloader import data_info
+
+class SizeDistributionTest(unittest.TestCase):
```
jamescrake-merani, line 9: For new code, I think we normally prefer to use Pytest for unit testing.  Pytest is already being imported so would it be possible for the unit test itself to be in Pytest?

jamescrake-merani (commented):
I haven't tested anything yet but just some initial thoughts on the code.

src/sas/sascalc/poresize/maxEnt_method.py
```diff
@@ -0,0 +1,390 @@
...
+            f_old = f.copy()                # preserve the last solution
+            f += xi.transpose().dot(beta)   # move the solution towards the solution, SB eq. 25
+        
+            # As mentioned at the top of p.119,
+            # need to protect against stray negative values.
+            # In this case, set them to RESET_STRAYS * base[i]
+            #f = f.clip(RESET_STRAYS * blank, f.max())
+            for i in range(n):
+                if f[i] <= 0.0:
+                    f[i] = RESET_STRAYS * first_bins[i]
```
krzywon, line 368: Likely faster, but untested: `f = np.array([x if x > 0.0 else RESET_STRAYS * y for x, y in zip(f, first_bins)])`

krzywon (commented):
Code review only

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
butlerpd, line 1: Not sure this is the change that is causing the problem @backmari, but I noticed in testing the installer (which looks really good btw!) that when first sending data to the perspective the data set gets plotted twice. Being on top of each other you cannot see it except that the legend shows it is there twice.

However, as soon as I change a background and it recalculates and replots that second data set (or at least the legend for it) disappears.  Not sure when that second plot is being drawn?

src/sas/sascalc/poresize/maxEnt_method.py
butlerpd: Perhaps, but the point is to keep that code as is (after losing additions at the top of the file). It is essentially taken from GSAS code.  It is not pretty by but it works and is tested. To fully understand it so that we can rewrite would take too much time. The goal would be to replace this module in the future.

test/poresize/data/Alumina_usaxs_irena_plot.jpg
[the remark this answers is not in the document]
butlerpd: stuff to lose though now that you mention it, it **might** be interesting to keep for documentation? we'll see. Have not gotten that far yet. but will not be in the test folder for sure

src/sas/sascalc/poresize/plottableHist.py
```diff
@@ -0,0 +1,45 @@
...
+    _yaxis = ''
+    _yunit = ''
+    
+    def __init__(self, x, y, dy=None, binWidth=None):
+        self.x = np.asarray(x)
+        self.y = np.asarray(y)
+        if dy is not None:
+            self.dy = np.asarray(dy)
+        if binWidth is not None:
+            if len(binWidth) > 1:
```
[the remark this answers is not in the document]
butlerpd: LOL .. as mentioned in my email about which files to worry about - this is not one of them. For the first iteration (ver 0.9 in SasView 6.1.0) we have now agreed not to try to include histogram plots. The "real" results anyway use scatter plot of the same data that is shown histogrammed.

src/sas/sascalc/poresize/media/gsaslicense2013.txt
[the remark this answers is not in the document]
butlerpd: should be as discussed with Jan and Bob.

src/sas/qtgui/Perspectives/SizeDistribution/UI/convert_all.bat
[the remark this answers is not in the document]
butlerpd: Thanks. We discussed but no knowing we erred on the side of caution because all other perspective directories have them.

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
backmari: I noticed too that the plot initially has the data plotted twice. I looked at the Invariant perspective for inspiration, and it has the same problem... Maybe there is some better example of a plot with more than one curves.
image

butlerpd: Interesting! You are right that is the only other place that has more than 2 data sets drawn at once ... and they both have the problem.  Hummmm... maybe something we leave for debugging later since it seems to be a broader problem

smk78: If you mean the double-plotting of the experimental data, it doesn't happen in Corfunc... :-)

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionLogic.py
```diff
@@ -0,0 +1,126 @@
...
+            if self._data.dy is not None and np.any(self._data.dy):
+                self.di_flag = True
+
+    def computeDataRange(self):
+        """
+        Wrapper for calculating the data range based on local dataset
+        """
+        return self.computeRangeFromData(self.data)
+
+    def computeRangeFromData(self, data):
```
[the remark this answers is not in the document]
backmari: nope there is not :slightly_smiling_face:

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionLogic.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            qmin = 0
+            try:
+                x = max(np.fabs(data.xmin), np.fabs(data.xmax))
+                y = max(np.fabs(data.ymin), np.fabs(data.ymax))
+            except (ValueError, TypeError):
+                msg = "Unable to find min/max of \n data named %s" % self.data.filename
+                raise ValueError(msg)
+            qmax = np.sqrt(x * x + y * y)
+        return qmin, qmax
```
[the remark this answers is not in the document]
backmari: The perspective currently doesn't allow loading 2D data, @butlerpd is this the right thing to do? If so, this 2D logic can be removed altogether. If not, then there is more work to be done to handle 2D :slightly_smiling_face:

src/sas/qtgui/Perspectives/SizeDistribution/UI/convert_all.bat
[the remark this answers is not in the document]
backmari: Thanks, done :heavy_check_mark:

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
```diff
@@ -0,0 +1,598 @@
...
+        self.is_calculating = True
+        self.enableButtons()
+        self.fit_thread = SizeDistributionThread(completefn=self.fittingCompleted)
+        self.fit_thread.queue()
+
+    def onFullFit(self):
+        self.is_calculating = True
+        self.enableButtons()
+        self.fit_thread = SizeDistributionThread(completefn=self.fittingCompleted)
+        self.fit_thread.queue()
```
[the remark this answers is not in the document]
backmari: This has been updated, they now differ by one bool parameter `full_fit`.

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
```diff
@@ -0,0 +1,598 @@
...
+            # Suppress the GUI update until the move is finished to limit model calculations
+            self.data_plot.slider_update_on_move = False
+            self.data_plot.slider_perspective_name = "SizeDistribution"
+            self.data_plot.slider_low_q_input = ["txtMinRange"]
+            self.data_plot.slider_high_q_input = ["txtMaxRange"]
+
+            GuiUtils.updateModelItemWithPlot(self._model_item, self.data_plot, title)
+            self.communicate.plotRequestedSignal.emit(
+                [self._model_item, self.data_plot], None
+            )
```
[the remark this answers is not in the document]
backmari: Good suggestion, this has been cleaned up.

butlerpd:
Note.. there remain some bugs and some debugging code that need cleanup but otherwise is ready to start reviewing in view of upcoming code freeze.

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionThread.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+
+            self.completefn(results)
+        except KeyboardInterrupt:
+            pass
+        except Exception:
+            if self.error_func is not None:
+                self.error_func(
+                    "Error in size distribution fit: %s" % sys.exc_info()[1]
+                )
```
krzywon, line 83: Exceptions are already handled in the base CalcThread class in `CalcThread._run()`.

krzywon (commented):
This works, as best as I can tell.

Changes that are required prior to the bug fix freeze:
- User documentation
- A full fit gives 10 `failed to converge` messages when convergence fails. Ideally, you would quit early in this case.
- Document unit tests
- Add type hints
- Add inline documentation which will be used to populate the developer documentation (sas.sascalc.poresize.maxEnt_method.maxEntMethod.MaxEnt_SB is a good example of what is desired)

Changes that could come for the next release:
- Linearize calculations using numpy where possible and remove the for (and nested for) loops
- Clean up the plotting or show the plot(s) as part of the perspective

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionThread.py
backmari: We are using this to add custom error handling, like re-enabling the fitting buttons in the GUI. Ah but I see now that we can initialize CalcThread with this function as `exception_handler` instead. I will fix that.

PR #3267, opened 2025-04-24
Include images icons and css in sasview package

bmaranville:
## Description

This PR adds a `MANIFEST.in` file at the root which controls which additional package data will be included in built wheels and pip installs of `sasview`

This fixes an issue where pip installing from source results in an unusable state, where `site-packages/sas/qtgui` is missing all the .png and .css files needed to run the application

## How Has This Been Tested?

I tried to start the sasview GUI from the command line after pip installing from source, and it failed.  Then I added this file, re-ran `pip install .`, and then tried starting the GUI, and it worked.

## Review Checklist:

**Installers**

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+recursive-include src/sas/qtgui *.png *.svg *.css


jamescrake-merani:
Now that we've merged the wheels into main, Stuart reckons this is no longer needed anymore. I'm keeping this PR open for now but if it does indeed turn out we don't need it then this PR should be closed.

bmaranville:
I agree - this can be closed

PR #3281, opened 2025-04-29
Let's build wheels

### LINKED ISSUE #3075: setup.py install does not install enough for sasview to work
**Describe the bug**
`setup.py install` should install sasview and have it work; running tests assumes that it does work and so this is called from CI.

However, `setup.py` doesn't know about any of the non-python files ("package-data" in setuptools parlance) that sasview requires. There's are hundreds of images, icons, stylesheets etc that are required at runtime.

This is papered over in `installers/sasview.spec` with a curated list of data files to include. Pushing this into the sasview package itself so that they land in the wheel will help clean up numerous pain points... including making sasview installable via `pip`.

**To Reproduce**
1.  create venv, enter it, install deps as normal
2. `git clone ...` 
3. `pip install .`
4. `sasview`
5. See error

**Expected behavior**
sasview installation in the venv is a working installation

**Errors**
`FileNotFoundError` for `sas/qtgui/images/SVwelcome.png` that results in sasview crashing immediately. (And if you fix that one, there are many more)

 - Version: git HEAD (main branch tested)

 - OS: Linux (but that's not relevant)

**Additional context**
Copying the files in manually lets sasview start - we're not a million miles from this working. The config really just needs to get into `pyproject.toml` (or at worst `setup.py`) and then wheels become a simple reality.

Crudely (and copying more that strictly necessary in order to get it working...):
```
$ find sasview/src/sas/qtgui/ -type d | cut -d/ -f3- | while read d; do mkdir -p venv/lib/python3.11/site-packages/$d ; done
$ find sasview/src/sas/qtgui/ -type f \! -name \*py | cut -f3- -d/ | while read f; do cp "sasview/src/$f" venv/lib/python3.11/site-packages/$f; done
```

llimeht:
## Description

We've talked a bit about how to simplify the build-system of sasview, to make CI simpler, building installers simpler, and to make the sasview code base much more like a normal Python code base. A big part of that is getting usable wheels out of sasview.

A standards-compliant build system for SasView means that the source needs to not have lots of special curation prior to building - `pip install .` should work without moving files around, checking out sibling projects into a special directory structure, or running other preparatory commands.

By making wheels that satisfy the build dependencies of sasview in the sasdata and sasmodels modules, sasview can then be built as desired

 - https://github.com/SasView/sasmodels/pull/642 
 - https://github.com/SasView/sasdata/pull/114

As part of this, we move to:

 - declarative build system
 - removing deprecated interfaces from code
 - being able to publish wheels

Fixes #3075 (and probably others)

## How Has This Been Tested?

Tested under Linux only

Known/potential issues:

 - testing needed under windows, macos
 - doc viewer doesn't work with these changes under linux (but then it doesn't current anyway? fixing that is a next step)
 - doc viewing/editing functions possibly currently broken - but can also be simplified as a next step
 - some sharp edges knocked off version string handling that should help with pre-releases (we fix this now because it was causing problems with testing throughout this bit of development work)
 - `build_sphinx.py` / `collect_sphinx_sources.py` can be further simplified in time.

## Review Checklist:

**Installers**


build_tools/requirements-test.txt
```diff
@@ -0,0 +1,6 @@
+pytest
```
jamescrake-merani, line 1: What is this file for?

jamescrake-merani (approved):
LGTM.

build_tools/requirements-test.txt
llimeht: Installing the test dependencies - it's used later in CI for that step. At the moment it's only pytest, but we likely want to see things like pytest-qt in there some time.

jamescrake-merani: Ah I see. That makes sense.

jamescrake-merani:
I'm merging this now so the release branch will be created. About to go for lunch but will carry this on when I'm back.

bmaranville:
Currently doing `pip install -e .` breaks because isolated builds are broken.

Is this intended to work in isolated mode?  If I do `python -m build` in sasview now, it fails because many support files needed for doing sphinx things don't exist in the isolated environment.

PR #3329, opened 2025-04-30
clear convergence plot before plotting new result

bmaranville:
## Description

Fixes a bug where convergence plot is not cleared between fits, leading to old axis labels and ticks not being removed (and accumulating making the plot unusable) 

## How Has This Been Tested?

Local testing - ran multiple fits without fix, seeing garbled axes on convergence plot.
After fix, convergence plot is cleared (like all other result plots) correctly.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/PlotView.py
+++ b/src/sas/qtgui/Utilities/PlotView.py
@@ -56,8 +56,10 @@ class FitResultView(QtWidgets.QWidget):
 
 class ConvergenceView(FitResultView):
     def plot(self):
+        self.figure.clear()
         best, pop = self.state.convergence[:, 0], self.state.convergence[:, 1:]
         plot_convergence(pop, best, self.figure)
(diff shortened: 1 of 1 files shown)


krzywon:
This works fine when running DREAM, but the convergence plots for LM are blank on the second and all subsequent fits.

bmaranville:
I think the LM fitter is just more likely to return immediately on subsequent fits because it's already within tolerance.  I saw the same thing as you, but if after the first fit you twiddle some parameter knobs and run LM again, it works fine.

krzywon:
If an immediate return happens and no convergence state is returned, couldn't you suppress the update?

bmaranville:
we could suppress the clear in the plot update if the new data is empty, for sure... but then how would a user know that the fit actually ran?  It would appear to them that pressing the "fit" button has no effect, because nothing changes (and it would look like they get the same convergence every time they press "fit")

PR #3330, opened 2025-05-01
Documentation viewer/regeneration, wheels, importlib.resources

llimeht:
## Description

Finding resources inside the module should be done with `importlib.resources` wherever possible, and not using file paths. Calculating locations using file paths fails when modules are inside zip files, wheels, manipulated by pyinstaller. The resources can even end up in different locations depending on the operating system and installation 'style' (all in one directory vs separate `/usr/bin` or `.../.venv/bin/` etc). 

The current code for locating the documentation makes incorrect assumptions about the proximity of the documentation files to the executable.

This PR adds code to obtain the documentation files firstly from the module using `importlib.resources` and then other locations as a fallback. 

Fixes # (I thought there were several bugs?)

This PR is needed in both the `release-6.1.0` branch and in `main`; I'm assuming that it will end up in `main` on its own via a merge, but please let me know if you'd like a parallel PR for `main` to help it land earlier.

## How Has This Been Tested?

Documentation locating, copying, and viewing code works from:

- unpacked source on linux with run.py
- pyinstaller on linux
- installed wheel on linux

The document generation code (manually calling `makedocumentation.py` was tested with

- unpacked source on linux
- installed wheel on linux

In principle this work is making the code more portable and so that should be enough... but it would be good to hear that the different sort of pyinstaller unpacking that happens on windows and macos works. When testing, you should delete your existing local cache of doc source and build to ensure that the relevant code is actually used. 

## Review Checklist:

**Installers**

### DIFF
--- a/run.py
+++ b/run.py
@@ -83,11 +83,6 @@ def prepare():
     #    with cd(root):
     #        subprocess.call((sys.executable, "setup.py", "build"), shell=False)
 
-    # Notify the help menu that the Sphinx documentation is in a different
-    # place than it otherwise would be.
-    docpath = joinpath(root, 'docs', 'sphinx-docs', '_build', 'html')
-    os.environ['SASVIEW_DOC_PATH'] = docpath
-
 import multiprocessing
 if __name__ == "__main__":
     multiprocessing.freeze_support()
--- a/src/sas/sascalc/doc_regen/makedocumentation.py
+++ b/src/sas/sascalc/doc_regen/makedocumentation.py
@@ -1,15 +1,15 @@
 """
 Creates documentation from .py files
 """
+import importlib.resources
 import logging
 import os
 import sys
 import subprocess
 import shutil
 
-from os.path import join, basename
 from pathlib import Path
-from typing import Union
+from typing import Sequence, Union
 
 from sas.sascalc.fit import models
 from sas.system.user import get_app_dir_versioned
@@ -17,7 +17,8 @@ from sas.system.user import get_app_dir_versioned
 from sasmodels.core import list_models
 
 
-PATH_LIKE = Union[Path, str, os.PathLike]
+PATH_LIKE = Union[Path, str, os.PathLike[str]]
+
 
 # Path constants related to the directories and files used in documentation regeneration processes
 APP_DIRECTORY = Path(get_app_dir_versioned())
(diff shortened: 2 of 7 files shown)


src/sas/sascalc/doc_regen/makedocumentation.py
```diff
@@ -33,26 +35,149 @@
...
+        if source_dir.exists():
+            shutil.copytree(source_dir, MAIN_DOC_SRC)
+        else:
+            logging.error("Could not find source for documentation")
+
+    if not MAIN_BUILD_SRC.exists():
+        if build_dir.exists():
+            shutil.copytree(build_dir, MAIN_BUILD_SRC)
+        else:
+            logging.error("Could not find pre-built documentation")
```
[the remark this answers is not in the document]
llimeht: codescene seems to see complexity in logging code quite often - my preference is to have the conditional and log things rather than leave it as a mystery as to what happened.

jamescrake-merani (commented):
The code changes look good to me, but I'm not able to get this to work on my machine. When I click on the help button, I get the following error.

```16:35:47 - INFO: Extracting docs from on-disk locations: source=docs/sphinx-docs/source-temp, build=build/doc
16:35:47 - ERROR: Could not find source for documentation
16:35:47 - ERROR: Could not find pre-built documentation
```

The error happens using the run.py file. I don't have the time to look into this more closely but I can take a closer look when I'm back from leave.

I'll need to test the Flatpak once thats built but I suggest we don't wait for that, and just merge once we know it works with the Windows, and Mac installers. As I need to sit down, and reconfigure the Flatpak when I get back from leave.

src/sas/sascalc/doc_regen/regenmodel.py
```diff
@@ -59,10 +59,14 @@
 
 from sas.sascalc.doc_regen.makedocumentation import MAIN_DOC_SRC, generate_html
 
-from typing import Dict, Any
+from typing import Any, Dict, Union
 from sasmodels.kernel import KernelModel
 from sasmodels.modelinfo import ModelInfo
 
+
+PATH_LIKE = Union[Path, str, os.PathLike[str]]
```
krzywon, line 67: This type now exists in three different places. I would suggeset importing it from `makedocumentation.py` so only one place needs to be updated if pathing changes in the future.

run.py
```diff
@@ -83,11 +83,6 @@ def prepare():
     #    with cd(root):
     #        subprocess.call((sys.executable, "setup.py", "build"), shell=False)
 
-    # Notify the help menu that the Sphinx documentation is in a different
-    # place than it otherwise would be.
-    docpath = joinpath(root, 'docs', 'sphinx-docs', '_build', 'html')
-    os.environ['SASVIEW_DOC_PATH'] = docpath
-
```
krzywon, line 90: This might be related to the issue noted by @jamescrake-merani. `run.py` does not require a developer to locally build, so the `build/docs` directory may not exist.

llimeht: I couldn't find any code that uses `SASVIEW_DOC_PATH` in sasview - hence removing it in this PR. The last mention of it in git (other than run.py) I can find is the removal of the wx sasgui (83b31818f).

I suspect that the answer is that in a fresh checkout, nothing has tried to build the docs at all when the user goes straight to `run.py`. Running the build machinery without build the wheel should be enough: `hatchling build --hooks-only`; I'll do some experimenting on this as part of thinking about the documentation updates that are needed.

PR #3331, opened 2025-05-01
Improvements in sasview version handling - internal version and versioned directory

llimeht:
## Description

A minor bugfix and a small improvement to do with sasview version handling:

- The import of `_version` from the hatch-vcs automation was incorrect - trivial fix to bring it in. 
- The versioned directory for config/cache should be only dependent on the base version (x.y.z) component and not on development markers (alpha, beta, dev).

This PR is aimed against `release-6.1.0` branch but is also needed in `main`. Please let me know if duplicating the PR is desired to get this fix synced sooner rather than later.

## How Has This Been Tested?

`run.py`:

- check the version string in title bar and logs 
- check what directory it wants to use when run

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/system/version.py
+++ b/src/sas/system/version.py
@@ -1,5 +1,5 @@
 try:
-    from _version import __version__
+    from ._version import __version__
 except ImportError:
     __version__ = "6.1.0a1"
 
--- a/src/sas/system/user.py
+++ b/src/sas/system/user.py
@@ -1,8 +1,12 @@
 import os
+from packaging.version import Version
 import shutil
 from pathlib import Path
 from platformdirs import PlatformDirs
-from sas.system.version import __version__ as sasview_version
+from sas.system.version import __version__
+
+
+sasview_version = Version(__version__).base_version
 
 # Create separate versioned and unversioned file locations
 PLATFORM_DIRS_VERSIONED = PlatformDirs("SasView", "sasview", version=sasview_version)


src/sas/system/version.py
```diff
@@ -1,5 +1,5 @@
 try:
-    from _version import __version__
+    from ._version import __version__
```
jamescrake-merani, line 2: Does this need to be a relative import?

jamescrake-merani (approved):
This seems sensible to me.

> This PR is aimed against release-6.1.0 branch but is also needed in main. Please let me know if duplicating the PR is desired to get this fix synced sooner rather than later.

We decided that the release branch is going to be merged into main immediately after the release. As this is a bug fix of a feature going into 6.1.0, it shouldn't be necessary to merge it into main at this time.

src/sas/system/version.py
llimeht: Yes, forgetting that was my initial error, with the exception handling meaning that the error wasn't seen.

llimeht:
Merged: the internal version reported by sasview in titlebar, logs, wheel is based on `git describe --tags` i.e. in this branch it is now: `v6.1.0a1-3-g00f0a7738` (unpacking: v6.1.0a1 + 3 commits, at commit id `00f0a7738`). Being able to uniquely identify which version is being discussed will help a lot. 

> We decided that the release branch is going to be merged into main immediately after the release. As this is a bug fix of a feature going into 6.1.0, it shouldn't be necessary to merge it into main at this time.

We might want to do two things in `main` before too much further work happens in `main`:
- include these two changes
- make a `v6.2.0.dev0` tag (and then a `v6.3.0.dev0` tag in main once the 6.2.0 release branch is cut)
- update `_version.py` to `v6.2.0.dev0` at that stage too? (this starts to make merge conflicts though)
- we should also consider putting something into `run.py` so that we have a consistent view of the version in checkouts; I can look at that in a bit

With that in place, when running artifacts from git main, the internal version reported in titlebar and logs will be automatically calculated as being part of the 6.2.0.dev0 development series. At the moment, it is calculated as being part of a 6.0.1 post-release (git main is currently `v6.0.1-1170-ge8edf81ca` for instance, because that's the last tag git can see in the history)

PR #3350, opened 2025-05-05
[Sasview v6.1.0] Model editor documentation and template generation fixes

### LINKED ISSUE #3293: User documentation required for model editor

### LINKED ISSUE #3296: Model Editor generates and extra 'parameters = [' line in front of the comment line
'The big issue is an error in the code that adds an extra `parameters = [` line in front of the comment line explaining the meaning of each column (and is never closed) when generating the python file. This happens every time apply is pressed making it impossible to actually get a working model ready.'

### LINKED ISSUE #3297: ER should be commented out by default
'ER should be commented out by default. With it on everything can be multiplied by S(Q). This will cause lots of confusion. However a comment above the commented out ER function explaining that uncommenting will allow it to be used with S(Q) but beware that the return value needs to be a useful value.' - @butlerpd

### LINKED ISSUE #3301: Default for the volume editor box should not be vol = 0
'Default for the volume editor box should not be vol = 0 (1st polydisperse parameter * 0.0) probably times 1.0 would be good since I(Q) will be divided by that value at some point... and the default anyway is volume = 1. A note in the comment however should mention that I(Q) will be divided by this value so the IQ equation needs to be written understanding that. It will be different than the equation they write if volume =1.' - @butlerpd

krzywon:
## Description

This resolves a number of issues with the new model editor found during testing of the original PR.

Refs #3291
Fixes #3293 - Documentation has been updated to include the new options in the easy model editor window, inclduing a new image of the latest model editor GUI\
Fixes #3297 - Ensures the ER() method is commented out when the model is generated
Fixes #3301 - Sets the default volume to be 1.0 in the form volume editor
No Issue - Adds `form_volume` and `radius_effective` to the parameter list when the model is flagged as a structure factor

## Review Checklist:

**Installers**


butlerpd (approved):
Testing functionality was severely hampered by the fact that the installer version now throws an error whenever saving (checking) a model. Basically it can no longer find `sasmodels` and so fails on the from` sasmodels.special import *`  I'm guessing this is due to changes in the CI? Still needs to be fixed.

Otherwise a couple of things that are likely to cause confusion:
* When entering data in the parameter section, returning to a box in order to remove it, it is not possible to use the delete or backspace key. Instead one has to actually type something new then backspace.
* on starting the `Add Custom Model` the `Generate Python model` checkbox is checked but can be unchecked and the `Generate C model template` is unchecked. This means the user can have both boxes unchecked until a new name is typed in the plugin name box. However, if that name is then deleted, the `Generate Python model` checkbox is checked but greyed out and can no longer be unchecked.  As far as I can tell there is no way to get to an illegal situation but it would be less confusing if the box started out greyed out as it should till a name is typed in that matches an existing python file.
* The editor box I think is sized to the display? but wheter or not that is the case, I am able, by adding polydisperse parameters which creates a new editing box for the volume equation, to lose sight of the new blank line which is a bit unnerving. It would be nicer I think if we fixed the size of the parameter entry box to 3 lines (the heading, the line being edited and the one below it?

That said I believe this fixes the issues it says it fixes so am approving the pull request thought the first issue of not finding sasmodels does need to be addressed

PR #3351, opened 2025-05-06
Speed up CI - faster python module installation with uv

llimeht:
## Description

The [uv tools](https://docs.astral.sh/uv/) are reimplementations of various Python package management tools in rust - they are between 10 and 100× faster for many operations. 

`uv` has lots of features and we might explore some of them later; @jamescrake-merani might like `uv lock` and `uv sync` for managing the dependency list, for example.

For the moment though, this PR is only looking at using `uv` as a pip-compatible package installer. It downloads and installs the wheels from pypi in exactly the same way as we are used to with `pip`, except it takes 3-5 s rather than 1-3 min (this time is quite variable).

Across the whole CI run, this shaves approx 2-4 min off CI with uv's faster dependency resolver and unpacking of wheels.

## How Has This Been Tested?

CI

## Review Checklist:

**Installers**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
         with:
           python-version: 3.11
 
-      ### Installation of some build-dependencies
+      ### Compilation of build-dependencies
 
       - name: Build wheel of ${{ matrix.package[0] }}
         run: |
@@ -91,6 +91,7 @@ jobs:
       MACOS_DMG_NAME: SasView6-${{ matrix.os }}.dmg
       # Only sign/notarize if the secrets will be available
       SIGN_INSTALLER: ${{ (! github.event.repository.fork && ! github.event.pull_request.head.repo.fork) && 'true' || '' }}
+      UV_SYSTEM_PYTHON: 1
 
     name: "${{ matrix.job_name }}"
 
@@ -110,10 +111,12 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-          cache: 'pip'
-          cache-dependency-path: |
-            **/ci.yml
-            **/requirements*.txt
+
+      - name: Install uv utilities to speed up python module installation
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+          prune-cache: false
 
       ### Installation of build-dependencies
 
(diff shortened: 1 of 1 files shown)


llimeht:
A decision about when/if to do this is needed - #3358 touches the same lines in `ci.yml`

 - if we want this for 6.1.0 image generation, then I can rebase it onto the `release-6.1.0` branch, fixing up the conflicts
 - if this is not for 6.1.0 then it shouldn't get merged at all until the `release-6.1.0` branch is merged back into `main` because it will make for a pile of conflicts in that merge; I can rebase it onto `main` at that time.

wpotrzebowski (approved):

rozyczko:
I tried using `uv` in our builds but the gain was minuscule. `pip install` times are not the limiting factor - the main issue is packaging with `pyinstaller`.
Not saying that using `uv` is bad - I use it myself for everything pip, but the applicability in this particular workflow is limited, I would say.

llimeht:
> I tried using `uv` in our builds but the gain was minuscule. `pip install` times are not the limiting factor - the main issue is packaging with `pyinstaller`. Not saying that using `uv` is bad - I use it myself for everything pip, but the applicability in this particular workflow is limited, I would say. ( https://github.com/SasView/sasview/tree/test_uv )

Speeding up CI doesn't need us to look at the slowest step and work only on that - we can look at any slow step that is on the critical (slowest) path through the parallel jobs. Installing the deps using pip is on that slowest path in building the installer images. Speeding up installation also gives unit test results back faster. So yes, pyinstaller is definitely slow, but pip is still slowing down CI.

pip calculating the dependencies, downloading and unpacking them can take between 30s and 3min on CI, depending on the architecture and the state of the cache. uv downloads about 4x faster (not relevant if the cache is hot) and unpacks about 20x faster.

In putting together this PR I did benchmarking using CI runs to see the differences. I also ran tests here on a development machine using `requirements-release-ubuntu-latest.txt` (making new venvs and using `--no-cache` options as needed):

| cache  |  pip | uv  |
|-------|-------|-----|
| cold   | 76s | 18s |
| hot  | 55s |  1s |

This machine is substantially faster than the CI workers, where these operations take 2-3× longer.

Incidentally, INNO to make the windows installer package (`iscc`) takes even longer than the pyinstaller step ... see #3352  #3356 #3362 for working to speed up that step. I concur that there are yet more steps we should look at speeding up.

jamescrake-merani:
Not exactly related to this PR, but I've mentioned before that I want to rewrite our developers guidance as it currently recommends Conda, and I think this can be replaced with uv. But this work will likely be post-release.

PR #3355, opened 2025-05-06
Suppress signature errors when CI run on PRs from forks

llimeht:
## Description

CI already knows not to try to sign/notarise the installers when running on push within a fork. This change ensures that the sign/notarise steps only happen when a PR comes from the `sasview/sasview.git` repository and not from a fork. This will suppress spurious errors about signing keys not being available in PRs.

Anyone using the unsigned installers likely needs to jump though additional hoops to enable the unsigned code, but they can't currently even get to the unsigned installers because the job fails prior to upload.

## How Has This Been Tested?

CI on this PR - it doesn't fail :)

(Checking that signing still works for 'internal' PRs is still needed)

## Review Checklist:

**Installers**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
     env:
       MACOS_DMG_NAME: SasView6-${{ matrix.os }}.dmg
       # Only sign/notarize if the secrets will be available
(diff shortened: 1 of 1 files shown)


krzywon (approved):

llimeht:
Thanks @krzywon! @jamescrake-merani do we also want this in `release-6.1.0` straight away?

PR #3356, opened 2025-05-06
Keep some excess files out of the distribution

llimeht:
## Description

As part of looking at the size of the installer and the speed of generating the installer, I've looked through the list of files included in it. In tweaking build backends, we've picked up some additional files that are unneeded, so we can exclude them from the installer. The `media` directories in particular add substantially to the disk footprint.

~~This PR probably needs to be rebased on top of some of the others in the queue (#3330, #3354, #3355 in particular?). I'll mark this as a __draft__ for until that has happened so that others know they don't need to do this work again.~~

## How Has This Been Tested?

Run of linux pyinstaller image; check documentation; check icons; check about box (i.e. places where missing resources would be an issue)

## Review Checklist:

**Installers**


krzywon (approved):
Thanks for the clarification, @llimeht. Approved

PR #3368, opened 2025-05-08
Update documentation about building from source

### LINKED ISSUE #3353: Update documentation on installing/developing sasview
**Describe the bug**

The existing `INSTALL.txt` is quite outdated (sasview 5 instructions, jenkins references; needs to be updated with new instructions)

There are also some instructions on the wiki that are conda-specific; other links are 404. Are there other places to update?

 - Version: 6.0.1a1 / git

 - OS: all

llimeht:
## Description

A first pass at updating the documentation to talk about tools to build/install that are no longer `setup.py`.

- I've not revised the sections encouraging the use of conda environments — I don't think they offer anything over a venv these days but push a lot of complexity and time at users to get conda working first, rather than just a simple Python environment. 
- I've not simplified the discussions about obtaining dependencies yet. We can now allow `pip` to do all this

A lot of complex instructions can in a future PR be replaced by `pipx install sasview` but that discussion can happen later - updating the basic documentation seemed more urgent.

I suspect we should also consider merging `INSTALL.md` into `README.md` to make a nicer and more informative landing page in github …. but that is for a future PR.

Fixes #3353 

## How Has This Been Tested?

The instructions in `INSTALL.md` are from #3354 which work for me and are reported to work for others. 

## Review Checklist:

**Installers**


INSTALL.md
```diff
@@ -0,0 +1,65 @@
...
+## Making a SasView Development Environment
+
+If you're familiar with working with developing in Python, then the
+very quick version is:
+
+```console
+$ git clone https://github.com/sasview/sasview/
+$ cd sasview
+$ python -m venv venv
+$ . venv/bin/activate
```
krzywon, line 31: Suggestions:
- Setting the venv to `venv` seems off. The name should be `sasview` or something similar
- Unless something changed recently, venv activation is different based on your OS. Maybe add a reference to the python venv site?

llimeht: My only experience on windows is still using bash as the shell (e.g. git bash) in which case the activation is the same. The `venv` module documentation didn't seem very useful. https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#create-and-use-virtual-environments looked better - should we link to that? I actually expected a reasonable number of people would be managing and activating the venv via their IDE and it was only dinosaurs like me who were doing it from the command line.

The venv docs say "conventionally named .venv or venv in the project directory" which is what I've always done. If it's more obvious to call it sasview then I can easily add that.

(Given that other developers currently have no way of knowing how to get a git checkout going unless they happen to have followed #3354, I'm quite happy to merge and improve in a subsequent iteration with more input from folks who use other operating systems)

krzywon: > https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#create-and-use-virtual-environments looked better - should we link to that

This was the documentation I was thinking of, so yes, please. I think adding this link should be enough to satisfy both of my suggestions. I'm going ahead and approving this PR, so feel free to merge once you've added the link.

> I actually expected a reasonable number of people would be managing and activating the venv via their IDE and it was only dinosaurs like me who were doing it from the command line.

I guess I'm a dinosaur, too!

krzywon (approved):
Pending inclusion of a link to python venv management

PR #3373, opened 2025-05-10
Start of the size distribution perspective documentation.

### LINKED ISSUE #3260: Write user documentation for the Pore Sized Distribution perspective
Write documentation explaining the theory and use of the perspective  in the `.media/sizedistribution.rst` file.

butlerpd:
## Description

This adds documentation for the new Size Distribution perspective. As with things like the invariant it starts with a general discussion of the method and its limitations followed by GUI use instructions.  This adds content to the file `sizedistribution_help.rst` and affects nothing else.

Fixes #3260 

## Review Checklist:

**Installers**


butlerpd:
The screen shots are not added, partly because there are some bugs left that will change those (e.g. diameter entry is listed in units of 1/A etc). However the text and references are now done in a first pass and can be reviewed. Hence removing the "draft" designation.

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+intensity in this case is given by the integral of the scattering intensity
+from every size present in the system. Basically a polydispersity integral of
+the monodisperse I(Q) but where the shape of the distribution can be anything
+and is most likely not possible to put in a parameterized analytical form. Thus
+the equation to fit is:
+
+.. math::
+    I(Q)= \Delta \rho^2 \int N_p(r) P(Q,r) dr
+
+Where $N_p(r)$ is the number density of particles of characteristic dimension
```
backmari, line 26:
```suggestion
Where :math:`N_p(r)` is the number density of particles of characteristic dimension
```

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+r, and P(Q,r) is the form factor for that characteristic dimension. For a
+sphere this translates as:
+
+.. math::
+    P(Q,r) = \left[
+        3V_p(\Delta\rho) \cdot \frac{\sin(Qr) - Qr\cos(qr))}{(Qr)^3}
```
backmari, line 32: Should the `qr` also be `Qr`?
```suggestion
        3V_p(\Delta\rho) \cdot \frac{\sin(Qr) - Qr\cos(Qr))}{(Qr)^3}
```

butlerpd: good catch! yes clearly

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+    I(Q)= \Delta \rho^2 \sum_{r_{min}}^{r_{max}} N_p(r) P(Q,r)
+
+Even so, fitting is over determined, particularly given the noise in any real
+data. Essentially this is an ill posed problem. In order
+to provide a reasonably robust solution a regularization technique is generally
+used. Here we implement only the most common MaxEnt (Maximum Entropy) method
+used for example by the famous CONTIN algorithm used in light scattering.
+
+.. note::
+    The assmumptions inherent in this method are:
```
backmari, line 63:
```suggestion
    The assumptions inherent in this method are:
```

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+
+Select a dataset and use the *Send To* button on the *Data Explorer* to load
+the dataset into the *Size Distribution* panel.
+
+INSERT IMAGE HERE!!!
+
+This will open the panel on the *Parameters* tab and plot the data to fit.
+The most important parameters to adjust at this point are the minimum and
+maximum diameters of the distribution. The calculation will only explore
+diameters in this range. It is important that the range not excessivly exceed
```
backmari, line 113:
```suggestion
diameters in this range. It is important that the range not excessively exceed
```

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+the limits imposed by the $Q$ range being fit.
+
+If the data is on absolute scale and a quantitave volume fraction is desired the
```
backmari, line 116:
```suggestion
If the data is on absolute scale and a quantitative volume fraction is desired the
```

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+contrast factor between the domain of interest and the matrix must be set
+correctly. Otherwise, the value is not important and can be left alone.
+
+The number of bins sets the number of sizes within the size range that will be
+calculated.
+
+The models section default is usually appropriate. Currently only the ellipsoid
+model is implemented. This may be expanded in the future to include cylinders
+for example. The aspect ratio for the ellipsoid however can be changed. The
+default of 1 is for a sphere. An aspect ration greater than 1 will yield a
```
backmari, line 126:
```suggestion
default of 1 is for a sphere. An aspect ratio greater than 1 will yield a
```

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+prolate ellipsoid while a value smaller than 1 is for an oblate ellipsoid.
+
+.. Warning::
+   The Size Distribution analyis assumes the data is properly background
```
backmari, line 130:
```suggestion
   The Size Distribution analysis assumes the data is properly background
```

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+*Quick fit*. The only difference between the two is that the first will only
+run the calculation once and produce the result.
+
+IMAGE INSERTED HERE OF QUICK FIT
+
+After a short time, the graph will be updated with the fit to the data using
+the resulting distribution, while a second plot will pop up showing the final
+distribution of sizes that are returned, giving the volume fraction (true or
+relative depending on whether the data are on absolute scale or not) of each
+size. Finally the *Ouput* section of the *Paremeters* tab will show the
```
backmari, line 146:
```suggestion
size. Finally the *Output* section of the *Parameters* tab will show the
```

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
```diff
@@ -5,15 +5,274 @@
...
+results including whether or not the fitting converged, the unormalized
```
backmari, line 147: normalized? unnormalized?

backmari (changes requested):

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
butlerpd: A good question. Probably normalized but need to look at the code. I think I thought it was unnormalized because of the very high chi^2 I was seeing but actually playing around with it more I think that it may in fact be a correct value for normalized!  Will check and correct if appropriate. Thanks!

yunliu01 (commented):
The help document is well written.
There are a few minor suggestions. I have discussed with @butlerpd directly about addressing them down the road. I therefore approved this before the changes are made.

A few places indicate that images will be inserted.
I guess that because the interface has been changing, those images are not yet inserted, but will be added later before finalizing the document.

Also, the role of weight factor adjustment is not very clear in the current draft.
After talking with @butlerpd , it seems that some change still needs to be made in the code for the weight factor. I feel that it may be useful to provide a little bit more discussion how the weight factor influences the calculation of Chi^2 after the final changes are made.

At the beginning of the document, P(Q,r) is used. But it may be useful to explicitly define r  as it is not clearly indicated in the current draft.

Also, for the equation of maximum entropy, it is useful to have the index, i, for the summation symbol.

Also, when define Ecc=Rpolar/Req, there are typos so that some characters do not appear to be a subscript.

yunliu01 (approved):

src/sas/qtgui/Perspectives/SizeDistribution/media/sizedistribution_help.rst
butlerpd: And the answer is..... in fact what seems to be returned is the unnormalized chi^2 though it is the normalized chi^2 that is used to test for convergence. I do not understand why they do this? It would be pretty easy to return the normalized chi^2 I think.  For now I will leave the documentation like this.

However, if and when we change the behavior we should not forget to change the documentation!

PR #3375, opened 2025-05-12
Fix bumps constrained fit

### LINKED ISSUE #3370: Error when running constrained fit
**Describe the bug**
#3366 allows the constrained fit window to launch after part one of the fitting refactor, but fits are now failing with the following error. This is likely related to bumps v1 attempting to coerce parameter values to floats, where we have always sent uncertainty quantities. The fix is to set `Parameter.slot` to the uncertainty quantity and `Parameter.val` to the raw floating point value.

>ERROR: Fitting failed: Traceback (most recent call last): File "sas\qtgui\Perspectives\Fitting\FitThread.py", line 17, in map_getattr return_value = getattr(classInstance, classFunc)(*args) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "sas\sascalc\fit\BumpsFitting.py", line 315, in fit problem.setp_hook() File "sas\sascalc\fit\BumpsFitting.py", line 232, in call self.update() File "M2.background=M3.background M2.scale=M3.scale M2.radius=M3.radius M1.radius=M2.radius M1.background=M2.background M1.scale=M2.scale", line 11, in eval_expressions File "bumps\parameter.py", line 451, in value return int(self.slot) if self.discrete else float(self.slot) ^^^^^^^^^^^^^^^^ File "uncertainties\core.py", line 2611, in raise_error raise TypeError("can't convert an affine function (%s)" TypeError: can't convert an affine function () to float; use x.nominal_value

**To Reproduce**
1. Launch sasview using an installer or code from #3366
2. Set up a series of fits and a constrained fit
3. Run the constrained fit
4. Wait for the first update
5. See error

**Expected behavior**
The fit should continue without errors.

 - Version: v6.1.0a1 with bumps v1.0.0b6

bmaranville:
## Description

Allows running constrained fits with new bumps version >= 1.0.0b6
Fixes #3370

## How Has This Been Tested?

Ran a test constraint fit with the fix, and it works without the error described in #3370 
**NOTE**: the convergence plot doesn't seem to work when running a consrtrained fit, which is I think a different issue.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1850,10 +1850,11 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         Given the fit results structure, pull out optimized parameters and return them as nicely
         formatted dict
         """
+        pvec = [float(p) for p in results.pvec]
         if results.fitness is None or \
             not np.isfinite(results.fitness) or \
-            np.any(results.pvec is None) or \
-            not np.all(np.isfinite(results.pvec)):
+            np.any(pvec is None) or \
+            not np.all(np.isfinite(pvec)):
             msg = "Fitting did not converge!"
             self.communicate.statusBarUpdateSignal.emit(msg)
             msg += results.mesg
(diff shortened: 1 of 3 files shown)


krzywon (commented):
This allows constrained fits to run again, but logger warnings are now triggered on line 360 of `BumpsFitting.py` complaining the `slot` object for constrained parameters are no longer `uncertainty` objects. A quick debug says they are now `bumps.parameter.Variable` objects.

bmaranville:
Ah, so this doesn't fix the problem.

PR #3376, opened 2025-05-12
Wait to update Size Distribution data plot until user presses Enter

### LINKED ISSUE #3277: Size Distribution GUI should wait for a <cr> before updating plots
when playing with background values, each recalculation takes a bit of time. It often happens that one might want to change several things at once. Having to wait after each number is entered can become annoying.  This was dealt with in the fitting perspective by making <cr> after entering a value force a compute while simply clicking away into another box only updates the value but does not force a recompute. Something similar here would be good

backmari:
## Description

- Remove model `dataChanged` connection
- Install an event filter for the background input widgets that should update the data plot. The event filter ensures that the plot is not updated until the user presses Enter.
- Update the background data and plot when the background input widgets are updated programmatically, e.g. after fitting the background.

Fixes #3277 

## How Has This Been Tested?

Tested locally in the dev environment by adding data to the Size Distribution perspective and verifying that the data plot does not get updated until the user presses Enter in one of the background input boxes or changes the state of the checkbox. The plot is not updated if the user deletes the text in the input box.

image

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
+++ b/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
@@ -171,6 +171,15 @@ class SizeDistributionWindow(QtWidgets.QDialog, Ui_SizeDistribution, Perspective
         # Local signals
         self.fittingFinishedSignal.connect(self.fitComplete)
 
+        # Event filters for plot background update
+        background_update_widgets = [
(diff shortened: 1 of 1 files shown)


butlerpd (approved):
Nice! Code LGTM (but what do I know) and certainly works as advertized, including the fact that the Q range boxes still update the sliders immediately as they should.

Ready to merge IMO

PR #3379, opened 2025-05-13
3288: Allow reparameterized parameters to be edited

### LINKED ISSUE #3288: Reparameterization editor: No way to edit a parameter once created
'The Edit Selected button actually does NOT allow editing the new parameter created. Instead it behaves exactly as "add a new parameter" (the plus). There is no way to edit a parameter once created.' @butlerpd

krzywon:
## Description

This populates the parameter editor widget when selecting edit parameter and updating the widget in the reparameterization editor.

Fixes #3288

## How Has This Been Tested?

Tested locally. Seems to work for all values in the window.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/ModelEditors/ReparamEditor/ReparameterizationEditor.py
+++ b/src/sas/qtgui/Utilities/ModelEditors/ReparamEditor/ReparameterizationEditor.py
@@ -198,12 +198,11 @@ class ReparameterizationEditor(QtWidgets.QDialog, Ui_ReparameterizationEditor):
         
         # Find the parameter item by using param_to_open and format as a dictionary
         param_properties = self.getParamProperties(param_to_open)
-        param_properties['highlighted_property'] = highlighted_property
         self.onAddParam()
         # Populate the editor with the parameter values
         if self.param_editor:
             self.param_editor.properties = param_properties
-            self.param_editor.qtree_item = selected_item
+            self.param_editor.qtree_item = param_to_open # Item to edit is always the parent of the selected_item!
             self.param_editor.onLoad()
             self.param_editor.returnEditedParamSignal.connect(self.updateParam)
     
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Utilities/ModelEditors/ReparamEditor/ReparameterizationEditor.py
```diff
@@ -199,8 +199,14 @@ def editSelected(self):
         
         # Find the parameter item by using param_to_open and format as a dictionary
         param_properties = self.getParamProperties(param_to_open)
-        param_properties['highlighted_property'] = highlighted_property  # TODO: Which property the cursor will start on
+        param_properties['highlighted_property'] = highlighted_property
         self.onAddParam()
+        # Populate the editor with the parameter values
+        if self.param_editor:
```
butlerpd, line 205: Not clear to me that self.onAddParam should be called here if we are meant to just be editing an exisiting paramewter? In fact testing the installer does do something hybrid: it opens a parameter editor Window which is now populated with the values from the highlighted parameter but the result is to add all those parameters as a sub parameter of the same name!  In other words, if I create a parameter named Re and enter values then come back to edit it, the new parameter box will now show the old settings plus a NEW setting, named Re which itself has a bunch of unnamed settings but that contain the new values.

butlerpd (changes requested):
Aside from the fact that it still does not work, I note that the editing behavior of the pop up editor is quite different from all the other panels in SasView and thus quite confusing.  That should be a separate issue though.

src/sas/qtgui/Utilities/ModelEditors/ReparamEditor/ReparameterizationEditor.py
krzywon: @butlerpd, `self.onAddParam` is probably a misnomer. The method generates the param editor window and not much else. I am looking into the other issue of creating a separate tree inside the original tree.

butlerpd:
Fixed last two (known) issues with the reparameterization widget.
* First, the model help button would throw an error if no model was loaded but then load the general help. Moved the show help to the right branch of the conditional so no error is thrown.
* The big one was to do with editing new parameters after they were created. Clicking on the description tag for example would bring up the parameter editor but if you changed some other values, as a user would be want to do, it then sometimes added those changes as new subitems to the description without changing the actual value of interest.  Other funny issues as well unless you clicked on the main parameter name. This two line change basically ignores which subitem was clicked on since it is completely irrelevant anyway.

This was tested on the development windows 11 machine and various combinations of editing parameters were tried without any problems, and the final saved model works.

butlerpd (approved):
removing my change request ... since it is now fixed

PR #3386, opened 2025-05-19
Customize plot needs a toggle for scientific notation

### LINKED ISSUE #3197: Customize Plot Needs a Toggle for Scientific Notation
SasView 6.0.0 introduced the ability to customize the plot axis labels which is a great addition! However, when ones chooses to increase the font size, as one might want to do for presentations or graphs for papers, the default of using the full number rather than scientific notation (i.e. 1000 instead of 1E3) becomes unwieldly for plots that range over too many orders of magnitude. Ideally, it would be helpful to add a checkbox to allow scientific notation instead.

3189
Add a checkbox to allow scientific notation vs the regular notation in the new dialog.

none - the above seems the simplest - but if there are better solutions that would be fine

**Additional context**

Image

rozyczko:
## Description

Added checkboxes for scientific notation on plot axes.

Fixes #3197

## How Has This Been Tested?

Local Win10

## Review Checklist:

**Installers**


butlerpd:
Cool.  I tested the pull request installer (as opposed to the push installer - is there a difference?) on a windows 11 machine. the logic seems to all work however the scientific notation being used seems different from the scientific notation being used by the default plotter - see attached figure. I am assuming there are different methods in matplotlib?

The right plot is the default when creating a new plot. The left is after using customize labels and asking for scientific notation.

SciNotationIssue

butlerpd:
So the correct form we need I think is the `LogFormatterSciNotation` class rather than the `ScalarFormatter` in order for the log scale to render "properly" (i.e. the way the default log plots render). So for example
```Python
self.ax.yaxis.set_major_formatter(ticker.LogFormatterSciNotation(labelOnlyBase=True))
```
I tested this locally and it works.  Question now is whether we should just automatically do this for log scale and use the `ScalarFormatter` for linear which would greatly simplify the changes needed and remove extra checkboxes or keep as it is but maybe change the option to read "log scale formatting" instead of "scientific notation." 

I've submitted a new PR with the first option. The second could be done by editing this PR but in this case we should change all the nomenclature from `scientific` to `logscale` or some such?

PR #3390, opened 2025-05-20
Poresize issue3280

achicks15:
## Description

The merge fixes issue 3280 for the Size Distribution perspective. It fixes the background power law and linear background fits not using the provided power law to fit. It also corrected the typo in the GUI perspective window. 

Requirements: I added the scipy.optimize package to do the fitting. 

Fixes # 3280

## How Has This Been Tested?

This was tested by successfully completing a quick and full max entropy fitting when subtracting both a power law and background. It was also tested in a local testing script with the same data. 

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/sascalc/size_distribution/SizeDistribution.py
+++ b/src/sas/sascalc/size_distribution/SizeDistribution.py
@@ -44,8 +44,13 @@ def add_gaussian_noise(x, dx, seed=None):
 
     return noisy_data
 
-def line_func(x,m,b):
-    return b+m*x
+def line_func(x, b, m):
+    ## y=A*x^m
+    ## ln(y) = ln(A) + m*ln(x)
+    ## b = ln(A)
+    ## m = m
+
+    return b + m*x
 
 def background_fit(data, power=None, qmin=None, qmax=None, type="fixed"):
     """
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionLogic.py
```diff
@@ -90,15 +90,15 @@ def fitFlatBackground(self, qmin, qmax):
         """
         Estimate the flat background
         """
-        background = background_fit(self.data, None, qmin, qmax)
-        return background[1]
+        background, background_err = background_fit(self.data, 0, qmin, qmax)
+        return np.exp(background)[0]
```
krzywon, line 94: I think ` return self.fitBackgroundScale(0, qmin, qmax)` would be easier to maintain, long-term. If the return of `background_fit` ever changes, it will only need to be updated in one place instead of two.

src/sas/sascalc/poresize/SizeDistribution.py
```diff
@@ -79,22 +83,37 @@ def background_fit(data, power=None, qmin=None, qmax=None, type="fixed"):
...
+        # Linearize the data for a linear or power law fit
+        
+        linearized_data = Data1D(np.log(data.x[idx]), np.log(fx[idx]), dy=sigma[idx]/fx[idx])
+        fit_func = lambda x,b:line_func(x, power, b)
+        init_guess = (linearized_data.y[0])
+
     else:
-        linearized_data = Data1D(data.x[idx], fx[idx], dy=None)
+        # Fit both the power and scale 
+        linearized_data = Data1D(np.log(data.x[idx]), np.log(fx[idx]), dy=sigma[idx]/fx[idx])
```
krzywon, line 95: This is duplicated inside both cases of the if/else block and can be moved out to before the if statement.

src/sas/sascalc/poresize/SizeDistribution.py
```diff
@@ -79,22 +83,37 @@ def background_fit(data, power=None, qmin=None, qmax=None, type="fixed"):
...
+        fit_func = line_func
+        init_guess = (4, linearized_data.y[0])
+
+    param_result, pcov = optimize.curve_fit(fit_func, linearized_data.x, linearized_data.y, init_guess, sigma = linearized_data.dy)
+    param_err = np.sqrt(np.diag(pcov))
 
+    if len(param_err) > 1: 
+        param_err[1] = np.exp(param_result[1])*param_err[1]
+    else:
+        param_err[0] = np.exp(param_result[0])*param_err[0]
```
krzywon, line 105: If the numpy error for the parameter error has at least two elements, only modify the 2nd element, otherwise only modify the first element? Why is this happening? Does the consumer of this error array know what to do with it?

Please add inline documentation for why this is necessary.

src/sas/sascalc/poresize/SizeDistribution.py
```diff
@@ -79,22 +83,37 @@ def background_fit(data, power=None, qmin=None, qmax=None, type="fixed"):
...
-    residuals = linearized_data.y - (result.slope * linearized_data.x + intercept)
-    sigma = np.sqrt(np.sum(residuals**2) / (n - 1))  # Sample standard deviation
+    #result = stats.linregress(linearized_data.x, linearized_data.y)
+    #intercept = np.mean(linearized_data.y - result.slope * linearized_data.x)
+    #n = len(linearized_data.x)
+    #residuals = linearized_data.y - (result.slope * linearized_data.x + intercept)
+    #sigma = np.sqrt(np.sum(residuals**2) / (n - 1))  # Sample standard deviation
     #std_dev_intercept = sigma * np.sqrt(np.sum(linearized_data.x**2) / (n * np.sum((linearized_data.x - np.mean(linearized_data.x))**2)))
     #mean_value = np.mean(numbers)
     #std_dev = np.std(numbers)
```
krzywon, line 115: Is there any reason to keep these lines?

krzywon (commented):
I've reviewed the code. Have not tested functionality.

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionLogic.py
butlerpd: Humm... would it not be easier to just call from SizeDistribution.py? instead of calling fitFlatBackground or fitBackgroundScale just call background_fit in sascalc directly?  

But maybe that is harder to understand? Not good coding practice? simpler would seem better to me?  I suspect this was set up before we really knew how it was all going to work? Just a thought

butlerpd (changes requested):
Functionality test on windows 10 works beautifully!!!

 However, besides some coding questions, release-6.1.0 will need to be merged into this branch locally because of the change of folder names (and method names). Another reason we wanted those changes sooner than later.  We will have to check again carefully after that merge to make sure nothing got accidentally changed before we merge this PR.

src/sas/sascalc/poresize/SizeDistribution.py
achicks15: No

src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionLogic.py
butlerpd: I think what @krzywon is suggesting is that `fitFlatBackground` and `fitBackgroundScale` have the same code and it would be better to unite them (basically the caller should pass the 0 I guess?

butlerpd:
actually it looks like some of those issue are already addressed so probably mergeable now unless you want to unite the two fitBackground methods?

krzywon:
I think, based on the recent changes and the desire for @butlerpd to have this change in order to work on another issue, I think my comments can be handled in a separate branch. I'm going ahead and merging this.

Issue #3278, opened 2025-04-29
Speed up startup of Size Distribution perspective


butlerpd:
It currently "takes a minute" for the size distribution to load when sending data to it. This is probably due to the fact that a default background is being subtracted and a new curve generated. I think that background subtraction makes no sense anyway till the user choose what they want. It would probably be faster and make more sense in a workflow anyway to just plot the data when sending to the perspective.

This could, maybe (?), have the side benefit of getting rid of the double plotting of the data?

butlerpd:
This was probably a bit of a misnomer. While not first tier issue, it is still more of a bug than a feature in that subtracting a background before the user even decides whether or not they want to do that, much less what that background should be **if** in fact they do want that, is not correct behavior.

PR #3395, opened 2025-05-20
Speed up startup of size distribution

backmari:
## Description

**Not working, do not merge**

Currently when data is loaded in the Size Distribution Perspective, we also plot a flat background at 1e-6. However, we would initially like to plot only the data and add background only if the user adds values in the background configuration.

This was a first attempt at only plotting the data, however, the logic called when emitting the signal `plotRequestedSignal` expects a list with at least two items, where the second probably should be a `Plottable` object. I tried passing in a dummy plot, but didn't like that solution.

Related to #3278

## How Has This Been Tested?

## Review Checklist:

**Installers**


butlerpd:
Seems like what we need to do is to replace the call to `self.plotData` at the end of the `setData` method with a different call? One that directly accesses `MainWindow.DateExplorerWindow.plotData` perhaps? The requestplot signal is I think essentially what we want but that works for a tabbed perspective only I think?

Anyway .. just some thoughts that may or may not be helpful from things I've stumbled across working on other PRs

PR #3397, opened 2025-05-20
Second try at fixing custom labels

### LINKED ISSUE #3197: Customize Plot Needs a Toggle for Scientific Notation

butlerpd:
This is a second attempt at fixing the problem reported in #3197. As discussed in PR #3386 comments, it turns out that using the log formatter should automatically take care of the issue with the tick labels on log scale. It also turns out that the linear scale has a similar problem. This is fixed by using the scaler formatter as in PR #3386.

## Description

This is a much simpler fix that implemented in PR #3386. It just adds the tick label formatting to the `onCustomizeLabel` method.
 
Fixes #3197

## How Has This Been Tested?

tested in the local dev environment of windows 10 machine. tried both x and y, both linear and log scales.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -5,6 +5,7 @@ import functools
 import copy
 import math
 import matplotlib as mpl
+import matplotlib.ticker as ticker
 import numpy as np
 import textwrap
 from matplotlib.font_manager import FontProperties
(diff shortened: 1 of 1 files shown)


butlerpd:
Note:  I think this is the best option to include in 6.1.0 as it is the least invasive to the existing code given how close we are to a release?

krzywon (approved):
Tested locally and the changes look fine. I think, for v6.1.0, this is better than #3386. The work in #3386 has a place, but is more of a feature than a bug fix.

jamescrake-merani:
In that case, I'm going to merge this PR.

PR #3399, opened 2025-05-20
Fix plot bounds

### LINKED ISSUE #3398: plot scaling is (mostly) broken
**Describe the bug**
Presumably due to bugs in matplotlib's autoscaling for some edge cases causing it to be massively off, sas.sascalc has implemented a manual override for all `Data1D` plots in a private method, `_plot_bounds`, of `plotter.py`.  The problem is that it is implemented for a linear scale only while most of the SasView plots are on log scales.

The result is that most log scale lower limit is set to a negative value.  Rather than throw an error, matplotlib appears to quietly convert it into a positive number. For many of our plots which start at ~ 0 that value does not lead to a glaring visual issue. However, the recently added size distribution perspective has a plot for which it becomes obvious on the x axis. For some reason, for typical scales which start in the 20 to 80 range and go up to 5 or 10k, up to 25% of the plot can be below the lowest data point in x.

Besides the new perspective needing it to be correct, it would seem to me that, to the extent that we are trying to bypass the auto scale of matplotlib, we should actually bypass it not revert to a random auto scale from matplotlib?

**To Reproduce**
Print statements can be placed appropriately in the `_plot_bound` method or a debugger can be used to examine the values.  Also, the size distribution perspective in the current version of the 6.1.0 release branch (pre alpha 2) can be run. the volume distribution has the diameter on the x axis but in log scale. In most cases, the volume distribution plot that appears after a fit will have a lot of white space below the lowest x value.

**Expected behavior**
In all cases a reasonable (5 to 10% of the plot range) should be left outside the data limits but no more. Further, the "white space" should be ~ evenly distributed around all four edges of the plot.

**Additional context**
I can think of two approaches to fix this.
1. Use an `if axex.get_scale == "log` section to compute the padding differently. This would probably be the easiest and quickest and possibly the most robust? But it is not clear that it would necessarily work for any scaling (matplotlib has a few others I think?).
2. use axes coordinates rather than data coordinates to compute the padding. This has the advantage that it should be universal (regardless of scale) and should provide exactly the same display "white space" on all sides. The disadvantage IMO is that I'm not sure how robust using all the transforms back and forth will be (or perhaps more accurately, how well we understand that complexity when implementing this).

butlerpd:
## Description

This implements the second suggestion in issue #3398.  The `_plot_bounds` private method of `plotter.py` was changed by converting the min and max data values to axes coordinates before calculating the padding and thence the desired plot bounds which are then converted back into data coordinates before being returned.
 
Fixes #3398 

## How Has This Been Tested?

It was tested on the local development windows 10 machine. verified that normal linear and log plots work, all log values are positive, and the size distribution plot, which revealed the issue in the first place also seems to work. 

## Review Checklist:

NOTE: actually the doc string for the `_plot_bounds` method was extensively updated and a few in line comments added to the code

**Installers**


backmari:
I did a quick test with `Alumina_usaxs_irena_input.csv`. I noticed that if I add a flat background, e.g. 0.12 and hit Enter, the y scale max is changed to 1e58 (!). Do you see the same?

krzywon:
I'm seeing the samething, @backmari. The chart below is what appeared when I clicked `Create New` from `P123_D2O_40_percent.xml` while using this branch.

image

butlerpd:
Hummm.... thanks! That is very odd. my testing was indeed with the Alumina data and it worked fine both for creating plot, sending to the data and then the linear y scale used for the distribution.  I'll need to investigate further

butlerpd:
It turns out that there was a deeper problem with log scales. The original problem identified here was that by subtracting a fixed amount the lower limit on a log scale can go negative which happens quite regularly and needed to be addressed.  This PR fixes that problem.

However, as shown here, the data **itself** can have negative values which will make the min value of the data negative even before subtracting anything for padding! Moreover, in this case, simply using the log scale approach of using 0.95 * data_min will not work either since that will still be negative. This is a harder problem to fix and will require finding the the smallest value of the data that is still positive IF we are on a log scale before applying any padding (whatever method we choose to use).

I will try to add that check keeping this method as the check will be necessary no matter what.

src/sas/qtgui/Plotting/Plotter.py
```diff
@@ -331,46 +331,101 @@ def has_nonempty_plots(self) -> bool:
...
+        # sure the error bars stay within the plot.
         for key in self.plot_dict:
 
             plot_data = self.plot_dict[key].view
 
+            # NOTE: we need to trim any negative values if the scale of the
+            # particular axis is log scale before looking for the minimum
+
+            # First the x axis
+            plot_data.x = [i for i in plot_data.x if i > 0] if self.ax.xaxis.get_scale() == "log" else plot_data.x
```
backmari, line 380: I tested a data set that included negative x and y values, and when using log scale the negative values had already been removed at this point, but I guess an extra check doesn't hurt.

backmari:
I tested locally with data sets with and without negative values and with and without dy data and all works great (at least in the Size Distribution perspective). The size distribution plot is now zoomed in to the right area.
If there is a specific GitHub issue for the matplotlib defect it could be good to add a link to that in the function docstring, but it's not a must have.

src/sas/qtgui/Plotting/Plotter.py
butlerpd: Interesting ... While I never tested x, the y axis problem we were seeing was due to negative values coming through? Perhaps it was to do only with the error bars? I can try checking that I guess

butlerpd:
Good catch @backmari. I have now checked all those permutations and you are correct that the `plot_data` automatically adjusts to only return positive value when on an a log scale but will add them back automatically if we change the scale from log to linear. Sweet!  I tried a slicer cut and saved the data with fold = off so that the data had negative x as well as positive. The number of x points is only 49 when plotted log log but as soon as I change scale to linear linear the number of points when it hits `plot_bounds` is 98 with all the negatives back. Likewise using apoferritin.

So the negatives I was seeing were all coming from y-dy!

My instinct would be to remove all the pointless lines. In fact, the only line needed after my first commit is the current lines 397-401. x does not need any checking so line 381 can be removed as can lines 393 and 406.

However, @rozyczko and @backmari you have a lot more software development experience than I. Would this be your recommendation also or is it somehow good form to include those lines for completeness even though there is no way they will be needed (IMO)

backmari:
I would remove lines 381, 393 and 406, since it looks like they are not needed and also because they may manipulate existing data structures. If you want to keep them I would assign the trimmed list to another variable, instead of updating `plot_data.x` and `plot_data.y`.

butlerpd:
Excellent point Marie! Had not considered that. Changes made and comments updated to reflect. Also, while I found some older stack overflows complaining about this scale problem while I was trying to figure out how to fix it, I cannot seem to find them now. On the other hand the reason it exists apparently is because of an example (it is rare of course) that came from a summer intern a few years ago issue #2904 and fixed in PR #2967. I'm not sure when the code was updated again to move things into `_plot_bounds`, but it seems like there have been numerous tweeks over time just looking at git blame. Anyway, I adjusted the docstring slightly as well.

Hopefully this is now ready to be approved and merged?

PR #3402, opened 2025-05-21
3309: Shape2SAS plotting enhancements

### LINKED ISSUE #3309: 3D plot aspect ratio needs tweaking (spheres look like ellipsoids)

krzywon:
## Description

This fixes the issue related to the 3D plot in the Shape2SAS window where spheres looked liked ellipsoids. I also added a few tweaks to the plots, mostly to allow them to expand as the window is expanded.

Fixes #3309

## How Has This Been Tested?

Locally, the aspect ratios for all of the 3D plot axis remains the same.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Calculators/Shape2SAS/ViewerModel.py
+++ b/src/sas/qtgui/Calculators/Shape2SAS/ViewerModel.py
@@ -44,10 +44,10 @@ class ViewerModel(QWidget):
         self.dict_series = {"Red": self.seriesRed, "Green": self.seriesGreen, "Blue": self.seriesBlue}
        
         self.scatterContainer = QWidget.createWindowContainer(self.scatter)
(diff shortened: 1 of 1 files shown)


butlerpd (approved):
The code looks reasonable to me and does add the ability to stretch and keep the aspect ratio.  Tested the pull request installer on a windows 11 machine and it works and the sphere is no longer flattened.

However, a very odd issue reveals itself when I had a cylinder of length 200 and offset in X by 100.  When doing this with the original version the sphere now looks spherical, however, in this PR it no longer looks spherical. In fact, looking end on both the cylinder diameter and sphere look squashed?

I am going to approve this as an improvement because of the stretching feature but with respect to the original issue of flattening a circle, that seems to have just moved the problem somewhere else?

PR #3407, opened 2025-05-22
3299: Include effective radius modes in model template

### LINKED ISSUE #3299: New version of ER that is required for use with C models
'Also there should be the new version of ER that is required for use with C models which should also be all commented out by default with explanation.' - @butlerpd

krzywon:
## Description

This adds the list for `radius_effective_modes` to the python plugin model template to allow for different ER calculations.

Fixes #3299

## Review Checklist:

### DIFF
--- a/src/sas/qtgui/Utilities/ModelEditors/TabbedEditor/TabbedModelEditor.py
+++ b/src/sas/qtgui/Utilities/ModelEditors/TabbedEditor/TabbedModelEditor.py
@@ -1002,6 +1002,12 @@ description = """{description}"""
 ER_VR_TEMPLATE = '''\
 # NOTE: If you want to couple this model with structure factors (S(Q)), please uncomment this section. This
 #     function will need to return a meaningful value to enable full structure factor compatibility.
+# 
+# The modes in which the effective radius can be applied. This list allows arbitrary values, but the index of the mode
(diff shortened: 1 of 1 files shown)


butlerpd:
After discussion with @pkienzle it turns out the `VR` method is no longer supported at all. However the `shell_volume` method is supported in both python and C models.  The templates need to thus be adjusted for this fact.

butlerpd:
After speaking to Jeff, I finished updating the templates based on above comments and a few other things discovered as I was reading everything even more carefully during edits. I think this should now be ready to be merged but since I made the final commits I leave it for someone else to merge.

src/sas/qtgui/Utilities/ModelEditors/TabbedEditor/TabbedModelEditor.py
```diff
@@ -1048,69 +1082,101 @@ def form_volume({args}):
...
-{{
-    switch (mode) {{
-    default:
-    case 1:
-    // Define effective radius calculations here...
-    return 0.0;
-    }}
-}}
+// uncomment this and provide appropriate functions for the effective
+// radius to be used IF this is to be ussed with structure factors.
```
rozyczko, line 1169: ussed --> used

butlerpd:
Thanks @rozyczko! seems I cannot spell so well -- but @smk78 could have told you that 😄

PR #3424, opened 2025-05-28
Power law background estimation in Size Distribution

### LINKED ISSUE #3279: Add option to power law background estimation in size distribution perspective
The background estimation tool currently allows for flat background and the scale for a power law of 4 background estimation. This is the most common need. However, there may be occasion where a different power law is required and the user may want to also fit the power.

Two obvious options are:
* Add a second button to the power law background estimation just like Irena: scale only and scale and power
* Add a checkbox for whether or not the power law gets fit when clicking the one button for power law fitting.

backmari:
## Description

Enable in the GUI fitting of the power in the power law background estimation.

Fixes #3279

## How Has This Been Tested?

Tested locally with `src/sas/sascalc/size_distribution/test_data/Alumina_usaxs_irena_input.csv`.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionLogic.py
+++ b/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionLogic.py
@@ -1,3 +1,4 @@
+import logging
 from typing import List
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


butlerpd:
WOW ... I was going to say that we should list it as a future issue, I suspect it is not caught in some othe rpespectives as well. Of course here, since we cannot use the sliders, the problem is a bit worse since you cannot see how many points are between the ghost sliders 😄.

Anyway .. I see you have made a fix. I'll test it next I guess. Thanks!

butlerpd (approved):
Right ... this looks only perfect!  The implementation looks good to me. The error is returned to the logfile/logwindow by sascalc as it should be along with the traceback error which is captured by the caller in a try/except loop. 

Tested the installer on windows 11 for all combinations .. even selecting a range outside the data range. Worked like a charm. Going to merge this as ready

PR #3429, opened 2025-05-28
make trust bars red and make the fit plot a line instead of points

### LINKED ISSUE #3428: Size Distribution plot issues
From first user review of perspective the two following items were identified:
* the fit line must be a line not a series of points - points mean data
* the trust zone bars should be a different color than the plot data

Note that the second was indeed the plan but we were not able to do this because of a bug in SasView or matplotlib that does not rotate the color if the marker symbol is "|" (vline).  Also the color attribute seemed to not work - this last bug has however been found and fixed in #3427 so this can now be addressed.

butlerpd:
## Description

Fixes #3428

## How Has This Been Tested?

Tested in local dev environment with fixes from  PR #3427

## Review Checklist:

**Installers**


backmari:
Looks good! The fit is now shown as a line on top of the data plot and the trust bars are red.
image

backmari (approved):

PR #3431, opened 2025-05-29
Reset window when sending new data to the Size Distribution

### LINKED ISSUE #3423: Sending new data to the size distribution perspective needs to clear everything out from the previous data (reset?)
**Describe the bug**
When evaluating the size distribution for more than one data set, sending the 2nd set to the perspective keeps vestiges of the previous one. In particular the fit data is not cleared. Worse it gets reassigned as a subplot of the new data set. Likewise all the options remain from the previous setting (including the dy choice).  Clicking Fit, recalculates the fit curve but the graph x scale still seems to remember the previous fit data.  This is bes seen when running a USAS data set followed by a SAS data set (very different q ranges)

Things can be reset by deleting the data from SasView before sending the next data set. This is not a good solution.

**To Reproduce**
1. Load the test Irena USAXS data set
2. send to the Size Distribution perspective
3. Run a quick fit
4. now load something like the`cyl_400_20.txt` data from the `example_data` (any other SANS q range data should work)
5. Send to the same perspective
6. See error
 
**Expected behavior**
IMO, the state of the perspective should be re-initialized prior to sending the new data set to it.

 - Version: 6.1.0a2+ (i..e. latest 6.1.0 branch)

backmari:
## Description

Reset window when sending new data to the Size Distribution. The existing plots for the old data remain, is that the expected behavior?

Fixes #3423

## How Has This Been Tested?

Tested locally using the steps in issue #3423

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
+++ b/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
@@ -439,6 +439,10 @@ class SizeDistributionWindow(QtWidgets.QDialog, Ui_SizeDistribution, Perspective
             msg = "Incorrect type passed to the Size Distribution Perspective"
             raise AttributeError(msg)
 
+        if self.logic.data_is_loaded:
+            # remove existing data and reset GUI
+            self.resetWindow()
+
         self._model_item = data_item[0]
         logic_data = GuiUtils.dataFromItem(self._model_item)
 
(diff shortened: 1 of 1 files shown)


butlerpd (approved):
Since we are not using tabs, the `removeData` function needed to be called to clear out the perspective but without the initial if statement.

Tested the installer on windows 11 and it works cleanly now.

PR #3434, opened 2025-05-30
Updating the size Distribution tests

butlerpd:
Cleaned up code and verified it all runs properly. Use values and data that converges and uncommented the convergence test. Test now actually does something (tests for convergence).

By visual inspection, the resulting values agree almost exactly with the Irena plot shown in Jan's video tutorial, https://www.youtube.com/watch?v=WMd0N9_5iJo,  using the exact values used there. Unfortunately, the compare file `Alumina_usaxs_irena_diameter_fit.csv` that was uploaded in Jan 2024 does not look at all like that Irena plot.

IrenaSD

SasViewSD

NOTE: not only are the two curves nearly quantitatively similar, the trust bars match!!!

Two things to do if we have time - or later as appropriate:
* run the data through Irena, as in the video, with these exact parameters, and obtain the output csv file to use in the test. It should be a reasonable match
* Change the assert  `pytest.approx`  but using abs with the error bar list (probably errors * 2 for 95 vs 68% confidence levels) as the tolerance

## How Has This Been Tested?

run pytest on the file on my local build machine and it passes (many times in fact)

## Review Checklist:

**Installers**


butlerpd:
This can be merged once the build is finished but that will be morning for me I'm afraid.. and the inversion is still breaking the builds so hard to verify all is well?

PR #3439, opened 2025-05-30
Size distribution should report normalized chi^2

### LINKED ISSUE #3410: Size Distribution should report normalized chi^2 not the unnormalized chi^2
The `maxEnt_method` ported from GSASII uses the normalized $\chi ^2$ to test for convergence but returns the **unnormalized** $\chi ^2$ to the caller and thus is what is currently reported by the GUI. This seems wrong. The GUI should report the more useful normalized version

backmari:
## Description

This changes the maximum entropy method to return the chi2 normalized by the number of bins. The style of the convergence message is changed to red and bold if the fit returned is not converged.

image

Screenshot from 2025-05-30 12-48-20

Fixes #3410

## How Has This Been Tested?

Tested locally what the message looks like with and without converged results. The converged results now has chi^2 close to 1.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
+++ b/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
@@ -794,7 +794,7 @@ class SizeDistributionWindow(QtWidgets.QDialog, Ui_SizeDistribution, Perspective
                 converge_msg = f"Full fit converged after on average {np.mean(result.num_iters):.1f} iterations"
             self.lblConvergence.setStyleSheet("color: black;")
         else:
-            converge_msg = "Not converged"
+            converge_msg = "Not converged! Try increasing the weight factor."
             self.lblConvergence.setStyleSheet("color: red; font-weight: bold;")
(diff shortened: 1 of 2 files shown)


src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
```diff
@@ -792,8 +792,10 @@ def updateStatistics(self, result):
                 )
             else:
                 converge_msg = f"Full fit converged after on average {np.mean(result.num_iters):.1f} iterations"
+            self.lblConvergence.setStyleSheet("color: black;")
         else:
             converge_msg = "Not converged"
```
butlerpd, line 797: Am thinking @ldschmitt might like to see more cautionary note? Maybe adding to the string? Something like:
"Not converged! Proceed with extreme caution - Try increasing the weight factor"
or maybe just
"Not converged!  Try increasing the weight factor"

or as is works too

butlerpd (approved):
Looks good I think. works in the installer for windows 11 as well. If adding to the red warning of not converged seems too much I'm happy for this to be merged as is.

PR #3444, opened 2025-06-02
Shape2SAS outstanding issues

### LINKED ISSUE #3413: Shape2SAS - deleting a column throws
**Describe the bug**
Using a counter for column delete is quite unintuitive - one would expect to be able to delete a column by just selecting it (its header).
Also, the default value of `0` is incorrect, since the first column has an index of `1`.

Clicking on `Delete column` apparently succeeds but there is also an exception thrown, so I'm not sure what's the final state of the widget.

```
2025-05-23 15:28:26,736 : ERROR : sas.qtgui.Perspectives.Fitting.FittingWidget (FittingWidget.py:158) :: Traceback (most recent call last):
  File "C:\Users\piotr\projects\sas\sasview\src\sas\qtgui\Calculators\Shape2SAS\DesignWindow.py", line 270, in deleteFromVariableTable
    elif row_pos[selected_column] == row_pos[-1]:
         ~~~~~~~^^^^^^^^^^^^^^^^^
IndexError: list index out of range
```

**To Reproduce**
1. Open Shape2SAS
2. Add a model or two
3. Select a valid column
4. Click on Delete column
5. See the exception

**Expected behavior**

 - Version: 6.1.0a1

 - OS: Win10

### LINKED ISSUE #3441: Shape2SAS throws when no points from shape are included in calculation
**Describe the bug**
As noted in https://github.com/SasView/sasview/pull/3406#issuecomment-2924318585, in the Shape2SAS tool, if two shapes completely overlap and the `Exclude Overlap` is selected, the tool throws an error because one of the set of points is empty.

**To Reproduce**
1. Open the Shape2SAS tool
2. Create two overlapping sphere objects
3. Click the `Plot` button

**Expected behavior**
The empty sequence should be ignored and no error should be thrown.

 - Version: 6.1.0a2

 - OS: All

**Additional context**
Error thrown:
> ERROR: Traceback (most recent call last): File "sasview\src\sas\qtgui\Calculators\Shape2SAS\DesignWindow.py", line 471, in onClickingPlot self.viewerModel.setPlot(modelDistribution, plotDesign) File "sasview\src\sas\qtgui\Calculators\Shape2SAS\ViewerModel.py", line 178, in setPlot minx = min(minx, min(distr.x[subunit])) ^^^^^^^^^^^^^^^^^^^^^ ValueError: min() iterable argument is empty

### LINKED ISSUE #3442: Shape2SAS typos
**Describe the bug**
As noted in https://github.com/SasView/sasview/pull/3406#discussion_r2107556722, Shape2SAS errors need to be spell-checked.

**Expected behavior**
Al words should be spelled correctly.

 - Version: 6.1.0a2

 - OS: All

krzywon:
## Description

This PR fixes a few issues related to the Shape2SAS tool.

Fixes #3413 - When a column or series of columns is selected, the first column is set as the one to be deleted.
Fixes #3441 - If one or more shapes are completely within other shapes and overlap is excluded, errors are no longer thrown.
Fixes #3442 - A couple of typos are fixed in the tool.

There are outstanding questions related to the first change in `DesignWindow.py` and all changes in `Shape2SAS.py` that might still need to be addressed.

## How Has This Been Tested?

Tested locally, and the issues are fixed.

## Review Checklist:

**Installers**


src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py
```diff
@@ -203,18 +204,22 @@ def getConstraintsToTextEditor(self):
...
-                    inputName = self.getTableName(column, row)
-
-                    #create parameter list
-                    parameter = [inputName, units[enumMember], inputVal, bounds[enumMember], 
-                                 types[enumMember], f"{inputName} for column {column + 1}"]
-        
-                    toTextEditor.append(parameter)
+                    # Only create MethodType if attr is callable
+                    if callable(attr):
+                        method = MethodType(attr, OptionLayout)
```
butlerpd, line 209: This is incorrect! As far as I can tell there is **no** valid reason for `attr` to not be callable. If it is, there is an error in the caller not here. Once could opt to trap it and raise an error but I'm not sure that is that useful here? At that rate we'd be putting try excepts around the entire codebase :-). However just masking the error seems like very bad coding technique? The error should be fixed not masked by ignoring the fact that the caller is sending incorrect information IMO.

Also, because we fixed the incorrect information being sent (indexed off by one) this added if statement will currently do nothing but
* It does not fix the problem it was meant to fix
* Much worse, it will make it much harder to find the error in the future should one occurs, or, even worse mask to the point that we release something with incorrect outputs.

For those interested, the error this time was because of the index error coming from starting at 1 or staring at 0 (fixed in) that led to the boolean list of lists of whether parameters are checked on not to be off by one. Basically the first parameter of shape two was moved to become the last parameter of shape 1 which now has an extra parameter that is never addressed. If any parameter is checked in shape 2 or beyond, of course they will now be in the wrong position which would have caused an error in the code generated except that it failed here because the last parameter of the last shape does not exist (it's first parameter having been moved to be the last of the previous shape) causing the` attr` to be empty and thus not callable.

src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py
```diff
@@ -349,7 +360,12 @@ def ifFitPar(value: Union[float, str], values: list[Union[float, str]], *args, *
         row, column = args #int, int
         
         if not value == "":
-            if conditionBool[column][row]:
+            # Add bounds checking to prevent IndexError
+            if (column < len(conditionBool) and 
+                row < len(conditionBool[column]) and 
+                column < len(conditionFitPar) and 
+                row < len(conditionFitPar[column]) and
+                conditionBool[column][row]):
```
butlerpd, line 368: Is there a valid reason column and row would be out of bounds?

If there is a valid reason fine (what is it?). If not this should either be removed and allow a traceback that can lead us to the actual error or raise an exception. Masking errors seems like very bad programming form that will end up biting us hard later

src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py
```diff
@@ -605,14 +634,11 @@ def getPluginModel(self):
...
         prPoints = int(self.constraint.variableTable.prPoints.text())
         modelName = self.constraint.variableTable.pluginModelName.text()
-        parNames = self.getAllTableNames(self.ifNoCondition)
-        checkedPars = self.checkedVariables()
         parVals = self.getStandardReadOfTableData()
 
-        #get chosen fit parameters
-        fitPar = self.getFitParameters()
+        # Define checkedPars before using it
+        checkedPars = self.checkedVariables()
```
butlerpd, line 640: Why are we getting rid of `parNames = ` and `fitPar =`?  that seems very wrong? Moving `checkedPars` from before `parVals =` might be reasonable but am guessing that forgettig to define these will have consequenses?

src/sas/sascalc/shape2sas/Shape2SAS.py
```diff
@@ -116,9 +116,38 @@ class SimulatedScattering:
...
+            valid_rotation.append(prof.rotation[i] if i < len(prof.rotation) else [0, 0, 0])
+            valid_rotation_points.append(prof.rotation_points[i] if i < len(prof.rotation_points) else [0, 0, 0])
+            valid_p_s.append(prof.p_s[i] if i < len(prof.p_s) else 1.0)
+    
+    if not valid_subunits:
+        raise ValueError("No valid subunits found with positive dimensions")
+    
+    x_new, y_new, z_new, p_new, volume_total = GenerateAllPoints(Npoints, valid_com, valid_subunits, 
+                                                  valid_dimensions, valid_rotation, valid_rotation_points, 
+                                                  valid_p_s, prof.exclude_overlap).onGeneratingAllPointsSeparately()
```
butlerpd, line 150: Question: this looks to be stripping all empty subunits prior calling generate points.  Is this redundant to the fix in `ViewerModel.py` above? If so my comment on `ViewerModel.py` are negated since then there would no longer be a valid reason for an empty subunit?

butlerpd (commented):
For a change I actually feel qualified to review the code😃.  Comments are in line.

butlerpd:
ahh.. and as I wrote those words the answer was suddenly blindingly obvious. Basically the only possible explanation is that model specific definitions were not complete somewhere. Turns out that is in `subunitTable.py`. Turns out that both bounds **and** type (type = "volume") were forgotten when writing this one out.

I double checked all the other models and they are all complete in their definitions. So really only r2 for the hollow sphere model. A stroke of luck that somebody tried that specific combo!

src/sas/sascalc/shape2sas/Shape2SAS.py
```diff
@@ -116,9 +116,38 @@ class SimulatedScattering:
...
+            valid_dimensions.append(prof.dimensions[i])
+            valid_com.append(prof.com[i] if i < len(prof.com) else [0, 0, 0])
+            valid_rotation.append(prof.rotation[i] if i < len(prof.rotation) else [0, 0, 0])
+            valid_rotation_points.append(prof.rotation_points[i] if i < len(prof.rotation_points) else [0, 0, 0])
+            valid_p_s.append(prof.p_s[i] if i < len(prof.p_s) else 1.0)
+    
+    if not valid_subunits:
+        raise ValueError("No valid subunits found with positive dimensions")
+    
+    x_new, y_new, z_new, p_new, volume_total = GenerateAllPoints(Npoints, valid_com, valid_subunits,
```
butlerpd, line 148: I just noticed this replaces prof which is essentially a method object: `ModelProfile` with a simple unconnected series of lists? Would it not be safer to ensure that `GenerateAllPoints`  gets the correct object? I've looked briefly at `GenerateAllPoints` which seems to only be expecting lists? Still, I wonder if this might make future work more confusing?  Just a question. I have no idea here.

butlerpd:
In tracking down the bug above I noticed a new behavior which I found (and would find as a user) enerving?

Basically somewhere in this PR, the blank rows have been added back to the plugin model list of parameters with names like param_4_0. I don't see where that was explicitly done and worry that it may be causing breakage we have not thoroughly tested yet?

krzywon:
Closing this in favor of #3446

PR #3446, opened 2025-06-03
Shape2SAS outstanding issues redux

krzywon:
## Description

This PR replaces #3444. It fixes the same issues as the previous PR, but removes some of the original changes that were in question.

Fixes https://github.com/SasView/sasview/issues/3413 - When a column or series of columns is selected, the first column in the selection is set as the one to be deleted.
Fixes https://github.com/SasView/sasview/issues/3441 - If one or more shapes are completely within other shapes and overlap is excluded, errors are no longer thrown.
Fixes https://github.com/SasView/sasview/issues/3442 - A couple of typos are fixed in the tool.
No issue - Fixes an issue with the hollow sphere where certain parameters were not being calculated.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py
+++ b/src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py
@@ -150,6 +150,7 @@ class DesignWindow(QDialog, Ui_Shape2SAS, Perspective):
         self.constraint = Constraints()
         self.subunitTable.add.clicked.connect(self.addToVariableTable)
         self.subunitTable.deleteButton.clicked.connect(self.deleteFromVariableTable)
+        self.subunitTable.table.clicked.connect(self.updateDeleteButton)
         self.constraint.variableTable.setConstraints.clicked.connect(self.setConstraintsToTextEditor)
         self.constraint.createPlugin.clicked.connect(self.getPluginModel)
         self.constraint.buttonOptions.reset.clicked.connect(self.onConstraintReset)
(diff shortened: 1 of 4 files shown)


src/sas/qtgui/Calculators/Shape2SAS/Tables/subunitTable.py
```diff
@@ -142,8 +142,8 @@ def hollow_sphere(self):
         """Return the hollow sphere dimensions"""
         name = {self.x: "R", self.y: "r"}
         units = {self.x: "Å", self.y: "Å"}
-        types = {self.x: "volume"}
-        bounds = {self.x: [0, inf]}
+        types = {self.x: "volume", self.y: "volume"}
+        bounds = {self.x: [0, inf], self.y: [0, inf]}
```
butlerpd, line 146: I believe this is only relevant to generating plugins. Should it be in this PR? It is correct and wont' hurt

src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py
```diff
@@ -268,6 +269,12 @@ def addToVariableTable(self):
         #set variables in variable table
         self.constraint.variableTable.setVariableTableData(names, column)
 
+    def updateDeleteButton(self):
+        selection_model = self.subunitTable.table.selectionModel()
+        selected_indexes = selection_model.selectedIndexes()
+        column = selected_indexes[0].column()
+        self.subunitTable.selected.setValue(int(column) + 1)
```
butlerpd, line 276: I thought this was meant to allow the table selection to control which column to delete. Two issues I found:
* it actually does not do that. no matter what combination I try the column deleted is the corresponding to the number in the box next to the `Delete column` button.
* Selecting the column turns every item in the column to a very dark blue background (making it hard to read the text) which can only be removed by deleting that column (using the number box value to match it). Otherwise, at best I can get it down to one cell with the dark background. It would be nice if I can just click again to unselect the column?
* Were this working, I feel like this method should be called either when the numerical value is changed OR the column highlighted so that the two can stay in sync? - but that may be too much at this point.

butlerpd (commented):
The code beyond the column selection looks good. To be fair it seems like the column code should work but ....

butlerpd (approved):
Yep .. that was it. I had not realized the table header was a separate event.  Now works like a charm.  I note you added a fix to the columns being too narrow. That also helps though as soon as I type a bigger number, very typical of using COMX which will be >100 to separate the shapes, it is not wide enough again :-( . But to be clear - this is not the time to try to improve that. Am approving and merging.

PR #3450, opened 2025-06-06
Fix for document download urls not working in QtWebEngine

### LINKED ISSUE #3448: Qt Webengine cannot read any files in html/_downloads
Besides the pdfs, all source files are compiled here. The html code for loading it is not anything I've seen or understand and Qt Webengine does not seem to be able to read it. However chrome seems to be able to do so?

This means that non of the links at the bottom of the model help files works at the moment. We should do something about it for the release

butlerpd:
## Description

We needed to add a line to capture the download request and then create the function that the signal connects to. That function just needs one line to accept the download. However, it does so completely silently so I also added a `logging.warning` message indicating what file is being downloaded in order to force the` logExplorer` window open and give the user feedback.

Finally, I re-enabled the local tutorials downloading

Fixes #3448 ... mostly

## How Has This Been Tested?

Tested on the local development windows 10 machine.

### NOTE:
This fixed the pdf and *.c files downloading. However, the logging indicates the *.py files are downloading but they do NOT in fact appear to be downloaded. I have yet to find any evidence of python files being treated special by QtWebEngine but...

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/DocViewWidget.py
+++ b/src/sas/qtgui/Utilities/DocViewWidget.py
@@ -153,10 +153,16 @@ class DocViewWindow(QtWidgets.QDialog, Ui_DocViewerWindow):
         self.show()
 
     def onDownload(self, download_item):
-        _filename = download_item.url().fileName()
-        logging.warning(f"downloading your file: {_filename} to your default download directory")
         download_item.accept()
-    
+        download_item.isFinishedChanged.connect(lambda: self.onDownloadFinished(download_item))
+
+    def onDownloadFinished(self, item):
+        _filename = item.downloadFileName()
+        if item.state() == item.DownloadState.DownloadCompleted:
+            logging.warning(f"your file: {_filename} was downloaded to your default download directory")
+        else:
+            logging.error(f"FAILED TO DOWNLOAD: {_filename} is unavailable")
+
     def regenerateIfNeeded(self):
         """
         Determines whether a file needs to be regenerated.


jamescrake-merani:
Ah ok so turns out I just misunderstood it. I didn't read Wojtek's message fully before I tried it, and turns out I had the same confusion as he did.

I think we should adopt Wojtek's suggestion in the warning message for now, but in future we should ideally pop up a directory choose dialog so the user can specify where exactly they want to save the tutorial file. This would be a task for the next release though, not this one.

butlerpd:
Right... and after infinite time proper (reasonable?) handling of files not downloading. Why it needs to be a lamda I still have no idea, but is clearly required.

jamescrake-merani:
> A modal pop up would be more obvious but I really hate those and it would start to annoy me very quickly I think.

I was moreso thinking of a pop up which asks you where to save the file. Another idea is that we could send a notification to the user's desktop. Though I'm not sure if QT has a way of natively sending notifications to a user's desktop.

smk78 (approved):
Functionality test on W11/x64 using CI #4093 _after_ having previously installed and run CI #4081 (see https://github.com/SasView/sasview/pull/3447).
Links download files as expected and a message appears in the Log Explorer.
image

https://github.com/SasView/sasview/pull/3447 is now merged to release-6.1.0 so this should be good to merge also once someone has checked the code.

wpotrzebowski (approved):
It looks good to me and  I think we can merge it as it is. The potential improvement would be  to provide a dialog to save a file to the chosen location. I've played little bit with it and it doesn't seem very complicated but given the timeline we have, I don't think we should add any new features.

butlerpd:
Yes, I think that would be more of a feature than this limited bug fix. But if it is not too hard maybe we can have that for a potential 6.1.1? Meantime I'll merge this

PR #3451, opened 2025-06-06
Fix version strings in release-branch builds

llimeht:
## Description

Two issues were spotted in the 6.1.0b1 builds:
- the pinned versions of the dependencies were not installed in the CI build for the tag
- the generated package had an ugly version string (`6.1.0b2.dev0+gd5f5c0082.d20250603` rather than `6.1.0b1`)

Unpacking this version string, it's a development version that would be leading towards `6.1.0b2` that is currently `0` commits past the previous tag, it is at the expected git hash (`d5f5c0082`), with an uncommitted modification that was spotted on that date (`20250603`). The cause of this ugly version string is that git tracked files were modified during build which the versioning then spots and tracks in the string — that's the perfect version string for a regular CI build, but a terrible version string for a release. 

The changes here:
- include tags in the logic for whether to use pinned dependencies
- avoid changing git tracked files so as to ensure that the generated versions are correct

## How Has This Been Tested?

- Look at which "Install Python dependencies" step runs.
- Look at [CI runs for a fake tag](https://github.com/llimeht/sasview/actions/runs/15458470279) (in my fork - you don't want fake tags in the sasview repo!); check the version of the sasview package in the filename for the wheel artifact and in the CI steps that handle the wheel, for example "Install sasview wheel".

These changes seem to be enough to get the CI run for the tag to have the correct version. 

A partial test of this will be on CI of this PR against the `release-6.1.0` branch - checking that "Install Python dependencies for release branches runs" and the version doesn't have a local modification indication in it.

The real test of this will be at 6.1.0b2.

## Review Checklist:

**Installers**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,10 @@ jobs:
       - name: Build sasview
         run: |
           python -m build --no-isolation --sdist --wheel
-          python -m pip install dist/sasview*whl
+
+      - name: Install sasview wheel
+        run: |
+          python -m pip install --no-deps dist/sasview*whl
 
       - name: Publish sdist and wheel package
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


jamescrake-merani (approved):
Code LGTM. I'm approving this even though its after the code freeze as the incorrect version strings would have been a major issue if we had released with this bug still in place.

PR #3458, opened 2025-06-11
Extend pyinstaller data for sas module to include all of sphinx output

### LINKED ISSUE #3455: python source files have been removed from _downloads causing all the links to break
**Describe the bug**
the model documentation lists all the source files at the bottom with links to each source so they can be easily examined. To do this apparently sphinx needs to move them into the html tree under a _downloads folder. However, at some point **AFTER** 6.0.1 only the c files (including library source files) are being placed in that directory while the python files are absent. This make that link fail

**To Reproduce**
1. Reproducing can be painful. You need to navigate to the docs folder which on windows is in `\Use\AppData\Local\SasView\6.1.0\doc\build\html\index.html.  Withe the above PR call up the documentation
2. navigate to USER DOCUMENTATION then Model Documentation then pick any model.
3. Scroll to the bottom of the model help to the Source section.
4. Click on each of the file links in turn
5. all the c files should appear in your download folder
6. the python files will not and, depending on your browser, should report  file not found error.
7. Doing the same with version 6.0.1 does not have that problem

**Expected behavior**
The file should be where the help is pointing

 - Version: 6.1.0b1

llimeht:
## Description

pyinstaller needs to be told what files to bundle with the module - this PR adds some more files to what pyinstaller will keep with the `sas` module.

Fixes #3455

## How Has This Been Tested?

Checked file list from pyinstaller output from CI. Installers not otherwise tested - those with the right hardware need to do that!

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/__pyinstaller/hook-sas.py
+++ b/src/sas/__pyinstaller/hook-sas.py
@@ -1,11 +1,19 @@
 # hook for pyinstaller to find all the extra parts of sas
 
+import os
 from pathlib import Path
 
 try:
     from PyInstaller.utils.hooks import collect_data_files
 
-    datas = collect_data_files('sas')
+    datas = collect_data_files(
+        'sas',
+        excludes=[
+            # handle the documentation trees manually
+            'docs/',
+            'docs-source/',
+            ],
+        )
 
     # add in additional data files
     base = Path(__file__).parent.parent.parent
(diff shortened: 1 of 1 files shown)


jamescrake-merani (commented):
I'm happy with the code. I'll approval when I'm able to test that its fixed the issue.

jamescrake-merani:
I built the installer locally on Windows, and this appears to work now so I'm approving, and merging.

I don't have access to our Mac at the moment (I'm at home now, and I'm not sure if its still configured for remote access). I will try to see if I can remote into it tomorrow, and test there.

PR #3465, opened 2025-06-12
Overide Qdialog closeEvent method to force kill(x) button to behave l…

### LINKED ISSUE #3461: document viewer window close button and kill (x) button do not do the same thing
The kill button, the X at the top right corner of the window calls the QtWidget closeEvent. The `Close` button calls the local `onClose` method then calls after checking for open threads self.close.

The two methods of closing the window should behave identically

butlerpd:
## Description

Overide `Qdialog` `closeEvent` method to force kill(x) button to behave like the `Close` button

Fixes #3461 

## How Has This Been Tested?

Added print statement in close method to verify both buttons call the method. That was not the case previously

## Review Checklist:

**Installers**


krzywon (commented):
This seems fine, but why are we pointing this at the release branch at this point? This is not a blocker

jamescrake-merani:
Agreed. I just realised I replied to the issue rather than the PR by accident, but this is what I said there:
> Is this a major issue? I'm not sure if it should go into 6.1.0 given the stage of release we're at.

butlerpd:
As I said, I was unsure how dangerous the current error really is. As I understand it nobody believes this will be a real problem so have removed the 6.1 label. However, if this is going into main there are more things that should be fixed (such as removing the line suggested above and its probably its counterparts elsewhere so turning this into a draft to prevent merging till I get around to doing that.

jamescrake-merani:
When the release branch is merged into main, I would suggest rebasing this branch onto main as well.

src/sas/qtgui/Utilities/DocViewWidget.py
```diff
@@ -6,6 +6,7 @@
 from typing import Optional
 
 from PySide6 import QtCore, QtWidgets, QtWebEngineCore
+from PySide6.QtGui import QCloseEvent
```
krzywon, line 9: There is now a merge conflict here due to the linting work. I think this is ready to be merged once the conflict is resolved.

krzywon (approved):

PR #3471, opened 2025-06-18
Updates to the release automation script

jamescrake-merani:
## Description

There are two main changes

- The script now uses the Python `csv` module, rather than parsing the `tsv` file manually. This means that quotations in the file will be handled properly.
- Some of the script broke on Linux (and presumably on macOS as well) because of how Unix file systems are case sensitive, as opposed to the Windows file system which is case insensitive. I've fixed this.

## How Has This Been Tested?

I ran the script, and pretty printed the contributors dictionary which came out as expected. I haven't tested the upload to Zenodo, however.=

## Review Checklist:

**Installers**

### DIFF
--- a/build_tools/release_automation.py
+++ b/build_tools/release_automation.py
@@ -7,6 +7,7 @@ import datetime
 import sys
 import os
 from pathlib import Path
+from csv import DictReader
 
 from sas.system.legal import legal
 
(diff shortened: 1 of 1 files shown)


krzywon (approved):
The changes to how version numbers are handled might also need to get overhauled here, but that can be for a different time. What was changed here seems reasonable.

PR #3473, opened 2025-06-19
Metadata Explorer

jamescrake-merani:
**This PR relates to the refactoring project, and is not intended to go into a release of SasView anytime soon.**

### DIFF
--- a/src/ascii_dialog/col_editor.py
+++ b/src/ascii_dialog/col_editor.py
@@ -1,0 +1,86 @@
+from PySide6.QtGui import QRegularExpressionValidator
+from PySide6.QtWidgets import QComboBox, QHBoxLayout, QWidget
+from PySide6.QtCore import Slot, Signal
+from sasdata.quantities.units import NamedUnit
+from sasdata.ascii_reader_metadata import bidirectional_pairings
+from ascii_dialog.column_unit import ColumnUnit
+from typing import cast
+
+class ColEditor(QWidget):
+    """An editor widget which allows the user to specify the columns of the data
+    from a set of options based on which dataset type has been selected."""
+    column_changed = Signal()
+
+    def __init__(self, cols: int, options: list[str]):
+        super().__init__()
+
+        self.cols = cols
+        self.options = options
+        self.layout = QHBoxLayout(self)
+        self.option_widgets: list[ColumnUnit] = []
+        for _ in range(cols):
+            new_widget = ColumnUnit(self.options)
+            new_widget.column_changed.connect(self.onColumnUpdate)
+            self.layout.addWidget(new_widget)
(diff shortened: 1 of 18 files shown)


src/sas/data_explorer_error_message.py
```diff
@@ -0,0 +1,17 @@
...
+        self.setIcon(QMessageBox.Icon.Critical)
+        self.setWindowTitle('Data Error')
+        if len(errors) == 1:
+            error = errors[0]
+            self.setText(str(error))
+        else:
+            full_error_msg = "Some data could not be removed due to the following errors:\n"
+            error_msg_set = set([str(err) for err in errors])
+            error_msg_str = '\n'.join(error_msg_set)
+            full_error_msg += f"\n{error_msg_str}"
```
rprospero, line 16: Optional suggestion: instead of using a `set`, we could use a `defaultdict` that starts at zero and keep track of the number of instances of each error.  This way we can inform the user of multiple occurances of an error.

I'm thinking of a user who sees that they see that error X occured, fixes error X, but then error X goes away because there's 300 instances of it and they have only fixed one.  The count would at least show them that the count has decremented.

src/sas/data_explorer_menu.py
```diff
@@ -0,0 +1,32 @@
...
+from src.sas.refactored import Perspective
+from dataclasses import dataclass
+
+# TODO: Could we perhaps enforce that, if the action is send_to, the action_data
+# should not be None?
+@dataclass
+class DataExplorerMenuAction():
+    action: Literal['remove', 'send_to']
+    action_data: Perspective | None = None
+
```
rprospero, line 15: We could do this with a bit of type trickery

```python
@dataclass
class DataExplorerRemoveAction():
    pass

@dataclass
class DataExporerSendToAction():
    perspective: Perspective

DataExplorerMenuAction = DataExplorerRemoveAction | DataExplorerSendToAction
```

then, instead

```python
match action.action:
    case "remove":
        ...
    case "send_to":
        ...
```

You can do

```python
match action:
    case DataExplorerRemoveAction():
        ...
    case DataExplorerSendToAction(perspective):
        ...
```

We can then create them with `DataExplorerRemoveAction()` and `DataExplorerSendToAction(perspective)`.  This allows to tools like MyPy to ensure that SendTo always have a perspective (as opposed to checking in the initialiser, which won't find the bug until runtime).

rprospero (approved):
Some optional suggestions, but otherwise things look good

jamescrake-merani:
Because of some rebasing I did, the diff for this PR includes some work that was done on the data explorer branch. I'm going to merge this now, but I will definitely take your comments on board as I think they are clear improvements.

PR #3479, opened 2025-06-25
Use bumps 1.0.1 for 6.1.0

jamescrake-merani:
Final bump dependency pins for 6.1.0


llimeht:
`build_tools/requirements-release-ubuntu-22.04.txt` too?

PR #3497, opened 2025-07-10
Fixes fatal access violation in Windows tests

### LINKED ISSUE #3491: pytest and numba cause a fatal access violation on Windows
When running the test script `utest_sas_gen.py` using pytest on Windows, the error "Windows fatal exception: access violation" results. Note that this only applies when using pytest, running the script in python does not cause the same error.

This is due to a known issue with numba: https://github.com/numba/numba/issues/8223#issuecomment-1175213066

"This is a known issue. Both pytest and Numba rewrite bytecode so that can result in strange errors. As a result it is not recommended to use pytest when testing Numba"

Therefore, we should find a way to ensure that we do not use numba when running these tests with pytest.

DrPaulSharp:
## Description

Pytest and numba cannot work together on Windows, so this PR introduces an autouse fixture to ensure the non-numba versions of routines are used when running pytest on Windows.

Fixes # (issue/issues)

Fixes #3491

## How Has This Been Tested?

Run both `python test/sascalculator/utest_sas_gen.py` and `pytest test/sascalculator/utest_sas_gen.py` on both linux (via WSL) and Windows. I have verified that the tests run without an access violation in all cases, and numba is used in each case except for the pytest run on Windows.

## Review Checklist:

**Installers**

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,8 +1,18 @@
+import os
+import platform
+
+import pytest
+
 from run import prepare
+from src.sas.system import config
 
 prepare()
 
-from src.sas.system import config
-
 config.override_with_defaults()
 
+
+@pytest.fixture(scope="session", autouse=True)
+def set_env():
+    # pytest and numba do not play nicely together on windows, see: https://github.com/numba/numba/issues/8223#issuecomment-1175213066
+    if platform.system() == "Windows":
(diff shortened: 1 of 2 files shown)


jamescrake-merani (approved):
Independently tested on my remote Windows desktop, and the problem is fixed there. Code changes look reasonable.

PR #3169, opened 2025-01-07
Fitting widget refactor

rozyczko:
## Description

This is the refactoring effort, the same as in https://github.com/SasView/sasview/pull/2651 but reparented to the current `main`.

- Polydispersity widget logic moved to PolydispersityWidget.py
- Polydispersity widget UI moved to PolydispersityWidgetUI.ui
- Magnetic widget logic moved to MagnetismWidget.py
- Magnetic widget UI moved to MagnetismWidgetUI.ui
- Moved common structures to FittingLogic.py

All issues raised by @krzywon in the original PR have been addressed.

Additionally, the files involved in the refactor were "improved" by
- adding type hints throughout
- adding style changes required by `ruff` (running `ruff check`)

This is part 1 of the refactoring, focused on the two biggest and most independent components.
Next, I will attempt to refactor out constraints to a separate module.

## Review Checklist:

[if using the editor, use `[x]` in place of `[ ]` to check a box]

**Documentation** (check at least one)
- [ X] There is **nothing** that needs documenting
- [ ] Documentation changes are **in this PR**
- [ ] There is an **issue** open for the documentation (link?)

**Installers**
- [ ] There is a chance this will affect the **installers**, if so
  - [X ] **Windows** installer (GH artifact) has been tested (installed and worked) 
  - [X ] **MacOSX** installer (GH artifact) has been tested (installed and worked) 

**Licencing** (untick if necessary)
- [x] The introduced changes comply with SasView license (BSD 3-Clause)

PR #3505, opened 2025-07-15
3500: Fix for PD plotting

### LINKED ISSUE #1587: 5.0 loader not finding polydispersity type ?
In my local build of ESS_GUI on Win 10, attached project (rename .txt to .json) should have Schulz on FitPage3, but is coming in as default Gaussian. 
[12June2020_3sets_test1.txt](https://github.com/SasView/sasview/files/4771849/12June2020_3sets_test1.txt)

### LINKED ISSUE #2438: Polydispersity settings not updating correctly

### LINKED ISSUE #2540: Polydispersity function is not being set when project file loads
**Describe the bug**
If a fitting project file is loaded containing a model using polydispersity, parameter values are correctly set, including for the polydispersity, but the polydispersity function type gets left on the Gaussian default _even though_ the required function is written into the json file.

**To Reproduce**
Load some data.
Create a model featuring polydispersity.
Change the polydispersity function to something other than Gaussian.
Fit.
Save Project.
Exit & restart SasVIew.
Load the project file.

**Expected behavior**
The polydispersity function type should be set to that in the project file.

 - Version: 5.0.5 (reported by User JenniferG), but also affects 5.0.6 (tested by @smk78).

 - OS: W10, but probably all platforms.

**Additional context**
See also:
https://github.com/SasView/sasview/issues/2353
https://github.com/SasView/sasview/issues/1772#issuecomment-768375426
https://github.com/SasView/sasview/issues/1587

### LINKED ISSUE #3248: Changing S(Q) selection with polydispersity active causes error and resets PD to gaussian
**Describe the bug**
When trying to fit 1D data (not replicated with theoretical data only) with a polydisperse radius and a structure factor the following error is given:
`ERROR: Traceback (most recent call last): File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1800, in onPolyModelChange IndexError: list assignment index out of range`
Which results in the polydispersity being changed to the default value (gaussian). Removing the structure factor results in the same error

**To Reproduce**
1. Load some 1D data and send to fit
2. Select Sphere, sphere for the model (also tested with cylinder, cylinder)
3. Tick the box for polydipersity and mark the radius as fittable in the polydispersity tab
4. Set polydispersity type to be not gaussian (tested with lognormal and schulz)
5. Click the fit button (note, changing polydispersity to be non-zero without fitting is insufficient)
6. Change your selection for structure factor (add or remove structure factor, tested with hardsphere, none, and hayter_msa)
7. Observe error posted above and reversion of polydipersisty to default values

**Expected behavior**
Structure factor should be applied without affecting polydispersity

 - Version: 5.0.4 and 6.0.1

 - OS: Windows 10

**Additional context**
Model seems to not be relevant, just that structure factor is updated after a fit which included a polydispersity with a type other than gaussian.

### LINKED ISSUE #3500: Polydispersity causes incorrect size distribution calculations when changing models
**Describe the bug**
When changing between models with polydispersity enabled, the calculated size distributions become nonsensically broad. Changing back to the original model does not recover the original computation. This is the case whether simulating data (when I first discovered it) but also when fitting data (I verified this). You can overcome this by switching off polydispersity or by changing both the polydispersity function and the value.

**To Reproduce**
1. Select a model to simulate. I chose Sphere/sphere.
2. Enable polydispersity, enter values, and calculate. I chose PD=0.08 schulz.
3. Change models and calculate. I chose Sphere/core_shell_sphere.
4. Return to the original model and calculate. The simulated data will not be the same.

**Expected behavior**
I do know that SasView does not remember the polydispersity function choice, but it does remember the PD value. The expected behavior is that the size distribution that is computed given the same combination of PD and function should be the same and that it should not depend on the path taken.

**Screenshots**
Here are screenshots of the steps in the "to reproduce" section.

Image

Image

Image

Image

Image

 - Version: 6.1.0 **only** (as far as I can tell)
 - I tested this in v5.0.6 and v.6.0.0 and do not see the issue.

 - OS: Windows 11

 - Installed: Windows installer

**Additional context**
I am not sure how common of an approach to fitting this in. It feels like it might be an edge case, as I don't know how often people switch around models when fitting, as I have done here. However, if you did follow this pathway to computation or fitting, then your computed curves would be incorrect in a way that was not easily visible (the PD and functions are the same).

krzywon:
## Description

This restores some of the code changed during the fitting widget refactor (#3169) to ensure the PD plots are correct when switching models, in any way. This 2nd commit (7c0569ccaadd12464add4a4846e15a113306d9ec) restores the PD function on model change, which wasn't present before.

Fixes #3500
(Possibly?) Fixes #3248 
(Possibly?) Fixes #2540 
(Possibly?) Fixes #2438
(Possibly?) Fixes #1587

## How Has This Been Tested?

I repeated the steps in #3500, selecting schultz and switching between sphere and core_shell_sphere. The PD distribution is nominally correct, and the PD function remains as schulz. I also added and changed the structure factor and the function remained schultz for each S(Q) change.

I have **NOT** tested loading a project, but this is the same method called during the project load process, so I am guessing this may also fix the issues related to that.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/PolydispersityWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/PolydispersityWidget.py
@@ -481,6 +481,14 @@ class PolydispersityWidget(QtWidgets.QWidget, Ui_PolydispersityWidgetUI):
             # Nsigs
             param_repr = GuiUtils.formatNumber(param_dict[param_name][5+ioffset], high=True)
             self.poly_model.item(row, 5+joffset).setText(param_repr)
+            # Function
+            param_repr = param_dict[param_name][6+ioffset]
+            self.poly_model.item(row, 6+joffset).setText(param_repr)
+            index = self.poly_model.index(row, 6+joffset)
+            widget = self.lstPoly.indexWidget(index)
+            if widget is not None and isinstance(widget, QtWidgets.QComboBox):
+                func_index = widget.findText(param_repr)
+                widget.setCurrentIndex(func_index)
 
             self.setFocus()
 
--- a/src/sas/qtgui/Perspectives/Fitting/PolydispersityWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/PolydispersityWidget.py
@@ -157,8 +157,11 @@ class PolydispersityWidget(QtWidgets.QWidget, Ui_PolydispersityWidgetUI):
             # PD[ratio] -> width, npts -> npts, nsigs -> nsigmas
             if model_column not in delegate.columnDict():
                 return
-            self.poly_params[parameter_name_w] = value
-            self.logic.kernel_module.setParam(parameter_name_w, value)
+            # Map the column to the poly param that was changed
+            associations = {1: "width", delegate.poly_npts: "npts", delegate.poly_nsigs: "nsigmas"}
+            p_name = f"{parameter_name}.{associations.get(model_column, 'width')}"
+            self.poly_params[p_name] = value
+            self.logic.kernel_module.setParam(p_name, value)
 
             # Update plot
             self.updateDataSignal.emit()


src/sas/qtgui/Perspectives/Fitting/PolydispersityWidget.py
```diff
@@ -478,6 +481,14 @@ def updateFittedValues(row: int) -> None:
             # Nsigs
             param_repr = GuiUtils.formatNumber(param_dict[param_name][5+ioffset], high=True)
             self.poly_model.item(row, 5+joffset).setText(param_repr)
+            # Function
+            param_repr = param_dict[param_name][6+ioffset]
+            self.poly_model.item(row, 6+joffset).setText(param_repr)
```
rozyczko, line 486: Setting the text on the combobox is not necessary.
Column 6 always contains a combo box (`addNameToPolyModel`)

rozyczko (approved):
Looks and behaves as expected with the caveat explained below

rozyczko:
This fixes the immediate issue of the combobox state not being restored on model change.

However, there's another issue here that needs addressing at some point.
Running a fit with polydisperse parameters checked adds an error column to the polydisp table.
This is OK when just running the fit - the min/max/npts/nsigs/function combobox columns get shifted and display ok.
But when a model is changed AFTER the fitting, the table is not properly reindexed.

The table headers are correctly not displaying the error, but the parameter values still contain the error column.
This means `Min` now shows the error, `Max` shows the min value etc. `Function` is now showing the value of Nsigs and `Filename` contains the combobox of functions.

Issue #3498, opened 2025-07-11
Enable Ruff for PRs


krzywon:
**Is your feature request related to a problem? Please describe.**
As discussed in https://github.com/orgs/SasView/discussions/3171, and agreed upon during a fortnightly meeting, we will be enabling a limited ruff implementation for PRs, and gradually enabling features over time. This issue is for the initial implementation.

Under discussion - do we run ruff on the entire code base and have one large PR to fix existing issues, or do we only do small fixes on a per-PR basis?

**Describe the solution you'd like**

- Enable ruff and a checker (pre-commit?)
- Configure it to check PRs and make changes using the diff
- Ensure it isn't run against the entire file (for now)

DrPaulSharp:
I have looked at the [graylint](https://github.com/akaihola/graylint) package, and the associated package [darker](https://github.com/akaihola/darker). These packages display ruff linter/formatter errors for only changed lines in a commit/pull request. However, the package runs very slowly (especially compared to ruff natively); uses a very old version of ruff - 0.2.0, the current version is 0.12.3; and is not customisable. Also, if we try to apply automatic fixes, these will be applied to the whole file anyway, and graylint runs the linter first and then does the comparison between different commits/branches. I'm not aware of any way to run fixes only after the comparison is done. These problems severely impact on how useful ruff is for improving code quality, so I'm not convinced by this offering.

I think using [pre-commit](https://pre-commit.ci/) on the CI is a better option, but this does have the drawback that it will apply to changed files in their entirety, rather than only changed lines. Given this, we should consider whether we want to run both the linter and formatter, or just the linter. Also, we should consider the ruleset we want to use. The default ruleset will probably serve us well for now, but rules can be easily added or removed if desired. Here is the full ruff [ruleset](https://docs.astral.sh/ruff/rules/).

One final thought I have is how to handle ruff rules with fixes available. If we want to focus only on those rules and leave aside unfixable rules, with our CI applying the fixes automatically, I'd be inclined to firstly have a PR applying linting fixes to the existing codebase. I am aware however that there are other branches that need to be rebased on `main` (the `refactor_24` branch is one example). If there are a lot of branches that require rebasing, we should proceed with care when it comes to making lots of changes to main. On the other hand, if the refactor branch is the only one we expect to rebase, then I think applying the fixes to both branches is the best way to proceed, so any future rebase is not further complicated by linting changes.

DrPaulSharp:
[...]

PR #3507, opened 2025-07-16
Applies automatic fixes from the default ruff ruleset

DrPaulSharp:
## Description

I have run `ruff check` on the codebase and applied the automatic fixes to the errors with such a fix available. These mainly concern unused imports (see #3498 for full details).

Begins work on issue #3498 

## How Has This Been Tested?

There are no changes to the functionality of the code. The CI should pass.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/system/console.py
+++ b/src/sas/system/console.py
@@ -159,9 +159,6 @@ def setup_console_simple(stderr_to_stdout=True):
 def demo():
     setup_console()
     if 0:
-        import win32
-        from win32 import win32console
-        from win32 import win32gui
         from win32 import win32process, win32api
         pid = win32process.GetWindowThreadProcessId(hwnd)
         handle = win32api.OpenProcess(win32con.PROCESS_QUERY_INFORMATION | win32con.PROCESS_VM_READ, False, pid[1])
--- a/src/sas/system/console.py
+++ b/src/sas/system/console.py
@@ -1,7 +1,8 @@
 """
 Windows console binding for SasView
 """
-import os, sys
+import os
+import sys
 import atexit
 
 def attach_windows_console():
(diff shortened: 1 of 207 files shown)


test/utest_sasview.py
```diff
@@ -12,7 +12,7 @@
 logger = logging.getLogger(__name__)
 
 try:
-    import xmlrunner
+    pass
```
jamescrake-merani, line 15: I don't think this is the intended behaviour of this test. It looks like its trying to fail if `xmlrunner` doesn't exist, as the test doesn't work without it. Except based on the CI, it looks like it does anyway. So my suspicion is that this whole `try ... except ...` block is no longer needed.

Anyhow, for tests that do have dependencies, we already have a separate requirements file for tests, which is also pointed to in the `pyproject.toml` file. So I don't think this sort of check is even necessary.

src/sas/system/console.py
```diff
@@ -159,9 +160,6 @@ def console_open(mode):
 def demo():
     setup_console()
     if 0:
```
jamescrake-merani, line 162: This whole if block can probably just be removed. `If 0` should always fail.

jamescrake-merani (changes requested):
LGTM. Just a couple of suggestions where I think we can go a bit further.

test/utest_sasview.py
DrPaulSharp: Yes, this is quite confusing. The next lines tell you to try installing `unittest-xml-reporting`, rather than `xmlrunner`. The `unittest-xml-reporting` package is also included in `requirements-test.txt`, so I agree that this section can be removed.

jamescrake-merani (approved):
All requested changes have been resolved. This is ready to merge.

PR #3509, opened 2025-07-16
Fix Invariant Test Warnings

krzywon:
## Description

Running the calculation unit tests locally gave a set of warnings:

> =============================== warnings summary ===============================
test/sasinvariant/utest_data_handling.py: 26 warnings
test/sasinvariant/utest_use_cases.py: 13 warnings
  /home/runner/work/sasview/sasview/src/sas/sascalc/invariant/invariant.py:363: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    err_matrix = math.fabs(residuals) * cov
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

One CI example that shows the warnings: https://github.com/SasView/sasview/actions/runs/16317859941/job/46088019662

## How Has This Been Tested?

These warnings are no longer shown when running the unit tests after this change. I also compared the calculations run in v6.1.0 (Windows release binary) to running from source. Using the same data file with the same Invariant parameter values, all calculated paramter values and errors matched between the two versions.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/sascalc/invariant/invariant.py
+++ b/src/sas/sascalc/invariant/invariant.py
@@ -360,7 +360,7 @@ class Extrapolator(object):
             try:
                 inv_cov = np.dot(A.transpose(), A)
                 cov = np.linalg.pinv(inv_cov)
-                err_matrix = math.fabs(residuals) * cov
+                err_matrix = np.fabs(residuals) * cov
                 err = [math.sqrt(err_matrix[0][0]), math.sqrt(err_matrix[1][1])]
             except:
                 err = [-1.0, -1.0]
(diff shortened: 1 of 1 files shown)


src/sas/sascalc/invariant/invariant.py
```diff
@@ -360,7 +360,7 @@ def fit(self, power=None, qmin=None, qmax=None):
             try:
                 inv_cov = np.dot(A.transpose(), A)
                 cov = np.linalg.pinv(inv_cov)
-                err_matrix = math.fabs(residuals) * cov
+                err_matrix = np.fabs(residuals) * cov
```
rozyczko, line 363: There's another `math.fabs(residuals)` invocation in line 349. Should it also be replaced?

(Maybe we should standardize on replacing all `math.` with `np.`?)

jamescrake-merani (approved):
This change seems reasonable to me. Since it removes the warning, and repeats the same results, I don't see any issues.

PR #3515, opened 2025-07-17
Use requirements-dev.txt instead of list of dependencies

krzywon:
A merge conflict between the release branch and main branch wasn't resolved correctly. This combines elements from both branches, as I believe they were expected.


jamescrake-merani (approved):
Apologies this was my fault. It definitely makes more sense to use the `requirements-dev.txt`

PR #3516, opened 2025-07-17
Dependabot (Option 1) - Move Release Yamls

krzywon:
Dependabot caught our release specific dependency files and attempted to increment versions on a few dependencies. This option moves the release-specific requirement files into their own sub-directory so they aren't caught by dependabot.


llimeht:
I was just thinking that we'd delete these files in `main` and they'd reappear quite close to the next release and in the release branch. Is there an advantage to keeping them around during the development cycle?

krzywon:
> Is there an advantage to keeping them around during the development cycle?

They act as a reminder for future release managers, they can be used to easily recreate the last release environment, regardless of what point in the release cycle we are at on a developers system, and release branches won't fail to build the first time they are created.

andyfaff:
It's not clear to me why this degree of version pinning is needed. I would hazard a guess that the majority of users would use the pre-built release binary, meaning that they never need the entire collection of specific versions. If the versions are recorded in the release binary that would allow one to easily recreate the binary again, without worrying about curating all the version pins.

krzywon:
@llimeht - all good points. I'll make a PR for Option 3 that deletes the release yamls.

@andyfaff - The pinning is for a number of reasons and not really for users at all. The primary reason is to minimize the potential for breaking changes late in the release process that are out of our control (think numpy 2.0). Another main reason is to have an exact record of the dependency list used in a particular release, which is useful for developers who are actively developing toward that release, and for fair principles in knowing the exact software calculations used for analysis. More of the reasoning we've used to go this route is in a discussion: https://github.com/orgs/SasView/discussions/3245.

krzywon:
After some discussion, assuming our primary requirements does not require pinning, the decision is the disable dependabot entirely in #3517. Closing this

PR #3521, opened 2025-07-18
Dependabot (Option 3) - Remove Release Yamls

krzywon:
Dependabot caught our release specific dependency files and attempted to increment versions on a few dependencies. This option removes the release yaml files, with the expectation that a release manager will create the yamls before creating the release branch.

Note - all options of this move from pip to uv.


krzywon:
After some discussion, assuming our primary requirements does not require pinning, the decision is to disable dependabot entirely in #3517. Closing this

PR #3534, opened 2025-07-23
Fetching of the license file from relative path

### LINKED ISSUE #3477: Shouldn't the licence agreement in the Windows installer include the BSD licence?
**Describe the bug**
Currently, the licence agreement window that first pops up when you run the Windows installer contains details about SasView's original funding source, participating facilities, and a request to acknowledge the software if its used in a publication. However, there doesn't seem to be any text from the BSD licence, which I'd expect to see in this section.

**To Reproduce**
1. Download a version of the SasView installer 
2. Launch it
4. See licence agreement

**Expected behavior**
See the BSD licence.

**SasView version:**
 - Version: 6.1.0b2

**Operating system:**
 - OS: Windows 11

rozyczko:
## Description

Added relative path to the correct license file to be used in the Windows installer.

Fixes #3477

## How Has This Been Tested?

Local test on GH built installer.

## Review Checklist:

**Installers**
  - [X] **Windows** installer (GH artifact) has been tested (installed and worked) 

### DIFF
--- a/build_tools/release_automation.py
+++ b/build_tools/release_automation.py
@@ -348,7 +348,6 @@ def main(args=None):
     license_line = legal.copyright
     update_file(SASMODELS_PATH / 'LICENSE.txt', license_line, 0)
     update_file(SASDATA_PATH / 'LICENSE.TXT', license_line, 0)
-    update_file(SASVIEW_PATH / 'installers' / 'license.txt', license_line, -1)
 
     sasview_issues_list = args.sasview_list
     sasmodels_issues_list = args.sasmodels_list


jamescrake-merani (approved):
@llimeht I'm not sure I agree. We have the licence in the documentation but there's no harm on having it in the installer too. It seems Git on Windows also shows the GPL, so I don't think this is unusual.

I've verified this works on my Windows desktop so I'm approving. You may want to remove the licence file in this PR before merging, or in a separate one.

Also I noticed that some files have this on the header:
```
"""
This software was developed by the University of Tennessee as part of the
Distributed Data Analysis of Neutron Scattering Experiments (DANSE)
project funded by the US National Science Foundation. 

See the license text in license.txt

copyright 2010, University of Tennessee
```

We should probably update this. But also, I'm not even sure if its necessary to keep this header at all. But that's probably a separate discussion to have.

llimeht:
It's _permitted_ to show the licence but it is absolutely not _required_. The only thing that is required is that the binary distribution have a copy of the licence which is what §2 of the licence says. It's a happily settled piece of free software legalese that the user is not required to click-through the licence; indeed the user is actually not even required to accept the licence as the licence controls distribution not usage. 

(and note that sasview wasn't even carrying a copy of its own licence... in violation of its own licence... until https://github.com/SasView/sasview/pull/3480)

I still think 2 changes are required here:

 - the request to give attribution to sasview was the most important part of the message in the windows installer and that is now lost - this needs to come back. 
 - `installers/license.txt` is also used in `build_tools/release_automation.py` which needs fixing.

llimeht:
> The point is to have consistent licensing text in both the installer display window and inside the app. This PR fixes the issue. 

If the installer displayed a difference licence to what sasview was actually released under then that would be an issue. However, that's not the case.

The installer does not currently display *any* licence text - it's not that it's inconsistent, it doesn't do it at all, and that's ok.

> The acknowledgement text is a separate

No, this PR creates the issue with the acknowledgement text, while not fixing a problem that I really don't think exists.

> we would need to add another click-through. 

No, as an alternative, we could not merge this PR, and then that wouldn't be needed.

> I would propose to merge this as-is, since it fixes the obvious issue of having different license wording between the two locations.

There isn't different licence text in two locations. There is the licence text in the source & documentation, and a request for acknowledgement in the Windows installer. That's self-consistent and OK, while this PR creates problems that then need to be fixed. If we really wanted we could add the licence text to that request for acknowledgement, (but it's not needed, it's just clutter).

The only inconsistency is that the file in git is called `installers/license.txt` and it could be renamed to anything else we want if that is confusing. Perhaps just `git mv installers/license.txt installers/acknowledgement.txt` instead?

krzywon:
I think #3680 covers everything in this PR while adding a user dialog. I would prefer the other PR over this one, IMO.

Issue #3522, opened 2025-07-18
example_data folder missing from Windows distribution


smk78:
**Describe the bug**
The \example_data folder is not included in the Windows distribution of 6.1.0!

@jamescrake-merani has independently checked for and verified this bug.

**To Reproduce**
1. Install the 6.1.0 release on a Windows machine
2. Open a File Explorer and navigate to the installation folder 
3. Observe that the \example_data folder is absent

**Expected behavior**
The \example_data folder should be present.

**Screenshots**

<img width="542" height="681" alt="Image" src="https://github.com/user-attachments/assets/9f0efc06-4946-43e6-a6a7-ea777cd46d3f" />

**SasView version (please complete the following information):**
 - Version: 6.1.0

**Operating system (please complete the following information):**
 - OS: Windows

**Installation method (please complete the following information):**
 - Installed: tagged release Windows installer

**Additional context**
The \example_data folder is present in 6.0.1 & 6.0.0 (in v5.x and earlier it was called test). This is a bug introduced in 6.1.0.

I presume that since no one else has flagged this (eg, at the recent ICNS SasView session) it is specific to Windows, but do not have the means to check other platforms (and IDAaAS@ISIS is offline at present). It would be good to have reports from other platforms.

In the run-up to the release of 6.1.0 there was discussion about the need for/size of \example_data in the distribution, and some house-keeping was performed, but to my knowledge there was no definitive decision to remove \example_data from 6.1.0? Thus what we are likely looking at here is either an erroneous commit (though surely that would affect all platforms?) or an issue with pyinstaller?

As \example_data is referenced in the Tutorials (and likely some Help Documentation too) I am labelling this as a blocker for 6.1.1.

krzywon:
I just looked at this and found it one level deeper in the file tree. Look in the `sas` directory and you should find `example_data` there. Regardless, we need to be sure the installation location is consistent across versions.

llimeht:
> Regardless, we need to be sure the installation location is consistent across versions.

I'd also add to that: consistency across installation methods (running from git, installing from source, from pip, from pyinstaller, from wheels, from flatpak, whatever is done on CI). 

In 6.0.x `example_data` was under *both* locations on CI but only one was shipped.

Making sure all the different routes end up at the same end point is really important to development, effective testing, the CI testing being meaningful, etc. That does mean that all the resources of the `sas` module need to be inside the `sas` directory.

Issue #3485, opened 2025-07-03
Installed docs should take precedence over development docs


krzywon:
**Describe the bug**
If a developer has been working on SasView and runs from source, the documentation is put into a location that uses the upcoming version of the app as a folder name. To accommodate potential doc edits, no existing documentation is currently overwritten when the app is launched, so anything new added to the docs won't be updated if any existing docs are found. Even worse, the development docs are put into the final release folder, not a development folder, so if a release is installed after developing, the latest docs aren't updated for the user.

**To Reproduce**
1. Run `run.py` from an older development branch (post 6.0.1, pre 6.1.0)
2. Close the program
3. Install the latest version (6.1.0)
4. Launch the installed app
5. Click on help
6. The help files from the development branch will be used instead of the installed docs.

**Expected behavior**
At least one of these three things should happen (or something else if others can think of one):

1. On install, the installer should always overwrite any existing documentation in the versioned folder to prevent conflicts.
2. The development branches should use the development version for the folder name, not the final version for documentation placement.
3. The doc editing is no longer going to be in the SasView app, so an app launch should *always* check for doc differences and overwrite any existing docs if changes have been made. Note, this might increase the launch time, so it isn't ideal.

**SasView version (please complete the following information):**
 - Version: 6.1.0 (at least, but likely also effects any 6+ version)

**Operating system (please complete the following information):**
 - OS: All

**Additional context**
Work around:
1. Navigate to the user documentation folder outlined in https://www.sasview.org/docs/user/qtgui/MainWindow/sasview_files.html#documentation
2. Delete the folder with the version you are running.
3. Launch the app and the documentation will be recreated.

llimeht:
>     On install, the installer should always overwrite any existing documentation in the versioned folder to prevent conflicts.

The installer can't actually do that - it runs as a single user (often admin), while sasview can run as a different user and the cached docs are stored in the user's profile.

>    The development branches should use the development version for the folder name, not the final version for documentation placement.

That was briefly the case and was causing other issues - but if we're good about making development tags in main to ensure that the version string is automatically incremented properly then this will work.

>   The doc editing is no longer going to be in the SasView app, so an app launch should always check for doc differences and overwrite any existing docs if changes have been made. Note, this might increase the launch time, so it isn't ideal.

[...]

Issue #3201, opened 2025-01-29
Example data is difficult to access on Mac


lucas-wilkins:
Example data, whilst present in the .app file is hard to get to. You can't see it from sasview, and MacOS makes copying it difficult. Is there a better way of distributing it.

butlerpd:
Thanks for creating this issue. I should point out that we actually had a mac person try to follow the tutorial at SAS2024 which used the example data and he could not find the data. He suggested it seemed like the example data did not ship with the mac installer but clearly that is not the case from what you say @lucas-wilkins . I should point out that he is an expert software developer and not a neophyte, so for regular users I think this means the data is effectively not available?

lucas-wilkins:
I was demonstrating SasView to my team, so this bug report comes from a software developer too!

jamescrake-merani:
I should also add to this that we don't actually currently ship the example data with the Flatpak, and I can't think of a place to put these files that would be obvious to the user since they would either be installing the Flatpak from the command line, or from their distribution's app store.

Issue #3062, opened 2024-08-17
Need easier way to access the Example Data folder


butlerpd:
**Is your feature request related to a problem? Please describe.**
The Example data can be very useful in trying to understand a new feature. However it is buried in the install directory which is not the most obvious place for a user to go looking. 

**Describe the solution you'd like**
There have been request to add it to the "load file" menu. Alternatively from a separate menu item?

**Additional context**
This was identified during the NIST CNR summer school. It would probably be nice to have but not clear to me how important it would be. Certainly it cannot be labeled a bug so I am not going to label it for release 6.0.0 despite original agreement

llimeht:
+1 to this. It's also useful for both automated and manual testing.

Getting a list of the examples that are shipped is likely another job for `importlib.resources`, which then saves needing to guess where things are installed and lots of potentially fragile file manipulations. It's also possible to write automated tests that verify that the resources were located. 

Not sure what the design vision is, but rather than a file picker widget, a list of the files with some comments about what each one is would be possible. It would be lovely to have a sentence about what each example is right next to it in the interface.

PR #3535, opened 2025-07-23
Various user files changes

krzywon:
## Description

This changes a number of items related to user files:
- All functions and constant PATHs were moved into sas.system.user so they are all in one location.
- On launch, example data files are copied from the installed/active directory into the OS-desired
- On launch, an attempt is made to move all files from the installed location to the desired user location. Nothing is written if the file already exists in the resulting path.

Refs #3522, #3485, #3201, #3062 - This PR may fix a few of these, but more testing is needed for each.

NOTE: The example data location still needs to be updated in other parts of the documentation.

## Review Checklist:

**Installers**


src/sas/system/user.py
```diff
@@ -124,3 +137,140 @@ def copy_old_files_to_new_location():
...
+        f_name = dpth / resource.name
+        if "__pycache__" in resource.name:
+            continue
+
+        if resource.is_dir():
+            # recurse into the directory
+            module_copytree(module, spth / resource.name, dpth / resource.name)
+        elif resource.is_file() and not f_name.exists():
+            logging.debug("Copied: %s", spth / resource.name)
+            with open(dpth / resource.name, "wb") as dh:
```
rozyczko, line 165: `dpth / resource.name` is already assigned to `f_name` - use it
also, maybe do the same for `spth / resource.name` as it is being used more than once.

src/sas/system/user.py
```diff
@@ -124,3 +137,140 @@ def copy_old_files_to_new_location():
...
+                dh.write(resource.read_bytes())
```
rozyczko, line 166: You might want to wrap this in an exception handler to catch things like permission issues

src/sas/system/user.py
```diff
@@ -124,3 +137,140 @@ def copy_old_files_to_new_location():
...
+        else:
+            logging.warning("Skipping %s (unknown type)", spth / resource.name)
+
+
+def copy_module_resources() -> bool:
+    """Obtain the source and build output from within the installed sas module"""
```
rozyczko, line 172: The name of the method does not correspond to its return type. You might want to describe the 'success' or 'failure' more explicitly in the docstring.

src/sas/system/user.py
```diff
@@ -124,3 +137,140 @@ def copy_old_files_to_new_location():
...
+        # such as when installed from the pyinstaller bundle prior to v6.1
+        source_dir = exe_dir / "doc" / "source"
+        build_dir = exe_dir / "doc" / "build"
+
+    elif (exe_dir.parent / "Frameworks" / "doc").exists():
+        # In the MacOS bundle, the executable and packages are in parallel directories
+        source_dir = exe_dir.parent / "Frameworks" / "doc" / "source"
+        build_dir = exe_dir.parent / "Frameworks" / "doc" / "build"
+
+    else:
```
rozyczko, line 206: It would probably be safer to `elif` this branch and use `else` to throw or at least send some info to warn the user that `source_dir` and `build_dit` can't be found.

rozyczko (commented):
Minor comments on the methods copied around. I know these were just copied, not written in this PR, but I made some suggestions anyway.
The rest looks OK.

`src/sas/sascalc/system/user.py` fails ruff with a minor issue. Please consider fixing.

krzywon:
> is copying the example_data to ~ a desirable thing? It's not like ~/AppData is an easily findable location on Windows (I'm curious whether that path is actually different under different types of Windows installation too)

From a MacOS perspective, absolutely. It's very difficult to find the example data (#3201). For other OSes, it's debatable.

PR #3539, opened 2025-07-24
Applies fixes for Ruff linting errors

### LINKED ISSUE #3498: Enable Ruff for PRs

DrPaulSharp:
## Description

Applies fixes for most (~75%) of the linting errors in the sasview codebase. This mostly concerns correcting import statements, either by adding those that are missing, or specifying routines instead of relying on star imports (i.e., `from numpy import *`)

Fixes #3498

## How Has This Been Tested?

The code passes its own tests. There should be no change in functionality.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingOptions.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingOptions.py
@@ -100,10 +100,10 @@ class FittingOptions(PreferencesWidget, Ui_FittingOptions):
         for option in bumps.options.FIT_FIELDS.keys():
             (f_name, f_type) = bumps.options.FIT_FIELDS[option]
             validator = None
-            if type(f_type) == types.FunctionType:
+            if isinstance(f_type, types.FunctionType):
                 validator = QtGui.QIntValidator()
                 validator.setBottom(0)
-            elif f_type == float:
+            elif isinstance(f_type, float):
                 validator = GuiUtils.DoubleValidator()
                 validator.setBottom(0)
             else:
@@ -245,9 +245,11 @@ class FittingOptions(PreferencesWidget, Ui_FittingOptions):
         """
         if current_fitter is None:
             current_fitter = self.current_fitter_id
-        if option_id not in list(bumps.options.FIT_FIELDS.keys()): return None
+        if option_id not in list(bumps.options.FIT_FIELDS.keys()):
+            return None
         option = option_id + '_' + current_fitter
-        if not hasattr(self, option): return None
+        if not hasattr(self, option):
+            return None
         return eval('self.' + option)
 
     def getResults(self):
--- a/src/sas/qtgui/Perspectives/Fitting/UnitTesting/ConstraintWidgetTest.py
+++ b/src/sas/qtgui/Perspectives/Fitting/UnitTesting/ConstraintWidgetTest.py
@@ -148,11 +148,11 @@ class ConstraintWidgetTest:
 
         # disable the tab
         widget.tblTabList.item(0, 0).setCheckState(0)
-        assert widget.tabs_for_fitting["test_tab"] == False
+        assert not widget.tabs_for_fitting["test_tab"]
         assert not widget.cmdFit.isEnabled()
         # enable the tab
         widget.tblTabList.item(0, 0).setCheckState(2)
-        assert widget.tabs_for_fitting["test_tab"] == True
+        assert widget.tabs_for_fitting["test_tab"]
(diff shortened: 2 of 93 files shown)


llimeht:
Great to get rid of some `*` imports!

Looks like there are some test failures to address prior to merging; given that the test suite is quite patchy in its coverage, it also can't be relied upon to spot all problems that can appear in what might otherwise look like simple refactoring.

src/sas/qtgui/Perspectives/Fitting/UnitTesting/ConstraintWidgetTest.py
```diff
@@ -148,11 +148,11 @@ def testOnTabCellEdit(self, widget, mocker):
 
         # disable the tab
         widget.tblTabList.item(0, 0).setCheckState(0)
-        assert widget.tabs_for_fitting["test_tab"] == False
```
krzywon, line 151: This and the 3 other changes in this file should be `_ is False` instead of `not _`. The latter checks the truthiness of the value, while the former matches `False` and only `False`.

src/sas/qtgui/Plotting/MaskEditor.py
```diff
@@ -218,7 +218,7 @@ def updateMask(self, mask):
         temp_mask = temp_mask / temp_mask
         temp_mask[mask] = temp_data.data[mask]
 
-        temp_data.data[mask == False] = temp_mask[mask == False]
```
krzywon, line 221: This is an array slice. It is looking for an explicit value in the array. This should go back to the original state.

src/sas/qtgui/Plotting/Masks/BoxMask.py
```diff
@@ -150,7 +150,7 @@ def _post_data(self):
         mask = Boxcut(x_min=x_min, x_max=x_max, y_min=y_min, y_max=y_max)
 
         if self.is_inside:
-            out = (mask(data) == False)
+            out = (not mask(data))
```
krzywon, line 153: `out` is an array of booleans. This and all other masking changes need to revert back to checking for explicit values.

src/sas/sascalc/pr/invertor.py
```diff
@@ -57,7 +55,8 @@ def help():
 
 class Invertor():
 
-    def __init__(self, logic: InversionLogic):
+    # "InversionLogic" type hint for "logic" removed to avoid circular import
+    def __init__(self, logic):
```
krzywon, line 59: We should keep type hints in most cases. To get around circular imports, and only use the import for type checking sake, you can try:

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from sas.qtgui.Perspectives.Inversion.InversionLogic import InversionLogic
```

test/size_distribution/MaxEntTest.py
```diff
@@ -64,7 +64,7 @@ def test_noRes(data1):
     # TODO: Need to understand how IRENA is reporting bins. It seems to be using bin 1 and last as the
     #       values input from the user while we are starting from the midpoint of what should be the
     #       the first bin if the min diameter is the bottom edge of the bin.
-    assert convergence[0][0] == True
+    assert convergence[0][0]
```
krzywon, line 67: A test checking for an explicit value? Not sure, but using `is True` might be better.

src/sas/qtgui/Perspectives/Fitting/FittingOptions.py
```diff
@@ -100,10 +100,10 @@ def assignValidators(self):
         for option in bumps.options.FIT_FIELDS.keys():
             (f_name, f_type) = bumps.options.FIT_FIELDS[option]
```
jamescrake-merani, line 101: Is this `f_type`  value still needed now? Could it be discarded with `_`?

jamescrake-merani (approved):
One minor comment, but still happy to approve.

I did quickly test by running the application, and I noticed that the icons are now missing. However, this problem seems to be present on `main` as well, so I don't think its this PR that caused that. I've raised a separate issue (#3551).

Issue #3310, opened 2025-04-30
Unit tests for the calculation portion of Shape2SAS


jamescrake-merani:

klytje:
Can you be more specific about what exactly "calculation portion" includes?

krzywon:
Currently, we have no unit test coverage for the shape2sas calculations run in the `src/sas/sascalc/shape2sas` package. This is asking for unit tests to ensure the calculations run by Shape2SAS are producing reliable and accurate results to be sure any future changes to the algorithms return the same, or similar values.

PR #3544, opened 2025-07-27
shape2sas plugin models

klytje:
## Description
This PR adds support for generating and fitting plugin models from the `shape2sas` model builder. 

### GUI
* The `to plugin model` button in the main `shape2sas` window has been re-enabled and fully implemented.
* The shape preview plot is now automatically updated upon making changes to the structure. I found the computational impact of this to be negligible, and it is a nice feature to have when building models. 
* The embedded log inside the constraint builder window is now fully functional, and reports common user errors in the constraints script (undefined vars, invalid constraints, python errors, ...)

### Constraints
Constraints are now fully supported. 
* Constraints are essentially just a block of standard Python code that is copy/pasted into the body of the intensity calculations of the plugin model script. 

Several convenience variables have been added to simplify the user code:
1. 'delta'-variants are defined for all shape parameters and may be used to easily constrain variables. 
    `dR1 = dR2` where dR2 is a fit parameter will propagate all changes in dR2 to dR1. Note how this is very different to writing `R1 = R2` (also a valid constraint) which will _fix_ their current values to be identical. Using delta variables allows the initial values to be different while still tracking changes.  
    
    A barbell (sphere 1, cyl 2, sphere 3) may therefore be constrained as: 
        `dR3 = dR1`
        `dL2 = -2*dR1`
    where R1 is a fit parameter. 
2. 'COM' may be used to easily link all center-of-mass parameters (`COMX, COMY, COMZ`). 
    `dCOM2 = dCOM1` is automatically expanded to `dCOMX2, dCOMY2, dCOMZ2 = dCOMX1, dCOMY1, dCOMZ1`. 

Most of the complexity of this PR comes from supporting delta parameters. 

### Efficiency
* The intensity calculations may now be performed with `AUSAXS`, which is _significantly_ faster (I measured 30x). I am not sure how feasible the fitting process is without this change. 
* To limit the plugin script size, delta variables are only defined for the shape parameters appearing in the constraint text. Though this is fine when using the GUI window to create the script, it means that manually modifying the script is not quite as simple. 
* The standard optimization algorithm (Levenberg-Marquardt) performs poorly for `shape2sas` fits. I am honestly quite surprised that this is the default option in SasView as I cannot imagine it performs well for any typical SasView problem.

## How Has This Been Tested?

This will require significant testing before being merged, though I don't have any good data to perform this testing with. The simulated data from `shape2sas` has too large error bars to be useful I think. 

## Todo
* [ ] Documentation
* [ ] Testing
* [ ] Clean up debug info from code

## Review Checklist:

**Installers**


llimeht:
> It is _not_ ideal that the automatic `ruff` fixes may break the build because it does not understand transitive imports. 

I'm surprised ruff isn't complaining about the `*` imports as a matter of style, particularly without `__all__` being explicit about what is to be imported. 

This might be a good opportunity to get rid of the * imports in this new code? (And of course progressively remove them in old code)

src/sas/sascalc/shape2sas/Models.py
```diff
@@ -0,0 +1,436 @@
...
+             + self.y_add * (np.cos(self.gam) * np.sin(self.beta) * np.sin(self.alpha) - np.sin(self.gam) * np.cos(self.alpha))
+             + self.z_add * (np.cos(self.gam) * np.sin(self.beta) * np.cos(self.alpha) + np.sin(self.gam) * np.sin(self.alpha)))
+
+        y_rot = (self.x_add * np.sin(self.gam) * np.cos(self.beta)
+             + self.y_add * (np.sin(self.gam) * np.sin(self.beta) * np.sin(self.alpha) + np.cos(self.gam) * np.cos(self.alpha))
+             + self.z_add * (np.sin(self.gam) * np.sin(self.beta) * np.cos(self.alpha) - np.cos(self.gam) * np.sin(self.alpha)))
+
+        z_rot = (-self.x_add * np.sin(self.beta)
+            + self.y_add * np.cos(self.beta) * np.sin(self.alpha)
+            + self.z_add * np.cos(self.beta) * np.cos(self.alpha))
```
pkienzle, line 98: Be sure the application of the Euler angles is consistent with other places in SasView. For my own version shape2sas in the sasmodels explore directory [here](https://github.com/SasView/sasmodels/blob/d5c04652e79532567afbf1df1a9cad457b16e606/explore/realspace.py#L29-L102) I'm using z-y-x convention for φ-θ-ψ, which is consistent with the convention I'm using for oriented particles in the sasmodels kernels.

src/sas/sascalc/shape2sas/Models.py
```diff
@@ -0,0 +1,436 @@
...
+        """Transforms the points"""
+        alpha, beta, gam = self.rotation
+        rotp_x, rotp_y, rotp_z = self.rotation_point
+        alpha = np.radians(alpha)
+        beta = np.radians(beta)
+        gam = np.radians(gam)
+        com_x, com_y, com_z = self.com
+
+        x, y, z = Rotation(x, y, z, alpha, beta, gam, rotp_x, rotp_y, rotp_z).onRotatingPoints()
+        x, y, z = Translation(x, y, z, com_x, com_y, com_z).onTranslatingPoints()
```
pkienzle, line 163: Your Rotation and Translation classes would be better written as functions. You create them, call them immediately, then through them away.

klytje:
> Ruff does complain about the * imports, the specific rule is [F405](https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/), which is part of our ruleset.

Personally I'd disagree with this rule - while wildcard imports from larger modules is obviously a terrible idea, I have here specifically _designed_ the Typing and Models modules for being imported in this manner. Explicitly listing every symbol in the import when we want everything is just unnecessary word salad.  But my opinion on this is not strong enough that I'd reject this suggestion, so I can change it. 

> The cause of the problems here was the use of the typing module for type hints
> ...
> We now have an ADR proposal for type hints https://github.com/orgs/SasView/discussions/3170. In this ADR it is explicitly stated that built-in types should be used instead of the typing module.

Again I'd like to point out that this is inherited from the existing shape2sas implementation. `ruff` broke existing code that I only moved somewhere else. I guess this will stop happening once the entire codebase has been updated to follow the new guidelines.

src/sas/sascalc/shape2sas/Models.py
klytje: The conventions differ. Shape2SAS uses the intrinsic ZYX convention, while you seem to be using ZYZ with exchanged angles (see the new test case [here](https://github.com/SasView/sasview/blob/61da50e625981f99dcb89b28e4710b56735396ba/test/sascalculator/utest_sas_gen.py#L311C2-L339C65)).

DrPaulSharp:
Alongside the comments from @butlerpd, @klytje could you please rebase this PR onto main, which will make reviewing more straightforward and ensure any potential merge conflicts are dealt with. Rebasing should also introduce the fix to the currently failing CI.

klytje:
@butlerpd I (attempted to) address all of @pkienzle's comments except for those related to improving the sampling efficiency. I'd rather not touch any of that stuff without an extensive test suite to validate against (#3310), which is out of scope for this PR. 

The only remaining issue is that of Euler rotation convention, where `shape2sas` is using a different (but more common) convention than the `sasview` kernel; see e.g. [the new unit test](https://github.com/SasView/sasview/blob/b68868273312da78591bf169a805fa3dbf3ff17e/test/sascalculator/utest_sas_gen.py#L311-L339) (which should probably be removed; we just need to decide on which convention to follow). 

We also have the issue of user testing, which is what has really been holding up this PR. I've done my best to make it as easy to use and robust as possible, but I need actual user feedback on these aspects.

src/sas/qtgui/Calculators/Shape2SAS/Constraints.py
```diff
@@ -55,75 +53,297 @@ def __init__(self, parent=None):
...
+                            constraints.append(node)
+            return params, imports, constraints
+
+        def extract_symbols(constraints: list[ast.AST]) -> tuple[list[str], list[str]]:
+            """Extract all symbols used in the constraints."""
+            lhs, rhs = set(), set()
+            lineno = {}
+            for node in constraints:
+                # left-hand side of assignment
+                for target in node.targets:
```
krzywon: The match in `node.targets` and `ast.walk` loops is exactly the same. Either find a way to combine the two loops, put the match/case into its own function/method, or both.

src/sas/qtgui/Calculators/Shape2SAS/Constraints.py
```diff
@@ -55,75 +53,297 @@ def __init__(self, parent=None):
...
-        return importStatement, parameters, translation, checkedPars
+            for node in ast.walk(tree):
+                match node:
+                    case ast.ImportFrom() | ast.Import():
+                        imports.append(node)
+
+                    case ast.Assign():
+                        if len(node.targets) != 1 or isinstance(node.targets[0], ast.Tuple) or isinstance(node.value, ast.Tuple):
+                            self.log_embedded_error(f"Tuple assignment is not supported (line {node.lineno}).")
+                            raise ValueError(f"Tuple assignment is not supported (line {node.lineno}).")
```
krzywon: This and a number of other errors are both logged and thrown. These could all be simplified using a functools wrapper - https://stackoverflow.com/questions/11420464/catch-exceptions-inside-a-class.

src/sas/sascalc/shape2sas/Models.py
```diff
@@ -0,0 +1,407 @@
...
+        #Show information about the model and its subunits
+        N_remain = []
+        for j in range(self.Number_of_subunits):
+            srho = rho[j] * self.p_s[j]
+            N_remain.append(N[j] - N_exclude[j])
+            print(f"        {N[j]} points for subunit {j}: {self.subunits[j]}")
+            print(f"             Point density     : {rho[j]:.3e} (points per volume)")
+            print(f"             Scattering density: {srho:.3e} (density times scattering length)")
+            print(f"             Excluded points   : {N_exclude[j]} (overlap region)")
+            print(f"             Remaining points  : {N_remain[j]} (non-overlapping region)")
```
krzywon: `logger.debug()` again?

butlerpd:
@klytje - I hope to do some quick functionality testing but after noting @jellybean2004 use of copilot for code review I thought I'd give it a shot -- sorry for using you as an experiment. It doesn't seem like it found much other than it suggests functions arguments should be `func(self, arg1, ..)` I have never exactly understood this detail of OO programming and when to use it and when not to. Please feel free to ignore if the suggestions make no sense.

Again apologies for being my guinea pig

butlerpd (approved):
I forgot to click approve. And looking at @pkienzle review his one reason to not click on "approve" was that the plugin generator button was disabled (hidden) which is now the expected behavior it seems this is essentially approved.

However, the confusion does speak to why editing the description of the PR would be quite helpful!! even for future digging when somebody is trying to figure out what was done. I would hope that the description gets edited before closing this PR?

src/sas/qtgui/Calculators/Shape2SAS/Constraints.py
```diff
@@ -55,75 +53,297 @@ def __init__(self, parent=None):
...
+
+        def as_ast(text: str):
+            try:
+                return ast.parse(text)
+            except SyntaxError as e:
+                # log most recent traceback error
+                all_lines = traceback.format_exc().split('\n')
+                last_lines = all_lines[-1:]
+                traceback_to_show = '\n'.join(last_lines)
+                logger.error(traceback_to_show)
```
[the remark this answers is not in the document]
klytje: I thought perhaps the traceback would be too long and uninformative for the average user. Not sure if it should be included or not.

src/sas/sascalc/shape2sas/Models.py
```diff
@@ -0,0 +1,407 @@
...
+
+        N_total = sum(N_remain)
+        print(f"        Total points in model: {N_total}")
+        print(f"        Total volume of model: {volume_total:.3e} A^3")
+        print(" ")
+
+        return x_new, y_new, z_new, p_new, volume_total
+
+    def onGeneratingAllPoints(self) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, float]:
```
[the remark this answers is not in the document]
klytje: I'm not a fan of the overall point generation logic and it should definitely be simplified. However, I think that would be wasted work as point generation will have to be completely refactored in a future PR anyway to solve the optimization issue.

src/sas/sascalc/shape2sas/TheoreticalScattering.py
```diff
@@ -0,0 +1,328 @@
...
+
+        ## calc h(r) with/without polydispersity
+        if polydispersity > 0.0:
+            Dmax = np.amax(dist) * (1 + 3 * polydispersity)
+            r_max = Dmax * ratio_rmax_dmax
+            r, hr_1 = self.generate_histogram(dist, contrast, r_max, Nbins)
+            N_poly_integral = 10
+            factor_range = 1 + np.linspace(-3, 3, N_poly_integral) * polydispersity
+            hr, norm = 0, 0
+            for factor_d in factor_range:
```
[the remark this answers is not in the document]
klytje: I don't see how `factor_range` can possibly be of type `int`?

src/sas/sascalc/shape2sas/structure_factors/StructureDecouplingApprox.py
```diff
@@ -0,0 +1,96 @@
...
+from sas.sascalc.shape2sas.HelperFunctions import sinc
+
+
+class StructureDecouplingApprox:
+    def __init__(self, q: np.ndarray,
+                 x_new: np.ndarray,
+                 y_new: np.ndarray,
+                 z_new: np.ndarray,
+                 p_new: np.ndarray):
+        self.q = q
```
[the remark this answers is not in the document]
klytje: How should it be formatted? The style seems to vary quite a bit within sasview. In C++ I'd do something like:
```
void func(
    [long line] x,
    [long line] y,
) {
    code;
}
```

klytje:
@krzywon Thanks for the review. I'll merge once we've resolved your open review issues; I left comments on the remaining ones.

klytje:
@pkienzle Thanks for re-reviewing. I have written down your suggestions and will open either an issue or discussion with them for future reference, when the implementation will have to be refactored to eliminate the randomness during optimization. 

Regarding the calculation efficiency: it is definitely doable, even for somewhat large collections of points. The only limitation is the inherent randomness of the generated points; see the previous comment on this issue. The constraints window button has been intentionally hidden until this issue has been resolved in a future PR.

src/sas/sascalc/shape2sas/ExperimentalScattering.py
```diff
@@ -0,0 +1,119 @@
...
+        ## simulate exp error
+        #input, sedlak errors (https://doi.org/10.1107/S1600576717003077)
+        #k = 5000000
+        #c = 0.05
+        #sigma = noise*np.sqrt((I+c)/(k*q))
+
+        ## simulate exp error, other approach, also sedlak errors
+
+        # set constants
+        k = 4500
```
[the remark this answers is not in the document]
klytje: I'm pretty sure these are part of the Virtual Scattering calculations (unrelated to this PR), so perhaps this change should be done in a separate PR.

krzywon:
On second look, I believe this is now ready. Unless there are objections, I'll merge this by the end of this week.

src/sas/sascalc/shape2sas/Models.py
pkienzle: Nevermind...I keep forgetting that shape2sas is already in sasview. Changing the interpretation of the angles is outside the scope of this PR.

pkienzle (approved):

PR #3552, opened 2025-07-28
Replace usage of connection proxy with the requests library.

### LINKED ISSUE #3420: New Version Available not working on MacOS
**Describe the bug**
While testing on macOS, I noticed the following error message on startup.

```
2025-05-28 10:36:47,354 : INFO : NewVersionAvailable (NewVersionAvailable.py:158) :: Error getting latest sasview version module 'plistlib' has no attribute 'readPlist'
```

**Expected behavior**
Not only should this error not happen, I also noticed that the log for it appears to be at the info level when really it should be at the error level instead.

**SasView version:**
 - Version: 6.1.0 Alpha 2

**Operating system:**
 - OS: macOS

jamescrake-merani:
## Description

This PR aims to fix #3420 by removing our reliance on the `ConnectionProxy` class, and instead use the `get` request function from the `requests` library.

## How Has This Been Tested?

Tested on `run.py`, and the wheel. Unfortunately, I haven't been able to test on the installer as I'm not sure if there is a way to fake the version to an older one to test the popup menu. Does anyone know how to do this? It might be useful if we add a debug version override flag for this purpose.

## Review Checklist:

**Installers**
  - [x] **Wheels** installer (GH artifact) has been tested (installed and worked)

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -629,7 +629,8 @@ class GuiManager:
         version number has been obtained. If the check is being done in the
         background, the user will not be notified unless there's an update.
 
-        :param version_str: version string
+        :param version_info: tuple of version string, download url, and the
+        version object.
         """
         version_str, download_url, _ = version_info
         try:
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -44,6 +44,8 @@ from sas.qtgui.Utilities.About.About import About
 
 # General SAS imports
 from sas.qtgui.Utilities.ConnectionProxy import ConnectionProxy
+from sas.qtgui.Utilities.SasviewLogger import setup_qt_logging
+from sas.qtgui.Utilities.NewVersion.NewVersionAvailable import get_current_release_version
 
 # from sas.qtgui.Utilities.DocViewWidget import DocViewWindow
 from sas.qtgui.Utilities.DocRegenInProgess import DocRegenProgress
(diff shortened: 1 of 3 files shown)


src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -640,26 +630,20 @@ def log_imported_packages(self):
         """
         PackageGatherer().log_imported_packages()
 
-    def processVersion(self, version_info):
+    def processVersion(self, version_str):
```
rozyczko, line 633: `version_str` is the return value from `get_current_release_version()`, which returns a tuple of (str, str, Version).
Here, you're treating this tuple as a string? Or as a dict in line 646?

src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
```diff
@@ -81,28 +80,26 @@ def cancel(self):
...
+        version_info = response.json()
+        logger.info("Connected to www.sasview.org. Received: %s", version_info)
 
         version_string = version_info["version"]
         url = version_info["download_url"]
 
         return version_string, url, parse(version_string)
 
-
+    except ConnectionError:
```
rozyczko, line 96: Shouldn't that be `requests.exceptions.ConnectionError`?

https://stackoverflow.com/questions/16511337/correct-way-to-try-except-using-python-requests-module

src/sas/qtgui/MainWindow/GuiManager.py
jamescrake-merani: version_info was originally a dictionary of the JSON fetched from the server. I was going to just make this method take in the string only, but it looks like I mistakenly read the return type of `get_current_release_version` as a string rather than a tuple. I've changed this so it takes the tuple in, and destructures it to get the necessary values.

src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
```diff
@@ -81,28 +80,26 @@ def cancel(self):
...
-    # Don't do anything if we can't reach the server
-    if response is None:
-        return None
-
     try:
-        content = response.read().strip()
-        logger.info("Connected to www.sasview.org. Received: %s", content)
-
-        version_info = json.loads(content)
+        response = requests.get(web.update_url)
```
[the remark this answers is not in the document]
jamescrake-merani: Good idea. I'll reuse the value we already had from the config.

jamescrake-merani:
I've rebased onto main. Will wait on CI, and then I'll merge.

PR #3554, opened 2025-07-28
Ensure libausaxs is downloaded when building installers

### LINKED ISSUE #3545: libausaxs is not included in installers
`libausaxs` is supposed to get downloaded during the build process through the `setup.py` script: 

https://github.com/SasView/sasview/blob/934d32c917e8256b85257318532c8af79bf3795a/setup.py#L97C1-L99C34

However, when going through the history, it seems like the `setup.py` build script was deprecated immediately after merging `libausaxs` support into the main branch: https://github.com/SasView/sasview/commit/824d7824ebb171b50be37cd010f96009d3b1e3f9

`libausaxs` is therefore not currently being included in the installers (I checked v6.1), and likely never was. 

---

The easiest fix would be to manually call `fetch_external_dependencies()` in the GitHub build scripts. However, seeing as the current approach to include `libausaxs` has already been criticized (https://github.com/SasView/sasview/issues/3449), perhaps it would be better to implement a more elegant solution.

klytje:
## Description

`libausaxs` should now be present and included in the installers. 

Fixes #3545, needed for #3544

## How Has This Been Tested?

The Linux installer has been tested and verified to now include `libausaxs`. The issue was that `libausaxs` was previously only being downloaded for test builds, but not for the installers, which were therefore created without it. As functionality was already verified in #2882 and tests are still passing (including the `libausaxs` test), I don't think we have to do anything further here. 

## Review Checklist:

**Installers**
- [x] This will affect the **installers**, so
  - [x] **Linux** installer (GH artifact) has been tested (worked) 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,10 +160,6 @@ jobs:
         run: |
           uv pip install dist/*whl
 
-      - name: Fetch external AUSAXS C library
-        run: |
-          python build_tools/get_external_dependencies.py
-
       - name: Install Python dependencies
         if: ${{ ! env.RELEASE_BRANCH_BUILD }}
         run: |
@@ -176,6 +172,10 @@ jobs:
         if: ${{ env.RELEASE_BRANCH_BUILD }}
         run: |
           python -m pip install --no-deps -r build_tools/requirements-release-${{ matrix.os }}.txt
+
+      - name: Fetch external AUSAXS C library
+        run: |
+          python build_tools/get_external_dependencies.py
 
       ### Actual building of sasview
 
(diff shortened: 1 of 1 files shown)


krzywon (commented):
For cleanliness, I would suggest this be a separate step in the CI file, not as part of the pip install.

```
      - name: Fetch AUSAXS C library
        run: |
          python build_tools/get_external_dependencies.py
```

klytje:
@jamescrake-merani Yes I agree. This PR is only for restoring functionality. Changing to pypi can be done in a future PR. I've started looking at creating a proper python wrapper for `libausaxs` [link](https://github.com/AUSAXS/pyAUSAXS), though it won't be done anytime soon.

krzywon (approved):
As noted by @jamescrake-merani, this is a reasonable short-term solution. Once this is on pypi, @klytje, please either let us know, or put in a pull request, that removes this step and adds ausaxs to our requirements file.

PR #3560, opened 2025-07-30
Adds Pyupgrade (UP) ruleset and rules SIM118 & SIM300 to ruff linter

### LINKED ISSUE #3498: Enable Ruff for PRs

DrPaulSharp:
## Description

This PR adds the pyupgrades ruleset to the ruff linter, which acts to remove python 2 features from the code. I have also added the rules SIM118, which changes `key in dict.keys()` to `key in dict`, and SIM300, which fixes Yoda conditions (e.g., `0 < q` is changed to `q > 0`). I have applied the automatic fixes for these rules. Note that this does not catch every instance, particularly for SIM118 (the automatic fix is only applied when the linter is certain that the type is a dict).

I have also set out the full set of rules we are using in the `pyproject.toml`, along with a link to the rules. I'm happy to include this elsewhere if there are any suggestions. This PR completes the task of adding rules to the ruff linter, although I am open to feedback.

This leaves a total of 274 ruff errors in SasView, which breaks down amongst the rulesets as:

- E: 86 errors
- F: 96 errors
- I: 0 errors
- UP: 36 errors
- SIM118: 56 errors
- SIM300: 0 errors

I have also chosen to ignore three rules, with the aim of keeping the number of errors manageable for anyone looking to tackle them. The ignored rules are:

- E501 -  line too long (leave this to the ruff formatter) (260 errors)
- UP008 - Use `super()` instead of `super(__class__, self)` (106 errors)
- UP031 -  Use format specifiers instead of percent format (504 errors)

I'm open to feedback on these choices.

Fixes #3498 

## How Has This Been Tested?

I have run the test suite, and the tests pass. I have also launched and inspected the GUI, and there are no immediate issues.

## Review Checklist:

**Installers**


build_tools/release_automation.py
```diff
@@ -240,7 +240,7 @@ def update_sasmodels_init(version):
     """
     init_file = os.path.join('sasmodels', 'sasmodels', '__init__.py')
     output_lines = []
-    with open(init_file, 'r') as f:
+    with open(init_file) as f:
```
krzywon, line 243: Personally, I'm not a fan of removing the explicit setting of read in the calls to open, but understand it's the default, so understand why this is.

src/sas/qtgui/UnitTesting/TestUtils.py
```diff
@@ -9,11 +9,11 @@ def WarningTestNotImplemented(method_name=None):
     Test name retrieved from stack trace.
     """
     if method_name is not None:
-        print(("\nWARNING: %s needs implementing!"%method_name))
+        print("\nWARNING: %s needs implementing!"%method_name)
```
krzywon, line 12: Is there a reason `%s` formatters were left when `"".format()` formatters were converted to f-strings?

src/sas/sascalc/fit/models.py
```diff
@@ -34,7 +33,7 @@ def plugin_log(message):
     """
     now = time.time()
     stamp = datetime.datetime.fromtimestamp(now).strftime('%Y-%m-%d %H:%M:%S')
-    with io.open(PLUGIN_LOG, 'a', encoding='utf-8') as out:
+    with open(PLUGIN_LOG, 'a', encoding='utf-8') as out:
```
krzywon, line 36: I believe this change should be fine at this point, but ant to be sure. @pkienzle, can this be changed to a bare `open` command?

krzywon (approved):
I think this should be acceptable. I do have a few comments and questions, but they shouldn't hold up this PR.

src/sas/qtgui/UnitTesting/TestUtils.py
DrPaulSharp: The format statements could be converted automatically, but there is not an automatic fix available for the `%s` formatters, and there are 504 instances of this in the code. I think this is something we will have to tackle more gradually. Note that this error is rule `UP031`, which I added to the list of rules to ignore.

DrPaulSharp: The fix safety is discussed here: https://docs.astral.sh/ruff/rules/printf-string-formatting/

krzywon:
I expect to merge this, assuming the CI is successful.

PR #3567, opened 2025-08-15
Adds routine "clearGuiValues" for the Inversion perspective

### LINKED ISSUE #3469: Inversion Perspective: Removing data doesn't clear the widget
**Describe the bug**
This is a bug resulting from the refactor of the Inversion perspective.

**To Reproduce**
1. Go to the Inversion Perspective
2. Send data to the perspective
3. Click the 'remove' button in the tab.
4. The data will be removed, but calculations like the estimated values will remain.

**Expected behavior**
All of the values in the widget should be removed. The widget should basically look exactly like it did when it was first created.  

**SasView version:**
 - Version: 6.1.0
 
**Operating system:**
 - OS: Linux (though unlikely to be platform specific)

DrPaulSharp:
## Description

This PR introduces a routine "clearGuiValues" to the inversion perspective. The purpose of this routine is to clear the calculated values in the perspective when a lone data entry is removed.

Fixes #3469 

## How Has This Been Tested?

I have followed the procedure in the issue and found that the values are removed and the widget looks like when first created.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
+++ b/src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
@@ -109,6 +109,9 @@ class InversionWidget(QWidget, Ui_PrInversion):
         self.batchResultsWindow: BatchInversionOutputPanel | None = None
         self.batch_dict: dict[str, Any] | None = None
 
+        self.input_boxes = [self.noOfTermsInput, self.regularizationConstantInput, self.maxDistanceInput,
+                             self.minQInput, self.maxQInput, self.slitHeightInput, self.slitHeightInput]
+
         self.updateGuiValues()
         self.events()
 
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Perspectives/Inversion/InversionWidget.py
```diff
@@ -290,6 +291,26 @@ def updateGuiValues(self):
             self.posFractionValue.setText(format_float(out.pos_frac))
             self.sigmaPosFractionValue.setText(format_float(out.pos_err))
 
+    def clearGuiValues(self):
+        self.noOfTermsInput.setText("")
```
jamescrake-merani, line 295: You could create a list of all the UI elements you want to reset, and then do a for loop over them, setting the text of each to empty string.

jamescrake-merani (changes requested):
Fixes the bug on my machine, but see suggestion.

PR #3569, opened 2025-08-18
Release 6.1.1 branch

jamescrake-merani:
**This is not to be merged into main until after the release.**

This PR is being made now mostly to make it easier to check the PR.


llimeht:
The changes from #3604 in `.github/workflows/ci.yml`, `.github/workflows/matrix.py` and `build_tools/requirements-release-windows-*.txt` where `windows-latest` is moved to `windows-2022` should be reverted in this merge or straight afterwards. That issue fixed in `main` via #3606.

DrPaulSharp:
@llimeht The CI fails on this pull request at the ruff stage, where it is checking for automatic fixes for linting violations. When `pre-commit-ci/lite-action` attempts to add a commit to the branch fixing linting violations, it is stuck in a pending status, with the message "Waiting for status to be reported — running".

> @DrPaulSharp we don't want to have to push more commits into the `release-6.1.1` branch, but merging this will immediately leave us with broken CI.
> 
> * how do we know why it is broken? (the ruff action unhelpfully says nothing more than `failed with exit code 1`; should we have `--verbose` added to the ruff args? both here and sasmodels - I wasted ages on this the other day with sasmodels CI failing at the ruff step)

I have run ruff locally on this branch, using versions 0.13.2 (latest version on September 30th, when the last commit was added to this branch) and 0.14.1 (the current latest version). For both versions, ruff reports that there are no automatic fixes available, and therefore the CI should pass.

Given this, I believe that the ruff action is failing erroneously. This would then explain why `pre-commit-ci/lite-action` is stuck as well, as it cannot make a commit where there are no changes, because no linting errors have been fixed. I have created a mirror of this PR in my SasView [fork](https://github.com/DrPaulSharp/sasview-test/pull/1). In it, I also find that the ruff action fails on the zenodo commit (`c5ac312`), but adding a commit where I simply add "verbose" as an argument to ruff sees the action pass. This suggests the failure of the ruff action is not due to linting errors on this branch. Unfortunately, I cannot say why the action is failing.

I agree that the output of the ruff action leaves a lot to be desired. You can see the effect of adding the verbose flag on my fork, it states the files that are being checked but not what the linting errors are. I'll make an issue to investigate this further and see if there's anything I can do.

> * as an aside about the ruff config, isn't `--exit-non-zero-on-fix` incompatible with `pre-commit-ci/lite-action`? if ruff finds things to complain about, it will cause CI to fail immediately and so the items don't get fixed in the next step?

No, this is not the case. Without the `--exit-non-zero-on-fix` option, the ruff action passes having applied the fixes. `pre-commit-ci/lite-action` then creates a new commit containing those fixes, and the tests are run on BOTH the original commit, and the commit from `pre-commit-ci/lite-action`. However, we do not want to waste time and resources running tests on a commit that is no longer the tip of the branch, so we use this option to stop the CI on the original commit, and allow it to proceed on the commit from `pre-commit-ci/lite-action`. If ruff exits with a non-zero status, it is still the case that pre-commit runs. see for example commit `ac269b5` on this branch where ruff fails but pre-commit runs afterwards.

> * we will need a PR to fix `main` straight away - do we know what to do and can it be done in a way that doesn't make merge conflicts with every single open PR?

If the merge still results in the CI failing, one possibility is to create a PR making the changes in the CI from `windows-2022` to `windows-latest`, as you noted above. If I'm right, this should resolve the issue.

Please let me know what you think and how you intend to proceed. I'm happy to make myself available if you need any assistance with this.

PR #3573, opened 2025-08-19
Adds local pre-commit hook for ruff linting auto fixes

DrPaulSharp:
## Description

This PR adds the config file for a local pre-commit hook to apply any automatic fixes available for ruff linting errors. This was requested in the SasView technical meeting on 8/8/25 #3171.

The hook is installed by running `pre-commit install` in the local SasView directory. The hook is set up to run only when pushing to the remote repository. Once installed, the hook will check for any ruff linting errors with an automatic fix available. If it finds any, it will prevent the branch from being pushed, and apply the fixes locally. It is then up to the user to add these fixes and amend their last commit, or add a new commit including the fixes. If ruff linting errors are found that do not have an automatic fix available, they are ignored. Hence, we are looking for the same errors as in the github actions on the remote repo.

I am open to feedback on how the hook works. The PR is a draft for the duration of the 6.1.1 code freeze.

I know that we need to make changes to a developer guide to explain this, could someone please point me to where would be most appropriate.

## How Has This Been Tested?

I have tested by adding unused imports to a source code file, and pushing to a fork of this repo. When I push, the hook fails, and the unused imports are removed.

## Review Checklist:

**Installers**


jamescrake-merani:
pre-commit should probably be added as a dependency, except I'm not sure where it should go. I tried adding it to `requirements-dev.txt`, but I think these only get installed during the build process, and aren't available to the developer.

DrPaulSharp:
> pre-commit should probably be added as a dependency, except I'm not sure where it should go. I tried adding it to `requirements-dev.txt`, but I think these only get installed during the build process, and aren't available to the developer.

Hmm, I think I'll add it there for now and see what everyone thinks is best.

build_tools/requirements-dev.txt
```diff
@@ -4,3 +4,4 @@ hatch-build-scripts
 hatch-requirements-txt
 hatch-sphinx
 hatch-vcs
+pre-commit
```
krzywon, line 7: `ruff` is not a dependency of `pre-commit`, so it is also needed here.

IMO, this requirements file is the right place for this.

DrPaulSharp: `ruff` is a dependency of the astral maintained hook used here, see https://github.com/astral-sh/ruff-pre-commit/blob/main/pyproject.toml.

I have set up a fresh virtual env and confirmed that this hook can be run without needing to install ruff explicitly, it's taken care of when running the hook for the first time.

krzywon: I was finally able to get this to trigger and it installed ruff automatically, so I agree, not necessary.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+default_install_hook_types: [pre-commit, pre-push]
+
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.12.9
+    hooks:
+      # Run the linter, applying any available fixes
+      - id: ruff-check
+        stages: [ pre-push ]
```
krzywon, line 9: This only triggers on push. I think the pre-commit stage would also be useful. Side note, I pushed the commit https://github.com/SasView/sasview/commit/ad11516c6ecc4a1f18a77d2ad49a7eee4c1188c0, which has a bad indent and no blank line at the end of the file, was able to be pushed without issue.

Things we will need for this:
- A setup procedure (pre-commit install, I believe is all that is needed)
- A list of checks pre-commit will attempt to correct
- Triggering pre-commit and what happens if a fix is found

pkienzle:
I have this .pre-commit-config installed and it doesn't seem to do anything.

For example, if I add a spurious space to the end of a line then do a commit the space is still there:
```console
$ git switch -c touch
$ pre-commit install
pre-commit installed at .git/hooks/pre-commit
pre-commit installed at .git/hooks/pre-push
$ vi src/sas/__init__.py
...
$ git commit -a -m "touch"                
[touch fbd70fe17] touch
 1 file changed, 1 insertion(+)
$ git show -p                             
2025-09-12 fbd70fe17 touch Paul Kienzle

diff --git a/src/sas/__init__.py b/src/sas/__init__.py
index e11cdc6af..871886c19 100644
--- a/src/sas/__init__.py
+++ b/src/sas/__init__.py
@@ -7,3 +7,4 @@ __all__ = ["config", "__version__"]
 # Load the config file
 config.load()
 
+import uncertainties     <= console showing a red block where the blank at the end of the line was commited
```

The pre-commit hook in the bumps repo seems to be working better:

```console
$ git switch -c touch
$ vi bumps/__init__.py
...
$ git diff
$ git diff            
diff --git a/bumps/__init__.py b/bumps/__init__.py
index cd3dd4d..0a5093d 100644
--- a/bumps/__init__.py
+++ b/bumps/__init__.py
@@ -19,6 +19,9 @@ except ImportError:
 
 __schema_version__ = "1"
 
+# unused import with extra spaces after 
+import uncertainties    <= this shows a red box in my terminal for the space after the end of the line
+
 
 def data_files():
     """
$ git commit -a -m "touch"
eslint...............................................(no files to check)Skipped
prettier.............................................(no files to check)Skipped
check for added large files..............................................Passed
check for merge conflicts................................................Passed
check yaml...........................................(no files to check)Skipped
fix end of files.........................................................Passed
trim trailing whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

Fixing bumps/__init__.py

ruff-format..............................................................Passed
$ git diff
diff --git a/bumps/__init__.py b/bumps/__init__.py
index cd3dd4d..95f7582 100644
--- a/bumps/__init__.py
+++ b/bumps/__init__.py
@@ -19,6 +19,9 @@ except ImportError:
 
 __schema_version__ = "1"
 
+# unused import with extra spaces after
+import uncertainties.     <= red box is gone because space was removed
+
 
 def data_files():
     """

```

Further evidence of the problem: despite having the hook on sasmodels, my recent PRs for sasmodels all contain changes to the 2-Yukawa model despite none of them touching the 2-Yukawa code.

The commit action on the CI should not be touching files that are not part of my commit. I hope this will not cause merge conflicts because the commit hook is applying the same changes on multiple branches. These can be difficult to resolve if you are new to the project because the code will likely be unfamiliar.

These problems may go away once all outstanding PRs have been merged and all files have been seen by ruff.

Ideally the linter would only touch the code blocks that I'm modifying, not the entire file. Are there flags for the linter to limit its action within the file? This is clearly not possible in general: if you are removing the only use of a symbol from the file and lint is removing unused imports, then the import block will be modified even if you haven't touched it.

Note: I'm a bit annoyed that lint it reformatting the 2-Yukawa model. I would prefer to keep it as close to the matlab original as possible so that we can do diffs on the code base. Lint changes may be a problem later if there are updates to the Matlab code that need to be ported.

PR #3578, opened 2025-08-21
Converts type comments to type hints

DrPaulSharp:
## Description

I have converted type comments in the codebase to type hints, as discussed in #3170. I used the tool [com2ann](https://github.com/ilevkivskyi/com2ann) for the initial conversion before a bit of tidying up.

## How Has This Been Tested?

The code builds correctly and the tests pass.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/AssociatedComboBox.py
+++ b/src/sas/qtgui/Perspectives/Fitting/AssociatedComboBox.py
@@ -1,4 +1,4 @@
-from PySide6 import QtWidgets
+from PySide6 import QtGui, QtWidgets
 
 
 class AssociatedComboBox(QtWidgets.QComboBox):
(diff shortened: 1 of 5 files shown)


src/sas/qtgui/Perspectives/Fitting/AssociatedComboBox.py
```diff
@@ -7,10 +7,9 @@ class AssociatedComboBox(QtWidgets.QComboBox):
 
     When the combo box's current text is changed, the change is immediately reflected in the associated item; either
     the text itself is set as the item's data, or the current index is used; see constructor."""
-    item = None  # type: QtGui.QStandardItem
+    item: QtGui.QStandardItem = None
 
-    def __init__(self, item, idx_as_value=False, parent=None):
-        # type: (QtGui.QStandardItem, bool, QtWidgets.QWidget) -> None
+    def __init__(self, item: QtGui.QStandardItem, idx_as_value: bool = False, parent: QtWidgets.QWidget = None) -> None:
```
jamescrake-merani, line 12: Do we want to keep the `-> None`? I normally leave it out if I'm type hinting a function that doesn't return anything. But I don't think we have anything written down as to what our style guide is on this.

src/sas/qtgui/Plotting/QRangeSlider.py
```diff
@@ -1,7 +1,11 @@
 """Double slider interactor for setting the Q range for a fit or function"""
+
 import numpy as np
+from PySide6.QtCore import QEvent
+from PySide6.QtWidgets import QLineEdit, QTextEdit
```
jamescrake-merani, line 5: I'm a bit worried about introducing a circular dependency here. The risk should be low, but I think best to wait until after the release until merging, and then we'll have plenty of time to deal with any issues that may crop up.

DrPaulSharp: I've already found two as a result of type hints (one in this PR and one previously) so we have to keep an eye out.

Issue #3580, opened 2025-08-21
Labels in the corfunc perspective not behaving properly


jellybean2004:
**Describe the bug**
In the Corfunc perspective, the Guinier label does not reposition itself when the sliders are moved.

**To Reproduce**
1. Go to the Corfunc perspective
2. Send data for it
3. Alter any slider
4. See that the 'Data' and the 'Porod' labels are repositioned to the centre of the range, but the 'Guinier' label remains stationary.

**SasView version (please complete the following information):**
 - Version: 6.1.1

**Operating system (please complete the following information):**
 - OS: Windows 11

**Installation method (please complete the following information):**
 - Installed: pip

PR #3587, opened 2025-09-03
Corfunc fix

### LINKED ISSUE #3579: Corfunc values resetting to default after new data sent to it.
**Describe the bug**
A bug report came through on the mailing list relating to the corfunc perspective. When new data is sent to the perspective, the guinier, and porod state/end regions are set back to their default values, where in previous versions their values remained the same.

**To Reproduce**
1. Go to the corfunc perspective.
2. Send data to it.
3. Alter any of the start/end points for the guinier, or porod region.
4. Send new data to the perspective.
5. Observe that the values have now reset.

**Expected behavior**
The start/end values for the guinier, porod region to remain the same as they were previously rather than being reset to default values.

**SasView version:**
 - Version: 6.1.0

**Operating system:**
 - OS:  Windows 11

**Installation method:**
 - Installed: Windows Pyinstaller distribution.

jellybean2004:
## Description

Corfunc Guinier and Porod start/end values no longer reset to default values when new data is loaded. A checkbox is included in the dialogue box, warning the user that new data will erase the current loaded data, which allows the user to reset to defaults. If a particular value is out of range for the new data, only that value will be reset to the default. 

Pops a dialogue box when the entered values are out of range, showing the users the max and min values. Rounds the value checked to 8sf to omit floating point precision errors

Fixed issues with the Corfunc perspective range:

- Users could go out of range by clicking outside, which has now been fixed.
- The textbox would remain red at times, even after the value is within range
- Inconsistencies between the two code files regarding the appropriate values
- Guinier label remaining static

Fixes #3579 and #3580 

## How Has This Been Tested?

Manually tinkered with the start/end values to test various cases, including if the data falls out of range. The reset check button was tested similarly.

## Review Checklist:

**Documentation**

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
+++ b/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
@@ -433,9 +433,6 @@ class CorfuncWindow(QtWidgets.QDialog, Ui_CorfuncDialog, Perspective):
 
     def check_extrapolation_entry(self, fits_enabled: list[str]):
         """ Disable Go button if extrapolation ranges empty or invalid """
-        # if fits_enabled == []:
-        #     self.disable_go_button("No fits enabled")
-        #     return
 
         if "background" not in fits_enabled:
             if self.txtBackground.text() == "":


jellybean2004:
> Nice work @jellybean2004 ! I've just done more functionality testing on CI build #4734 and your changes are good.
> 
> Playing with the interface some more I do have some additional comments:
> 
> * the message 'Reset Q range to defaults' on the popup when loading in a new dataset could perhaps be confusing? The 'Q range' you mean here is actually the limits for the start/end of the Guinier/Porod extrapolations. Whereas Users might think you mean the Q range of the data (and thereby get confused). I wonder if some alternative text such as 'Reset extrapolation ranges to dataset defaults' might be better?
> * I'm also wondering about the message in the popup when the background box goes red. 'Background is higher than minimum extrapolated value' is stating the programmatic condition. I'm wondering if something like 'Subtracting background value will result in negative intensities' would be better. Or, if keeping your text, could we also add the value of the extrapolation minimum to the message, just to clue the User in?
> * Lastly, I see you have now set a condition that if Fit Background AND Fit Guinier AND Fit Porod are all unchecked then the 'Go' button is greyed out. Which is great. But if you check any one of those boxes you can click 'Go'. Even if the A, B, K or sigma parameters are unspecified (which I presume internally means they are zero???) and thereby get a nonsense extrapolation as I showed in my earlier comment. So what we need is additional logic that ALSO checks if a box is ticked but the relevant parameters are unspecified. Does that make sense (if not, come back to me)? Ideally, we could also have a popup explaining why they can't click 'Go'; 'No extrapolation parameters set' or something like that.

@smk78 
Thank you for your feedback again. 

I agree with the first two points on the error messages. I have pushed commits implementing your suggestions.

I have also improved the logic, checking the enabling/disabling of the Go button. It now checks if the values are set. If the conditions are not met, it sets the button text to 'GO (disabled)', and when you click on it, it gives a pop-up giving you the reason why.

Let me know what you think of these new changes.

smk78:
Thanks @jellybean2004 . I've just tested CI build #4753.

The textual changes to the popup dialogs look good.

However, there are some issues with the checkbox-Go button-parameter logic. Let me see if I can walk you through them...

**Instance 1**
So here's the perspective on first running it:
image
There are no extrapolation parameters set but all 3 Fit boxes are checked and Go is enabled. This is ok because on clicking Go all five parameters will be auto-determined.

However, when you click Go it immediately goes disabled because it sees no extrapolation parameters are set:
image
So at the moment the advantage of fitting is being withheld.

ALSO, I notice you are now greying out parameter boxes in some situations, in which case in this instance all 5 should be greyed out as they are not being set manually.

**Instance 2**
Now I untick one of the boxes (Guinier in this example):
image
Go is now disabled (and if you click it it tells you why - good!) because there are no Guinier parameter values. So in this instance (Fit Guinier unchecked) they need to be given manually but you'll notice that the Guinier parameters are greyed out. So what is implemented here is actually the reverse logic! :smile:

And what I've just said about Fit Guinier here applies to Fit Background and Fit Porod also.

There is also a bug, because if I re-check Fit Guinier the Go button still says disabled:
image
In fact, you cannot now process any data in Corfunc unless you a) restart the program or, b) provide values for all 5 extrapolation parameters in order to re-enable Go (although once you correct Instance 1 option b) will no longer be available).

**Instance 3**
Here I fudged around Instance 1 by supplying parameters in order to enable Go but then let Corfunc fit everything:
image
What I then did was uncheck the Fit buttons. As I did that the corresponding parameter box(es) greyed out/become non-editable - which is what is required (good!) - but once ALL 3 boxes are unchecked, _even though_ all the parameter boxes are populated, you notice Go is disabled.

And the popup dialog is non-sensical because even though there might not be any fits selected, the program has a full house of extrapolation parameters.

**Required Behaviour**
- All 3 Fit boxes checked AND all parameter boxes empty:
  - corresponding parameter boxes greyed out/non-editable
  - Go enabled
  - processing performed
  - parameter boxes updated after fits

- All 3 Fit boxes checked AND all parameter boxes populated:
  - corresponding parameter boxes greyed out/non-editable
  - Go enabled
  - processing performed
  - parameter boxes updated after fits as required

- All 3 Fit boxes unchecked AND all parameter box(es) empty:
  - corresponding parameter boxes white/editable
  - Go disabled (message if clicked)
  - processing not performed

- All 3 Fit boxes unchecked AND all parameter box(es) populated:
  - corresponding parameter boxes white/editable
  - Go enabled
  - processing performed

- Any 1/2 Fit box(es) unchecked
  - corresponding parameter box(es) white/editable
  - other parameter box(es) greyed out/non-editable
  - Go only enabled and processing performed if corresponding parameter box(es) are populated
  - other parameter box(es) updated after fits as required

Sorry, bit long-winded but hope this makes sense?

jellybean2004:
@smk78 Thank you again for your feedback. I realised I implemented the opposite functionality of what was needed - I thought checking the box meant that the user should input the values 😅 

But I have corrected it now. It should behave as expected.

smk78:
Hi @jellybean2004 . Have just tested CI build #4760. (NB: I had to re-run the CI job as the one your commit triggered didn't create any artefacts. Odd, but I also notice several other commits tinkering with the CI setup so maybe you were just unlucky with the timing... Though I also note one of the Ubuntu builds is failing).

Anyway, the Go button logic looks good now! :smile: Well done.

Two other small things I've noticed:

When you load a new dataset into Corfunc on top of an existing dataset, all the extrapolation parameters should be reset, I think? But at present the background is being kept. Is there a reason for that?

And on my screen the control part of the perspective window (as opposed to the graph part) is not quite wide enough to fully display the text string 'Use inflection point':
image
Is it the same on your screen? Dragging & resizing the perspective window doesn't fix this.

smk78 (approved):
Functionality tested CI build #4779 on W11/x64. LGTM! 👍

DrPaulSharp (approved):
All looks good to me too - I have one small thing to look at, then merge when ready!

PR #3601, opened 2025-09-18
Qwedge

### LINKED ISSUE #3221: The Q wedge slicer has defaults to fold=false when it should be true
**Describe the bug**
The sector and box averaging slicers have the option of setting a `fold` attribute to `True` or `False`. The default is generally true and the two slicers allow setting the attribute when calling the calculator and the attribute can be set in the GUI. This attribute sets the behavior to either treat -Q and +Q the same (so they get averaged together creating a plot that goes from 0 to +Qmax), `fold=true`, or to treat -Q separately from + Q (so they do NOT get averaged together and the plot goes from a negative Qmin to a positive Qmax), `fold=false`

The Wedge slicer in Q uses the same classes as the sector averaging and calls the same calculator which is expecting a value for `fold`. However, by definition of the wedge currently there is no mirror wedge in -Q  Thus the fold attribute is not currently accessible to the user (correct behavior) but is set to `fold=false` incorrect behavior.

**To Reproduce**

**Expected behavior**

**Screenshots**

 - Version: [e.g. 5.0.4]

 - OS: [e.g. Windows 10]

**Additional context**

jellybean2004:
## Description

Sets the fold for Wedge Slicer to True, so the Wedge Averaging in Q does not create data in the negative Q space. 

Fixes #3221 

## How Has This Been Tested?

Tried out a few data sets to see if the expected behaviour is produced.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
@@ -40,7 +40,6 @@ class WedgeInteractor(BaseInteractor, SlicerModel):
         self.qmax = max(self.data.xmax, np.fabs(self.data.xmin), self.data.ymax, np.fabs(self.data.ymin))
         self.dqmin = min(np.fabs(self.data.qx_data))
         self.connect = self.base.connect
-        
 
         # Number of points on the plot
         self.nbins = 100


butlerpd (approved):
Code looks good to me with the question caveat. However conflicts need to be resolved before merging.

functionality tested on windows 11 and works as advertized.  However note that the fold parameter has been removed from the `Slicer Paremeter` menu. I do not see how that is relevant to this pull request and will create another issue

PR #3602, opened 2025-09-19
Dependabot: Include all action updates in one PR

krzywon:
## Description

This bundles all github action updates, made by Dependabot, into a single PR. This will ensure we aren't in a situation where a series of major updates that depend on one another aren't pushed into different PRs, each of which is failing, without knowing if they will work together well (see #3588, #3589, #3590, #3591, and #3592 as an example).

## Review Checklist:

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -7,3 +7,7 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every week
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


llimeht:
CI failure appears to be that `windows-latest` is in the process of being changed from `windows-2022` to `windows-2025` and the latter does not include InnoSetup.

PR #3608, opened 2025-09-23
Explicitly specifies latest version of ruff in CI

### LINKED ISSUE #3605: ruff config generates warnings on CI runs
**Describe the bug**
The current usage of ruff generates a warning annotation on all CI runs. It's generally nice to avoid warnings as they are a nag to fix a problem that we need to fix, such as an action version being deprecated, an image being withdrawn from use soon.

```
ruff
Could not parse version from /home/runner/work/sasview/sasview/pyproject.toml. Using latest version.
```

**To Reproduce**
1. Check any recent CI run

**Expected behavior**
No warning annotations on CI - clean run

 - Version: git main and release-6.1.1 branches

DrPaulSharp:
## Description

Fixes #3605 

## How Has This Been Tested?

Running the CI shows no warning.

## Review Checklist:

**Installers**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: astral-sh/ruff-action@v3
         with:
+          version: "latest"
           args: "check --fix-only --exit-non-zero-on-fix"
       - uses: pre-commit-ci/lite-action@v1.1.0
         if: always()


DrPaulSharp:
> Thanks @DrPaulSharp! Looking at #3573, does that also need some config to say which version of ruff should be used? (and should it be the same config via the action's `version-file` config?)

The ruff pre-commit hook specifies the version of ruff to use in its own `pyproject.toml`. https://github.com/astral-sh/ruff-pre-commit/blob/main/pyproject.toml Note that 0.13.1 is the latest version.

> I have little experience with ruff, but over the years I've seen other linters get more strict and so a `latest` version suddenly moves the goal posts by suddenly generating new CI failures that get in the road of development. Is that a potential issue here that would reward picking a particular version of ruff rather than just the latest?

Given that the pre-commit hook uses the latest version, and we are only using a subset of the full set of ruff rules (as set out in SasView's `pyproject.toml` https://github.com/SasView/sasview/blob/fc4a5c6a10442c32b773998d987771ae46fb2863/pyproject.toml#L203), and then only applying automatic fixes, with everything else done manually as and when a developer so chooses, I'm happy to keep to the latest version here.

llimeht (approved):

Issue #3559, opened 2025-07-30
Use AUSAXS pypi distribution once it exists


krzywon:
**Is your feature request related to a problem? Please describe.**
Integrating AUSAXS into our code base requires a separate step in our CI to download the library. A python wrapper is in the works, but won't be ready 'for a while'.

**Describe the solution you'd like**
Ideally, AUSAXS would be available on pypi (python wrapper) and the library could be added to our requirements file. The CI step to download it and the python script used to fetch it should also be removed.

**Additional context**
See #3554 for further discussion.

PR #3634, opened 2025-10-08
Replace direct AUSAXS library integration with the pyAUSAXS wrapper

klytje:
## Description

As discussed in #3559 and #3554, this PR replaces the current direct `AUSAXS` shared library integration with calls to the new `pyAUSAXS` wrapper I designed specifically for this integration. 

**Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.**

We should verify that the `pyAUSAXS` library correctly installs on real systems beyond the GitHub runners. Someone who's more competent than me at building Python wheels should also verify that the [three wheels available from pypi](https://pypi.org/project/pyausaxs/#files) are general enough to be installable on all relevant systems. 

I have removed the `c60.pdb` system from Debye testing and relaxed the tolerance to 3% on the remaining systems. Previously, I was compiling higher-precision custom `AUSAXS` versions for SasView, but `pyAUSAXS` uses the default, lower-precision release, which was failing these tests. 

## Review Checklist:

**Installers**
  - [x] **Linux** installer (GH artifact) has been tested (installed and worked)

### DIFF
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -17,7 +17,7 @@ numpy
 packaging
 periodictable
 platformdirs
-pyausaxs
+pyausaxs==1.0.4
 pybind11
 pylint
 pyopencl


test/sascalculator/utest_sas_gen.py
```diff
@@ -251,7 +258,7 @@ def test_debye_impl(self):
             errs = (external - analytical)/analytical
             different_entries = 0
             for val in np.abs(errs):
-                self.assertLessEqual(val, 0.02, "Ensure that the error is acceptable.")
+                self.assertLessEqual(val, 0.03, "Ensure that the error is acceptable.")
```
llimeht: This (and the later test) might be better written with `numpy.testing.assert_array_less`  or `numpy.testing.assert_allclose`

https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_less.html
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html

test/sascalculator/utest_sas_gen.py
```diff
@@ -251,7 +258,7 @@ def test_debye_impl(self):
...
                 if val != 0:
```
llimeht: Given how floating-point comparisons have a habit of kicking one in the backside... I'd expect that this would often be just slightly off zero. Is there a closeness threshold that makes sense instead?

llimeht (approved):
Some comments on minor improvements but looks good. Thanks!

test/sascalculator/utest_sas_gen.py
```diff
@@ -225,19 +225,26 @@ def test_debye_impl(self):
...
+
         from sas.sascalc.calculator.ausaxs import ausaxs_sans_debye, sasview_sans_debye
 
         rng = np.random.default_rng(1984)
+        ausaxs = AUSAXS()
 
-        if not ausaxs_sans_debye.ausaxs_available():
-            self.assertTrue(False, "AUSAXS library not found, test cannot be run.")
+        # ensure the library is available and ready to run on all CI systems
+        if not ausaxs.ready():
```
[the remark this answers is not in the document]
klytje: I originally added this check because otherwise the implementation silently uses the built-in SasView implementation, which trivially passes the following comparisons against itself. 

Now, the `AUSAXS()` constructor hooks into the library and invokes a simple test method to ensure the integration works. Should this integration for some reason fail, e.g. if an old Windows machine which does not support AVX instructions is used, the import succeeds but the test call segfaults and `ausaxs.ready()` is set to false. 

I think it is useful to keep this test around: we expect it to always pass, but it may catch issues which would otherwise go undetected (due to the automatic fallback). Perhaps when #3194 is merged in we can remove this, as we'd then have other tests depending on `AUSAXS` without fallbacks.

PR #3666, opened 2025-10-30
Make ruff more verbose about what it doesn't like

### LINKED ISSUE #3641: Can we get more output from the ruff github action
When running ruff in the CI, the output on failure is simply a message `ruff failed with exit code 1`. Is there a way for the action to give us the usual ruff output, explaining the linting errors it has fixed, or any other error?

llimeht:
## Description

Tweak the ruff options to make it show what it doesn't like in the files rather than just exiting with failure.

Fixes #3641

## How Has This Been Tested?

CI failures in another PR turning to verbose diffs ([logs](https://github.com/llimeht/sasview/actions/runs/18934250033/job/54057203634)).

```
--- src/sas/qtgui/MainWindow/GuiManager.py
+++ src/sas/qtgui/MainWindow/GuiManager.py
@@ -60,8 +60,7 @@
 # General SAS imports
 from sas.qtgui.Utilities.SasviewLogger import setup_qt_logging
 from sas.qtgui.Utilities.WhatsNew.WhatsNew import WhatsNew
-from sas.system import web
-from sas.system import HELP_SYSTEM
+from sas.system import HELP_SYSTEM, web
 from sas.system.user import create_user_files_if_needed
 from sas.system.version import __release_date__ as SASVIEW_RELEASE_DATE
 from sas.system.version import __version__ as SASVIEW_VERSION

--- src/sas/system/__init__.py
+++ src/sas/system/__init__.py
[... snipped ...]

Would fix 4 errors (132 additional fixes available with `--unsafe-fixes`).
```

## Review Checklist:

**Installers**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,9 @@ jobs:
       - uses: astral-sh/ruff-action@v3
         with:
           version: "latest"
-          args: "check --fix-only --exit-non-zero-on-fix"
+          args: "--version"
+      - run: ruff check --diff --exit-zero
+      - run: ruff check --fix-only --exit-non-zero-on-fix
       - uses: pre-commit-ci/lite-action@v1.1.0
         if: always()
         with:


DrPaulSharp:
Note that `output-format` is not required as `full` is the default: https://docs.astral.sh/ruff/settings/#output-format

DrPaulSharp:
> > Note that `output-format` is not required as `full` is the default: https://docs.astral.sh/ruff/settings/#output-format
> 
> It's the default for when you normally run ruff from the command line, but not in the action. The action sets the environment variable:
> 
> > Set RUFF_OUTPUT_FORMAT to github
> 
> which then prevents any useful output to stdout.

Ahh, I stand corrected on that, we should add that into the diff clause in that case.

> > ```
> >   - uses: astral-sh/ruff-action@v3
> >     with:
> >       version: "latest"
> >       args: "check --diff --exit-zero"
> >   - run: ruff check --fix-only --exit-non-zero-on-fix
> > ```
> 
> I understand the intention here, but is it weird to use the action to run ruff once and then run it from a `run` step afterwards? Does it make sense to run it the same way each time? i.e. either with the action both times or from the command line both times?

The complication here is that the default behaviour of the action is to run `ruff check` after installation unless the arguments are changed (see https://github.com/astral-sh/ruff-action?tab=readme-ov-file#use-to-install-ruff). To solve all of this without running an initial (unnecessary) `ruff check` we could try:

```
 - uses: astral-sh/ruff-action@v3
    with:
      version: "latest"
      args: "--version"
  - run: ruff check --diff --exit-zero --output-format=full
  - run: ruff check --fix-only --exit-non-zero-on-fix
 ```

PR #3689, opened 2025-11-06
Prevent crashes and log errors if docs can't be found.

llimeht:
## Description

Calculating the location of the docs at runtime can generate exceptions if the docs can't be found. They should never not be found, of course, ... but....

Given that development environments are probably the most likely to have this broken and contributor camp is looming, this is a quick and urgent fix to make sure that new contributors don't end up yak-shaving their development environment more than needed.

The commits here:
 1. prevent crash on startup if the docs can't be found to be copied around; logging an error if this happens
 2. drop an unused `_tutorialLocation` member that would lead to a crash on startup
 3. protect other places the help system is started, having them log an error rather than dump a backtrace to console; these don't actually cause sasview to crash, they are just ugly for the user

## How Has This Been Tested?

1. Start sasview and test that view docs still works from both FitPage and Help menu
2. Make loading docs fail by either editing the "docs" entries in `sas/system/user.py` or by deleting them (from `dist-packages/sas` if you are working in an editable install as recommended)
3. start sasview, see a log entry on startup and log entries when you attempt to view docs:

```
ERROR : sas.system.user (user.py:196) :: Documentation could not be found! Help documentation will not be available in SasView.
ERROR : sas.qtgui.Perspectives.Fitting.FittingWidget (FittingWidget.py:1666) :: Help documentation was not found.
ERROR : sas.qtgui.MainWindow.GuiManager (GuiManager.py:361) :: Help documentation was not found.
```

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/MainWindow/GuiManager.py
+++ b/src/sas/qtgui/MainWindow/GuiManager.py
@@ -357,6 +357,10 @@ class GuiManager:
         """
         Open a local url in the default browser
         """
+        if not HELP_SYSTEM.path:
+            logger.error("Help documentation was not found.")
+            return
+
         counter = 1
         window_name = "help_window"
         # Remove leading forward slashes from relative paths to allow easy Path building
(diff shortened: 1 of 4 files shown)


DrPaulSharp (approved):
This works as expected.

PR #3737, opened 2025-11-11
Warn user before starting scattering calcs on large structures

### LINKED ISSUE #2915: Need a way out of the GSC
For the heck of it I tried a 7M atom (nuclear) PDB file in the GSC.

It loaded in ~30s.

Rendering the 3D drawing was acceptably fast.

Doing the Debye Full Avg calculation went on and on (I gave up at 10 mins) using ~90% CPU.

There is no Stop button on the GSC. We do have one in the FitPage. There is a Cancel button but it doesn't do anything and it certainly doesn't kill the thread. Also, even if you Close/kill the GSC window the calculation thread is still active and eats so much system resource that you can't Quit SasView. The only option is to kill the process.

I think we either need a proper Stop button (that kills the thread), or at least a popup window when you click Compute if the loaded coordinate file is beyond a certain size ("this calculation may take hours, are you sure you want to continue").

klytje:
## Description

The user will now receive a popup if a GSC calculation is requested with large input structures. This will continue to appear on every such request until the user accepts it, after which it will no longer be shown.

Fixes #2915

## How Has This Been Tested?

Manually verified functionality by running with various different input structures, ranging from 50k to 500k atoms. 

**Installers**

### DIFF
--- a/src/sas/qtgui/Calculators/GenericScatteringCalculator.py
+++ b/src/sas/qtgui/Calculators/GenericScatteringCalculator.py
@@ -1402,7 +1402,7 @@ class GenericScatteringCalculator(QtWidgets.QDialog, Ui_GenericScatteringCalcula
         popup.setText(msg)
         popup.setWindowTitle("Large input structure")
         popup.setDefaultButton(QMessageBox.Cancel)
-        result = popup.exec_()
+        result = popup.exec()
         flag_continue = (result == QMessageBox.Ok)
         self.warned_user_large_calc = flag_continue
         return flag_continue
@@ -1415,7 +1415,7 @@ class GenericScatteringCalculator(QtWidgets.QDialog, Ui_GenericScatteringCalcula
         try:
             # create the combined sld data and update from gui
             sld_data = self.create_full_sld_data()
-            if not self.check_large_calculation(sld_data.data_length): return
+            if hasattr(sld_data, 'data_length') and not self.check_large_calculation(sld_data.data_length): return
 
             # TODO: implement fourier transform for meshes with multiple element or face types
             # The easy option is to simply convert all elements to tetrahedra - but this could rapidly


rozyczko (approved):
Looks good. It's important to tell users their requests are sometimes too tall :)

Minor code comments only.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1367,6 +1369,44 @@ def create_rotation_matrices(self):
...
+    def check_large_calculation(self, num_points):
+        """
+        Generate a popup warning the user about a large calculation
+        :return True if the calculation should continue, False otherwise.
+        """
+
+        if self.warned_user_large_calc or num_points < 50_000: return True
+        # timings for an i7-1165G7 4-core CPU from 2020:
+        # 50k: 1.8s | 100k: 6.3s | 200k: 28.7s | 300k: 65.6s | 400k: 117.2s | 500k: 181.2s
+        # multiply by 2 & round to get a conservative estimate
```
wpotrzebowski: It may be good to keep this numbers as a configurable variable. It is now being tuned for the current calculator but we may potentially set different limits if in the future we for example change approach

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1367,6 +1369,44 @@ def create_rotation_matrices(self):
...
+                       "Do you wish to proceed?")
+                break
+
+        flag_continue = False
+        popup = QMessageBox(standardButtons = QMessageBox.Ok | QMessageBox.Cancel)
+        popup.setIcon(QMessageBox.Warning)
+        popup.setText(msg)
+        popup.setWindowTitle("Large input structure")
+        popup.setDefaultButton(QMessageBox.Cancel)
+        result = popup.exec_()
```
wpotrzebowski: I am not an expert on Qt but exec_() seems to becoming obsolete.  exec() using setters is probably the safest across PySide/PyQt flavors and future versions

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1375,6 +1415,8 @@ def onCompute(self):
         try:
             # create the combined sld data and update from gui
             sld_data = self.create_full_sld_data()
+            if not self.check_large_calculation(sld_data.data_length): return
+
```
wpotrzebowski: Not sure if this can be an issue. The check uses sld_data.data_length. If any codepath returns an object without that attribute, you’d get an AttributeError?

wpotrzebowski (commented):
Generally looks good to me. Some minor suggestion.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
klytje: Added a `hasattr` check for the `data_length` attribute. Currently all codepaths _should_ have this attribute, though of course it's good to check.

PR #3742, opened 2025-11-11
Limits python version to <3.14

### LINKED ISSUE #3738: PySide6 support issue for Python 3.14
While attempting to install SasView with Python v 3.14, I found out that the PySide6 library does not support this Python version. However, the downgrade to Python 3.13.9 solved this issue, therefore, we should limit the Python version to < 3.14 until the PySide6 will support this Python 3.14.

DrPaulSharp:
## Description

Fixes #3738. The change is ultimately academic, as the build system fails on the requirements for PySide6 before we get to the `requires-python` section, but I think it will be helpful for users/developers to see in SasView itself that we cannot support python 3.14 until PySide6 starts supporting it.

## How Has This Been Tested?

Attempted to install SasView with python 3.14. When commenting out pyopengl and pyside6 in the build system requirements, I get the desired error `ERROR: Package 'sasview' requires a different Python: 3.14.0 not in '<3.14,>=3.12'`

## Review Checklist:

**Installers**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dynamic = [
 ]
 description = "SasView application"
 readme = "README.md"
-requires-python = ">=3.12"
+requires-python = ">=3.12,<3.14"
 license = { text = "BSD-3-Clause" }
 authors = [
     {name = "SasView Team", email = "developers@sasview.org"},


DrPaulSharp:
Hmm, this is very strange. When installing with `uv`, it works as you say. However, when installing with pip, I get the error I described (and shown below). I note that [pyside6](https://pypi.org/project/PySide6/) itself does state its requirements are "<3.14,>=3.9". Given that we recommend in `INSTALL.md` to use pip rather than `uv`, we should probably consider adding this restriction. However, I would like to understand why there is this discrepancy between `pip` and `uv`.

image

rozyczko:
pip doesn't yet have PySide6 for 3.14, but conda-forge already does
https://anaconda.org/conda-forge/pyside6/6.9.3/download/win-64/pyside6-6.9.3-py314h2c9462b_1.conda

Maybe `uv` is pulling from conda-forge behind the scenes, just like `pixi` does.

butlerpd:
> I also have sasview successfully running with Python 3.14 here. One just needs to remove `numba` from the requirements.txt (numba does not yet have 3.14 support, tracked [here](https://github.com/numba/numba/issues/9957)). Looking at the code, numba seems to be an optional dependency, so perhaps it should be moved into some sort of extras group anyway.

I thought numba is required for the GSC (and maybe for Inversion?) That is why it takes so infinitely long to load the first time?

llimeht:
> `numba` should NOT be removed. As @butlerpd said, it is being used in both GSC and P(r).

I get that we wouldn't want to release sasview without numba working and bundled into the installers, and I've got no thoughts to change that. But that's a long way from saying that sasview doesn't work without numba. 

However, I'm also currently happily analysing SANS data with Python 3.14 (without numba). Most of sasview just works with some bits that I don't having reduced functionality. In most Python projects, that would be reflected as an optional dependency — that's why extra dependency groups exist in python packaging via PEP-735 and friends.. Being able to do testing with newer interpreters and newer packages is important to being able to upgrade sasview to those packages in the future, particularly given the extremely limited state of unit tests and functionality test on this code base.

PR #3753, opened 2025-11-13
Update install instructions, Readme, and add resources

TLO218:
## Description

Update the Readme and make it feel more inviting 

Fixes # (issue/issues)
Gives a quickstart guide to newcomers on how to install the packages, Links to most usefull resources and tries to  gives example code to test the installation

## How Has This Been Tested?
Provided example is used to test different install instrutions

## additional maitanance
This adds a new `resources.md` file along with an image for the `sasview-hello_world` example. By using `Install.md` and `Resources.md` with Pandoc to generate the corresponding web pages, we can simplify maintenance and minimize the risk of outdated information over time.

currently it also mentions the issue with launching the gui throught sas as it's only been introduced after the 6.1.1 release.
 
## Review Checklist:

**Install instructions** (check at least one)
- [x] test user installation with uv and pip
- [x] test instucitons with the dev version

**discussion**
- [x] should there be a badge talking about test coverage or CI
- [x] discuss folder structure
- [ ] should python file be provided in tutorials repo and a test case be added for this in the ci
- [ ]  discuss which citations and papers should be added to the end


RESOURCES.md
```diff
@@ -0,0 +1,38 @@
...
+- [The full user documentation](/docs/user/user.html) including model documentation. This is available as a [PDF download](/downloads/SasViewDocumentation_4.2.2.pdf) for version 4.2.2
+- Contact us at: [help@sasview.org](mailto:help@sasview.org) to report bugs (in which case please also attach a SasView log file, see below) or for help in using the software
+- Try the [Users Forum Archives](http://lists.sasview.org/pipermail/users) or even email the forum at [users@sasview.org](mailto:users@sasview.org). To join the Users Forum, click [here](http://lists.sasview.org/cgi-bin/mailman/listinfo/users)
+- Check our [release notes](http://www.sasview.org/docs/user/RELEASE.html) for Known Issues
+- Visit our [issues list](https://github.com/SasView/sasview/issues?q=is%3Aissue) to view reported bugs, things being worked on or possible future ehancements
+- For the adventurous, many new features and bug fixes are available in the [latest builds](http://builds.sasview.org) and may have what you are looking for
+
+## Model Marketplace
+
+In case you are want to use models not provided in Sasview by default or you want to share your models with others please check out our model Marketplace.
```
llimeht: be good to link to the marketplace

README.md
```diff
@@ -1,7 +1,169 @@
...
+
+### Install using `pip` inside the working directory:
+
+```shell
+python -m venv .venv # create the environment
+source .venv/bin/activate # activate the environment
+
+python -m pip install sasview
+
+python -m sasview # launch gui
```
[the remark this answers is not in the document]
llimeht: No, `-m` means run a Python module, and there is no `sasview` module at all, so that doesn't work at all. There is, however, a `sas` module that has a `__main__` which is is what is needed. 

So the current situation is that  `python -m sasview` doesn't work anywhere and `python -m sas` works

Yes, we could offer a `sasview` module as well as a `sas` module one day, but we do not currently.

Providing an executable is a completely separate item. The sasview distributable provides a `sas` module (so `python -m sas` works) and a `sasview` script, so running `sasview` as a command works.

.gitconfig
llimeht: This file should not be in the PR

uv.lock
llimeht: This file is unrelated and should not be in the PR

.gitignore
llimeht: This change shouldn't be part of the PR

llimeht:
I've pushed a few fixups and additional info to the the branch. I think it's better than what is currently there so worth merging - thank you @TLO218!

We might want to consider moving the scripting example elsewhere, or at least putting some screenshots before it, but that can happen later, and getting the installation instructions updated is more urgent.

llimeht (approved):

PR #3754, opened 2025-11-13
Show the correlation stats as a table

### LINKED ISSUE #2505: SasView GUI should provide access to the DREAM output on fit parameters
A user points out that the documentation (https://www.sasview.org/docs/user/qtgui/Perspectives/Fitting/optimizer.html#fit-dream ) suggests that the DREAM output includes "confidence intervals" (though @pkienzle points out there are technically "credible intervals" rather than "confidence interval"). As shown in that documentation the output table should look something like:
DREAM_Output

However, there is no way to retrieve that table from the SasView GUI.

Note that this information is available to the SasView GUI code, it is just not being presented to the user in anyway. There are 3 ways that this information should probably be made available.
* This table should be made a part of the report when a fit uses a DREAM fit
* There should be a way to pop up a table with this information in the GUI. Two obvious options would be:
  * as a tab on the bumps plot outputs
  * from the top level menu where you can currently access the table of batch  results
    * The extra columns should be added to the batch grid table in the case of a DREAM batch fit
* The output of parameters (currently under the `edit` menu) should output the extra columns from a DREAM fit either as a .txt,  or in Tex or CSV formats.

**Additional context**
@pkienzle has provided the following script than can be cut and pasted into the python shell/editor of the SasView GUI to obtain that information

```python
from pathlib import Path
from PyQt5 import QtWidgets
from bumps.gui.uncertainty_view import UncertaintyView
from bumps.dream.stats import var_stats, format_vars, save_vars

app = QtWidgets.QApplication.instance()
state = app.activeWindow().findChild(UncertaintyView).plot_state[0]
stats = var_stats(state.draw())
print(format_vars(stats))
save_vars(stats, Path('~/saved.json').expanduser())
```

rozyczko:
## Description

Added a correlation stats table in the DREAM's Result Panel tabbed widget

Fixes #2505 

## How Has This Been Tested?

Local dev build tests

Resulting table:

Untitled

## Review Checklist:

- [X ] There is **nothing** that needs documenting

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/PlotView.py
+++ b/src/sas/qtgui/Utilities/PlotView.py
@@ -105,8 +105,9 @@ class CorrelationTable(QtWidgets.QWidget):
             tuple: (headers, data_rows) where headers is a list of column names
                    and data_rows is a list of lists containing the formatted data.
         """
-        from bumps.dream.stats import format_vars, parse_var, var_stats
         import re
+
+        from bumps.dream.stats import format_vars, parse_var, var_stats
 
         # Get formatted statistics from the uncertainty state
         draw = self.state.uncertainty_state.draw()


wpotrzebowski (commented):
It works as expected, however I have some questions related to nomenclature. 
It currently says Correlation Table but I think Statistics will be more appropriate here (we also have Correlation plot and it can be bit misleading in this context too).Also "best" for the parameter can probably be changed to something more accrurate. I think in Bayesian teminology  it is a "point estimate". Maybe keeping "best" but adding  tooltip will work.

butlerpd:
Indeed, I think the table is produced by bumps so the nomenclature in the table would have to be changed there I think? we could put an issue into bumps perhaps?

wpotrzebowski:
The column headings are setup on sasview end but indeed it would be good to be consistent with bumps

PR #3755, opened 2025-11-13
Make the data operation window resizable and update documentation

### LINKED ISSUE #3752: "Data Operation" tool window not resizable
**Problem description**
This might be a problem when one is working with files with really long name. In this case they are not displayed entirely in the ComboBox.

**Solution description**
Resizable window, that can be expanded in width.

Cristian2299:
## Description

The window of Data Operation utility is now resizable and the order of "Data1", "Data2", and "Output Data" has been change to have the output on the right. The modifications only concerns the UI. The documentation has been updated with a screenshot of the new layout.

Fixes #3752 

## How Has This Been Tested?

Tested on windows 10. Tested with example files and experimental data set. Operation and output results works as expected.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
+++ b/src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
@@ -125,7 +125,7 @@ class DataOperationUtilityPanel(QtWidgets.QDialog, Ui_DataOperationUtility):
 
 
     def onCompute(self):
-        """ perform calculation """
+        """ Perform calculation """
         # set operator to be applied
         operator = self.cbOperator.currentText()
         # calculate and send data to DataExplorer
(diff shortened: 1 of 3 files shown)


Cristian2299:
@rozyczko Thanks for the help. I tried it on my machine and it's working as it should do, for both the windows installer and the dev version. @drnpaul I saw that on some machines is working as I was expecting (on mine and the one of @conanr-create for instance), but in other cases, like yours, not. Honestly, I don't have an explanation for that.

src/sas/qtgui/Plotting/PlotterBase.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, parent=None, manager=None, quickplot=False):
         # Set the layout and place the canvas widget in it.
         layout = QtWidgets.QVBoxLayout()
         # FIXME setMargin -> setContentsMargins in qt5 with 4 args
-        #layout.setContentsMargins(0)
+        layout.setSpacing(0)
```
krzywon: This doesn't seem to effect anything, but it seems out of scope for this PR.

krzywon (approved):
Major improvement to the data operation panel! Tested locally and all 3 data panels expand equally. I like the new left-to-right flow of the window (as it should be), and the ability to preview the resulting data set before sending it to the data explorer.

butlerpd:
I forgot to mention that I am only able to resize the `Data Operation` panel width but not its height. Is that the intended behavior?  Also the minimum width of the panel seems pretty generous on my computer? Is that intended? or just a feature of my large screen?

butlerpd (approved):
Very smooth! I like it. I'll try to make some new issues based on the other comments that are really not part of the work here. The Reset button still does not have a tool tip (all other now do), but it is pretty self-explanatory so won't hold up the merging just for that

This should be ready to merge IMO

PR #3758, opened 2025-11-14
Adds multiple simultaneous slicers support

jellybean2004:
## Description

Implemented support for multiple simultaneous slicers on 2D plots, allowing users to create, manage, and switch between different slicer instances.

- A unique ID is assigned to each slicer instance so they can be viewed simultaneously on the 2D fitting plot, and each generates an independent plot.
- A list of all active slicers is added in the Slicer Parameters window so users can select and tweak them individually. The selected slicer gets automatically updated when a slicer is added or modified in the 2D plot. Users can also delete individual slicers from there.
- Slicers have different colours to help differentiate them.

## How Has This Been Tested?

Functionality tested in the program, works as expected.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter2D.py
+++ b/src/sas/qtgui/Plotting/Plotter2D.py
@@ -479,7 +479,7 @@ class Plotter2DWidget(PlotterBase):
             child = item.child(item_index)
             if child.hasChildren():
                 self._recurse_plots_to_remove(child)
-            if not (isinstance(child.data(), (Data1D, Data2D))
+            if (isinstance(child.data(), (Data1D, Data2D))
                     and hasattr(child.data(), 'type_id')
                     and slicer_type_id in child.data().type_id
                     and child.rowCount() > 0
--- a/src/sas/qtgui/Plotting/SlicerParameters.py
+++ b/src/sas/qtgui/Plotting/SlicerParameters.py
@@ -151,7 +151,6 @@ class SlicerParameters(QtWidgets.QDialog, Ui_SlicerParametersUI):
         """
         Create and initially populate the list of slicers with radio button behavior
         """
-        current_slicer_text = self.getCheckedSlicer()  # Already returns string now
 
         self.lstSlicers.clear()
 
@@ -330,9 +329,9 @@ class SlicerParameters(QtWidgets.QDialog, Ui_SlicerParametersUI):
             if self.active_plots.keys():
                 self.parent.setSlicer(slicer=slicer)
                 # Reset combo box back to "None" after setting slicer
-                self.cbSlicer.blockSignals(True)
-                self.cbSlicer.setCurrentIndex(0)
-                self.cbSlicer.blockSignals(False)
+                self.cbSlicerReplace.blockSignals(True)
+                self.cbSlicerReplace.setCurrentIndex(0)
+                self.cbSlicerReplace.blockSignals(False)
         self.onParamChange()
 
     def onGeneratePlots(self, isChecked):
--- a/src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
@@ -287,10 +287,9 @@ class BoxInteractor(BaseInteractor, SlicerModel):
             new_plot.ytransform = "y"
             new_plot.yaxis("\\rm{Residuals} ", "/")
 
-        # new_plot.id = (self.averager.__name__) + self.data.name
         # Create id to remove plots after changing slicer so they don't keep
         # showing up after being closed
-        new_plot.type_id = "Slicer" + new_plot.id
+        new_plot.type_id = "Slicer" + self.data.name
         new_plot.is_data = True
         item = self._item
         if self._item.parent() is not None:
--- a/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
@@ -96,7 +96,6 @@ class WedgeInteractor(BaseInteractor, SlicerModel):
         Clear the slicer and all connected events related to this slicer
         """
         self.averager = None
-        # self.clear_markers()
         self.outer_arc.clear()
         self.inner_arc.clear()
         self.radial_lines.clear()
--- a/src/sas/qtgui/Plotting/UI/SlicerParametersUI.ui
+++ b/src/sas/qtgui/Plotting/UI/SlicerParametersUI.ui
@@ -195,16 +195,16 @@
            </widget>
           </item>
           <item>
-           <layout class="QHBoxLayout" name="horizontalLayout_2">
+           <layout class="QHBoxLayout" name="horizontalLayout_create">
             <item>
-             <widget class="QLabel" name="label">
+             <widget class="QLabel" name="lblSlicerCreate">
               <property name="text">
                <string>Create new slicer:</string>
               </property>
              </widget>
             </item>
             <item>
-             <spacer name="horizontalSpacer_2">
+             <spacer name="horizontalSpacer_create">
               <property name="orientation">
                <enum>Qt::Horizontal</enum>
               </property>


src/sas/qtgui/Plotting/Slicers/BaseInteractor.py
```diff
@@ -59,45 +59,44 @@ def clear_markers(self):
...
-            h.remove()
-        if self.markers:
-            self.base.connect.clear(*self.markers)
-        self.markers = []
+        try:
+            for h in self.markers:
+                h.remove()
+            if self.markers:
+                self.base.connect.clear(*self.markers)
+            self.markers = []
```
rozyczko: You probably want to keep deleting the markers on exception. Currently, the first exception will stop subsequent removals. Consider wrapping just `h.remove()` with try..except

src/sas/qtgui/Plotting/Slicers/BaseInteractor.py
```diff
@@ -119,16 +118,29 @@ def onLeave(self, ev):
...
         self.clickx, self.clicky = ev.xdata, ev.ydata
         self.save(ev)
+        try:
+            # Case 1: this interactor is the slicer (base is the plotter)
+            if hasattr(self.base, 'notifySlicerModified'):
+                self.base.notifySlicerModified(self)
+            # Case 2: this interactor is a child of the slicer (base is the slicer)
+            elif hasattr(self.base, 'base') and hasattr(self.base.base, 'notifySlicerModified'):
+                self.base.base.notifySlicerModified(self.base)
+        except Exception:
```
rozyczko: This is dangerous! What kind of exceptions are we expecting? If they are benign, let's catch them with proper signatures and add comments on why we drop them.

src/sas/qtgui/Plotting/SlicerParameters.py
```diff
@@ -128,6 +134,62 @@ def updatePlotList(self):
...
+        """
+        checked_slicer = None
+        if self.lstSlicers.count() != 0:
+            for row in range(self.lstSlicers.count()):
+                item = self.lstSlicers.item(row)
+                isChecked = item.checkState() != QtCore.Qt.Unchecked
+                slicer = item
+                if isChecked:
+                    checked_slicer = slicer
+        return checked_slicer
```
rozyczko: Since there's only one checked item, you can simplify this

```python
        for row in range(self.lstSlicers.count()):
            item = self.lstSlicers.item(row)
            if item.checkState() != QtCore.Qt.Unchecked:
                return item
        return None
```

src/sas/qtgui/Plotting/SlicerParameters.py
```diff
@@ -216,6 +284,30 @@ def onSlicerChanged(self, index):
...
+            slicer_name = item.text()
+
+            # Update the parameter model to show this slicer's parameters
+            if slicer_name in self.slicer_models:
+                self.setModel(self.slicer_models[slicer_name])
+            elif slicer_name in self.parent.slicers:
+                # Get the model from the slicer object
+                slicer_obj = self.parent.slicers[slicer_name]
+                if hasattr(slicer_obj, '_model'):
+                    self.slicer_models[slicer_name] = slicer_obj._model
```
rozyczko: it would probably be safe to also check if `slicer_name` is in `slicer_models`, since you're querying a parent

```python
if hasattr(slicer_obj, '_model') and slicer_name in self.slicer_models:
```

src/sas/qtgui/Plotting/SlicerParameters.py
```diff
@@ -437,6 +568,25 @@ def onHelp(self):
         url = "/user/qtgui/MainWindow/graph_help.html#d-data-averaging"
         self.manager.parent.showHelp(url)
 
+    def updateSlicersList(self):
+        """
+        Update the slicers list when slicers are added or removed
+        """
+        self.setSlicersList()
+
+    def checkSlicerByName(self, slicer_name):
```
rozyczko: Please add a docstring for this method

src/sas/qtgui/Plotting/Slicers/AnnulusSlicer.py
```diff
@@ -69,9 +66,18 @@ def clear(self):
         Clear the slicer and all connected events related to this slicer
         """
         self.clear_markers()
-        self.outer_circle.clear()
-        self.inner_circle.clear()
-        self.base.connect.clearall()
+        try:
+            self.outer_circle.clear()
```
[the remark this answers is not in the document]
jellybean2004: This was a precaution I added as I was having issues when clearing the slicers, but I reckon it was coming from just `remove`s. Will remove these now.

jellybean2004:
@krzywon, I have changed the slicer colours to be different, so it is easier to differentiate. I have also added the functionality of individual slicer deletion, which can be done from the slicer parameter window. 

@rozyczko, I have implemented most of your suggestions. I will now work on refactoring out the ID generation, as it is the same for each slicer. 

Please let me know if there is anything else I could add, or something I missed.

src/sas/qtgui/Plotting/Plotter2D.py
```diff
@@ -411,81 +453,102 @@ def updateSlicer(self):
...
+
+    def _removeSlicerPlots(self):
+        """
+        Clear the previous slicer plots
         """
         # Clear current slicer
-        if self.slicer is not None:
-            self.slicer.clear()
+        # if self.slicer is not None:
+            # self.slicer.clear()
```
krzywon: If this is obsolete, please remove.

butlerpd (commented):
Nice work though I have not looked closely at the code yet. This is mainly a functionality review based on window 11 testing of the installer.
* `Clear slicers` does not clear the box sum box on the plot so once a box sum is slicer is activated, the box on the 2D plot is permanent and we have to close the whole plot before another slicer can be used cleanly. Not sure if this predates this PR in which case another issue might be appropriate, but prior to this PR a new slicer removed the old anyway?
* There are really irritating white lines that randomly appear (apparently when another window passes over the plot?) but this seem to be there from before this pull request? In which case an issue might be the better place to worry about this.
* I like that all slicers, even the annulus (and box sum though IMO we should eventually remove that as a thing) allows for multiple slicers. The color coding helps, but it is a bit unintuitive that the new slicer falls directly on top of the old one. To the newbie it looks like nothing happened? maybe just offset the interactors by a small fraction from the first one each time. If the third is on top of the second it won't matter as by that point it should be clear? and most people will move the second to where they want it before asking for a third slicer?
* Alternatively I guess the offset could increment each time?  This is more complicated for the annulus which does not have a center that ties the two sides together. Not sure how to deal with that, but I think the colloid orientation people will use the multiple annulus a lot.
* Related, the top color is the color for the last slice. However, if one grabs the center point and moves the ROI, it is the very first one that gets moved. This is very counterintuitive. At first I thought that resizing the one I moved was not working as I assumed it should be updating the last 1D plot (the one on top).
* Clearing slicers does not clear the 1D plots. Should it? I think so, specially now that we may have several plots from one 2D plot. Also @ChristianWoegerbauer work to promote the slicer plot to a regular data set, if the user wants to maitain that data they presumably will have pressed send to fitting?
* It is not clear to me that there is a use case for having two different types of slicers on the plot at once? And if the normal desire is to change slicer types it might be useful to revert that behavior to clearing the slicers and creating the new one. That said, all the user has to learn to do is to `clear slicers` before asking for a different slicer type so maybe it is OK to leave as is for now?

butlerpd:
Another question: what happens if I add enough slicers to go through the list of 15 defined colors? do the colors recycle back at the beginning?

butlerpd:
Using the long flight to do a more thorough functionality review. The following is the result

FIXED
* After a 2nd slicer is added, manually moving it now moves the topmost (last) slicer. 👍 

PROBLEMS:
* If one or more slices are created before launching the `Edit Slicer Parameters` panel 
  * The top half of the new editor panel shows the first slice as selected (its radio button is highlighted) but the parameters shown in the bottom box are for the last slice. Normally the bottom box should match the selected slice from the top box and it does except for this startup case.
  * Also, the `create new` slicer dropdown in the top box shows the type selected instead of `none` and new slices of the same type cannot be created from the panel until a different type of slicer is chosen. Once reset, creating a new slice from the top panel of the `Edit Slicer Parameters` works and will add a new slice, make its radio button selected and return to showing `none` in ehe `create new` drop down box, thus allowing multiple slices of the same type again.
  * Choosing `none` in the create new box when it is not already `none` (i.e. in this edge case described above since it is otherwise always `none`)  effectively does a "clear slicers" before closing the panel altogether.
* I somehow managed at one point (after several different types of slicers and several of each and deleting some 1D plots to see things) to get to a place where clear slicer cleared all but two sector slices. They were apparently ghosts as `clear slicers` was no longer a choice. Adding another slicer and trying to `clear slicers` again only cleared the new slicer. This may need more investigation as I could not quickly reproduce.

OTHER THINGS
* There is no way to remove all the subplots created either from the data explorer OR by clearing slicers etc. I found that subplots cannot be deleted from the `DataExplorer` in general. Only the top level data can be deleted. Doing so is the only way to delete all subplots. So this would be a different issue probably unless we think `clear slicers` should do that? Not sure that would be the desired behavior but thought I'd raise the question.
* However, and this may have been discussed previously, I wonder it clearing slicers should not at least remove all the associated plots as part of the effort to keep from cluttering up the window? Since they are now permanent subplots in the `DataExplorer` (as discussed above) people can still plot them whenever they want? Maybe a question to ask at a fortnightly meeting?
* The batch tab only allows picking the available 2D plots. That means that clicking apply will act on ALL the 1D plots associated with the selected 2D plots. There is no way to pick only some 1D plots. Is that a problem? That may be the desired behavior but worth bringing up as a question I figure.
* I thought there was a third tab being added? Or am I thinking of something else? or a discussion where the answer was no?

butlerpd:
This all looks good to me now. I was able to throw an error at one point but was never able to reproduce it.  I think it is ready to merge from the functionality side. I note Jeff still has comments?  Also I think the documentation needs more work. It is not at all clear to me. I suspect it is time to add a screen shot or two and a bit more detail about the context menu vs the slicer parameter window. Actually, that whole plotting data/model section (graph_help) probably needs a facelift.

Given that I would vote for merging this and add a documentation upgrade as a new issue.

src/sas/qtgui/Plotting/Plotter2D.py
```diff
@@ -411,81 +458,105 @@ def updateSlicer(self):
...
+        Get the next color for a new slicer and increment the index
         """
-        # Clear current slicer
-        if self.slicer is not None:
-            self.slicer.clear()
+        color = SLICER_COLORS[self._slicer_color_index % len(SLICER_COLORS)] # so colours don't overflow
+        self._slicer_color_index += 1
+        return color
 
+    def _removeSlicerPlots(self):
```
krzywon:
```suggestion
       def _recurse_plots_to_remove(self, item):
        temp_plots_to_remove = []
        slicer_type_id = 'Slicer' + self.data0.name
        for item_index in range(item.rowCount()):
            child = item.child(item_index)
            if child.hasChildren():
                self._recurse_plots_to_remove(child)
            if not (isinstance(child.data(), (Data1D, Data2D))
                    and hasattr(child.data(), 'type_id')
                    and slicer_type_id in child.data().type_id
                    and child.rowCount() > 0
                    and child.child(0).text()):
                        temp_plots_to_remove.append(item.child(item_index))

        for plot in temp_plots_to_remove:
            item.removeRow(plot.row())

    def _removeSlicerPlots(self):
        """
        Clear the previous slicer plots
        """
        # Clear the old slicer plots so they don't reappear later
        if not hasattr(self, '_item'):
            return
        item = self._item
        if self._item.parent() is not None:
            item = self._item.parent()
        self._recurse_plots_to_remove(item)
```

src/sas/qtgui/Plotting/SlicerParameters.py
```diff
@@ -289,6 +408,46 @@ def onApply(self):
...
+        """
+        Delete the current slicer
+        """
+        # Pop up a confirmation dialog
+        reply = QtWidgets.QMessageBox.question(self, 'Delete Slicer',
+                                             'Are you sure you want to delete this slicer?',
+                                             QtWidgets.QMessageBox.Yes | QtWidgets.QMessageBox.No,
+                                             QtWidgets.QMessageBox.No)
+        if reply == QtWidgets.QMessageBox.No:
+            return
```
krzywon: Good use of defaults here! This will catch the window closing by the ESC key, or the window X.

krzywon (approved):
I think this might be ready to merge! One code suggestion that I'll leave to @jellybean2004 to implement or not.

A future improvement I could see: plot all slicers on the same plot (differentiating phi vs Q of course), but this might be best once the plotting refactor is better integrated.

PR #3766, opened 2025-11-15
3762 error in data operation null outputs

### LINKED ISSUE #3762: Error appearing when generate output dataset containing zero or negative values in "Data Operation" window
**Describe the bug**
When tring to subtract the same dataset in the Data Operation tool, pressing the compute button will throw an error. The preview is loaded, but it shows an empty plot. The new output dataset is created normally.
The error is:
`Traceback (most recent call last): File "sas\qtgui\Calculators\DataOperationUtilityPanel.py", line 147, in onCompute File "sas\qtgui\Calculators\DataOperationUtilityPanel.py", line 430, in updatePlot File "sas\qtgui\Plotting\Plotter.py", line 305, in plot File "matplotlib\axes\_base.py", line 3707, in set_xbound File "matplotlib\axes\_base.py", line 3828, in set_xlim File "matplotlib\axis.py", line 1216, in _set_lim File "matplotlib\axes\_base.py", line 3749, in _validate_converted_limits ValueError: Axis limits cannot be NaN or Inf`

**To Reproduce**
1. Go to Data Operation tool
2. Select a dataset for Data1 and Data2, being Data2 with the same or higher Intenisty than Data1
3. Select the subtraction operator and then click the "Compute" button

**Expected behavior**
I would probably expect some warning in the log explorer instead of the raw error.

**Screenshots**

Image

 - Version: 6.1.1

 - OS: Windows 10

 - Installed: Windows installer

**Additional context**

Cristian2299:
## Description

A condition was added to make sure the user cannot select the same dataset in both Data1 and Data2 comboboxes in the Data Opeartion tool. In the main version, if the user performs the subtraction of the same 1D data, an error raise. Now, the user cannot perform this action, and when the same 1D data is chosen in combobox2 a message appears in the log explorer.

Fixes #3762 

## How Has This Been Tested?

Tested on windows, version 6.1.2.dev

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
+++ b/src/sas/qtgui/Calculators/DataOperationUtilityPanel.py
@@ -135,7 +135,7 @@ class DataOperationUtilityPanel(QtWidgets.QDialog, Ui_DataOperationUtility):
             # set operator to be applied
             operator = self.cbOperator.currentText()
             # calculate and send data to DataExplorer
-            output = None   
+            output = None
             try:
                 data1 = self.data1
                 data2 = self.data2
@@ -329,7 +329,7 @@ class DataOperationUtilityPanel(QtWidgets.QDialog, Ui_DataOperationUtility):
                 self.cbData2.setStyleSheet(BG_RED)
                 logger.error('Cannot compute 2D data of different lengths')
                 return False
-                
+
             else:
                 self.cbData1.setStyleSheet(BG_WHITE)
                 self.cbData2.setStyleSheet(BG_WHITE)


rozyczko (changes requested):
This works only in one case:
1. Select dataset 1
2. Select subtraction operation
3. Select dataset 2 (identical to 1)

The message is shown and the behavour is correct
However, any other combination breaks the pattern and we see the same exception as in the original. Eg.

1. Select dataset 1
2. Select dataset 2 (identical to 1)
3. Select subtraction operation

Cristian2299:
The check for the same dataset was moved now at the beginning of the compute button.

rozyczko:
After discussing this with other people it seems that the consensus is to allow this kind of operations, but make sure the exceptions aren't shown explicitly. Showing an empty chart (or e.g. showing a straight line if we do `set1/set`) is a perfectly reasonable request.

Let's just try to hide the exception (or try to find how to work around it) rather than prohibit the user from choosing this option.

Issue #3624, opened 2025-09-30
Magnetic parameter minimum limits not editable some times


krzywon:
**Describe the bug**
As noted in #3581, the minimum limits for magnetic parameters are not modifiable at times. This is not 100% repeatable, so this might need further investigation.

https://github.com/SasView/sasview/pull/3581#issuecomment-3347137051

**To Reproduce**
1. Load a 2D data file and model that can use magnetic fitting (sphere...)
2. Change to a different model
3. Attempt to modify the lower limit of a magnetic parameter.
4. The minimum limit is not recognized.

**Expected behavior**
All parameter values should be modifiable at any point outside of during the fit.

**SasView version (please complete the following information):**
 - Version: 6.1.1RC

PR #3768, opened 2025-11-15
Revert the polydisp table to the original state on model change

rozyczko:
## Description

Running a fit with polydisperse parameters checked adds an error column to the polydisp table.
This is OK when just running the fit - the min/max/npts/nsigs/function combobox columns get shifted and display ok.
But when a model is changed AFTER the fitting, the table is not properly reindexed.
THis PR fixes the issue.

Fixes # 3536

## How Has This Been Tested?

Local win 10 testing

## Review Checklist:

[if using the editor, use `[x]` in place of `[ ]` to check a box]

**Documentation** (check at least one)
- [ X] There is **nothing** that needs documenting
- [ ] Documentation changes are **in this PR**
- [ ] There is an **issue** open for the documentation (link?)

**Installers**
- [ ] There is a chance this will affect the **installers**, if so
  - [ ] **Windows** installer (GH artifact) has been tested (installed and worked) 
  - [ ] **MacOSX** installer (GH artifact) has been tested (installed and worked)
  - [ ] **Wheels** installer (GH artifact) has been tested (installed and worked)

**Licensing** (untick if necessary)
- [x] The introduced changes comply with SasView license (BSD 3-Clause)

PR #3772, opened 2025-11-16
unify behaviour of ViewDelegate when changing models

dehoni:
## Description

The minimum limits for magnetic and polydispersity parameters are not modifiable after fitting and changing between models.

This PR fixes https://github.com/SasView/sasview/issues/3624

Load a data file and model that can use magnetic or polydispersity fitting (sphere...)
Change to a different model
Attempt to modify the lower limit of a magnetic or polydispersity parameter.
The minimum limit is not modifiable and the entries of the table seem to have shifted to the left, such that the unit row show a number .
Expected behavior
All parameter values should be modifiable at any point outside of during the fit and the table should keep the information at the correct place.

SasView version (please complete the following information):

Version: 6.1.1RC

## Review Checklist:

**Installers**


rozyczko:
I think I addressed this very issue in https://github.com/SasView/sasview/pull/3768 ;)

Issue #3741, opened 2025-07-17
Error for dispersity for multishell modells


dehoni:
Every model with selectable number of shells larger than 1 throws an error, e.g. using core-multishell with more than 1 shell.
Trying to use PD one the shell no2, 3, etc gives the error:

ERROR: Traceback (most recent call last): File "sas\qtgui\Perspectives\Fitting\PolydispersityWidget.py", line 248, in File "sas\qtgui\Perspectives\Fitting\PolydispersityWidget.py", line 308, in onPolyComboIndexChange File "C:\SasView-6.1.0\sasmodels\sasview_model.py", line 816, in set_dispersion raise ValueError("%r is not a dispersity or orientation parameter" ValueError: 'thickness[n_shells]' is not a dispersity or orientation parameter.

If you want to reproduce the error:
1. select a model like sphere>core_multi_shell
2. choose number of shell to be larger than 1
3. Tick PD and set parameter for shell no 2, you need to set distribution function away from default Gaussian to obtain the error.

Version: Sasview 6.1 WindowsInstaller, Windows 11, Version 6.1 release and 6.1.1dev19+g260ab6d56 (that fixes some PD issues https://github.com/SasView/sasview/pull/3505)

PR #3779, opened 2025-11-17
Fixing name for dispersity parameter for multishell modells

dehoni:
## Description

onPolyComboIndexChange takes the parameter name as thickness[n] instead of the correct value, e.g. for shell 2 this should be thickness2, such that sasmodels can handle it. A fix needs to ensures to pass on an expanded parameter name like thickness2 to kernel_module.set_dispersion instead of the thickness[n] which triggered the ValueError.

Fixes https://github.com/SasView/sasview/issues/3741

## How Has This Been Tested?

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -1,5 +1,6 @@
 import copy
 import functools
+import logging
 import math
 import textwrap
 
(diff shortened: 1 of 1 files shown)


butlerpd:
I have not had a chance to test this but my first question would be how this might affect saved experiment files using a multishell model?

PR #3786, opened 2025-11-22
Support for SAXS fitting (v2)

klytje:
## Description

After long discussions on how to support SAXS fitting in SasView in the short term, it was decided to extend the generic scattering calculator with a 'SAXS fitting' option. 

When this option is selected, the `Compute` button text is replaced with `Generate plugin model`. Loading a `Nuclear data` file with this option enabled circumvents the usual loading logic, and instead forwards the selected file path to `AUSAXS`. This serves two purposes: 
1. `AUSAXS` also supports `mmCIF` files, which is the new standard in biological scattering, and
2. we avoid having to extend/modify the `PDB` loading logic with the additional information required for SAXS calculations. 

Clicking the `Generate plugin model` will then create a new `SAXS fit (<filename>)` model, which is automatically selected & opened for the user in the fit panel. The SAXS data may then be loaded in & fitted. 

Though a little awkward, I found this implementation to work decently well, at least for a short-term solution. In the long run, I'd like to have the following:
1. The ability to load both the SAXS data & structure file directly into the fitting panel (see my discussion [here](https://github.com/orgs/SasView/discussions/3739)). 
2. An optional initialization function for plugin models, which is invoked once before the actual Iq calculations start. I need this to avoid attaching the loaded structure file as a property to the Iq function. Having it float in the module namespace would make it load on SasView startup, which is also not ideal.
3. Custom plugin model interfaces in the fit panel. There's a lot of additional options available in `AUSAXS` which are currently difficult/impossible to expose to the user. Though this could also be implemented through a dedicated SAXS fitting window, for (1) to work, this is required. 
4. We could potentially extend the `SasView` file loaders to support `mmCIF` & full `PDB` data loading, if we want to avoid relying on `AUSAXS` for this. Alternatively, we could continue to rely on `AUSAXS` to parse the files, but store the data on the `SasView` side. Or just keep it as is. I'm indifferent on this issue. 

This PR replaces #3194.

## Review Checklist:

**Documentation**
Do we need documentation for this? There's links to the `AUSAXS` paper in the generated plugin model file itself, if people are interested. 

**Installers**


src/sas/qtgui/Calculators/SAXSPluginModelGenerator.py
```diff
@@ -0,0 +1,65 @@
...
+parameters = [
+    # name, units, default, [min, max], type, description
+    ['c', '', 1, [0, 100], '', 'Solvent density'],
+    #['d', '', 1, [0, 2], '', 'Excluded volume parameter']
+]
+
+###
+import pyausaxs as ausaxs
+
+structure_path = "{structure_path}"
```
wpotrzebowski: This maybe problematic on Windows. Backslashes in the path (e.g. C:\Users\Foo\) will be interpreted as escape sequences. Forcing raw string should help: structure_path = {structure_path!r}

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -704,13 +723,20 @@ def loadFile(self):
...
             if load_nuc:
-                f_type = """
-                    All supported files (*.SLD *.sld *.pdb *.PDB, *.vtk, *.VTK);;
-                        SLD files (*.SLD *.sld);;
-                        PDB files (*.pdb *.PDB);;
-                        VTK files (*.vtk *.VTK);;
-                        All files (*.*)
-                """
+                if self.model.type is ComputationType.SAXS:
+                    f_type = """
```
krzywon: To cover all cases without a nested if and no repetition (this needs a little modification):
```suggestion
cif = " *.CIF *.cif"
cif_files = "CIF files (*.CIF *.cif);;\n        "

use_cif = self.model.type is ComputationType.SAXS

f_type = f"""
    All supported files (*.SLD *.sld *.pdb *.PDB, *.vtk, *.VTK{cif if use_cif else ""});;
        SLD files (*.SLD *.sld);;
        PDB files (*.pdb *.PDB);;
        VTK files (*.vtk *.VTK);;
        {cif_files if use_cif else ""}All files (*.*)
"""
```

build_tools/requirements.txt
```diff
@@ -16,7 +16,7 @@ numpy
 packaging
 periodictable
 platformdirs
-pyausaxs==1.0.4
```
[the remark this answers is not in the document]
klytje: I agree. I decided to pin it because I have not yet settled on the final API for `pyAUSAXS`, and would prefer not to break all `SasView` builds when/if I make changes to it. I will remove this pin once the API is finalized.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -658,6 +659,24 @@ def change_is_avg(self):
         self.checkboxLogSpace.setEnabled(self.is_avg)
         self.checkboxPluginModel.setEnabled(self.is_avg)
 
+        # set the type of calculation
+        match self.cbOptionsCalc.currentIndex():
```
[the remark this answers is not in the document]
klytje: Default case now raises an exception to inform developers they forgot to handle their new case.

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
klytje: Good suggestion. I'm not sure if we should change it though, as the SAXS fitter is only supposed to live in the GSC window temporarily (however long that may be). I have therefore tried to make the overall changes to the file as minimal as possible (like my nested `if` which can easily be removed again).

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -633,7 +634,7 @@ def update_cbOptionsCalc_visibility(self):
             self.checkboxLogSpace.setEnabled(not self.is_mag)
 
 
-    def change_is_avg(self):
+    def change_computation_type(self):
```
krzywon: Nitpick: It's often a bad idea to completely replace a method, in case there are direct calls to it that may have been missed.

```suggestion
    def change_is_avg(self):
        self.change_computation_type()

    def change_computation_type(self):
```

krzywon:
I tested today and am getting a warning when trying to create the plugin model. I think this is a Win-specific issue. The path saved in the plugin model uses backslashes causing a unicode error.

When I open the file generated in the model editor, I get the following error:

> 2026-01-13 14:56:52,113 : WARNING : sas.sascalc.fit.models (models.py:166) :: Failed to load plugin '\\sasview\\SasView\\plugin_models\\ausaxs_saxs_plugin.py'. See \sasview\SasView\plugin_models\plugins.log for details
2026-01-13 14:57:07,276 : ERROR : sas.qtgui.Utilities.ModelEditors.TabbedEditor.TabbedModelEditor (TabbedModelEditor.py:448) :: Error building model: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape (<unknown>, line 21)
2026-01-13 14:57:07,278 : ERROR : sas.qtgui.Utilities.ModelEditors.TabbedEditor.TabbedModelEditor (TabbedModelEditor.py:454) ::     structure_path = "\sasview\src\sas\example_data\coordinate_data\diamond.pdb"

src/sas/qtgui/Calculators/GenericScatteringCalculator.py
```diff
@@ -1412,6 +1445,48 @@ def onCompute(self):
...
+                    # load the model into the parameter table
+                    fw.cbModel.setCurrentIndex(midx)
+                    fw.onSelectModel()
+
+                    # make sure the perspective window is visible and focused
+                    self.close() # close the calculator window to highlight the changes to the fitting window
+                    per.show()
+
+            except Exception:
+                pass
```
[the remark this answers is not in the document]
klytje: What about something like this? It's not critical that this part succeeds, which is why I wrapped it in a try/catch clause.  
`logger.warning("Could not bring the newly generated plugin model into focus in the Fitting window. Please report this issue.")`

klytje:
`pyausaxs` is no longer pinned as it was causing issues with the debian sasview packaging.

klytje:
I have now added support for processing files with unknown residues, but I really do not think we should enable this by default. It needs an explicit toggle so users are aware of what is actually being done with their data, which is currently not possible. 

Regarding automatically closing the GSC: I tried disabling this, and I found that to be an even worse experience, as there is no feedback that something actually happened. I think that, though it may not be ideal, closing the GSC & opening the fitting panel with the model already selected gives much clearer feedback. And it would be the next natural step to take for the user anyway.

krzywon (approved):
I think this is ready. I think the two flags added to the SAXS model file need more (any?) documentation, but that shouldn't hold this PR up.

Let's wait to merge this until after the biweekly call next week.

PR #3789, opened 2025-12-01
Invariant Refactor

### LINKED ISSUE #1496: Refactor invariant perspective
At present the invariant calculation has several limitations, including requiring a knowledge of the contrast term from which it computes the volume fraction but not the other way round. Other problems include lack of input parameter uncertainty propagation and the immediate computation of the (incorrect) invariant as soon as the perspective is called.
   
TODO task list for this issue:
- [ ] Refactor the Invariant to accept either the vol fraction or the contrast factor as inputs and compute the other using the invariant. 
- [ ] Accept uncertainty values along with the parameter itself (Porod, contrast, vol fraction) and propagate the through to the uncertainty on computed values. Currently, only the uncertainty on intensity is all that is propagates.
- [ ] Do not supply default values for automatic computation of the Invariant. Currently the contrast factor defaults to 1e-6 and an incorrect invariant is immediately calculated and displayed in the GUI.
- [ ] Edit documentation to match changes

jellybean2004:
## Description

Note: This branch is based on Andrew Jackson's (@ajj) branch, which started refactoring the perspective to incorporate volume fraction as an input.

- Changes the UI for Invariant perspective:
 - The first tab contains all the input/output for the invariant/contrast/volume fraction calculation
 - The second tab contains the extrapolation options and parameters
 - The extrapolation slider has been replaced with the slider from Corfunc Perspective (the gradient bar in the extrapolation tab), and the slicers on the plot are just a visual guide
 - Progress bars from the status window are also visible on the extrapolation tab
 - This also replaces the 'number of points' input with a Q value input
 - The data plot is now shown upon loading data for easier visualisation

- Adds SLD contrast calculation from volume fraction
 - Users can now choose between the two corollary calculations

Known issues:
- ~~The plot displayed after loading the data shows the data points twice (overlapping). This had to be done due to how the plotting works, and has been left unresolved in anticipation of the plotting refactor~~
    Resolved: the initial plot now uses the functionality of the `Create New` button at the bottom of Data Explorer. Extrapolated plots still use the invariant plotting, replacing the initial plot.
- The sliders on the plot can be moved, but they do not change the extrapolation values, rendering them nonfunctional. If one changes the sliders on the plot and runs a calculation, one will notice that the sliders move back to their initial positions (as defined by the slider on the tab). This is also left unresolved for the same reason as above

Fixes #1496 

## How Has This Been Tested?

Manually ran the software with different datasets to check the functionality, and compared with the old version to reproduce the same results.

## Review Checklist:

- [x] There is an **issue** open for the documentation #3813 

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
+++ b/src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
@@ -327,6 +327,7 @@ class InvariantWindow(QtWidgets.QDialog, Ui_tabbedInvariantUI, Perspective):
         if self.high_extrapolation_plot:
             self.high_extrapolation_plot.plot_role = DataRole.ROLE_DEFAULT
             self.high_extrapolation_plot.symbol = "Line"
+            self.high_extrapolation_plot.custom_color = "#2ca02c"
             self.high_extrapolation_plot.show_errors = False
             self.high_extrapolation_plot.show_q_range_sliders = True
             self.high_extrapolation_plot.slider_update_on_move = False
@@ -342,6 +343,7 @@ class InvariantWindow(QtWidgets.QDialog, Ui_tabbedInvariantUI, Perspective):
         if self.low_extrapolation_plot:
             self.low_extrapolation_plot.plot_role = DataRole.ROLE_DEFAULT
             self.low_extrapolation_plot.symbol = "Line"
+            self.low_extrapolation_plot.custom_color = "#ff7f0e"
             self.low_extrapolation_plot.show_errors = False
             self.low_extrapolation_plot.show_q_range_sliders = True
             self.low_extrapolation_plot.slider_update_on_move = False


src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -1,21 +1,25 @@
...
 
 import numpy as np
 from PySide6 import QtCore, QtGui, QtWidgets
 from twisted.internet import reactor, threads
 
 import sas.qtgui.Utilities.GuiUtils as GuiUtils
+from sas.qtgui.Perspectives.Corfunc.CorfuncSlider import CorfuncSlider
+from sas.qtgui.Plotting import PlotterData
 from sas.qtgui.Plotting.PlotterData import Data1D, DataRole
+from sas.sascalc.corfunc.calculation_data import ExtrapolationInteractionState, ExtrapolationParameters
```
DrPaulSharp: We should look to extract these common routines from the `Corfunc` perspective so that we do not introduce cross-perspective dependencies.

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -129,47 +136,91 @@ def __init__(self, parent=None):
...
+
+        if state and self.rbLowQFix_ex.isChecked():
+            self.txtLowQPower_ex.setEnabled(True)
+        else:
+            self.txtLowQPower_ex.setDisabled(True)
+
+        if state and self.rbHighQFix_ex.isChecked():
+            self.txtHighQPower_ex.setEnabled(True)
+        else:
+            self.txtHighQPower_ex.setDisabled(True)
```
DrPaulSharp: Can this code be simplified to:

```
if state:
    self.txtLowQPower_ex.setEnabled(self.rbLowQFix_ex.isChecked())
    self.txtHighQPower_ex.setEnabled(self.rbHighQFix_ex.isChecked())
```

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -129,47 +136,91 @@ def __init__(self, parent=None):
...
     def set_high_q_extrapolation_lower_limit(self, value):
         n_pts = len(self._data.x) - (np.abs(self._data.x - value)).argmin() + 1
-        item = QtGui.QStandardItem(str(int(n_pts)))
-        self.model.setItem(WIDGETS.W_NPTS_HIGHQ, item)
-        self.txtNptsHighQ.setText(str(n_pts))
+        self._high_points = n_pts
 
-    def enabling(self):
-        """ """
+    def enabling(self) -> None:
```
DrPaulSharp: I think `enableStatus` would be a better name here.

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -365,24 +455,53 @@ def calculate_thread(self, extrapolation: str) -> None:
...
         if self._high_extrapolate:
-            function_high = "power_law"
+            function_high: str = "power_law"
             if self._high_fit:
                 self._high_power_value = None
+            elif self._high_fix:
+                self._high_power_value = float(self.model.item(WIDGETS.W_HIGHQ_POWER_VALUE_EX).text())
+            else:
+                logger.debug("High Q extrapolation: no fit/fix option selected, using default power value")
+                self._high_power_value = DEFAULT_POWER_VALUE
```
DrPaulSharp: Is it possible to hit this clause - surely one of fit/fix must always be selected?

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -365,24 +455,53 @@ def calculate_thread(self, extrapolation: str) -> None:
...
+                    n_pts_high: int = len(self._data.x) - idx
+
+                    # Ensure n_pts_high is within valid bounds
+                    n_pts_high = max(1, min(n_pts_high, len(self._data.x)))
+
+                    if n_pts_high != self._high_points:
+                        self._high_points = n_pts_high
+
+            except Exception as ex:
+                logger.debug(f"Could not convert Porod start to n_pts_high: {ex}")
```
DrPaulSharp: Again, what precise exception do you expect?

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -397,34 +516,58 @@ def calculate_thread(self, extrapolation: str) -> None:
...
-            volume_fraction = "ERROR"
-            volume_fraction_error = "ERROR"
-        reactor.callFromThread(self.update_model_from_thread, WIDGETS.W_VOLUME_FRACTION, volume_fraction)
-        reactor.callFromThread(self.update_model_from_thread, WIDGETS.W_VOLUME_FRACTION_ERR, volume_fraction_error)
+        if self.rbContrast.isChecked() and self._contrast:
+            try:
+                volume_fraction, volume_fraction_error = self._calculator.get_volume_fraction_with_error(
+                    self._contrast, extrapolation=extrapolation
+                )
+            except Exception as ex:
```
DrPaulSharp: What exception???

src/sas/qtgui/Perspectives/Invariant/InvariantPerspective.py
```diff
@@ -397,34 +516,58 @@ def calculate_thread(self, extrapolation: str) -> None:
...
+            reactor.callFromThread(self.update_model_from_thread, WIDGETS.W_CONTRAST_OUT_ERR, "")
+
+        # Surface Error calculations
+        if self._porod:
+            # Use calculated contrast if in volume fraction mode, otherwise use input contrast
+            contrast_for_surface = contrast_out if self.rbVolFrac.isChecked() and self._volfrac1 else self._contrast
+            if contrast_for_surface:
+                try:
+                    surface, surface_error = self._calculator.get_surface_with_error(contrast_for_surface, self._porod)
+                except Exception as ex:
```
DrPaulSharp: What exception?

src/sas/sascalc/invariant/invariant.py
```diff
@@ -933,6 +933,49 @@ def get_volume_fraction(self, contrast, extrapolation=None):
             msg = "Could not compute the volume fraction: inconsistent results"
             raise RuntimeError(msg)
 
+    def get_contrast(self, volume, extrapolation=None):
+        """
+        Compute contrast is deduced as follows: ::
+
+            q_star = 2*(pi*contrast)**2* volume(1- volume)
```
DrPaulSharp: Change to `q_star = 2*(pi*contrast)**2 * volume*(1- volume)` to be clearer.

DrPaulSharp (commented):
Looking good so far! Alongside the code comments here, I have a number of suggestions from my functionality tests:

- The text boxes for the q ranges look editable, but they (correctly) are not. A better visual cue to this would be helpful.
- The type used for units appears bold for all except the cm<sup>-1</sup> used for the background. All units should look like this if possible.
- When entering "2e-6" for the contrast, it turns red halfway through. You probably want to change to an `editingFinished` signal here.
- For the output boxes, I think it would make sense to grey out the contrast box when contrast is the input, and similarly for the volume fraction (i.e, mirroring how the input boxes look)
- The second tab is called "Page", I'd suggest "Extrapolation" makes more sense.
- In the extrapolation tab, why do we have the "name" text box? Would it make more sense to have the name of the dataset from the first tab?
- The text boxes for the Extrapolation limits are too narrow.
- The radio buttons for the low-q extrapolation are a bit squashed.
- The extrapolation box goes wider to the left than the Invariant chart box.
- !! Given the above points, I'd suggest making the perspective window a bit wider. It's narrower than Corfunc, so we don't need to worry too much about space. !!
- The high q extrapolation is a power law rather than Porod - note that Porod only applies when the power is equal to 4. See https://www.sasview.org/docs/user/qtgui/Perspectives/Invariant/invariant_help.html, https://www.sasview.org/docs/user/qtgui/Perspectives/Corfunc/corfunc-technical.html and lines 256-265 of `sascalc/invariant/invariant.py`.
- The low q extrapolation can be either power law or Guinier, but the text box and slider always say Guinier.
- !! Again, given the above, I think switching the text labelling the text boxes and on the slider to "low-q" and "high-q" makes more sense.
- There is always a power displayed for the low q extrapolation, even when Guinier is used. Can we extract and display the Guinier parameters, as we do in Corfunc?
- Note that some comments in `sascalc/invariant/invariant.py` incorrectly state that Guinier puts the radius of gyration to a power of 2/3 - this should be corrected.
- In the extrapolation tab, the "A-1" at the end of the ranges text boxes should have the -1 as a superscript.
- We need to think about the format of Guinier for this perspective compared to Corfunc (note the docs use equaivalent, but different parameters). This is one to discuss more widely.
- For the output, I think it makes sense to have both volume fractions to make things completely clear.
- If one does a specific surface calculation, but then removes the Porod constant, the box turns red and the value from the last calculation is maintained. Instead, the box should stay white and the value cleared when the calculation button is pressed.
- I would move the radio buttons for contrast and volume fraction to the right hand side, so they are clearly grouped with the text boxes.
- The error for a surface calculation is always None.
- You've already noted this, but simply running a calculation reduces the two datasets plotted to one.
- If a volume fraction outside 0-1 is entered the calculation cannot be performed and the boxes turn red (correctly). I think a message of some kind explaining why would be useful. What do you think?
- When I delete the data, I get an error: "Traceback (most recent call last): File "C:\Users\gnn85523\projects\sasview\src\sas\qtgui\MainWindow\GuiManager.py", line 1404, in showPlot self.filesWidget.displayData(plot) File "C:\Users\gnn85523\projects\sasview\src\sas\qtgui\MainWindow\DataExplorer.py", line 1156, in displayData main_data = GuiUtils.dataFromItem(plot_item.parent()) ^^^^^^^^^^^^^^^^ AttributeError: 'NoneType' object has no attribute 'parent'"

Good luck with these, let me know if you need a hand.

src/sas/qtgui/Perspectives/Invariant/UI/TabbedInvariantUI.ui
```diff
@@ -6,452 +6,658 @@
...
+          </property>
+          <property name="readOnly">
+           <bool>true</bool>
+          </property>
+         </widget>
+        </item>
+        <item row="5" column="0">
+         <widget class="QLabel" name="lblInvariantTotalQ">
+          <property name="text">
+           <string>Invariant Total [Q]:</string>
```
DrPaulSharp: This line should read: `<string>Invariant Total [Q*]:</string>`

DrPaulSharp:
@butlerpd I'd be interested to hear your thoughts on #3825 and how that impacts this PR.

jellybean2004:
> > * The sliders on the plot can be moved, but they do not change the extrapolation values, rendering them nonfunctional. If one changes the sliders on the plot and runs a calculation, one will notice that the sliders move back to their initial positions (as defined by the slider on the tab). This is also left unresolved for the same reason as above
> 
> The sliders are added in the `InvariantWindow.plot_result()`. Remove `show_q_range_sliders` and any other plot attribute that starts with `slider` if they aren't needed any longer.

We want to keep the sliders on the plot as a visual guide, so they are needed for that, but as per previous discussions, we want to disable them from updating the extrapolation parameters.

PR #3806, opened 2025-12-08
Slicer plot stacking

jellybean2004:
## Description

These changes implement a plot stacking feature for slicers, allowing multiple slices from the same slicer type to be displayed on the same plot window instead of creating separate windows, along with UI controls to manage and delete slicer-generated plots.

- Added a "Stack Plots" checkbox to the slicer parameters dialogue box, which, when checked, adds plots of the same slicer type to the same canvas
- Stacking set set to True by default, ie. slicers append to the same 1D plot.
- Added a "Slicer Plots" tab with a list widget to display plots created by slicers
- Added a "Delete" button to remove selected slicer plots
- Created the `StackableMixin` class in SlicerUtils.py to provide plot stacking functionality, and modified `AnnulusInteractor`, `SectorInteractor`, and `WedgeInteractor` to inherit from it
- Implemented `onDeleteSlicerPlots()` to delete selected slicer plots:
    - Shows confirmation dialogue
    - Closes plot windows
    - Removes from PlotHelper, active plots, and the manager's plot widgets
    - Cleans up from data structures

To use this feature:

- Create a 2D plot
- Open *Slicer Parameters* dialog (right click plot > `Slicers` > `Edit Slicer Parameters`)
- Notice the `Stack Plots` checkbox (checked by default)
- Add slicers to the plot
- Notice that slicers of the same type add the data points to the same 1D plot instead of creating new plots
- Disable stacking to create individual plots for every new slicer added

## How Has This Been Tested?

Manually ran the software to test functionality.

## Review Checklist:

- [x] There is an **issue** open for the documentation (link?)

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/BoxSlicer.py
@@ -2,16 +2,15 @@ import logging
 
 import numpy
 
-import sas.qtgui.Utilities.GuiUtils as GuiUtils
 from sas.qtgui.Plotting.PlotterData import Data1D
 from sas.qtgui.Plotting.SlicerModel import SlicerModel
 from sas.qtgui.Plotting.Slicers.BaseInteractor import BaseInteractor
-from sas.qtgui.Plotting.Slicers.SlicerUtils import generate_unique_plot_id
+from sas.qtgui.Plotting.Slicers.SlicerUtils import StackableMixin, generate_unique_plot_id
 
 logger = logging.getLogger(__name__)
 
 
-class BoxInteractor(BaseInteractor, SlicerModel):
+class BoxInteractor(BaseInteractor, SlicerModel, StackableMixin):
     """
     BoxInteractor plots a data1D average of a rectangular area defined in
     a Data2D object. The data1D averaging itself is performed in sasdata
@@ -28,6 +27,8 @@ class BoxInteractor(BaseInteractor, SlicerModel):
     def __init__(self, base, axes, item=None, color="black", zorder=3, direction=None):
         BaseInteractor.__init__(self, base, axes, color=color)
         SlicerModel.__init__(self)
+        StackableMixin.__init__(self)
+
         # Class initialization
         self.markers = []
         self.axes = axes
@@ -170,6 +171,9 @@ class BoxInteractor(BaseInteractor, SlicerModel):
         self.horizontal_lines.save(ev)
         self.center.save(ev)
 
+    def _get_slicer_type_id(self):
+        return f"BoxSlicer{self.direction}{self.data.name}"
+
     def _post_data(self, new_slab=None, nbins=None, direction=None):
         """
         post 1D data averaging in Qx or Qy given new_slab type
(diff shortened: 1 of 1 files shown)


src/sas/qtgui/Plotting/Plotter2D.py
```diff
@@ -309,6 +311,46 @@ def getActivePlots(self):
...
+        slicer_plots = {}
+
+        if not hasattr(self, 'manager') or not hasattr(self.manager, 'active_plots'):
+            return slicer_plots
+
+        # Search through all active plots
+        for plot_id, plot_window in self.manager.active_plots.items():
+            if hasattr(plot_window, 'data') and plot_window.data is not None:
+                # Get the data (could be a list or single item)
+                data_list = plot_window.data if isinstance(plot_window.data, list) else [plot_window.data]
```
krzywon: `data_list = list(plot_window.data)` to simplify?

src/sas/qtgui/Plotting/Plotter2D.py
```diff
@@ -309,6 +311,46 @@ def getActivePlots(self):
...
+        that were created by slicers from this 2D plot.
+        """
+        slicer_plots = {}
+
+        if not hasattr(self, 'manager') or not hasattr(self.manager, 'active_plots'):
+            return slicer_plots
+
+        # Search through all active plots
+        for plot_id, plot_window in self.manager.active_plots.items():
+            if hasattr(plot_window, 'data') and plot_window.data is not None:
```
krzywon: To save some indentation later, a suggestion (though what plot window doesn't have the data attr?):

```
if not hasattr(plot_window, 'data') or not plot_window.data:
    # No data loaded (catches the cases where data == None and data == [])
    continue

data_list = ___
...
```

krzywon (commented):
Functionally, this seems to work as advertised. I can add multiple slicers to a 2D plot and, with the `stack` feature enabled, they will all show up on the same 1D plot. 

Code-wise, see my comments.

Other notes:
- This needs documentation.
- Stack is not the most intuitive name, and could be implied to add or average the plots together instead of viewing them all on the same plot. That said, I don't have a better name for it, so it's good enough for now.
- I would argue the stacking feature could be a global user preference. In addition, this could be enabled/disabled per plot with how you are doing it here. (Yes, both...)
- I would also argue the stacking feature should be the default behavior.

src/sas/qtgui/Plotting/Slicers/SlicerUtils.py
```diff
@@ -36,3 +44,194 @@ def generate_unique_plot_id(base_id: str, item) -> str:
...
+                if hasattr(plot_window, 'data') and plot_window.data is not None:
+                    # Get first data item
+                    data_list = plot_window.data if isinstance(plot_window.data, list) else [plot_window.data]
+                    if len(data_list) > 0:
+                        first_data = data_list[0]
+                        # Check if it's from the same slicer type and parent
+                        if (hasattr(first_data, 'type_id') and
+                            first_data.type_id is not None and
+                            first_data.type_id == type_id):
+                            return plot_window
```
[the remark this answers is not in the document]
jellybean2004: `first_data.type_id == type_id` is matching the selected slicer type, so it is needed, I believe it is needed. 
I will simplify the block and remove other redundancies

JoGaudet:
I tested the “Stack Plots” functionality for the slicer parameters. The two wedge slicers, the annulus slicer, and the sector slicers all appear to work as intended, allowing multiple slicers of the same type to be added to a single graph. No issues were found when modifying the various slicer parameters when multiple plots were created, etc.

However, I noticed that the Box interactor slicers (X and Y) do not support “Stack Plots.” Is this intentional? My impression is that it would be beneficial for the Box interactors to support this functionality as well.

JoGaudet:
One suggestion, which is related to one of Jeff's comments is that we could name the option for stacking as "overplot" instead of stack.

PR #3833, opened 2025-12-22
Generalize Corfunc Slider

### LINKED ISSUE #3824: Refactor the Corfunc slider to be more general
Following the merge of #3789, the slider used in Corfunc will also be used directly by the Invariant perspective. To avoid cross-perspective dependencies, and enable greater functionality for the slider, it should be moved out of Corfunc and refactored into a general slider - perhaps called the "extrapolation slider". One possible place for the slider to live is in `qtgui/Utilities`

jellybean2004:
## Description

This refactors the Corfunc slider into a reusable component so it can be shared by Invariant, making the slider generic and reusable for components that need different constraint semantics (e.g., Invariant needs to permit extrapolation beyond the raw data).

- Takes `CorfuncSlider.py` out of `src/sas/qtgui/Perspectives/Corfunc`
- Renames to `ExtrapolationSlider.py` and places in `src/sas/qtgui/Utilities`
- Moves utility classes from `src/sas/sascalc/corfunc/calculation_data.py` to `src/sas/sascalc/util.py`
- Updates all imports/usages to match the new locations and names

**Behavioural change (boundaries)**

Introduces two distinct boundary pairs used by different callers:

- Data range — the actual start/end of data points (existing boundary used by Corfunc).
- Extrapolation range — a broader fixed range used by Invariant (10e-5 … 10).

**Usage differences:**

Corfunc: uses Data range only (no functional change except location/imports).
Invariant: uses both:

- Low-Q end and High-Q start are constrained within the Data range.
- High-Q end extends into the Extrapolation range.

Fixes #3824 

## How Has This Been Tested?

Ran the program and tested the slider functionality in both perspectives.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
+++ b/src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
@@ -842,9 +842,6 @@ class CorfuncWindow(QtWidgets.QDialog, Ui_CorfuncDialog, Perspective):
         self.model.setItem(WIDGETS.W_QCUTOFF,
                            QtGui.QStandardItem(format_string%state.point_3))
 
-        # Check validity of the text boxes
-        # self.correct_extrapolation_values()
-
     def on_extrapolation_slider_changing(self, state: ExtrapolationInteractionState):
         """ Slider is being moved about"""
         self._q_space_plot.update_lines(state)
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,4841 @@
+version = 1
+revision = 3
+requires-python = ">=3.12"
+resolution-markers = [
+    "python_full_version >= '3.13'",
+    "python_full_version < '3.13'",
+]
+
+[[package]]
+name = "accesscontrol"
+version = "7.3"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "acquisition" },
+    { name = "authencoding" },
+    { name = "btrees" },
+    { name = "datetime" },
+    { name = "extensionclass" },
+    { name = "persistence" },
+    { name = "restrictedpython" },
+    { name = "transaction" },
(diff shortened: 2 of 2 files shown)


src/sas/qtgui/Perspectives/Corfunc/CorfuncPerspective.py
```diff
@@ -10,27 +10,26 @@
...
-# pylint: disable=import-error, no-name-in-module
-from sas.qtgui.Perspectives.Corfunc.CorfuncSlider import CorfuncSlider
 from sas.qtgui.Perspectives.Corfunc.ExtractionCanvas import ExtractionCanvas
 from sas.qtgui.Perspectives.Corfunc.IDFCanvas import IDFCanvas
 from sas.qtgui.Perspectives.Corfunc.QSpaceCanvas import QSpaceCanvas
 from sas.qtgui.Perspectives.Corfunc.RealSpaceCanvas import RealSpaceCanvas
 from sas.qtgui.Plotting.PlotterData import Data1D
+
+# sas-global
+# pylint: disable=import-error, no-name-in-module
```
DrPaulSharp: We're moving away from using pylint, so could you please remove these comments seeing as you're modifying the line.

jellybean2004:
@DrPaulSharp thank you for your review.

I have updated the slider input validation logic in both Invariant and Corfunc to account for edge cases and disable the buttons if the values are invalid. I have also rewritten the logic in Corfunc to match that of Invariant. So, instead of popping a dialog box on press, the button is just greyed out. What do you think of this?

I will also look into Pydantic a bit more.

DrPaulSharp (commented):
Ok, I've found another problem a bit like the last one. In the invariant perspective, if you put an invalid value into low-Q end or high-Q start, it snaps back to the min/max allowed value, but the calculate button is not re-enabled.

Issue #3690, opened 2025-11-07
Consider using pixi as the package manager


rozyczko:
**Is your feature request related to a problem? Please describe.**
Pixi is built on uv, so it is VERY fast.

It uses a pixi.toml (or pyproject.toml if integrated) to declare environments, dependencies, and tasks.
The workflows don’t have to hard-code dependency installation steps (uv sync, uv pip install -r requirements.txt, etc.).
It also allows for use of `tasks`, simplifying the workflow files logic.

We (ESS) use it extensively and have built complex workflows around it

PR #3842, opened 2026-01-20
Add basic Pixi config in `pyproject.toml` and developer instructions

backmari:
## Description

Adds a basic Pixi configuration to the `pyproject.toml` and instructions for setting up your developer environment using Pixi. The reason for adding Pixi as an option is that many institutions are moving from conda to Pixi due to conda licensing changes.

This is related to #3690, but does not resolve the issue. There is more work to do to e.g. move the CI to use Pixi, but that may require some discussion.

TODO (in a separate PR?): Figure out how to build the docs using a Pixi task and have the docs changes reflected in the GUI.

## How Has This Been Tested?

I tested that I can create the developer environment, run the GUI, and run the unit tests:

```
pixi shell
sasview
pixi run test
```

I have **only** tested on Linux.

## Review Checklist:

**Installers**


pixi.lock
```diff
@@ -0,0 +1,6920 @@
...
+  default:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    indexes:
+    - https://pypi.org/simple
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
+    packages:
+      linux-64:
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
```
llimeht: We're spent a lot of time consolidating all our lists of packages down to (approximately) 1 place in the requirements.txt and this is now duplicating all that information. That feels like the wrong direction to take.

Not only is it duplicating the information, it is pinning every single dependency at a particular version which is something we decided not to do. These pins actually also pin the python version - they list Python 3.14 wheels in lots of places, but that's not the agreed version for release at this stage.

How do these get updated?

INSTALL.md
```diff
@@ -87,7 +89,39 @@ debugging software, e.g.:
...
+```shell
+# Clone the repository
+git clone https://github.com/sasview/sasdata/
+git clone https://github.com/sasview/sasmodels/
+git clone https://github.com/sasview/sasview/
+
+# Enter the developer environment
+# (This will create (or reuse) a local Pixi environment with all required dependencies.)
+cd sasview
+pixi shell
```
llimeht: Developers will need to know where this is - the same details as for the venv approach will need to be offered. A simple list of commands wasn't considered enough detail hence the overly-long explanation of what they are doing in the venv section.

The same caveats about 'almost all packages' being precompiled is needed here too - including some packages that will actually need to be compiled (pycairo, for instance).

INSTALL.md
```diff
@@ -87,7 +89,39 @@ debugging software, e.g.:
...
+
+# Start the GUI
+sasview
```
llimeht: The two sets of instructions shouldn't have different entry points: `sasview` vs `python -m sas`.

INSTALL.md
```diff
@@ -87,7 +89,39 @@ debugging software, e.g.:
...
+
+# Run tests
+pixi run test
+
+# Exit the developer environment
+exit
+```
+
+Note that `sasdata`, `sasmodels` and `sasview` are installed in editable mode,
+meaning that any changes are immediately available during development.
```
llimeht: "any _code_ changes" - a short caveat about changes to docs and resources would be a appropriate, with a link to the longer explanation on that.

llimeht (changes requested):
There are a few conceptual things that need to be figured out collectively - particularly around pinning of dependencies. cf. https://github.com/orgs/SasView/discussions/3831

.gitignore
```diff
@@ -81,3 +81,6 @@ tests.log
 /installers/dist
 *.exe
 installers/credits.html
+# pixi environments
+.pixi/*
+!.pixi/config.toml
```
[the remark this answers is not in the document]
backmari: that line came along when I copied this from somewhere else, but will remove to avoid confusion!

INSTALL.md
```diff
@@ -87,7 +89,59 @@ debugging software, e.g.:
...
+
+```shell
+# Clone the repository
+git clone https://github.com/sasview/sasdata/
+git clone https://github.com/sasview/sasmodels/
+git clone https://github.com/sasview/sasview/
+
+# Enter the developer environment
+# (This will create (or reuse) a local Pixi environment with all required dependencies.)
+cd sasview
```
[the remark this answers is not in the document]
backmari: I made a change to extract this part into a section "Prerequisites". Please see if you think it makes sense.

krzywon (approved):
Code and documentation look good, but I'll leave testing the setup procedure to someone more familiar with `pixi`.

butlerpd:
@llimeht your hold on this dates back to Jan I was wondering but everything seems to have been resolved? There is an approving review as well so wondering if you had further comments?

.gitattributes
```diff
@@ -7,3 +7,5 @@ sasview/test/* linguist-language=Other
 *.h text eol=lf
 *.cpp text eol=lf
 *.rst text eol=lf
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true -diff
```
llimeht, line 11: Should `pixi.lock` instead be in `.gitignore` since it is not being added?

INSTALL.md
```diff
@@ -3,7 +3,7 @@
 Whether you're installing SasView to use as a tool for your research or
 because you're wanting to work on the code, it is recommended that you
 work inside a Python virtual environment of some sort.
-A `venv` or a `conda` are both popular choices.
+A `venv` or a `Pixi` are both popular choices.
```
llimeht, line 6: missing word "environment" or some other rewording? (venv is both a tool and an environment type, pixi is a tool that manages environments)

pyproject.toml
```diff
@@ -237,4 +237,19 @@ ignore = ["E501",  # line too long (leave to formatter)
 "sas" = ["sasdata", "sasmodels"]
 
 [tool.ruff.lint.isort]
-section-order = ["future", "standard-library", "third-party", "sas", "first-party", "local-folder"]
\ No newline at end of file
+section-order = ["future", "standard-library", "third-party", "sas", "first-party", "local-folder"]
+
+[tool.pixi.workspace]
+channels = ["conda-forge"]
```
llimeht, line 243: The edit to `INSTALLmd` says that it will grab the packages from pypi but this says conda-forge?

llimeht:
> but everything seems to have been resolved? 

There is three very fundamental questions that I've not seen any resolution to (but might have been discussed at a developer meeting?):

 - why make installing it look harder than it is by providing multiple different sets of instructions; if pixi is useful to document, perhaps it should be in a separate file. We have already received bug reports from folks who have looked at INSTALL.md, tried to simplify it, and then reported that it didn't work.
 - there are quite literally dozens of projects like pixi appearing at the moment; why pixi? are we going to document all of them as each person picks their favourite? (noting that we don't include how to build linux packages for any distros or flatpak within the installation instructions either)
 - what happens when (not if) installation via `pip` and `pixi` generate different outcomes (and bugs)? This is both a pragmatic question (the more different routes in we have, the thinner the testing is, and more problems we will be surprised by later.) and a policy question (if we have to choose, how do we then do so? there are other useful things that I could have done in CI that I didn't to maintain compatibility with `pip`, since it has a special place as the benchmark tool).

I also see that there's no test that the package is installable via pixi. Should there be? CI currently tests building the wheel and installing it via pip.

backmari:
@llimeht Apologies for the late reply. The answer to "why Pixi" is that multiple facilities have moved to Pixi after the Anaconda licensing changes. However, I completely agree that the most important thing is having one well tested install method that is also exercised in the CI. Since I don't see a strong reason right now to rewrite the CI to use Pixi, I am okay with closing this. I will bring it up in the next technical meeting.

PR #3846, opened 2026-01-24
modify the truthiness check

### LINKED ISSUE #3053: Phi and theta stuck at zero in 2D theory calculations

rozyczko:
## Description

In the `param_remap_to_sasmodels_convert` method, the condition `if not value:` incorrectly treated the value 0.0 as false. When orientation parameters like theta or phi were set to 0, they were being replaced with `np.nan`, breaking subsequent fits.

Fixes #3053 

## How Has This Been Tested?

Local tests

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingController.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingController.py
@@ -84,6 +84,11 @@ class FittingController:
         qmin = self.widget.q_range_min
         qmax = self.widget.q_range_max
 
+        # Update the model with extra parameters (polydispersity, magnetism)
+        # This ensures that values from the PolydispersityWidget and MagnetismWidget
+        # are applied to the kernel_module before fitting
+        self.widget.updateKernelModelWithExtraParams(model)
+
         # Gather all parameters to fit
         params_to_fit = copy.deepcopy(self.widget.main_params_to_fit)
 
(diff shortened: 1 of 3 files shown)


krzywon:
@DrPaulSharp, when you are able, could you compare the behavior outlined in [#3053 comment](https://github.com/SasView/sasview/issues/3053#issuecomment-3814069532) in both this branch and in main? I want to be sure others are able to see the difference in behavior before merging.

DrPaulSharp:
Hi @krzywon I've tried this, and on this branch I do find that the values change on a fit, but the errors are still extremely large:
image

dehoni:
The fit converges also for magnetic angles, but the error is out of bounds like @DrPaulSharp has seen.
image

PR #3889, opened 2026-03-02
Fixes bug in scaling for updating 1D plots

### LINKED ISSUE #3886: Folding option for slicers doesn't update properly
**Describe the bug**
Following the merge of #3878, we now have option to fold or unfold the 1D plot obtained from averaging a 2D data using the different slicers such as the sector slicer. However, whenever the slicer parameters are updated then the negative side of the unfolded data disappears. 

**To Reproduce**
1. Load 2D data and plot it.
2. Right click on the 2D plot and select "Edits slicer parameters' under the "slicers" option.
3. Under the "Create new slicer" tab, select "Sector Interactor". This will produce a 1D folded data.
4. Within the "Slicer Parameters" box, make sure to unclick the fold option value. 
5. Right-click on the 1D plot and select "change scale" then choose a linear scale for x (which is needed to allow visualization of the negative Q value on the 1D plot).
6. Right click on the 1D plot and set the "Scale range" from negative to positive value. This example uses Xmin=-0.12 and Xmax= 0.12 A^-1. See Screenshot below.

Image

7. Within the "Slicer Parameters" tab or directly within the 2D plot, you can change the Delta_Phi or Phi value of the sector slicer, and you will see that the negative side of the 1D plot disappears. See screenshot below where I changed the phi value from 60 to 66.6.

Image

 
9. **Expected behavior**
The full range of the unfolded 1D data should remain plotted whenever we update the slicer parameters. 

Windows 11

DrPaulSharp:
## Description

When updating 1D plots, the transformation properties of the data are now also updated. The issue described in #3886 occurred because data presented on a log scale was correctly transformed to a linear scale, but when the data was updated it was transformed to log and then linear again, causing negative q values to be omitted.

Fixes #3886

## How Has This Been Tested?

Followed the procedure described in #3886 and observed the problem is solved.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/Plotter.py
+++ b/src/sas/qtgui/Plotting/Plotter.py
@@ -746,6 +746,8 @@ class PlotterWidget(PlotterBase):
         new_plot.custom_color = selected_plot.custom_color
         new_plot.markersize = selected_plot.markersize
         new_plot.symbol = selected_plot.symbol
+        new_plot.xtransform = selected_plot.xtransform
+        new_plot.ytransform = selected_plot.ytransform
         self.removePlot(id)
         self.plot(data=new_plot)
         # Apply user-defined plot range


jellybean2004 (approved):
Looked at the code and tested functionality against main. All looks good.

PR #3910, opened 2026-03-26
Safe flag

### LINKED ISSUE #1576: refactor update_model handling in slicer models
The class `sas.qtqui.Plotting.SlicerModel.SlicerModel` uses an unsafe style (set flag, do something, reset flag) which risks leaving the flag set when "do something" raises an exception.  This is not yet known to cause bugs, but should still be refactored to a safer style.

Here are some things you could do instead:
* Avoid having the `update_model` flag in the first place.  The less state you maintain inside your classes the happier you will be.  In this case, have setParams() work without "updating the model", and in the case where you do need the model update (maybe in a callback from a drag event?) call the model updater after setParams.
* Pass the flag as a parameter to the method.  If you can't cleanly separate setParams from model update, then pass the update_model flag to setParams() so it can check it as needed.
* Protect the flag with a resource context.  This would set the flag before entering the setParams code and reset it when it leaves, even if it leaves with and exception.  Something like:
```python
from contextlib import contextmanager
class SlicerModel:
    @contextmanager
    def suppress_update(self):
        self.update_model = False
        yield
        self.update_model = True
    def setParamsFromModel(self):
        ...
        with self.suppress_update():
            self.setParams(params)
```
* Protect the flag with a try-finally block.  If you only need to protect a resource in one or two places then you can skip the context manager and explicitly reset the flag after the call even if the code exits with an exception:
```python
def setParamsFromModel(self):
    ...
    try:
        self.update_model = False
        self.setParams(params)
    finally:
        self.update_model = True
```

The first option is best, but it is the most work since you have to fix all the slicers.

_Originally posted by @pkienzle in https://github.com/SasView/sasview/pull/1567_

jellybean2004:
## Description

Fixes #1576 by implementing a shared context manager that restores the previous flag value safely, wires it into current call sites, and adds tests to prove restoration on exceptions.

## How Has This Been Tested?

Ran the new tests and tested manually.

## Review Checklist:

**Documentation**

**Installers**

### DIFF
--- a/src/sas/qtgui/Plotting/UnitTesting/SlicerModelTest.py
+++ b/src/sas/qtgui/Plotting/UnitTesting/SlicerModelTest.py
@@ -28,7 +28,7 @@ class SlicerModelTest:
         '''Assure that SlicerModel contains pure virtuals'''
         model = SlicerModel()
         with pytest.raises(NotImplementedError):
-            model.setParams()
+            model.setParams({})
         with pytest.raises(NotImplementedError):
             model.setModelFromParams()
 
--- a/src/sas/qtgui/Plotting/UnitTesting/SlicerModelTest.py
+++ b/src/sas/qtgui/Plotting/UnitTesting/SlicerModelTest.py
@@ -64,3 +64,27 @@ class SlicerModelTest:
         # Check the new model. The update should be automatic
         assert model.model().rowCount() == 3
         assert model.model().columnCount() == 2
+
+
+    def testSetModelFromParams_blocks_itemChanged(self, model):
+        """Programmatic model writes should not emit itemChanged."""
+        emissions = []
+        model.model().itemChanged.connect(lambda *args: emissions.append(args))
+
+        model.setModelFromParams()
+
+        assert emissions == []
+
+    def testSetModelFromParams_restores_signal_state_on_exception(self, model, monkeypatch):
+        """Signal blocking should be restored even if model population fails."""
+        def boom(value):
+            if value == 2:
+                raise RuntimeError("boom")
+            return str(value)
+
+        monkeypatch.setattr("sas.qtgui.Plotting.SlicerModel.GuiUtils.formatNumber", boom)
+
+        with pytest.raises(RuntimeError):
+            model.setModelFromParams()
+
+        assert model.model().signalsBlocked() is False
--- a/src/sas/qtgui/Plotting/SlicerModel.py
+++ b/src/sas/qtgui/Plotting/SlicerModel.py
@@ -16,19 +16,23 @@ class SlicerModel:
         Set up the Qt model for data handling between controls
         """
         parameters = self.getParams()
-        self._model.removeRows(0, self._model.rowCount())
-        # Crete/overwrite model items
-        for parameter in list(parameters.keys()):
-            item1 = QtGui.QStandardItem(parameter)
-            if isinstance(parameters[parameter], bool):
-                item2 = QtGui.QStandardItem(parameters[parameter])
-                item2.setCheckable(True)
-                item2.setCheckState(QtCore.Qt.Checked if parameters[parameter] else QtCore.Qt.Unchecked)
-            else:
-                item2 = QtGui.QStandardItem(GuiUtils.formatNumber(parameters[parameter]))
-            self._model.appendRow([item1, item2])
-        self._model.setHeaderData(0, QtCore.Qt.Horizontal, "Parameter")
-        self._model.setHeaderData(1, QtCore.Qt.Horizontal, "Value")
+        blocked = self._model.blockSignals(True)
+        try:
+            self._model.removeRows(0, self._model.rowCount())
+            # Crete/overwrite model items
+            for parameter in list(parameters.keys()):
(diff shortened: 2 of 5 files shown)


src/sas/qtgui/Plotting/UnitTesting/SlicerModelTest.py
```diff
@@ -64,3 +64,48 @@ def testSetParamsFromModel(self, model):
...
+            def getParams(self):
+                return self.params
+
+            def setParams(self, par):
+                raise RuntimeError("boom")
+
+        model = FailingModel()
+        model.setModelFromParams()
+
+        with pytest.raises(RuntimeError):
```
pkienzle: Need to make sure `update_model` is False here for the assert below check that it is True after the `setParamsFromModel` call even if the that call raises an exception.

pkienzle (commented):
I think there is a simpler way to do things. The problem we are trying to avoid is that setting the value in the model parameter box from the interactor triggers the same callback as typing the value in the box.

Instead of using the update_model flag we may be able to block the signal in `SlicerModel.setModelFromParams` using `blocked = self._model.blockSignals(True)` in a try-block with `self._model.blockSignals(blocked)` in the finally-block.

`blockSignals` is already being used in `MultiSlicerBase._synchronized_movend`, though without the `try...finally` structure. This can probably be removed if we block signals in `SlicerModel.setModelFromParams`.

jellybean2004:
Hi @pkienzle! Thank you for your review, and sorry for taking time to get back on this.

I have implemented the changes you suggested, which include using `try-finally` blocks with `blockSignals`. Please let me know what you think.

pkienzle:
Slicer is not updating from dialog.

Try showing a box slicer and typing values for parameters into the dialog. I find that the box does not update until the cursor hovers on one of the slicer handles. Worse, it's not where the handle is displayed, but rather where it will appear after it is moved. To see this, move the box out in qx then set the qx location of the box center. Move the cursor horizontally until the box suddenly springs to the new position. This occurs at the new qx position.

I'm testing on a mac with pyside 6.9.1

jellybean2004:
> Slicer is not updating from dialog.
> 
> Try showing a box slicer and typing values for parameters into the dialog. I find that the box does not update until the cursor hovers on one of the slicer handles. Worse, it's not where the handle is displayed, but rather where it will appear after it is moved. To see this, move the box out in qx then set the qx location of the box center. Move the cursor horizontally until the box suddenly springs to the new position. This occurs at the new qx position.
> 
> I'm testing on a mac with pyside 6.9.1

I could replicate the problem on Windows 11 -- it seems to be only in box sum, with all other slicers working fine. But I found the same issue on main, so it is not introduced by changes here. 


I will see if I can find a fix for this regardless. For now, I will raise an issue for this.

pkienzle (approved):

pkienzle:
SasView is gaslighting me. I was about to report that tabbing to the next field after changing a value in the slicer dialog does not update the slicer. I went to confirm that clicking in another field also does not update the slicer (a check I'm sure I've done before, but I wanted to confirm it before creating a new ticket), and suddenly both tabbing and clicking are working. 

I'm wondering if the signal is somehow getting stuck blocked for some time? I haven't been able to reproduce this problem.

While trying to clear the slicers and start afresh I got this error:
```python
11:22:15 - ERROR: Traceback (most recent call last):
File ".../sasview/src/sas/qtgui/Plotting/Plotter2D.py", line 292, in onClearSlicer
    self.slicer.clear()
    ~~~~~~~~~~~~~~~~~^^
File ".../sasview/src/sas/qtgui/Plotting/Slicers/BoxSum.py", line 182, in clear
    self.widget.closeWidgetSignal.emit()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
RuntimeError: Signal source has been deleted
```

Trying again from restart, updates work and the slicer can be deleted.

And again, closing the dialog before clearing slicers raises the above error. It seems to be reproducible:
* restart sasview app
* open a 2D dataset
* plot [create new] button
* Right-Click on graph > Slicers > Box Sum
* Close slicer dialog
* Right-Click on graphs > Slicers > Clear slicers

pkienzle (changes requested):
Trying to unapprove

jellybean2004:
Hi @pkienzle! The issue you reported should now be resolved from #3970.

If a slicer plot, or the dialog box in the case of box sum, is closed before clearing the slicer, the slicer gets auto-deleted.

pkienzle:
@jellybean2004 What's with the force-push?

It leads to surprises when pulling your PRs to check the latest version.

Are you rebasing on main?

pkienzle (approved):
Spurious errors from test for number of named colormaps unrelated to this PR (fixed in #4000).

Code runs, with updates tied between the parameter dialog and the interactor.

pkienzle:
Ubuntu latest tests were stuck on install for over 2 hrs. Hopefully this was a hiccup in the CI infrastructure and not a recurring problem.

jellybean2004:
@pkienzle Yes, I've been rebasing on main and `git push force-with-lease`

PR #3913, opened 2026-04-01
Sasbdb load

wpotrzebowski:
## Description

This branch adds Load from SASBDB: users can enter a SASBDB dataset ID, fetch metadata and intensity data from the SASBDB REST API, download a local copy, load it through the existing File data pipeline, and enrich the loaded Data1D objects with SASBDB-derived metadata (sample, instrument, publication, Rg, I(0), Dmax, MW, etc.) where the API provides them.

**Implementation highlights**

- sasbdb_api: HTTP client for https://www.sasbdb.org/rest-api using requests — metadata fetch, URL discovery for intensity files, streamed download to a temp path, robust parsing into SASBDBDatasetInfo (dataclass). Validates 7-character entry codes (e.g. SASDN24) before calling the API; does not guess download URLs when metadata lacks a file link.
- SASBDBDownloadDialog: Simple UI (dataset ID, Download, progress, status, Help/Cancel); validates ID up front; on success returns filepath + parsed info for loading.
sasbdb_loader: Loads the downloaded file via filesWidget.readData, then populate_metadata() merges API fields into data.sample, data.run, data.instrument, data.title, and meta_data['SASBDB_*'] keys. Surfaces a warning if readData returns no datasets.
- sasbdb_display: Formats metadata for dialog status, logging, and Data Explorer summary panels. Dict-line cleanup applies only to SASBDB datasets so non-SASBDB summaries are unchanged.
- GuiManager.actionLoad_SASBDB: Wired from File → Load from SASBDB… (MainWindowUI.ui); thin handler that opens the dialog and delegates to load_downloaded_dataset.
- GuiUtils: retrieveData1d / retrieveData2d call append_sasbdb_data_summary() when SASBDB_code is present — SASBDB display logic lives in the SASBDB package, not in GuiUtils.
- Docs: User menu/tools Sphinx sources and in-app help sasbdb_download_help.rst.

## How Has This Been Tested?

- **Manual (recommended):**
  1. **File → Load from SASBDB**; enter a known valid SASBDB entry ID.
  2. Confirm download completes, data appears in **Data Explorer**, and plots load as for a normal ASCII/dat file.
  3. Inspect dataset details / summary text for **SASBDB** fields (code, Rg, publication, etc.) when returned by the API.
  4. Try an **invalid** ID and confirm a clear error (404 / network).
  5. **Offline**: confirm failure is handled without crashing (message or log).

---

## Review Checklist

- [ ] **Security / network**: Timeouts, HTTPS-only usage, and user-visible errors for failed downloads are acceptable.
- [ ] **Metadata mapping**: SASBDB → `Sample` / `Source` / `meta_data` keys do not break loaders or perspectives.
- [ ] **`SASBDBDownloadDialogUI.ui`** is accompanied by regenerated Python UI (`convertUI.py` / project build) if required.
- [ ] **Documentation** (`menu_bar.rst`, `tools.rst`, `sasbdb_download_help.rst`) matches the menu label and workflow.
- [ ] **`requests`** remains available in the installed dependency set for end users (not only dev tooling).

**Installers**


src/sas/qtgui/Utilities/SASBDB/sasbdb_api.py
```diff
@@ -0,0 +1,184 @@
...
+        if ext.lower() in _KNOWN_EXTENSIONS:
+            return ext
+
+    if isinstance(metadata, dict):
+        for field_name in ("file_type", "fileType", "format", "data_format"):
+            file_type = metadata.get(field_name)
+            if file_type is None:
+                continue
+            file_type = str(file_type).lower()
+            if "csv" in file_type:
```
jamescrake-merani: Instead of having these repetitive if statements, I think it would be better if this information was in some data structure. For example, you could have:

```python
file_types = [(["csv"], ".csv"), (["txt", "text"], ".txt"), (["dat", "data"], ".dat")]
```

And then have a for loop like:

```python
for types, file_extension in file_types:
    for type in types:
        if type in file_type:
            return file_extension

```

This way, if you had to add another file type, you would just add it to the list.

src/sas/qtgui/Utilities/SASBDB/SASBDBDownloadDialog.py
```diff
@@ -0,0 +1,90 @@
...
+        QtWidgets.QMessageBox.warning(self, "Download Error", message)
+
+    def getDownloadedFilepath(self) -> str | None:
+        return self.downloaded_filepath
+
+    def getDatasetInfo(self) -> SASBDBDatasetInfo | None:
+        return self.dataset_info
+
+    def onHelp(self):
+        from sas.qtgui.Utilities import GuiUtils
```
jamescrake-merani: Is there a reason you are doing an import here, and not at the top of the file?

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -765,6 +766,24 @@ def actionLoad_Data_Folder(self):
         """
         self.filesWidget.loadFolder()
 
+    def actionLoad_SASBDB(self):
+        """
+        Menu File/Load from SASBDB
+
+        Opens a dialog to download and load a dataset from SASBDB.
+        """
+        from sas.qtgui.Utilities.SASBDB.sasbdb_loader import load_downloaded_dataset
```
jamescrake-merani: Is it necessary for these to not be at the top of the file?

src/sas/qtgui/Utilities/SASBDB/sasbdb_display.py
```diff
@@ -0,0 +1,332 @@
...
+
+def is_sasbdb_data(data) -> bool:
+    meta = getattr(data, "meta_data", None) or {}
+    return "SASBDB_code" in meta
+
+
+def metadata_summary(info: SASBDBDatasetInfo) -> str:
+    """Format dataset metadata for dialog status and logging."""
+    lines = []
+    if info.title:
```
jamescrake-merani, line 19: I'm not very comfortable with how long, and repetitive this function is but I also can't suggest any easy way of fixing it. 

What I would say though is that it might be better if this is a method within the `SASDBDatasetInfo` class. You could then use a decorator to give each field a full string name to be printed, and then have a print function within the same class to loop over every field in the dataclass, check if its not none, and print it if so.

src/sas/qtgui/Utilities/SASBDB/sasbdb_display.py
```diff
@@ -0,0 +1,332 @@
...
+        lines.append(f"Concentration: {info.concentration} {info.concentration_unit}")
+    if info.buffer_description:
+        lines.append(f"Buffer: {info.buffer_description}")
+    if info.instrument:
+        lines.append(f"Instrument: {info.instrument}")
+    if info.wavelength:
+        lines.append(f"Wavelength: {info.wavelength} {info.wavelength_unit}")
+    if info.temperature:
+        lines.append(f"Temperature: {info.temperature} {info.temperature_unit}")
+    if info.rg is not None:
```
jamescrake-merani, line 42: Isn't `info.rg is not None` equivalent to `info.rg`? If so, you could probably use the former on this line (and in the lines below it) to make this more consistent.

src/sas/qtgui/Utilities/SASBDB/sasbdb_display.py
```diff
@@ -0,0 +1,332 @@
...
+        if rg_error is not None:
+            text += f" ± {rg_error:.2f}"
+        return text + " Å"
+    rg_text = f"Rg: {rg:.2f}"
+    if rg_error:
+        rg_text += f" ± {rg_error:.2f}"
+    return f"{rg_text} Å"
+
+
+def _format_i0_line(i0, i0_error=None, style="label") -> str:
```
jamescrake-merani: This appears to have the same pattern as the `_format_rg_line` function. Would it be better to abstract this logic into one function?

src/sas/qtgui/Utilities/SASBDB/sasbdb_display.py
```diff
@@ -0,0 +1,332 @@
...
+                sample_parts.append(detail.replace("Concentration: ", ""))
+            elif detail.startswith("Buffer:"):
+                sample_parts.append(detail.replace("Buffer: ", ""))
+
+    if sample_parts:
+        return f"Sample: {' | '.join(sample_parts)}"
+    return None
+
+
+def _structural_lines_from_meta(meta: dict) -> list[str]:
```
jamescrake-merani, line 241: Some of these functions have the same issue I mentioned in `metadata_summary`, and perhaps could be solved in a similar way.

src/sas/qtgui/Utilities/SASBDB/SASBDBDownloadDialog.py
```diff
@@ -0,0 +1,90 @@
...
+SASBDB Dataset Download Dialog.
+"""
+
+import logging
+import os
+import tempfile
+
+from PySide6 import QtWidgets
+
+from .sasbdb_api import SASBDBDatasetInfo, downloadDataset, validateDatasetId
```
jamescrake-merani: I think it would be best to use absolute imports rather than relative.

src/sas/qtgui/Utilities/SASBDB/sasbdb_api.py
```diff
@@ -0,0 +1,184 @@
...
+        return filepath, dataset_info
+    return None, dataset_info
+
+
+def _normalizeDatasetId(dataset_id: str) -> str | None:
+    if not dataset_id:
+        return None
+    normalized = dataset_id.strip().upper()
+    if len(normalized) != 7 or not _SASBDB_ID_PATTERN.match(normalized):
+        logger.warning("Invalid SASBDB dataset ID: %r", dataset_id)
```
jamescrake-merani: Should this be logged as a warning considering that the user is also given a message box with the error as well?

Because at the moment, when I type in an incorrect identifier, I get both the error message, and the warning in the console:

Image

jamescrake-merani (changes requested):
I've made a few comments on the code I'd like you to take a look at.

I've tested with a couple of IDs. The first one worked, but the second one ([SASD2B2](https://www.sasbdb.org/data/SASD2B2/)) had an error that the identifier was invalid, but I'm not sure it is? Unless I am doing something wrong.

Also, are you intending these changes to go into the next major release? There are some things in here that would need to change with the refactoring project, but if its intended to go into the next release then we probably shouldn't do that work now.

Let me know if anything in the review is unclear. I realise some of the changes I've talked about are not straight forward. If it helps, I'm happy to jump on a meeting to discuss, or try to do some of the changes myself.

wpotrzebowski:
@jamescrake-merani Thanks for the thorough review and for testing with multiple IDs.

Yes — this is intended for the next major release (or one after that). I’ve addressed the correctness/UX items and the smaller cleanups from your comments. Larger structural changes (dataclass/decorator display model, async download worker, moving the client out of qtgui) I’m deliberately deferring so we don’t fight the upcoming refactor.

On SASD2B2: that was a real bug. Validation was too strict (SAS + 2 letters + digits). It’s now SAS + 4 alphanumeric characters, which matches entries like SASD2B2.

jamescrake-merani (approved):
Ok thanks for your response. This all makes sense to me, and I'm happy to approve this PR in its current state.

Issue #3884, opened 2026-02-24
Automated DOI generation


krzywon:
As agreed on by the SasView community, all subpackages will have their own DOIs. This will require a few changes to SasView

Tasks required:

- [ ] Port the DOI generation (`build_tools/release_automation.py`) process into the CI to automate DOI generation
- [ ] Update the README to include the latest citation information
- [ ] Update the documentation (and help docs) to include citation information

Approved proposal: https://github.com/orgs/SasView/discussions/3876

PR #3918, opened 2026-04-08
Release automation script rework

### LINKED ISSUE #1840: Further develop release automation scripts
build_tools/release_automation.py covers already few places for streamlining release process but it can be further developed and should be reviewed.

### LINKED ISSUE #3489: Release Automation Script: A list of issues must be provided
This is despite the fact that we no longer use the script to generate the release notes, as the functionality integrated in GitHub has replaced this.

krzywon:
## Description

This revamps the release automation script used to create the Zenodo DOI.

Changes:

- Automates the DOI generation process for sasdata and sasmodels, in addition to sasview.
- Generalizes most of the functions that were separate for sasview vs. sasmodels. Metadata updates are different, and will likely remain different due to differences in repos.
- Removes all release note generation routines, in favor of the GitHub automated releases notes generated by the website.
- Documentation and type hints updated to match the latest functionality.
- Updates the version number is in many places in the repo. The only outstanding place a version needs changing is in `config.py` for the `LAST_WHATS_NEW_HIDDEN_VERSION` and `LAST_UPDATE_DISMISSED_VERSION`
- Updates the copyright year is in multiple spots.

Fixes #3489
Refs #3884 (second bullet)
Fixes #1840

## How Has This Been Tested?

Tested locally and the majority of files listed in https://github.com/SasView/sasview/wiki/DevNotes_DevGuide_ReleaseChecklistTemplate that need to be updated are updated.

## Review Checklist:


build_tools/release_automation.py
```diff
@@ -163,34 +225,53 @@ def generate_zenodo(sasview_data, zenodo_api_key):
...
-
-
-def update_sasmodels_init(version):
+        file.write(version_template.format(version, year))
+
+    # Update the citation information
+    with open(citation_filename, 'w') as file:
+        file.write(acknowledgement_template.format(doi, release_manager))
+
+    # Update the Pyinstall config with the version and the year
```
llimeht: This is innosetup config not pyinstaller config (just for the comment)

build_tools/release_automation.py
```diff
@@ -163,34 +225,53 @@ def generate_zenodo(sasview_data, zenodo_api_key):
...
+    for line in fileinput.input(iss_file, inplace=True):
+        if line.startswith('#define MyAppVersion'):
+            print(f'#define MyAppVersion "{version}"', end='\r')
```
llimeht: \r line endings does not look right here. The current file is unix file endings (\n) and a mixture of them will not be fun.

build_tools/release_automation.py
```diff
@@ -163,34 +225,53 @@ def generate_zenodo(sasview_data, zenodo_api_key):
...
-    version_filename = os.path.join(system_directory, "version.py")
-    zenodo_filename = os.path.join(system_directory, "zenodo.py")
+    build_tools_directory = SASVIEW_PATH / 'build_tools'
+    installers_directory = SASVIEW_PATH / 'installers'
+    flatpak_directory = build_tools_directory / 'application_metadata'
+    system_directory = SASVIEW_PATH / "src" / "sas" / "system"
+    version_filename = system_directory /"version.py"
+    citation_filename = system_directory / "citation.py"
+    iss_file = installers_directory / 'installer.iss'
+    flatpak_manifest = flatpak_directory / 'org.sasview.sasview.metainfo.xml'
```
llimeht: Not wild about these `flatpak_*` variable names - these is a quite generic freedesktop metadata files and used by other linux packaging. The metainfo file is also not the flatpak manifest of course, which is the yml/json elsewhere.

build_tools/release_automation.py
```diff
@@ -163,34 +225,53 @@ def generate_zenodo(sasview_data, zenodo_api_key):
...
+            print(f'#define MyAppVersion "{version}"', end='\r')
+        elif line.startswith('#define MyAppPublisher'):
+            print(f'#define MyAppPublisher "(c) 2009 - {year}, UTK, UMD, NIST, ORNL, ISIS, ESS, ILL, ANSTO, BAM, TU Delft, and DLS"', end='\r')
+        else:
+            print(line, end='')
+
+    # Update the flatpak manifest with the version and date
+    for line in fileinput.input(flatpak_manifest, inplace=True):
+        if line.strip().startswith('<release '):
+            print(f'    <release version="{version}" date="{date}"></release>', end='\r')
```
llimeht: \r again - this file is definitely unix file endings \n

build_tools/release_automation.py
```diff
@@ -292,57 +351,41 @@ def main(args=None):
...
+            new_sasdata_doi = generate_zenodo(sasdata_data, zenodo_api_key)
+
+    # Regardless of previous checks and changes, apply changes to all packages (especially for the copyright year)
+    update_sasview_metadata(sasview_version, new_sasview_doi, release_manager)
+    update_sasmodels_metadata(sasmodels_version, new_sasmodels_doi, release_manager)
+    update_sasdata_metadata(sasdata_version, new_sasdata_doi, release_manager)
 
     # Pull the license from a know location
-    license_line = legal.copyright
+    license_line = legal.copyright + "\r"
```
llimeht: \n

build_tools/release_automation.py
```diff
@@ -163,34 +225,53 @@ def generate_zenodo(sasview_data, zenodo_api_key):
...
+    iss_file = installers_directory / 'installer.iss'
+    flatpak_manifest = flatpak_directory / 'org.sasview.sasview.metainfo.xml'
 
-    year = datetime.datetime.now().year
+    now = datetime.datetime.now()
+    date = now.strftime("%Y-%m-%d")
+    year = now.year
 
+    # Update the version in version.py
     with open(version_filename, 'w') as file:
```
llimeht: I'm not sure that we want to overwrite the version string in `version.py` that has been calculated from the git tag, particularly not in a file where it is intended to commit it to git.

docs/sphinx-docs/source/conf.py
```diff
@@ -15,6 +15,9 @@
 import sys
 
 import sphinx
+from packaging.version import parse
+
+from sas.system.version import __version__ as sasview_version
```
llimeht: Importing the package while it is being built makes me very nervous - there is a circularity about this. It currently works because of some PYTHONPATH manipulation in the hatch-sphinx config (`pyproject.toml` `[[tool.hatch.build.targets.wheel.hooks.sphinx.tools]] build` table), but that could be fragile.

build_tools/release_automation.py
```diff
@@ -52,6 +48,55 @@
...
+}
+
+version_template = \
+"""try:
+    from ._version import __version__
+except ImportError:
+    __version__ = "{0}"
+
+__release_date__ = "{1}"
+__build__ = "GIT_COMMIT"
```
llimeht: Is it time to get rid of that `__build__` member (it's unreferenced, and we'd use the `git describe` info that is in the version string / version tuple instead.

(and perhaps `__version_tuple__` should be imported+exported as well, since other things seem to be wanting the version of sasview at a granular level in `src/sas/system/config/config_meta.py` and `src/sas/system/user.py`)

docs/sphinx-docs/source/conf.py
```diff
@@ -86,9 +89,10 @@
 # The version number must follow StrictVersion rules as outlined
 # in http://epydoc.sourceforge.net/stdlib/distutils.version.StrictVersion-class.html
 # The short X.Y version.
-version = '6.1'
+parsed_version = parse(sasview_version)
```
llimeht: In `sas.system.user` the code uses `Version(sasview_version)`. I wonder if we should use the exact same method each time so that no-one needs to go and re-read the docs to find that they are actually the same (as I just did)

build_tools/release_automation.py
```diff
@@ -254,34 +335,12 @@ def update_acknowledgement_widget():
...
-        if response.status_code != 200:
-            return []
-        title = response.json()['title']
-        issue_titles.append(f"#{issue}, {title}")
-    return issue_titles
-
-
 def parse_args():
     parser = argparse.ArgumentParser('Script to automate release process')
     parser.add_argument('-v', '--sasview_version', required=True)
```
llimeht, line 337: Are we at the stage where these 3 values are introspectable? (or is that for the next batch of improvements some future time). We're more or less getting to standardised ways of reading the version strings for each project.

llimeht (changes requested):
Great strides forward in fixing things that can confuse in the release process :)

build_tools/release_automation.py
llimeht, line 16: While touching this code, is it time to modernise this `argparse` usage - put all the help text into the `parser.add_argument()` calls and let argparse build and show the help text as needed (also avoiding the `hasattr(args, 'help')` at the end)?

build_tools/release_automation.py
krzywon: Not exactly. This script is run before the final tags are created, but still needs to generate the Zenodo records with the exact version number. Without them, the implicit version may not be correct.

build_tools/release_automation.py
krzywon: I'm not against this, but I'm not very familiar with how `argparse` auto-generates the help documentation, so I'm going to leave this as-is for now.

build_tools/release_automation.py
```diff
@@ -63,18 +110,25 @@
...
+            `python ./sasview/build_tools/release_automation.py [options]
         """
         logging.error(msg)
-CONTRIBUTORS_FILE = SASVIEW_PATH / 'build_tools' / 'contributors.tsv'
+SASVIEW_CONTRIBUTORS_FILE = SASVIEW_PATH / 'build_tools' / 'contributors.tsv'
+SASDATA_CONTRIBUTORS_FILE = SASDATA_PATH / 'contributors.tsv'
+SASMODELS_CONTRIBUTORS_FILE = SASMODELS_PATH / 'build_tools' / 'contributors.tsv'
 
 
 class SplitArgs(argparse.Action):
```
llimeht: `SplitArgs` is now unused - it was only for the comma-separated bugs lists, so could be deleted

llimeht (approved):

krzywon:
Part of todays discussion during the technical meeting included this PR and where the Zenodo generation script should live:

1. Keep this as-is, despite sasmodels and sasdata. Pros: Porting this to other repos would create code duplication and changes in one place would require changes in all others, which no one wants. Cons: SasModels and SasData releases would require something from SasView.
2. Have this live in sasdata which is imported by both sasmodels and sasdata, but does not import either. This means it would be available for any release of a dependent project. Pros: Single location. Cons: Most people work in sasview, so it might not be obvious this exists.
3. Move this to a separate repo/package that can be imported by other repos (sasutils...). All codelets and modules that are used by multiple repos, include example files, ruff rules, example data, project templates, etc. could live in this repo. Pros: single location for everything. Cons: Another repo to maintiain.

PR #3923, opened 2026-04-16
Use native browser for help - rebased against current main

rozyczko:
## Description

This is a preliminary version of the QtWebEngine-less SasView rebased against current main.
Original PR: https://github.com/SasView/sasview/pull/3847
All calls have been converted to consistent system browser calls.
References to QtWebEngine were removed.
Tests updated (although we don't run these tests, anyway)

Fixes #3511 

## How Has This Been Tested?

Locally win11

## Review Checklist:

**Installers**
- [X] There is a chance this will affect the **installers**, if so


krzywon (approved):
I'm approving this and will merge it before our biweekly call next week.

There is one minor thing that can either be fixed here, or that I'll create an issue for after merging. `https://www.sasview.org/docs/v6.1.2/` which is opened during unit tests, doesn't exist but does resolves as a 404 page. Older documentation is found in `https://www.sasview.org/docs/old_docs/X.Y.Z/`. In the future, if we move to a different web server system, the URL might not resolve, and would break the unit test.

PR #3926, opened 2026-04-19
check if the doc location is correct. Needs installer.

rozyczko:
## Description

Update to the doc webview for proper doc URL.


pkienzle:
If there are existing versions of sasview out in the wild looking for help at `docs/v{version}`, moving the URL to  `old_docs/{version}` will break them all.

Can the underlying problem (whatever it was) be fixed by redirecting `docs/index.html` to `docs/v{current_version}/index.html` on the doc server and leave the existing code pointing to `docs/v{version}/index.html` for all versions?

PR #3928, opened 2026-04-23
Normalise wedge slicer phi to (-180,180]

jellybean2004:
## Description

This change updates `WedgeSlicer.py`, so the wedge phi parameter field in the slicer parameters window follows the same display convention as the plotted slice. 

Generally, phi is restricted to (-180, 180]. When phi > 180 (negative value), but delta phi brings the wedge below 180 (positive value), the plot range shows values > 180 to show an unbroken plot. To this end, checks are implemented to match the field value to the range displayed on the plot.
For example, if phi = 190 (-170) but delta_phi = 15, the plot range will be 175-205, and the phi value displayed will be 190, not -170.

Fixes #3914

## How Has This Been Tested?

Manually tested on Windows 11.

## Review Checklist:

**Documentation**

**Installers**

**Licensing**

### DIFF
--- a/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
+++ b/src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
@@ -9,6 +9,16 @@ from sas.qtgui.Plotting.Slicers.SectorSlicer import LineInteractor
 from sas.qtgui.Plotting.Slicers.SlicerUtils import StackableMixin, generate_unique_plot_id
 
 
+def _normalize_phi_for_wedge_display(phi_deg, delta_phi_deg):
+    """Normalize phi to match wedge display convention across +/-180 boundary."""
+    phi_normalized = np.mod(phi_deg - 180.0, 360.0) - 180.0
+    # Handle the case where phi is negative, but the wedge extends across the -180 boundary.
+    # Display phi as a positive angle above 180 to match the plot range.
+    if phi_normalized < 0 and (phi_normalized - delta_phi_deg) < -180.0:
+        phi_normalized += 360.0
+    return phi_normalized
+
+
 class WedgeInteractor(BaseInteractor, SlicerModel, StackableMixin):
     """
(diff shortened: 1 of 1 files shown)


jellybean2004:
@pkienzle would you be willing to review this PR? 

Please let me know if you are happy with the normalisation method used here.

src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
```diff
@@ -9,6 +9,16 @@
 from sas.qtgui.Plotting.Slicers.SlicerUtils import StackableMixin, generate_unique_plot_id
 
 
+def _normalize_phi_for_wedge_display(phi_deg, delta_phi_deg):
+    """Normalize phi to match wedge display convention across +/-180 boundary."""
+    phi_normalized = np.mod(phi_deg - 180.0, 360.0) - 180.0
+    # Handle the case where phi is negative, but the wedge extends across the -180 boundary.
+    # Display phi as a positive angle above 180 to match the plot range.
+    if phi_normalized < 0 and (phi_normalized - delta_phi_deg) < -180.0:
+        phi_normalized += 360.0
```
pkienzle, line 18: Minor quibble: As I was dragging the center line around the circle I was surprised that φ was not negative on the lower half plane, but instead depended on the width of the phi arc. I would drop this clause and just use phi_normalized.

pkienzle (approved):

src/sas/qtgui/Plotting/Slicers/WedgeSlicer.py
jellybean2004: @pkienzle, I did this to keep the phi displayed in parameters consistent with the plot range. 
If I remove this clause, for cases where the wedge crosses into the positive quadrant, the plot will show a positive value, while the parameter will show a negative value in the range (-180, 180].
So, should I keep this to keep the two values consistent?

PR #3931, opened 2026-04-28
Adds dummy err_data to residual data for slicers to work

### LINKED ISSUE #3765: Attempting to slice a residual plot throws
**Describe the bug**
Trying to slice a residual plot leads to exception.

**To Reproduce**
1. Load a 2D dataset
2. Try slicing it with any slicer
3. See exception

```python
sasdata\data_util\manipulations.py", line 896
err_data = data2D.err_data[np.isfinite(data2D.data)] 
```
Location of the exception differs from slicer to slicer, so possibly the fix is required in more than one place

**Expected behavior**
Should not throw

 - Version: 6.1.1 dev

 - OS: Win10

 - Installed: dev

jellybean2004:
## Description

Trying to average a residual plot using slicers would throw errors. This is because slicers expect an `err_data` array, which is set to `None` for residuals. 

This PR fixes that by adding a dummy `err_data` of ones for residual plots.

Fixes #3765 

## How Has This Been Tested?

Manually tested by attempting to reproduce the issue.

## Review Checklist:

**Documentation**

**Installers**

**Licensing**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
@@ -595,7 +595,8 @@ def residualsData2D(reference_data, current_data, weight):
     residuals.qx_data = current_data.qx_data
     residuals.qy_data = current_data.qy_data
     residuals.q_data = current_data.q_data
-    residuals.err_data = None
+    # Slicers expect residuals to have err_data, so add dummy error data of ones
+    residuals.err_data = numpy.ones(len(residuals.data))
     residuals.xmin = min(residuals.qx_data)
     residuals.xmax = max(residuals.qx_data)
     residuals.ymin = min(residuals.qy_data)


src/sas/qtgui/Perspectives/Fitting/FittingUtilities.py
```diff
@@ -595,7 +595,8 @@ def residualsData2D(reference_data, current_data, weight):
     residuals.qx_data = current_data.qx_data
     residuals.qy_data = current_data.qy_data
     residuals.q_data = current_data.q_data
-    residuals.err_data = None
+    # Slicers expect residuals to have err_data, so add dummy error data of ones
+    residuals.err_data = numpy.ones(len(residuals.data))
```
krzywon, line 599: Wouldn't `numpy.zeros` make more sense?

jellybean2004: In `manipulations.py`, a zero value in `err-data` triggers a fallback, and `data` is used as a proxy. So I used ones.

pkienzle: χ² = Σ (f(x) - y)²/Δy² so Δy=1 is a natural value to use in the absence of any information about the data uncertainty. Parameter uncertainties can be normalized by √(χ²/DOF) to account for improperly scaled error bars. I'm not doing this by default because this assumes that there is no systematic error.

Note: in the broader stats world where measurement uncertainty is unknown, data is generally drawn without uncertainty. Whether the result of SasView (i.e., the plotting infrastructure and the bumps interface) can handle the missing error bars is a separate question. They probably should, since fixed Δy = 1 will make for ugly plots.

What does sasdata do with datasets that don't have any associated uncertainty? Does it return them as None? In which case, the correct fix is to leave err_data as None and fix the "isfinite" masking.

Note 2: If this is acting on a normalized residual, then the Δy is supposed to be one. It should be put there by whatever generated the residual, though, and not be introduced by the slicer.

pkienzle:
I'm not even able to get to a 2D residual plot. Choosing exp18_14_igor_2dqxqy.dat I can't send it to fitting without seeing the error:
```
 ERROR: Traceback (most recent call last):
File ".../sasview/src/sas/qtgui/MainWindow/DataExplorer.py", line 889, in sendData
self._perspective().setData(data_item=selected_items, is_batch=self.chkBatch.isChecked()) 
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'setData'

During handling of the above exception, another exception occurred:
File ".../sasview/src/sas/qtgui/MainWindow/DataExplorer.py", line 891, in sendData
msg = "%s perspective returned the following message: \n%s\n" % (self._perspective().name, str(ex)) 
                                                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'name'
```

It also fails for 1D datasets. I suspect the problem is outside the scope of this PR, but the result is that I can't test the fix.

krzywon:
> I'm not even able to get to a 2D residual plot. Choosing exp18_14_igor_2dqxqy.dat I can't send it to fitting without seeing the error:
> 
> ```
>  ERROR: Traceback (most recent call last):
> File ".../sasview/src/sas/qtgui/MainWindow/DataExplorer.py", line 889, in sendData
> self._perspective().setData(data_item=selected_items, is_batch=self.chkBatch.isChecked()) 
>       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NoneType' object has no attribute 'setData'
> 
> During handling of the above exception, another exception occurred:
> Traceback (most recent call last):
> File ".../sasview/src/sas/qtgui/MainWindow/DataExplorer.py", line 891, in sendData
> msg = "%s perspective returned the following message: \n%s\n" % (self._perspective().name, str(ex)) 
>                                                                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NoneType' object has no attribute 'name'
> ```
> 
> It also fails for 1D datasets. I suspect the problem is outside the scope of this PR, but the result is that I can't test the fix.

Odd. That means you have no active perspective. Can you plot the data?

What OS are running and are you testing with the built binary or directly from source?

jellybean2004:
> I'm not even able to get to a 2D residual plot. Choosing exp18_14_igor_2dqxqy.dat I can't send it to fitting without seeing the error:
> 
> ```
>  ERROR: Traceback (most recent call last):
> File ".../sasview/src/sas/qtgui/MainWindow/DataExplorer.py", line 889, in sendData
> self._perspective().setData(data_item=selected_items, is_batch=self.chkBatch.isChecked()) 
>       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NoneType' object has no attribute 'setData'
> 
> During handling of the above exception, another exception occurred:
> Traceback (most recent call last):
> File ".../sasview/src/sas/qtgui/MainWindow/DataExplorer.py", line 891, in sendData
> msg = "%s perspective returned the following message: \n%s\n" % (self._perspective().name, str(ex)) 
>                                                                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NoneType' object has no attribute 'name'
> ```
> 
> It also fails for 1D datasets. I suspect the problem is outside the scope of this PR, but the result is that I can't test the fix.

@pkienzle @krzywon, I tried this data file, and I faced no issues. The original file I tested on was `SLIC010.dat`, but I tried it out on a few other files, and didn't have any problems.

krzywon (approved):
Based on @pkienzle's comment related to errors, I am approving this. The issue noted seems to be a localized issue which is beyond the scope of this PR.

pkienzle:
I was messing with bumps and that broke the fitting perspective.

It is not incorrect to introduce err_data=1 in this code, assuming the reference data is a theory function with no uncertainty. The 1D residuals function does not define dy, so it is a little odd that we have it for 2D.

The question still stands whether it is better add the uncertainty here or to fix the slicers to handle data without uncertainty. If we keep the current fix, then please make sure that all sasdata 2D loaders define err_data = 1 for all datasets which do not otherwise define uncertainty.

krzywon:
> The question still stands whether it is better add the uncertainty here or to fix the slicers to handle data without uncertainty. If we keep the current fix, then please make sure that all sasdata 2D loaders define err_data = 1 for all datasets which do not otherwise define uncertainty.

The rewrite of the slicers would be a much larger change. I will ensure this change is made as a part of the refactor project, but this fix is a good enough for now type of change.

The data loaders set `err_data` to None when no values are provided, so they will be treated the same as the residuals are here.

@pkienzle, I think this is ready to be merged. Please do if my answers are sufficient.

PR #3932, opened 2026-04-28
Upload Binaries On Tag

krzywon:
## Description

This adds a new workflow that uploads binaries to releases when tags are created.

## How Has This Been Tested?

TBD - Needs CI and a tag to be generated

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,10 +41,6 @@ jobs:
           done
           mv SasView-Installer-ubuntu*/SasView6.tar.gz ./SasView-${{ github.event.release.tag_name }}-Ubuntu22.04.tar.gz
 
-      - name: Clean out old release
-        run: |
-          gh release delete nightly-build --cleanup-tag --yes || true
-
       - name: Upload Nightly Build Installer to GitHub releases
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+          mv SasView-Installer-windows*/setupSasView.exe ./SasView-${{ github.event.release.tag_name }}-Win64.exe
+          mv SasView-Installer-macos*/SasView6-*.dmg ./SasView-${{ github.event.release.tag_name }}-*.dmg
+          for file in $(ls *.dmg); do
+              mv $file ${file/SasView6-/SasView-nightly-MacOSX-}
+          done
+          mv SasView-Installer-ubuntu*/SasView6.tar.gz ./SasView-${{ github.event.release.tag_name }}-Ubuntu22.04.tar.gz
+
+      - name: Clean out old release
+        run: |
+          gh release delete nightly-build --cleanup-tag --yes || true
```
llimeht: Guessing that this whole step isn't needed since we wouldn't be re-using tags?

krzywon: Whoops, yes! I'll remove that. Copy/paste...

PR #3942, opened 2026-05-06
Remove .coverage file

krzywon:
#3897 added a pytest `.coverage` file to the repo. I'm not sure if this was intended, but this isn't something we should include.


jellybean2004 (approved):
I must have added it by accident. Thanks for pointing it out!

PR #3943, opened 2026-05-06
3927: Use correct URL for past versions documentation

### LINKED ISSUE #3927: Unit test calls an incorrect URL
**Describe the bug**
In #3923, a new unit test was introduced that opens the website `https://www.sasview.org/docs/v6.1.2/`. That page doesn't exist but does resolves as a 404 page, which allows the test to pass. Older documentation is found in `https://www.sasview.org/docs/old_docs/X.Y.Z/`. In the future, if we move to a different web server system, the test URL might not resolve, and would break the unit test.

**Expected behavior**
The correct URL should be used.

krzywon:
## Description

This updates the URLs used for past version of our documentation. This also updates the help tests to use an older version so future developers don't expect they need to update this version for every release.

__One question: Should `get_current_release_version()` be in `sas.system.web`, where I have it now, or would it make more sense to be in `sas.system.version`?__

Fixes #3927

## How Has This Been Tested?

Units tests pass

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
+++ b/src/sas/qtgui/Utilities/NewVersion/NewVersionAvailable.py
@@ -2,8 +2,7 @@ import logging
 import webbrowser
 from copy import copy
 
-import requests
-from packaging.version import Version, parse
+from packaging.version import parse
 from PySide6.QtCore import QSize
 from PySide6.QtGui import QIcon
 from PySide6.QtWidgets import (
@@ -19,8 +18,8 @@ from PySide6.QtWidgets import (
 )
 
 from sas import config
-from sas.system import web
 from sas.system.version import __version__ as current_version_string
+from sas.system.web import get_current_release_version
 
 logger = logging.getLogger("NewVersionAvailable")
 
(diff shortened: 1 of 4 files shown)


krzywon:
This is already being worked on by @rozyczko in #3926. Sorry!

DrPaulSharp:
I'll hold fire on the review then.

krzywon:
@rozyczko - could you compare this PR to your own in #3926 and let me know what I should do with this, when you have a moment, please?

pkienzle:
I noticed that the constructed URL to old docs contains // in the path: `https://www.sasview.org/docs/old_docs/6.1.2//index.html`

Anyway, wouldn't it be better to handle a redirect from `docs/index.html` to `docs/v{current_version}/index.html` on the server so that sasview installations out in the wild with the existing path don't break?

krzywon:
> Anyway, wouldn't it be better to handle a redirect from `docs/index.html` to `docs/v{current_version}/index.html` on the server so that sasview installations out in the wild with the existing path don't break?

The feature @rozyczko and I are trying to fix was introduced after v6.1.3. The URL in question has only seen the 2 most recent alpha releases, so now is the time to fix this.

pkienzle:
Okay.

FWIW, I like the initial choice of url as `docs/v{version}` better than `docs/old_docs/{version}`. That way you never have to query the web to see what is the most recent version.

krzywon:
After a discussion last week, some of this work has been ported into #3926 and I am now closing this PR.

PR #3947, opened 2026-05-07
Nightly Build: Cancel on Concurrent

krzywon:
## Description

This adds 'cancel on concurrent workflows' option to the nightly build workflow.

## How Has This Been Tested?

CI...

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -21,6 +21,11 @@ jobs:
         os: [ ubuntu-latest]
     needs: reuse_main_ci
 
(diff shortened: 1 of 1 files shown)


llimeht:
The current nightly flow runs on every merge into `main` and I wonder if that's actually the thing that should be fixed? (i.e. have `nightly-build.yml` run... umm... nightly, via a cron expression?)

krzywon:
During todays discussion, we should switch to a truly nightly build. Closing this and will submit an issue to build nightly, rather than on pushes to main.

PR #3951, opened 2026-05-14
Pins pyausaxs to earlier than 1.1.4

DrPaulSharp:
Recent CI runs fail due to changes made in pyausaxs version 1.1.4. This PR pins the version to earlier than that until the problems are resolved.


jellybean2004 (commented):
CI is now passing -- good to merge.

jellybean2004 (approved):

PR #182, opened 2018-09-17
Create Release Notes option in Help Menu and popup box

smk78:
What does __revision__ do (in releasebox.py)? It was in acknowledgebox.py that I used as a template (that @krzywon wrote)...

PR #3963, opened 2026-05-22
Single kernel call for volume and intensity

### LINKED ISSUE #3418: Fix problem of hard coding in Size Distribution
Currently, the SizeDistribution calculator has a separate function to calculate the volume. Basically `ellipse_volume`. However the volume and the intensity can both be returned from a single call.  I would propose doing this in `sizeDistribution().generate_model_matrix`. This would streamiline and generatize the code

code snippet to help do this would be:
```
import numpy 
from sasmodels.core import load_model_info, build_model
from sasmodels.direct_model import call_Fq

modelname = 'ellipsoid' # this should eventually be whatever is passed
q = np.logspace(-3, -0.2, 200)  # q should be data.q 
# pars should depend on the model so probably create the pars for the chosen model somewhere else eventually?
pars = {
     "radius_polar": 40.0,
     "radius_equatorial": 25.0,
}

# Fsq = Intenstiy. for hollow objects Volume may in fact return shell_volume but that is
# a complexity that will probably never be needed
model_info = load_model_info(modelname)
model = build_model(model_info)
calculator = model.make_kernel((q,))
F, intensity, R_eff, Volume, V_ratio = call_Fq(calculator, pars)

```

jellybean2004:
## Description

Removes hard-coded/duplicated volume math and ensures intensities/volumes come from the model kernel.

Replaces the previous per-bin DirectModel/analytic volume approach with a kernel-based single call.
Builds the sasmodels kernel (from model info) and calls `call_Fq` for each bin to retrieve Fsq and volume. Compute intensity per bin as `intensity = (scale/volume) * Fsq + background` and store per-bin volumes in `self._volumes`.

Fixes #3418 

## How Has This Been Tested?

Tested functionality -- same as before.

## Review Checklist:

**Documentation**

**Installers**

**Licensing**


src/sas/sascalc/size_distribution/SizeDistribution.py
```diff
@@ -483,24 +485,45 @@ def generate_model_matrix(self, moddata: Data1D) -> None:
...
         }
 
-        kernel = DirectModel(moddata, model)
-
         intensities = []
+        volumes = []
+
+        # Build a single Kernel to compute both intensity and volume per bin
+        model_info = load_model_info(self.model)
+        model_obj = build_model(model_info)
```
pkienzle: Line 486: `model = load_model(self.model)` is equivalent to line 499-500. Use `model_obj = load_model(self.model)` instead.

src/sas/sascalc/size_distribution/SizeDistribution.py
```diff
@@ -483,24 +485,45 @@ def generate_model_matrix(self, moddata: Data1D) -> None:
...
+            scale_val = p.get("scale", 1.0)
+            background_val = p.get("background", 0.0)
+            combined_scale = scale_val / volume
+
+            intensity = combined_scale * Fsq + background_val
+            intensities.append(intensity)
+
+            # Volume ratio is 1 for solid particles
+            if volume_ratio is not None:
+                volume *= volume_ratio
```
pkienzle: volume_ratio from call_Fq is never None.

src/sas/sascalc/size_distribution/SizeDistribution.py
```diff
@@ -483,24 +485,45 @@ def generate_model_matrix(self, moddata: Data1D) -> None:
...
-            pars["radius_polar"] = bin * self.aspectRatio
-            intensities.append(kernel(**pars))
+            p = pars.copy()
+            p["radius_equatorial"] = bin
+            p["radius_polar"] = bin * self.aspectRatio
+
+            call_pars = p.copy()
+            _, Fsq, _, volume, volume_ratio = call_Fq(calculator, call_pars)
+
+            # Compute intensity using kernel convention: combined_scale = scale / shell_volume
```
pkienzle, line 498: This is duplicating code from within `sasmodels.kernel.Kernel.Iq`.  This isn't avoidable in the current implementation, but we should consider this use case when addressing [sasmodels #182](https://github.com/SasView/sasmodels/issues/182).

The idea is to return intermediate and derived values from the calculation back to the caller. The current hack to get $P(Q)$ and $S(Q)$ from `P@S` is that sasmodels.product sets up a `kernel.results()` function that returns a dictionary with keys P(Q), S(Q), volume, volume_ratio and radius_effective after calling `kernel.Iq()`. We could extend this to the base `Kernel` class in sasmodels. Then using `call_kernel` instead of `call_Fq` we can get the volume and volume ratio from `calculator.results()`.

jellybean2004:
Hey @pkienzle! Thanks for your review! I have addressed your comments in my recent commit.

pkienzle:
_volumes is never used. Is this preparation for a future change?

jellybean2004:
> _volumes is never used. Is this preparation for a future change?

I somehow managed to miss one commit. The goal in the issue was to replace `ellsipse_volume` with `self._volumes`, and I have now done that.

pkienzle:
Don't scale volume by volume ratio. Intensity scales with the volume fraction of the material in the shell. Volume ratio is for determining the average volume enclosed by the shape when computing structure factors. Since shape is fixed as ellipsoid then volume ratio is one and you won't see a problem. Maxent on vesicles would be a problem, which you can verify by comparing to maxent on a core-shell model.

With background fixed at zero and scale fixed at one you don't need to include them in the intensity calculation. It doesn't hurt, but it doesn't help either.

The `model_matrix` and `_volumes` attributes have the same status, either both private or both public.

Outside this PR there are a number of things that could change.

Linear systems don't need levenberg-marquardt. For background_fit you can use np.polyfit with deg=0|1 and w=1/σ.

PEP8 recommends short snake_case rather than CamelCase for module names, so `size_distribution.py` and `maxent.py`.

PEP8 recommends CamelCase for class names and snake_case for functions and methods, such as `maxent.DecisionHelper.chi_now`.

Shouldn't be using a builtin function like `iter` as a for loop index: `for iter in range(IterMax):` => `for step in range(maxiter):`

In `class SizeDistribution`, you don't need to use properties for `scale`, `background`, etc. with the setter assigning the internal value unchecked and the getter returning the value unaltered. Simple attributes work well for these.

The weight type setter in size distribution is suspect. It is calling update weights with default None for the full data, whereas the prep maxent call is using a trimmed q range for the data. Easiest if changing weight type does nothing special since the prep maxent call provides the data for weights.

More generally, the size distribution controls aren't going to change during analysis, so they should be set as arguments to the constructor rather than assigned as attributes.

SizeDistributionThread is accessing "private" `sd._binDiff`. If it is needed by the caller then the attribute shouldn't be tagged with a leading underscore.

pkienzle:
Please fix the merge conflicts and remove the volume ratio scaling. Code style can be put in as a separate issue if you want to minimize changes for this PR.

pkienzle (approved):

Issue #3827, opened 2025-12-16
"Send to Fitting" button causes errors in other perspectives


DrPaulSharp:
**Describe the bug**

In #3763, a button labelled "send to fitting" has been added to all plots produced by the fitting, inversion, invariant and size distribution perspectives.

For the fitting perspective, clicking on the button on a plot produced by the perspective adds data entries to the data explorer, and sends all of those additional data objects to the fitting perspective, placing them in additional tabs within the perspective. For the inversion perspective, the behaviour is similar, although the data is sent to the inversion perspective instead (the text of the label should probably be changed to make this clearer).

However, the Corfunc, Invariant and size distribution perspectives do not have the functionality for multiple tabs. In this case, when pressing the "send to fitting" button, the best case scenario is that warnings are issued regarding clearing data from the perspective as multiple datasets are loaded (note that the code bypasses the checks introduced in #3587 to ensure multiple datasets are not sent to these perspectives), the worst case scenario is that errors are triggered, and in the case of the refactored invariant perspective in #3789, I have observed that SasView crashes.

It should also be noted that it is possible to create plots using one perspective, switch to another perspective, and then click the "send to fitting" button. In this case the data is sent to the currently active perspective, making it easy to send fitting data to inversion, for example.

It is clear that the behaviour of this feature across different perspectives has not been properly thought through or tested. At this point, I would recommend at the very least that the "send to fitting" button should be disabled whenever the perspective is changed to prevent it raising errors. However, my position remains as I expressed clearly at the contributor camp, which is that the functionality of the "send to fitting" button should be to add the data to the data explorer, with the user then choosing whether to send to the data to the perspective or not. Adding tabs to perspectives required writing complex code that was always going to be prone to errors. If the button's functionality is changed to just send the data to the data explorer, it is likely that the problems I have listed will not occur. However, if one of these solutions cannot be implemented then I think we should consider reverting the PR.

The example below shows an error when switching between the tabbed perspectives.

**To Reproduce**
1. Load the 1D data file "98289.txt"
2. Send the data to the fitting perspective
3. Fit the data using a sphere model and click "compute/plot"
4. Switch to the inversion perspective
5. On the residuals plot, click "send to fitting"
6. See error: 

**Expected behavior**

The send to fitting button should not load fitting data in the invariant perspective - indeed it should not load data into **any** perspective.

**Screenshots**

[...]

PR #3999, opened 2026-06-10
Add documentation for CustomToolbar

### LINKED ISSUE #3792: Documentation for the custom NavigationTool button
1D plots have now a new button to send all data directly to Fitting and stores those plots as new data in the DataExplorer. The documentation for that routine is still required.

ChristianWoegerbauer:
## Description

Documentation added to CustomToolbar

Fixes #3792

## How Has This Been Tested?

- Documentation-only change
- Manually reviewed for correctness and consistency with code

## Review Checklist:

- [x] There is an **issue** open for the documentation (https://github.com/SasView/sasview/issues/3792#issuecomment-3664651065)

**Installers**

### DIFF
--- a/src/sas/qtgui/MainWindow/media/graph_help.rst
+++ b/src/sas/qtgui/MainWindow/media/graph_help.rst
@@ -500,4 +500,4 @@ The plots for the individual slicers will be stacked.
 
 .. ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
 
-.. note:: This help document was last modified by Sujaya Shrestha, 24 December 2025.
+.. note:: This help document was last modified by Christian Wögerbauer, 12 June 2026.
--- a/src/sas/qtgui/MainWindow/media/graph_help.rst
+++ b/src/sas/qtgui/MainWindow/media/graph_help.rst
@@ -197,6 +197,13 @@ Getting data coordinates
 Clicking anywhere in the plot window will cause the current coordinates to be
 displayed in the status bar at the very bottom-left of the SasView window.
 
+Sending data to Explorer
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Clicking the button sends the current plot data to the Data Explorer.
+This feature is primarily useful for generated 1D plots during slicing, as 
+data must be selected in the Explorer before it can be sent to the Fitting tool.
+
 .. ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
 
 Dataset Menu options
--- a/src/sas/qtgui/Plotting/PlotterBase.py
+++ b/src/sas/qtgui/Plotting/PlotterBase.py
@@ -43,7 +43,6 @@ class CustomToolbar(NavigationToolbar):
     def sendToDataExplorer(self):
         for item in self.parent.data:
             GuiUtils.communicator.freezeDataNameSignal.emit(item.name)
-            break  # short fix but it is not the soluction for the problem
 
 class PlotterBase(QtWidgets.QWidget):
     #TODO: Describe what this class is
--- a/src/sas/qtgui/Plotting/PlotterBase.py
+++ b/src/sas/qtgui/Plotting/PlotterBase.py
@@ -19,17 +19,21 @@ DEFAULT_CMAP = mpl.cm.jet
 
 class CustomToolbar(NavigationToolbar):
     """
-    Custom toolbar adding a "Send to Data Explorer" action.
-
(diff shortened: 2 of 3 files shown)


ChristianWoegerbauer:
I'm not sure how detailed the documentation needs to be.

krzywon:
@ChristianWoegerbauer - we have two places for documentation. The doc strings that you are have added here and the user documentation. The doc strings are parsed out of the source during our build process and converted to restructured text. This text is then converted to html. The user documentation lives in the source as an rst file and is built into html, also during the build process. Both end up in the shipped documentation, but the user documentation is typically used more than the developer docs.

Best practices for doc strings:
- Surround everything in triple quotes (✅)
- Give a general description of what the module is supposed to do in general terms, e.g. `This method calculates and returns the scattering cross section of the formula passed to the method.`
- Add the parameters in the format `:param <param_name>: (<type>) A description to allow future developers to know what the parameter is used for` (✅)
- Add the return in the format `:return: (<type>) A description to allow future developers to know what is returned`

Suggestions for general documentation:
- User documentation for plotting features should go into `src\sas\qtgui\MainWindow\media\graph_help.rst`
- Add a section to this file that describes what the feature is, why it was added, and, most importantly, what the user can expect when it is used.
- Reading through what is already in the file should give you a sense of what level we are looking for when it comes to documenting features.

src/sas/qtgui/MainWindow/media/graph_help.rst
jellybean2004, line 1: Could you also edit the note at the bottom of this file saying that you modified it?

src/sas/qtgui/Plotting/PlotterBase.py
```diff
@@ -37,6 +45,7 @@ def add_custom_button(self):
     def sendToDataExplorer(self):
         for item in self.parent.data:
             GuiUtils.communicator.freezeDataNameSignal.emit(item.name)
+            break  # short fix but it is not the soluction for the problem
```
jellybean2004: Is this a fix for a bug you encountered?

src/sas/qtgui/Plotting/PlotterBase.py
```diff
@@ -18,15 +18,23 @@
...
         self.parent = parent
-        self.add_custom_button()
+        self.addAction_SendToExplorer()
 
-    def add_custom_button(self):
-        # I have been told that a Button is better
-        # But all NavigationToolbar interactions are Actions
-        # This way all can be called with:
+    def addAction_SendToExplorer(self):
+        # All actions can be called with:
```
jellybean2004: Should this be a doc string?

jellybean2004 (changes requested):
Thank you for taking the time to write this. The docs look good. I have made a few suggestions to finalise it so I can put it in the release.

You added a `break` in `sendToDataExplorer`. Is this fixing a bug you encountered?

src/sas/qtgui/Plotting/PlotterBase.py
ChristianWoegerbauer: In https://github.com/SasView/sasview/issues/3827 a bug was found. This line should temporary fix it. 

When I originally wrote the code, it worked correctly, but there was a bug in another part of the codebase that affected its behavior. The attempted bug fix merely changed the behavior to work around the issue rather than actually solving it. As a result, the root cause remained and was still visible through the multiple emissions.

The `break` statement ensures that the button behaves as intended, but it limits processing to only the last dataset in `self.parent.data`. The original intention of the `for` loop was to send all datasets in the list, which is required when multiple slices are displayed in the 1D plot.

The real bug is that the list contains far more data models than intended. Every time, for example, the sector slicer is updated, the 1D data model is added again to `self.parent.data`. This causes the list to grow indefinitely if the slicer is used repeatedly.

Once that underlying bug is fixed, the `break` statement can simply be removed, and the original functionality will work correctly again.

src/sas/qtgui/Plotting/PlotterBase.py
ChristianWoegerbauer: no, i think i that can be removed

src/sas/qtgui/Plotting/PlotterBase.py
jellybean2004: I see. Since this PR is for documentation, I wanted to add this to the beta release. The patch makes it slightly more concerning to be put in the beta release. 

Could I please ask you to make a separate PR for the patch, and put just the doc changes in this one? With that it will be ready to merge.

jellybean2004 (changes requested):
Please see comment: https://github.com/SasView/sasview/pull/3999#discussion_r3414076380

src/sas/qtgui/Plotting/PlotterBase.py
DrPaulSharp: This statement should not be included, instead see #4003 for how we will go about a fix.

jellybean2004:
I want to put this in the beta, which I have been holding off waiting for this, so I am going to remove the `break` addition to keep it just a doc addition PR.

jellybean2004 (approved):

PR #4009, opened 2026-06-25
Adds a trust bar for size distribution result plots

### LINKED ISSUE #3271: Add an IRENA style trust level bar to the Size Distribution results plot
Provide Trust Bar a la IRENA. This simply indicates that diameters much beyond 2Pi on Qmin or Qmax are out of the Q range used and should be ignored. Essentially something like a z amplitude color bar for a 2D plot but spanning Dmin to Dmax on the x axis with 2 break points, each with some width. So 3 colors: green=good; yellow = careful ; red = bad.  Further we don't want sharp transitions in color so a smoothing function from one to the other will be needed. Again, see IRENA plots.

Corfunc built such a function which could probably be used. Ideally this should be a feature added to the plotters? It seems it might be used in more places than one

jellybean2004:
## Description

Fixes #3271

This PR adds a Trust Bar to the Size Distribution perspective, inspired by IRENA-style plots. The bar visually indicates which diameter region is reliable based on the fitted Q-range, using a smooth red/yellow/green gradient.

### Key changes

- Adds a horizontal trust bar above the size distribution plot.

- Uses two physical trust boundaries derived from the fit range as defined in `SizeDistributionLogic`:
  - D_low = 1.8 * pi / Qmax
  - D_high = 0.95 * pi / Qmin

- Displays:
  - red for out-of-range / unreliable diameters
  - yellow for transition regions
  - green for the trusted region

- A small helper class (TrustBar) handles:
  - drawing the gradient bar
  - clearing it
  - updating it when the x-axis limits change

- The bar stays aligned when the plot is zoomed or rescaled.

## How Has This Been Tested?

Manually tested functionality.

## Review Checklist:

**Documentation**
- [x] There is an **issue** open for the documentation #4016 

**Installers**

**Licensing**

### DIFF
--- a/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
+++ b/src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
@@ -83,7 +83,6 @@ class SizeDistributionWindow(QtWidgets.QDialog, Ui_SizeDistribution, Perspective
         self.backgd_subtr_plot: Data1D | None = None
         self.fit_plot: Data1D | None = None
         self.size_distr_plot: Data1D | None = None
-        self.trust_plot: Data1D | None = None
 
         self.model = QtGui.QStandardItemModel(self)
         self.mapper = QtWidgets.QDataWidgetMapper(self)
@@ -526,7 +525,6 @@ class SizeDistributionWindow(QtWidgets.QDialog, Ui_SizeDistribution, Perspective
         self.backgd_plot = None
         self.fit_plot = None
         self.size_distr_plot = None
-        self.trust_plot = None
         self.setupModel()
         self.enableButtons()
         self.clearStatistics()
@@ -636,7 +634,7 @@ class SizeDistributionWindow(QtWidgets.QDialog, Ui_SizeDistribution, Perspective
         qmin_fit = float(self.txtMinRange.text())
         qmax_fit = float(self.txtMaxRange.text())
 
-        self.size_distr_plot, self.trust_plot = self.logic.newSizeDistrPlot(result, qmin_fit, qmax_fit)
+        self.size_distr_plot = self.logic.newSizeDistrPlot(result, qmin_fit, qmax_fit)
         if self.size_distr_plot is not None:
             # set the trust range for the size distribution plot
             self.size_distr_plot.show_trust_bar = True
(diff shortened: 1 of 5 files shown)


src/sas/qtgui/Perspectives/SizeDistribution/SizeDistributionPerspective.py
```diff
@@ -635,15 +635,26 @@ def fitComplete(self, result: MaxEntResult) -> None:
...
+
         self.size_distr_plot, self.trust_plot = self.logic.newSizeDistrPlot(result, qmin_fit, qmax_fit)
         if self.size_distr_plot is not None:
+            # set the trust range for the size distribution plot
+            self.size_distr_plot.show_trust_bar = True
+            trust_range = self.logic.computeTrustRange(qmin_fit, qmax_fit)
+            self.size_distr_plot.trust_range = {
+                "d_low": trust_range[0],
+                "d_high": trust_range[1],
+            }
```
butlerpd, line 645: This code is in the wrong place. It in fact already exists in `SizeDistributionlogic`, The problem is that currently returns a 2 point plot as `self.trust_plot` (see line 639). the code in `SizeDistributionlogic.newSizeDistrPlot` can be considerably simplified by getting rid of all the complicated cruft used to kludge the two lines (the `trust_plot`) and just return `d_low` and `d_high` directly in lieu of the `trust_plot` which we no longer need.  This also solves the problem of having the two kludgy trust lines on top of the nice trust bar we should have had to start.

I suspect that also would allow us to get rid of the `DataRole.ROLE_SIZE_DISTRIBUTION` specific role and all the infrastructure that depends on it. As I recall that was done specifically to deal with the vline 2 point data? One would need to do a search to make sure it is properly extracted.

butlerpd (changes requested):
This works very smoothly whether stretching the graph or changing scales. A GREAT addition that was desired from the beginning.  I do have one question I do not fully comprehend though. While the right hand red trust line matches the new trust bar transition, the left hand red line seems to be well into the trust bar's red region instead of lined up with the transition. As far as I can tell both numbers are computed from the exact same code? but maybe used differently?

Otherwise of course those two kludgy trust lines that were created as an easy stop gap in lieu of the trustbar should be removed (see comment on code). Fortunately it was not done yet or I'd have missed the fact that the two do not match. That matching needs a bit more investigation.

Once they match and the red trust lines (trust_plot) are removed this should be good to go.

jellybean2004:
I found a better way to implement this, which is much simpler and works much better with the log scale.

Will change to that instead.

jellybean2004:
@butlerpd I have implemented a different approach to creating the trust bar, and it now aligns properly with the lines. Could you please test this one out and let me know if all looks good? I will then remove the lines.

Also, I don't think the data role can be removed, as it is handling other things, not the lines.

butlerpd (approved):
The bars now line up indeed. I also tried zooming on the plot and it worked as well. I see what you mean by the old method now. That was an image I guess? which explains a bit of why it shrank a bit differently on log vs linear scale?

As for the ROLE, you may be right. I'm vaguely remembering that it if a regular plot it was wanting to overlay the original data maybe on the plot by default? In which case it needs to stay.

I also so the changes you reverted so we could check how the lines matched the bar (thanks for doing that) and they looked right to me. So if you go ahead and reinstall them this should be ready to merge.

butlerpd:
Changes checked and approved. I note that you opted to remove the calculation of the trust region from the logic.py completely and go back to your implementation in sizedistribution.py  That works too.

Feel free to merge at any time. Look forward to seeing all 3 size distribution improvements together.

Issue #4016, opened 2026-06-29
Add docs for trust bar in Size Distribution


jellybean2004:
PR #4009

Will do this after I also rework the UI layout of the perspective to make it fit in the screen without needing to scroll.

PR #4017, opened 2026-06-29
Update Size Distribution UI to fit in screen

jellybean2004:
## Description

The size distribution perspective UI has been reorganised to fit on the screen without needing to scroll. 

This includes reorganisation, renaming and removal of UI elements wherever necessary.

Will update screenshots in documentation once UI changes have been reviewed and finalised.

## How Has This Been Tested?

Manually tested.

## Review Checklist:

- [x] There is an **issue** open for the documentation #4016 

**Installers**

**Licensing**

### DIFF
--- a/src/sas/qtgui/Perspectives/SizeDistribution/UI/SizeDistributionUI.ui
+++ b/src/sas/qtgui/Perspectives/SizeDistribution/UI/SizeDistributionUI.ui
@@ -7,11 +7,11 @@
     <x>0</x>
     <y>0</y>
     <width>577</width>
-    <height>634</height>
+    <height>586</height>
    </rect>
   </property>
   <property name="sizePolicy">
-   <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+   <sizepolicy hsizetype="Fixed" vsizetype="Preferred">
     <horstretch>0</horstretch>
     <verstretch>0</verstretch>
    </sizepolicy>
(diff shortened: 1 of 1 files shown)


DrPaulSharp:
For the parameters tab, here's the before and after:
Screenshot 2026-07-13 151930
Screenshot 2026-07-13 151608

DrPaulSharp:
and for the options tab:
Screenshot 2026-07-13 151941
Screenshot 2026-07-13 151735

DrPaulSharp:
Based on this, I would suggest:

- The parameters tab is hugely improved, I think it would be worthwhile to remove the space above `ChiSq` in the `Output` section.
- I think the `Background` section of the `Options` tab is less clear as a result of the changes. I could suggest how to fiddle with it, but I'm wondering if the best option is to make a new `Background` tab, what do you think?

DrPaulSharp:
On that subject, could you please clarify to me what the `Fit Flat Background`/`Fit Power Law` buttons do compared to the `Quick Fit`/`Full Fit` buttons please?

jellybean2004:
> Based on this, I would suggest:
> 
> * The parameters tab is hugely improved, I think it would be worthwhile to remove the space above `ChiSq` in the `Output` section.
> * I think the `Background` section of the `Options` tab is less clear as a result of the changes. I could suggest how to fiddle with it, but I'm wondering if the best option is to make a new `Background` tab, what do you think?

Thanks for your feedback @DrPaulSharp.

The space above `ChiSq` is where the number of iterations appears after a full fit is run, so it isn't really a blank space even though it seems so.

I considered adding a separate tab for the background options, as it is still quite cluttered, but it proved challenging because of how fiddly the designer app is... or my lack of patience to cope with its fiddliness... I could give it another shot.

butlerpd (approved):
NICE job of tightening the UI!! I generally hate tabs which is why we only had 2 but I think in this case it really does call for a 3rd. 

My only quibble here is that we should move the background to be the second tab since that will almost always need to be used. The options that are left are mostly fine left alone EXCEPT the weight factor which is a bit awkward. It really is a fudge factor multiplying the error bar factor (tsk tsk -- but seems to be needed for SAXS data in particular.

The only two options I see would be to move it to the "distribution parameters" on the front page (the only place with some space .. next to contrast) or as a separate box on background.  Plenty of space there but has nothing to do with background. 

So bottom line, if we move the background tab to be the second tab we are good. The only other thing might be to move the weight factor to "parameters" but it really should sit inside the weighting box. So I'd leave it for now unless somebody else has a better idea?

jellybean2004:
> NICE job of tightening the UI!! I generally hate tabs which is why we only had 2 but I think in this case it really does call for a 3rd.
> 
> My only quibble here is that we should move the background to be the second tab since that will almost always need to be used. The options that are left are mostly fine left alone EXCEPT the weight factor which is a bit awkward. It really is a fudge factor multiplying the error bar factor (tsk tsk -- but seems to be needed for SAXS data in particular.
> 
> The only two options I see would be to move it to the "distribution parameters" on the front page (the only place with some space .. next to contrast) or as a separate box on background. Plenty of space there but has nothing to do with background.
> 
> So bottom line, if we move the background tab to be the second tab we are good. The only other thing might be to move the weight factor to "parameters" but it really should sit inside the weighting box. So I'd leave it for now unless somebody else has a better idea?

Thanks for your review @butlerpd! 

I could move the weight factor to the bottom of the weighting. I thought it was related to |sqrt (1 Data)|, so I put them together. 

image

What do you think?

butlerpd:
Actually, You were correct to start with that the weight factor basically is a multiplication of whatever "error bar" you choose to use. So I think the original positioning was better.

Also it doesn't gain us anything to move it to a different place in that box. What I was trying to do was move all the things that a user almost always has to touch into the first two tabs.  That way people only touch two tabs most of the time. Unfortunately it seems that the error bars (however you choose them) are often too tight which is why the "weight factor" gets touched quite frequently as I understand it.

No worries. I see the background tab is moved up which is the easy thing we can do for now. I think it better to move the weight factor back to its original position but otherwise good to merge

PR #4026, opened 2026-07-02
getting example data from sasdata 

dehoni:
## Description

This PR fixes #3982 by getting example data from sasdata. Currently it seems that the installer only takes the example data folder shipped with sasview.

## How Has This Been Tested?

Click load data. Navigate to the example_data folder of sasview and check that folder like 1D and 2D data exist.

## Review Checklist:

[if using the editor, use `[x]` in place of `[ ]` to check a box]

**Documentation** (check at least one)
- [X ] There is **nothing** that needs documenting
- [ ] Documentation changes are **in this PR**
- [ ] There is an **issue** open for the documentation (link?)

**Installers**
- [ ] There is a chance this will affect the **installers**, if so
  - [ ] **Windows** installer (GH artifact) has been tested (installed and worked) 
  - [ ] **MacOSX** installer (GH artifact) has been tested (installed and worked)
  - [ ] **Wheels** installer (GH artifact) has been tested (installed and worked)

**Licensing** (untick if necessary)
- [x] The introduced changes comply with SasView license (BSD 3-Clause)

PR #4033, opened 2026-07-07
Adds example data from SasData repository

DrPaulSharp:
This PR is functionally the same as #4026, but with the branching properly resolved.


jellybean2004 (approved):

llimeht:
@jellybean2004, @DrPaulSharp I'm confused - wasn't this supposed to be merged into `release_6.2.0` branch? and it can land in whatever merging-back-then-main branches later. btw `git` will normally spot that a commit has been already applied on another branch and skip around it, so reverts are not needed.

jellybean2004:
Hi @llimeht.

The problem was that the original PR targeted the `merge_release_6.2.0` branch from an older base, so it was not the right place for the fix. I reverted that merge to remove the incorrect change, and then Paul opened a fresh PR based on main, so the fix is now going in through the correct path.

I have been cherry-picking commits into the release branch to put them into the next (pre-)release.

PR #4035, opened 2026-07-08
Updates navigation icons for the What's new window

### LINKED ISSUE #3304: Use arrows rather than shortened previous/next
'It would probably be better to just use arrows rather than to shorted Previous.
Arrows will make the meaning of the buttons less ambiguous, I think, mimicking what we see in e.g. web browsers' - @rozyczko 

In my opinion, this isn't essential to have before the bug fix code freeze, and thus should be low priority.

jellybean2004:
## Description

Changes "Prev" and "Next" to arrow buttons as suggested. Also adds a page count.

Fixes #3304 

## How Has This Been Tested?

Manually tested in the build.

## Review Checklist:

**Documentation**

**Installers**

**Licensing**

### DIFF
--- a/src/sas/qtgui/Utilities/UI/WhatsNewUI.ui
+++ b/src/sas/qtgui/Utilities/UI/WhatsNewUI.ui
@@ -58,7 +58,7 @@
      <item>
       <spacer name="horizontalSpacer">
        <property name="orientation">
-        <enum>Qt::Horizontal</enum>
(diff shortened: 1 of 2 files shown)


DrPaulSharp (approved):
This looks good, just one point to consider - is it worth setting up such that pressing the < button on page 1 takes you to page N, and vice versa? It seems like that would be more convenient to me.

PR #4054, opened 2026-07-13
Reorganises Shape2SAS UI for plots

### LINKED ISSUE #4044: Shape2SAS window is neither resizeable or scrollable
Not specific to 6.2.0b2.
But found whilst testing 6.2.0b2 on W11/x64.

**Describe the bug**
The Shape2SAS window opens maximised and is neither resizeable or scrollable.

Image

This becomes a problem when trying to use it on a small screen, for example a laptop, as you cannot access the 'action' buttons to make it do anything!

**To Reproduce**
Click Tools > Shape2SAS Calculator.

**Expected behavior**
The window should at least be scrollable.

And if one is not present, there should also be a Close button for consistency across SasView.

jellybean2004:
## Description

Changes the layout of plots in the tool window to be tabbed instead of vertically stacked to give plots more space.

Also reformats and cleans up some of the code in the changed files.

Fixes #4044 

## How Has This Been Tested?

Manually tested.

## Review Checklist:

**Documentation**
- [x] Documentation changes are **in this PR** -- will add as final commit after design finalised.
- [ ] There is an **issue** open for the documentation

**Installers**

**Licensing**

### DIFF
--- a/src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py
+++ b/src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py
@@ -2,7 +2,7 @@
 import re
 from types import MethodType
 
-from PySide6.QtCore import Qt
+from PySide6.QtCore import Qt, QTimer
 from PySide6.QtWidgets import (
     QCheckBox,
     QDialog,
@@ -76,7 +76,6 @@ class DesignWindow(QDialog, Ui_Shape2SAS, Perspective):
         self.modelTabButtonOptions.horizontalLayout_5.insertWidget(1, self.checkTheoreticalScattering)
         self.plot = QPushButton("Plot")
         self.modelTabButtonOptions.horizontalLayout_5.insertWidget(1, self.plot)
-        self.modelTabButtonOptions.horizontalLayout_5.insertWidget(1, self.plot)
 
         self.line2 = QFrame()
         self.line2.setFrameShape(QFrame.VLine)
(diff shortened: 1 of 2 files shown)


jellybean2004:
The new commit adds a third tab, which shows both the viewer and the scattering profile vertically stacked.

Instead of creating duplicates, the tabs reuse the plots.

DrPaulSharp (commented):
This looks good overall, I have a few comments.

- The `Delete Column` button has got squashed by the new tab layout. Have a think about how best to fit all of these parts in.
- The tabs help with making the graphs larger. What I think would be a good ideal would be to disable the `Scattering Profile` and `Combined` tabs when `Include Scattering` is False. Also, when `Plot` is pressed and `Include Scattering` is True, we should automatically switch to the `Combined` tab, so the user can see the changes to both plots (or that the scattering plot has appeared in the first instance).
- I'm keen to see whether the window is now at an appropriate size on @smk78 laptop.

Image

klytje:
Looks good. In addition to @DrPaulSharp's comments, I have the following: 
1. I think there is too much focus on the two plots in the new layout: for me, the default size only fits 5 subunits before horizontal scrolling is required. Since managing those data cells is the main purpose of the tool, I feel it'd be better to give them more space by default. This would probably also solve the squished "delete column" and "exclude overlap" texts. 
2. It would be nice if the divider between the data columns & the plots could be grabbed and shifted. No idea how difficult that'd be to implement. 
3. Perhaps enabling "include scattering" should trigger a plot update. That will basically always be the next move anyway for the user, and give direct feedback on what the button actually does. 
4. The generated P(q) plot needs to be of higher resolution: it looks very pixelated in the "scattering profile" view on larger displays.

jellybean2004:
Thank you both for your feedback and suggestions, @DrPaulSharp and @klytje!

I have updated the UI to give the data columns a bit more space, so no buttons are squished. I have also put the data columns and plot area in a `QSplitter`, so the divider between them is moveable.

The scattering plot is no longer embedded through `QGraphicsView.fitInView()`. Instead, it is drawn directly with Matplotlib on a `FigureCanvas`, which fixes the pixelation problem.

The `Include Scattering` checkbox now enables the two tabs with the scattering plot, which are disabled by default. It also triggers a plot update. There is a noticeable lag here, carried forward from the replotting.

Please let me know if there is anything else.

DrPaulSharp:
Hi @jellybean2004, this is looking good. I'd just like you to take another look at the header of the left hand panel. I've made the plots as big as possible and everything on the left is squashed up again - in particular, you can't see which column you might be deleting! What do you think is the best approach to aid clarity here?

image

Please also remove the debugging print statements on lines 295, 297 &  305 of `src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py`

jellybean2004:
> Hi @jellybean2004, this is looking good. I'd just like you to take another look at the header of the left hand panel. I've made the plots as big as possible and everything on the left is squashed up again - in particular, you can't see which column you might be deleting! What do you think is the best approach to aid clarity here?
> 
> image
> Please also remove the debugging print statements on lines 295, 297 & 305 of `src/sas/qtgui/Calculators/Shape2SAS/DesignWindow.py`

I have set the minimum width to stop the widgets from being squashed. What do you think?

DrPaulSharp (approved):
Yeah, this is looking good, a big improvement!

PR #4060, opened 2026-07-22
Remove pins from os-specific requirements files

krzywon:
## Description

The OS-specific files should be reverted to unpinned when in the main branch, as per our ADR https://github.com/orgs/SasView/discussions/3831.

This will also ensure @dependabot does not create new PRs.


jellybean2004 (approved):

llimeht:
It does leave me feeling nervous that the 6.2.0 installer is going to go out containing a pile of packages with published CVEs that IT departments will scan, find, and then get grumpy about.

PR #4067, opened 2026-07-25
fix the fit_started flag enablement. #3949

### LINKED ISSUE #3949: clicking "stop fit" on a fit page while a constrained fit is running starts a new fit
**Describe the bug**
I'm seeing interleaved fits running in the sasview log viewer. I think that is happening because "stop fit" on a fit page tab actually starts a new fit instead of stopping the running fit.

**To Reproduce**
1. Load '1d_data/latex_smeared'
2. Send to fitting
3. Set models to on each page to sphere. click radius and scale as fit parameters
4. Select constrained fit from the Fitting menu bar
5. Constrain M1.scale = M2.scale and M1.radius=M2.radius
6. Select the M2 tab and set scale, radius and background as fitting paramters
7. Select the M1 tab and set background as fitting parameters
8. Select the const. Si... tab and click fit
9. Select View log explorer from the menu bar
10. Select the M2 tab and click "stop fit"
11. Observe the log window and see if there are interleaved fits

**Expected behavior**
"stop fit" should stop the running fit, not start a new fit thread.

 - Version: main branch 2026-05-07 3326c208b

 - OS: mac os x Tahoe 26.4.1

 - Installed: git checkout

rozyczko:
## Description

When a constrained fit starts, ConstraintWidget emits `fittingStartedSignal`, and the perspective calls `disableInteractiveElements()` on each participating fit page but never set that page's `fit_started` flag.
Clicking "Stop fit" on the M2 page ran FittingWidget.onFit(). Since fit_started was set to False, it started a brand-new single-page fit while the constrained fit was still running...

`FittingWidget.onFit` now checks whether the page is running C/S fit
If so, it delegates the stop to the constraint tab instead of launching a new fit.

Fixes #3949

## How Has This Been Tested?

Locally on win11

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
@@ -393,12 +393,14 @@ class ConstraintWidget(QtWidgets.QWidget, Ui_ConstraintWidgetUI):
         """
         if not self.is_running:
             return
-        self.is_running = False
+        # stop the fit thread
+        self.calc_fit.stop()
+
         #re-enable the Fit button
         self.cmdFit.setStyleSheet(BG_DEFAULT)
         self.cmdFit.setText("Fit")
-        # stop the fit thread
-        self.calc_fit.stop()
+
+        self.is_running = False
         # re-enable the fit pages participating in the fit
         self.parent.fittingStoppedSignal.emit(self.getTabsForFit())
         self.parent.communicator.statusBarUpdateSignal.emit("Fitting cancelled.")
--- a/src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
@@ -292,12 +292,7 @@ class ConstraintWidget(QtWidgets.QWidget, Ui_ConstraintWidgetUI):
         """
         # Stop if we're running
         if self.is_running:
-            self.is_running = False
-            #re-enable the Fit button
-            self.cmdFit.setStyleSheet(BG_DEFAULT)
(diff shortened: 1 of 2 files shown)


src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
```diff
@@ -392,6 +387,22 @@ def onFit(self):
         self.parent.communicator.statusBarUpdateSignal.emit('Fitting started...')
         self.is_running = True
 
+    def stopFit(self):
+        """
+        Attempt to stop the running constrained/simultaneous fit
+        """
+        if not self.is_running:
+            return
+        self.is_running = False
```
krzywon: I would suggest stopping the fit before resetting flags and GUI elements. Ideally, stopping the fit would also return a success flag, similar to how you changed the constrained fit, that we could use to set `self.is_running`, but that is probably out of scope for this PR.

```suggestion
        # stop the fit thread
        self.calc_fit.stop()
        
        #re-enable the Fit button
        self.cmdFit.setStyleSheet(BG_DEFAULT)
        self.cmdFit.setText("Fit")
        
        self.is_running = False
```

krzywon:
Testing, this seems to fix the issue, but I want to test more before approving.

src/sas/qtgui/Perspectives/Fitting/ConstraintWidget.py
rozyczko: The single-page branch already does calc_fit.stop() first, then resets the GUI.
My new code in the Constrained widget does the reverse (it just preserves the ordering of the code from the original onFit).
I guess doing it consistently as you're proposing is worth doing.

krzywon (approved):

PR #4072, opened 2026-07-29
Don't calculate magnetic models if magnetism is not toggled.

### LINKED ISSUE #4071: Magnetism is always calculated for 2D data, even if magnetism is off
**Describe the bug**

Magnetism is always used for sasview models with sld parameters even if the magnetism box is not checked.

sasmodels decides whether to use 1D, 2D or 2D magnetic models based on the shape of the q inputs and the presence of non-zero `sld_..._M0` parameters. Since sasview initializes M0 to non-zero values, the magnetism calculation is always triggered for 2D data.

**To Reproduce**
1. Create a sphere model
2. Click 2D and magnetic
3. Select magnetic tab and modify parameters `sld_mtheta=90`. You can set `sld_M0=10` to make the effect more obvious.
4. Calculate and plot
5. Switch back to the model tab and turn off magnetism
6. Calculate and plot

**Expected behavior**
After turning off magnetism the scattering should no longer be asymmetric

**Screenshots**

Image

 - Version: 6.2.0rc1

 - OS: any

 - Installed: any

**Additional context**

The issue is that the sasmodels looks for a non-zero `..._M0` parameter to determine if the model is magnetic:

https://github.com/SasView/sasmodels/blob/81249db87885a4480c7951c28119840320d17dc7/sasmodels/details.py#L298-L306

The sasview shim passes the magnetism flag to the calculator unchanged:

https://github.com/SasView/sasmodels/blob/81249db87885a4480c7951c28119840320d17dc7/sasmodels/sasview_model.py#L727-L737

The GPU calculator checks if it is 1D, or 2D + magnetic  or 2D to call the appropriate kernel:

https://github.com/SasView/sasmodels/blob/81249db87885a4480c7951c28119840320d17dc7/sasmodels/kernelcl.py#L616 

The DLL calculator is a little different, with the 1D vs. 2D logic coming from `DllModel.make_kernel` and the non-magnetic/magnetic logic in `DllKernel._call_kernel`.

For some time now (since 2018), the `..._M0` parameter has been set to a non-zero value in SasView, even when the system is not magnetic:

https://github.com/SasView/sasview/blob/f24ed24d5312f390df51a9d7e2fe20b1707011f9/src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py#L79-L88

Because `..._mtheta` defaults to zero, the fact that `Imagnetic` was being called for 2D models rather than `Iqxy` was not visible.

The quick fix is to remove the initialization of the `..._M0` parameters. These should anyway be zero except for a one or two magnetic parameters. This makes the 2D models use `Iqxy` by default, only switching to `Imagnetism` when magnetism is checked and M0 is set for one of the SLDs. When magnetism is unchecked the M0 parameters all need to be reset to zero.

It also fixes the problem with a `cylinder+power_law` model, which otherwise complains that pure python models do not support magnetism.

Another approach is to update the sasview model shim so that the call to Iq takes a magnetic flag. Then the call to the kernel can ignore the magnetic flag set by `call_details` in the call to the kernel. This will require a number of changes to qtgui and sascalc to manage the flag, and to the sasview model shim in sasmodels to receive it. This should be done as part of a sascalc refactor that removes the sasview shim and calls sasmodels directly.

pkienzle:
## Description

Leaves the M0 values as their default (zero) when initializing the model.

Sets the magnetic parameters in the model to zero if magnetism is not active. This causes sasmodels to use Iqxy rather than Imagnetic to compute the 2D models.

This incidentally fixes the problem with a mixed magnetic and pure python model, which otherwise failed in sasmodels because the pure python model don't yet handle magnetism.

Fixes #4071

## How Has This Been Tested?

Start sasview with a sphere model, toggle 2D and plot. Verify that there is no asymmetry.

Toggle magnetism and check that M0 are all zero.

Set sld_M0 to 10, set sld_mtheta to 90 and plot. Shows asymmetry.

Toggle magnetism and plot. Shows no asymmetry.

Toggle magnetism and plot. Shows asymmetry again from values stored in the magnetism widget.

Load 2d magnetic data set and send to fitting.

Toggle radius and scale fit.

Toggle magnetism, and toggle sld_M0 and sld_mphi fit.

Start the fit. Verify that radius, scale, sld_M0 and sld_mphi are being updated.

Stop the fit, toggle magnetism and plot. No asymmetry.

Save analysis as non magnetic.

Toggle magnetism and save analysis as magnetic.

Restart sasview and load nonmagnetic analysis. Plot shows no asymmetry.

Toggle magnetism and plot. Shows asymmetry.

Restart sasview and load magnetic analysis. Plot shows asymmetry.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
@@ -75,7 +75,10 @@ class MagnetismWidget(QtWidgets.QWidget, Ui_MagnetismWidgetUI):
         if not self.logic.model_parameters:
             return
         self._magnet_model.clear()
-        # default initial value
+        # Add the magnetic parameters to the table.
+        # 2026-08-13 PAK: Initialize M0 values with defaults from the model (typically zero).
+        # The previous code initialized them to 0.5, 1.0, 1.5, etc., though this was
+        # not obvious since magnetic theta defaulted to zero.
         for param in self.logic.model_parameters.call_parameters:
             if param.type == 'magnetic':
                 self.addCheckedMagneticListToModel(param, param.default)
--- a/src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
@@ -76,16 +76,9 @@ class MagnetismWidget(QtWidgets.QWidget, Ui_MagnetismWidgetUI):
             return
         self._magnet_model.clear()
         # default initial value
-        m0 = 0.5
         for param in self.logic.model_parameters.call_parameters:
-            if param.type != 'magnetic':
-                continue
-            if "M0" in param.name:
-                m0 += 0.5
-                value = m0
-            else:
-                value = param.default
-            self.addCheckedMagneticListToModel(param, value)
+            if param.type == 'magnetic':
+                self.addCheckedMagneticListToModel(param, param.default)
 
         FittingUtilities.addHeadersToModel(self._magnet_model)
 
@@ -154,10 +147,12 @@ class MagnetismWidget(QtWidgets.QWidget, Ui_MagnetismWidgetUI):
             self.updateDataSignal.emit()
 
     def updateModel(self, model: Any | None = None) -> None:
-        # add magnetic parameters if asked
-        if self.isActive and self._magnet_model.rowCount() > 0:
+        """
+        Set model magnetism parameters from widget if magnetism is active, otherwise zero them.
+        """
+        if self._magnet_model.rowCount() > 0:
             for key, value in self.magnet_params.items():
-                model.setParam(key, value)
+                model.setParam(key, value if self.isActive else 0.)
 
     def toggleMagnetism(self, isChecked: bool) -> None:
         """
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -1391,7 +1391,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         category = self.cbCategory.currentText()
         # Check if the user chose "Choose category entry"
         if category == CATEGORY_DEFAULT:
-            # if the previous category was not the default, keep it.
+            # if the previous  was not the default, keep it.
             # Otherwise, just return
             if self._previous_category_index != 0:
                 # We need to block signals, or else state changes on perceived unchanged conditions
@@ -1412,8 +1412,16 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         if self.model_data is not None:
             # Store any old parameters before switching to a new category
             self.page_parameters = self.getParameterDict()
-        # Wipe out the parameter model
+        # Wipe out the parameter model and related sub-models
         self._model_model.clear()
+        # Clear polydispersity state
+        self.polydispersity_widget.poly_model.clear()
+        self.polydispersity_widget.poly_params = {}
+        self.polydispersity_widget.poly_params_to_fit = []
+        # Clear magnetism state
+        self.magnetism_widget._magnet_model.clear()
+        self.magnetism_widget.magnet_params = {}
+        self.magnetism_widget.magnet_params_to_fit = []
         # Safely clear and enable the model combo
         self.cbModel.blockSignals(True)
         self.cbModel.clear()


krzywon (approved):
Tested and fixes the issue. Code seems okay

dehoni:
Tested and it fixes the issues [4071]( https://github.com/SasView/sasview/issues/4071). Also fixes the error "NotImplementedError: Magnetism not implemented for pure python models" when trying to fit in 1D a combined model of  form factor with a shape-independent python model.
The error "NotImplementedError: Magnetism not implemented for pure python models" now only appears in the 2D mode and when magnetism is selected, which is OK and reasonable.

The old code incremented M0 (0.5, 1.0, 1.5...). I guess the reason was to initialise non-zero parameters to help users to see effects directly. In some sense, it is an improvement that the user now manually sets specific magnetic parameter values as suitable for the system.

src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
```diff
@@ -154,10 +147,12 @@ def onMagnetModelChange(self, top: QtCore.QModelIndex, bottom: QtCore.QModelInde
...
     def updateModel(self, model: Any | None = None) -> None:
-        # add magnetic parameters if asked
-        if self.isActive and self._magnet_model.rowCount() > 0:
+        """
+        Set model magnetism parameters from widget if magnetism is active, otherwise zero them.
+        """
+        if self._magnet_model.rowCount() > 0:
             for key, value in self.magnet_params.items():
-                model.setParam(key, value)
+                model.setParam(key, value if self.isActive else 0.)
```
dehoni, line 159: If a user switches between models, the old M0 values from the previous model could theoretically bleed through (tested by switching between spheres and vesicles). Is it possible to reset to the default?

src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
```diff
@@ -76,16 +76,9 @@ def setMagneticModel(self) -> None:
...
-            if param.type != 'magnetic':
-                continue
-            if "M0" in param.name:
-                m0 += 0.5
-                value = m0
-            else:
-                value = param.default
-            self.addCheckedMagneticListToModel(param, value)
+            if param.type == 'magnetic':
+                self.addCheckedMagneticListToModel(param, param.default)
```
dehoni, line 84: I propose to document why the initialisation was changed with a docstring like:
"""
    Set magnetism values on model.
    M0 parameters are initialized to their default values (typically zero).
    This ensures 2D models use Iqxy by default. Magnetic calculations are only
    triggered when magnetism is explicitly enabled and M0 is set to a non-zero value.
"""

src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
pkienzle: Other parameters (sld, sld_solvent, radius, scale, background) are also preserved when switching models.

To change this behaviour, open a PR with the following removed:

https://github.com/SasView/sasview/blob/02130e435755625f26c74913a50e6517e92f1967/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py#L1194-L1196

src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
pkienzle: Magnetism is now suppressed when magnetism is off regardless of the M0 values active in the GUI. I do this by checking if magnetism is active, setting the magnetism parameter value to zero if it is not.

src/sas/qtgui/Perspectives/Fitting/MagnetismWidget.py
krzywon: @dehoni, when the model changes, all model parameters are purposely copied/pasted into the new model to allow easier comparison between models. The main use case for this is one model didn't fit very well, but is close, so you switch to a similar model, and don't want to remember the parameter values you just had.

Can you give a use case where you would not want to copy/paste the values into the new model?

dehoni: @krzywon , it is fine to copy parameters over from one form factor model to another when staying inside the same category. Changing the category the nuclear model parameters are reset to their default values.
However, the Magnetism tabs stays when changing categories, and can be even selected and manipulated with no "model name" selected.
Switching to a shape independent model with 2D mode on: Magnetism is greyed. Going back to a form factor model the magnetism tab is set back to the default values (I would assume that this is the case all the time, when changing categories).

pkienzle: Polydispersity on radius is also preserved when switching category from sphere to cylinder.

I put in a fix but I'm not happy with it. It would be better to have a single copy of the parameters which gets copied when switching models within the category but cleared when the category changes. If the forms are rendered from the parameters then it would handle all three cases.

For now I'll leave it as technical debt.

krzywon:
@dehoni will recheck functionality to verify the latest changes are consistent with previous behavior

dehoni:
I am happy how polydispersity and magnetism is cleared now between categories and dissimiliar models. 

image
I have found a similiar issue where parameter are not properly cleared with the orientation parameters switching around between ellipsoid models in 2D mode. For example, in the attached image visible as double overlay in the function dropdown.

pkienzle:
While trying to reproduce the overwrite problem I noticed that a lognormal distribution throws an error when used for orientation parameters (see https://github.com/SasView/sasmodels/issues/757). I won't try to fix that here.

pkienzle:
I moved the combobox overlay problem to #4105 so that it doesn't block this PR.

krzywon:
I agree with what @pkienzle said and the combobox mirroring can be fixed separately. I think all other items noted by @dehoni are now resolved. I plan to merge this Friday if there are no objections.

PR #4076, opened 2026-08-04
Fix polydispersity plots

### LINKED ISSUE #4066: polydispersity plots persist even after turning polydispersity off or making the width zero
**Describe the bug**
.

**To Reproduce**
1. Make sure your preferences have "show distribution plots" on
2. load data and choose a model (I uses appoferritin and core-shell sphere)
3. Do a fit  - no distribution plot comes up
4. activate polydispersity and give at least one parameter a broad dispersity (I gave it 0.3)
5. Fit again. Dispersity plot(s) appear as they should
6. Set the dispersity to zero and delete the plots
7. Fit again
8. Dispersity plots pop back up. However the fit clearly does not include polydispersity
9. Uncheck the polydispersity box on the front page and try again (after deleting plots)
10. Dispersity plots show up again.

**Expected behavior**
They should not pop as the distribution should be a delta function. WORSE .. these plots show the last non-zero distribution used.

What seems to be happening is that when the plots at the end of a fit are called all subplots associated with that fit panel are shown. Because the distribution plots are not deleted when dispersity is removed they continue to plot showing false information.

The correct answer should be to delete those. If one changes the width or even the type of distribution (e.g. Gaussian vs Schulz ) the dispersity subplots are overwritten. deleting is the appropriate way to overwrite with zero distribution. I realize that may lead to tortuous logic which should be avoided but ....

 - Version: 6.02b3

 - OS: WIndows 11

 - Installed: windows installer

DrPaulSharp:
## Description

The routine `FittingUtilities.plotPolydispersities` correctly ignores polydispersity plots if there is no polydispersity for the parameter in question. However, this causes the bug outlined in #4066 when polydispersity is set to zero, or disabled, as the previous plots persist. This PR acts to set the plots are DELETABLE when these circumstances arise, resulting in them being cleaned up by exisiting code.

Fixes #4066

## How Has This Been Tested?

Follows procedure in #4066, results are as expected.

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -76,6 +76,7 @@ if not hasattr(SasviewModel, 'get_weights'):
     SasviewModel.get_weights = get_weights
 
 logger = logging.getLogger(__name__)
+polydispersity_plot_name = "{} {} polydispersity {}"
 
 class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
     """
@@ -684,6 +685,7 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         self.polydispersity_widget.cmdFitSignal.connect(lambda: self.cmdFit.setEnabled(self.haveParamsToFit()))
         self.polydispersity_widget.updateDataSignal.connect(lambda: self.updateData())
         self.polydispersity_widget.iterateOverModelSignal.connect(lambda: self.iterateOverModel(self.updateFunctionCaption))
+        self.polydispersity_widget.deletePlotSignal.connect(self.setPlotDeletable)
         self.polydispersity_widget.toggledSignal.connect(self.onPolyToggled)
         self.magnetism_widget.cmdFitSignal.connect(lambda: self.cmdFit.setEnabled(self.haveParamsToFit()))
         self.magnetism_widget.updateDataSignal.connect(lambda: self.updateData())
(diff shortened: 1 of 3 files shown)


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -2519,9 +2543,9 @@ def _appendPlotsPolyDisp(self, new_plots: list[Any], return_data: dict, fitted_d
         """
         for plot in FittingUtilities.plotPolydispersities(return_data.get('model', None)):
             data_id = fitted_data.id.split()
-            plot.id = "{} [{}] {}".format(data_id[0], plot.name, " ".join(data_id[1:]))
+            plot.id = "{} [{} polydispersity] {}".format(data_id[0], plot.name, " ".join(data_id[1:]))
```
jamescrake-merani: I see in other parts of the branch that you have changed some old format calls to f-strings, so could the same not be done here as well? I think it would look a bit neater.

jamescrake-merani (approved):
I'm approving this with one minor comment regarding the code which I'd like you to take a look at, but if you don't think this is a change that should be done in this branch then feel free to merge.

I would say however that it would be good if @butlerpd could confirm this fix works as well given he was person to originally raise this issue. I would not merge this branch until this is done.

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
DrPaulSharp: Yep, good call. I forgot to change this one.

butlerpd (approved):
Tested reasonably extensively as requested and I agree the issue is resolved. There is only one quirk which I think comes from the way we have implemented the compute/plot button and not from this PR. Basically, if one changes the PD from 0 to something and hits fit, the PD plot pops up. However if instead one hits return or the compute button, the curve recalculated and plots correctly but the PD plot does NOT pop up. Only if you hit the compute button a second time (and any subsequent computes) will the PD plot pop up.

Code looks reasonable too but @jamescrake-merani James probably did a better job of reviewing it already.

Ready to merge

PR #4082, opened 2026-08-06
attempting to fix #4047

### LINKED ISSUE #4074: composite model failure
**Describe the bug**
Elliot Gilbert from the ACNS reported that an error occurred when trying to run a composite model created by adding a power law plus two Gaussians following the `help` documentation. @butlerpd reproduced the error.  The error occurs when choosing that model from the dropdown. In other words creating the model saves correctly, SasView then recognizes it as a valid plugin to the extent that it shows up in the plugin category dropdown, but as soon as that model is chosen the following error is thrown and no model is actually loaded.  

```python
10:45:20 - INFO: --- SasView session started, version 6.1.2, 2025 ---
10:45:20 - INFO: Python: 3.11.9 (tags/v3.11.9:de54cf5, Apr 2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
10:47:41 - ERROR: Traceback (most recent call last): File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1418, in onSelectModel File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1589, in respondToModelStructure File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 2364, in SASModelToQModel File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 2412, in fromModelToQModel File "", line 244, in basename File "", line 213, in split TypeError: expected str, bytes or os.PathLike object, not NoneType
```

The plugin model file @butlerpd used was:
```python
from sasmodels.core import load_model_info
from sasmodels.sasview_model import make_model_from_info

model_info = load_model_info('power_law+gaussian_peak+gaussian_peak')
model_info.name = 'PowerPlusTwoGaussian'
model_info.description = 'power_law + gaussian_peak + gaussian_peak'
Model = make_model_from_info(model_info)
```

**To Reproduce**
1. create the above plugin. I used the Add/multiply feature to first create a power law plus Gaussian. I then opened that model in the model editor and added the second Gaussian as described in the help and shown above. Then saved
2. Now, from within the SasView GUI, use the `Plugin Models` Category and then choose the model just created
3. Result, no model is loaded and the error is thrown

**Expected behavior**
The model should load correctly as any other model

 - Version: Still present in 6.2.0 but initially detected in 6.1.2

 - OS: Windows 11 for sure

 - Installed: from installer on release page

rozyczko:
## Description

Mixture/composite model infos are created in memory, not read from a file, so `model_info.filename` is None.
`make_model_from_info` copies that None onto the generated class.
The fix proposed: in `find_plugin_models()`, after loading each plugin, fill `model.filename` with the actual plugin file path when it's None.

Fixes #4074 

## How Has This Been Tested?

Local Win11

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
+++ b/src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
@@ -2048,7 +2048,11 @@ class FittingWidget(QtWidgets.QWidget, Ui_FittingWidgetUI):
         if kernel_module is None:
             # mismatch between "name" attribute and actual filename.
             curr_model = self.models[model_name]
-            name, _ = os.path.splitext(os.path.basename(curr_model.filename))
+            filename = getattr(curr_model, 'filename', None)
+            if filename is None:
+                logger.error("Can't find the model file for %s", model_name)
+                return
+            name, _ = os.path.splitext(os.path.basename(filename))
             try:
                 kernel_module = generate.load_kernel_module(name)
             except ModuleNotFoundError as ex:
--- a/src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py
+++ b/src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py
@@ -247,6 +247,27 @@ class FittingWidgetTest:
         #Test what is current text in the combobox
         assert fittingWindow.cbCategory.currentText(), "None"
 
+    def testSelectModelWithoutFilename(self, widget, mocker, caplog):
+        """
+        Selecting a model whose class carries no filename (e.g. a mixture
+        model built programmatically with make_model_from_info) must log an
+        error instead of raising TypeError in os.path.basename(None).
+        Regression test for issue #4047.
+        """
+        class DummyMixtureModel:
+            name = "dummy_mixture"
+            filename = None
+
+        widget.models["dummy_mixture"] = DummyMixtureModel
+        # force the name-based lookup to fail so the filename fallback runs
+        mocker.patch.object(FittingWidget.generate, 'load_kernel_module',
+                            side_effect=ModuleNotFoundError("dummy_mixture"))
+
+        with caplog.at_level(logging.ERROR):
+            widget.fromModelToQModel("dummy_mixture")
+
+        assert "Can't find the model file for dummy_mixture" in caplog.text
+
     def testSignals(self, widget):
         """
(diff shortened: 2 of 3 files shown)


krzywon:
Are both this and https://github.com/SasView/sasmodels/pull/749 needed to fix the issue?

rozyczko:
> Are both this and [SasView/sasmodels#749](https://github.com/SasView/sasmodels/pull/749) needed to fix the issue?

No, the sasmodel fix is just complementary. The fix here addresses the main issue and should be merged.

butlerpd (approved):
Thanks @rozyczko. I did the functional test and this does fix the problem I reported.

Like @krzywon I had some questions about the number of places the fix is being implemented (sasmodels `sasview_model,py`; sasview `FittingWidget.py` and sasview `models.p`y). All are essentially catching the same thing. After looking at it in more depth I agree that is the best approach for the moment though in principle I think the sasmodels fix should be all that is needed. Alternatively, the change in `models.py` should be all that is needed, but I believe our code in now convoluted enough that there may multiple paths to failure and defense in depth may be our best option.

In the longer term we need a refactor of our model interface to ditch the `sasview_model.py` shim and work directly with the sasmodels native API.  As part of that we need to re-evaluate IMO the sasview assumption (which seems to diverge from the sasmodels assumption) that all models exists as files on disk. Part of our problems have stemmed from the equating `model.name` and `model.filename` in the minds of various contributors over the years. A model that is created on the fly for example (a composite model for example) does not necessarily need to be written to disk, and if it isn't it will not have a filename which may legitimately be set to None.

One question this raises: Should we not be updating the unit tests to cover this by editing `src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py`?

rozyczko:
> One question this raises: Should we not be updating the unit tests to cover this by editing `src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py`?

This would be a weird case to test for, but for completeness, I added a test.

butlerpd:
Sorry @rozyczko I meant it more as a question. What I was thinking was that given the disconnect between sasmodels approach (models do not need to files on disk) and sasview (models are always files on disk) should we put something in to guard against somebody accidentally committing something that again confuses those two. To be frank, I'm not sure there is given this fix should deal with anything that falls through?

Either way we should merge this important fix IMO

PR #4087, opened 2026-08-18
Flatpak automated update script

jamescrake-merani:
## Description

This provides a script to automatically update the Flatpak manifest files which should be run every release.

## How Has This Been Tested?

The script uses PEP 723 to define its dependencies, so you shouldn't need to install them manually. The easiest way to run this script is with uv, as uv will automatically create a separate environment with the correct dependencies installed.

You need to have the [flatpak pip generator](https://github.com/flatpak/flatpak-builder-tools/tree/master/pip) script on your path. You should also make sure the `requirements-release-ubuntu-latest.txt` file has _pinned_ dependencies generated through `uv pip freeze`.

The script will output to the `outputs` directory, though you can change this using the `--output-dir` flag. Once the script has run, you should replace the files in the Flatpak repo with the files the script has generated, and try to build the Flatpak. The Flatpak build should succeed. If it doesn't, this might be because a new dependency was added, and should use the prebuilt wheel. If this is the case, you should add that file to the `prefer_wheels.txt` file. Any other errors indicate a problem with this script which should be reported, as the script aims to reduce manual intervention as much as possible.

During testing, you can try different sets of dependency pins. Perhaps from a previous release, or also from the CI.

A known issue is that there seems to be some issues building the Flatpak with the output of `pip freeze` (which unfortunately is not exactly equivalent to `uv pip freeze`), so you should use pins generated from `uv pip freeze` for now.

## Review Checklist:

**Documentation**
- [x] There is an **issue** open for the documentation (link to be added)

**Installers**

**Licensing**


jamescrake-merani:
A couple things to note:

- This appears to be failing the ruff check. I'll fix this tomorrow.
- I've tested this before, and it produces a working build but I made some last minute changes which I will test tomorrow morning.

build_tools/flatpak_update_script.py
```diff
@@ -0,0 +1,203 @@
...
+
+def process_requirements_file(file_path: Path, output_directory: Path) -> Path:
+    """This function is needed to remove references to other requirement files
+    because the Flatpak builder tool needs to read each file in an isolated
+    container. This function completely strips these references because SasView
+    can be built without them. Later, its also possible the requirements can
+    just be copied where these references appears."""
+    with open(file_path) as f:
+        lines = f.readlines()
+    new_lines = [l for l in lines if "-r " not in l and "-e " not in l]
```
krzywon, line 122: Would you also want to remove empty lines and comments, or are these fine for later use?

`new_lines = [l for l in lines if l and not l.startswith("#") and "-r " not in l and "-e " not in l]`

build_tools/flatpak_update_script.py
```diff
@@ -0,0 +1,203 @@
...
+    new_path = output_directory / Path("processed_requirements") / file_path.name
```
krzywon: `Path("processed_requirements")` -> `"processed_requirements"`

build_tools/flatpak_update_script.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def install_sdk(qt_version: str):
+    subprocess.call(["flatpak", "--user", "-y", "install", f"org.kde.Sdk//{qt_version}"])
+
+
+def check_platform() -> bool:
+    if system() != "Linux":
+        print(
+            "ERROR: You appear to be running this script outside of Linux. "
```
krzywon, line 177: Nitpicky wordsmithing: The word Linux appears 5 times in two sentences. It's also capitalized four of those times.

build_tools/flatpak_update_script.py
```diff
@@ -0,0 +1,203 @@
...
+            ["flatpak-pip-generator", "--help"],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+    except FileNotFoundError:
+        print(
+            "Error: you need to have the program 'flatpak-pip-generator' on your path. See the README for more information.",
+            file=stderr,
+        )
+        exit(1)
```
krzywon: Suggestion: add a second check here `check_flatpak_reqs()` that handles the try/except block to simplify the entire block and allow future reqs without polluting main.

```
if not check_platform() or not check_flatpak_reqs():
    exit(1)
```

build_tools/flatpak_update_script.py
```diff
@@ -0,0 +1,203 @@
...
+    requirements_dev_file = process_requirements_file(Path("requirements-dev.txt"), output_dir)
+    requirements_file = process_requirements_file(Path("requirements-release-ubuntu-latest.txt"), output_dir)
+    flatpak_requirements_file = process_requirements_file(Path("flatpak_requirements.txt"), output_dir)
```
krzywon: See my previous comment.

krzywon (commented):
I don't have a working Linux system handy, so I haven't tested, but see my comments.

build_tools/flatpak_update_script.py
jamescrake-merani: The requirements parser should be able to handle the comments, as I think it uses the same logic that pip would use to parse it. The reason why these lines have to get removed is because the script looks at each requirements file in isolation, and it can't refer to other ones so any attempt to do so will make it crash.

build_tools/flatpak_update_script.py
jamescrake-merani: I've capitalised the Linux that wasn't. I do get what you mean about the repetitiveness, but I'm not sure what would be a good way of resolving it.

build_tools/flatpak_update_script.py
jamescrake-merani: Good suggestion. I went further than this, and instead rolled all the verification into one `check_requirements`. I've also rolled the files into a separate dataclass so I can move a lot of that out of the main function.

jamescrake-merani:
A check is needed to see if Flatpak is installed, as right now it only checks for the Flatpak builder tools.

PR #3974, opened 2026-06-01
Allow pytest from project base directory

pkienzle:
## Description

Allows pytest to run from the project base directory.

The unused sas.sascalc.simulations and sas.sascalc.realspace packages are not scanned for tests.

Fixes the glob pattern to look for files with "test" or "Test" in their name. The previous version was looking for `*Test*py`, which wasn't finding the utest files in the test directory on case sensitive filesystems. Instead, a second `u*py` glob pattern was introduced to find these tests, but that picks up non-test files such as `util.py` and `urls.py`.

Fixes #3892

## How Has This Been Tested?

Run `pytest` from the base directory of the project and check that it includes both `*Test*` files from qtgui and `utest*` files from the test subdirectory. Verify that there are no errors from the tests.

## Review Checklist:

[if using the editor, use `[x]` in place of `[ ]` to check a box]

**Documentation** (check at least one)
- [x] There is **nothing** that needs documenting
- [ ] Documentation changes are **in this PR**
- [ ] There is an **issue** open for the documentation (link?)

**Installers**
- [ ] There is a chance this will affect the **installers**, if so
  - [ ] **Windows** installer (GH artifact) has been tested (installed and worked) 
  - [ ] **MacOSX** installer (GH artifact) has been tested (installed and worked)
  - [ ] **Wheels** installer (GH artifact) has been tested (installed and worked)

**Licensing** (untick if necessary)
- [x] The introduced changes comply with SasView license (BSD 3-Clause)

PR #4099, opened 2026-08-27
allow pytest to run locally on an editable install

### LINKED ISSUE #3892: Pytest throws when running tests against the base sasview directory
**Describe the bug**
As noted in #3867, a few directories in sasview throw errors when running pytest against them. Either these directories need to be added to the norecurse list, the errors need to be fixed, or the test directories need to be specified in `pyproject.toml`. I think the last option is the cleanest and best long-term.

Directories causing issues:

- `docs/sphinx-docs/source-temp/user/models/src/`: A couple `import file mismatch` errors
- `src/sas/sascalc/simulation/pointsmodelpy/tests/`: No module named `sasModeling` and no module named `enthought`
- `src/sas/webfit/`: Multiple `no module named django` errors

**To Reproduce**
1. Run `python -m pytest .` from the base sasview direectory
2. Multiple errors are thrown and unit tests stop early.

**Expected behavior**
All tests should run and no errors should be thrown.

**Additional context**
This will not effect our CI, which only runs tests against `./test/` and `./src/sas/qtgui/` (once #3867 is merged) but might cause confusion for new contributors trying to run the tests.

pkienzle:
## Description

Resource tests involving docs were failing locally on an editable install.

The pip install process for the editable install is compiling the docs into the `site-packages/sas/docs` directory but it isn't creating `site-packages/sas/__init__.py` so it is not showing up as the package location (nor should it). Instead, I've extended the sasview resource manager so that it looks in the site-packages directory if it cannot find a resource through `importlib.resources`.

This PR is based on #3974 (3892-pytest-errors branch) allowing pytest to run from the root of the source tree.

Fixes #3892

## How Has This Been Tested?

To reproduce, create a new python environment for sasview and install from the git clone with `pip -e .` or `uv pip -e .`.

Try running `pytest test/system/utest_resources.py`. It should fail if you are not using the 3892-pytest-errors-local branch.

This needs to be tested in the various execution environments: mac install, windows install, linux install, linux flatpack, debian install, pip install, pip install -e ., and the various git workflows.

## Review Checklist:

**Installers**
- [X] There is a chance this will affect the **installers**, if so

### DIFF
--- a/src/sas/system/resources.py
+++ b/src/sas/system/resources.py
@@ -43,6 +43,8 @@ import importlib.resources
 import itertools
 import logging
 import re
+import shutil
+import site
 import tempfile
 from collections.abc import Generator
 from pathlib import Path, PurePath
@@ -136,7 +138,7 @@ class ModuleResources:
             yield tmp_path
 
     def path_to_resource(self, src: str | PurePath) -> Path:
-        """Provide the filesystem path to a file resource
+        """Provide the filesystem path to a file resource (not a directory)
 
         If the resource is already available on the filesystem, then provide
         the path to it directly; if it is not available, then a FileNotFoundError
@@ -283,24 +285,28 @@ class ModuleResources:
 
     def _path_to_resource_adjacent(self, src: str, accept: _ResType = _ResType.FILE) -> Path | None:
         """calculate the filesystem path to the recorded resource if it exists"""
-        resource = importlib.resources.files(self.module) / src
-
         # importlib.resources will always return something... but it might
-        # not exist, so the is_file() check is needed.
+        # not exist, so the is_file() check is needed. There is no node.exists() method.
         # If the resource is on disk, then importlib.resources.readers.FileReader
         # will return a pathlib.Path object directly to the resource which is
-        # all that is needed here
+        # all that is needed here.
+        def check_node(node):
+            return (node.is_file() and accept == _ResType.FILE) or (node.is_dir() and accept == _ResType.DIR)
 
-        result = None
-
-        if isinstance(resource, Path):
-            if (resource.is_file() and _ResType.FILE in accept) or (resource.is_dir() and _ResType.DIR in accept):
-                result = resource
-
-        if result:
+        resource = importlib.resources.files(self.module) / src
+        if check_node(resource):
             logger.debug("Found adjacent: %s", resource)
+            return resource
 
-        return result
+        # For editable distributions, the docs and example data are compiled into
+        # subdirectories of site-packages/sas, so check for resources there as well.
+        for path in site.getsitepackages():
+            resource = Path(path) / self.module / src
+            if check_node(resource):
+                logger.debug("Found adjacent: %s", resource)
+                return resource
+
+        return None
 
     def _extract_resource_adjacent(self, src: str, dest: Path) -> bool:
         """extract the adjacent resource (if it exists) to the destination filename
@@ -309,8 +315,8 @@ class ModuleResources:
 
         This method should should transparently access resources in zip bundles
         """
-        resource_path = importlib.resources.files(self.module) / src
-        if not resource_path.is_file():
+        resource_path = self._path_to_resource_adjacent(src)
+        if not resource_path:
             return False
 
         if not dest.parent.exists():
@@ -326,37 +332,16 @@ class ModuleResources:
 
         dest must be a directory
         """
-        # normalise the representation
-        spth = Path(src)
-        src = str(spth)
-
-        dpth = Path(dest)
-        dpth.mkdir(exist_ok=True, parents=True)
-
-        found = False
-        try:
-            for resource in importlib.resources.files(self.module).joinpath(src).iterdir():
-                f_name = dpth / resource.name
-                s_name = spth / resource.name
-                if "__pycache__" in resource.name:
-                    continue
-
-                if resource.is_dir():
-                    # recurse into the directory
-                    found |= self._extract_resource_tree_adjacent(s_name, f_name)
-                elif resource.is_file():
-                    if not f_name.exists():
-                        logger.debug("Found adjacent: %s", s_name)
-                        with open(f_name, "wb") as dh:
-                            dh.write(resource.read_bytes())
-                    found = True
-                else:
-                    logger.warning("Skipping %s (unknown type)", str(s_name))
-        except FileNotFoundError:
-            # specified path does not exist or is not a directory
+        resource = self._path_to_resource_adjacent(src, accept=_ResType.DIR)
+        if not resource:
             return False
 
-        return found
+        with importlib.resources.as_file(resource) as source_path:
+            destination_path = Path(dest)
+            destination_path.mkdir(exist_ok=True, parents=True)
+            shutil.copytree(source_path, destination_path, dirs_exist_ok=True)
+
+        return True
 
 
 def _clean_path(src: str | PurePath) -> str:


src/sas/system/resources.py
```diff
@@ -283,24 +285,28 @@ def _copy_resource_recorded(self, resource: importlib.metadata.PackagePath, dest
...
-            if (resource.is_file() and _ResType.FILE in accept) or (resource.is_dir() and _ResType.DIR in accept):
-                result = resource
-
-        if result:
+        resource = importlib.resources.files(self.module) / src
+        if check_node(resource):
             logger.debug("Found adjacent: %s", resource)
+            return resource
+
+        # For editable distributions, the docs and example data are compiled into
```
llimeht, line 301: In an editable install, the docs are installed and so they are explicitly listed in the `dist-info/RECORD` file. That makes them 'recorded' resources not 'adjacent' resources. This is the wrong function to be looking for them - they should already be discovered by the `_path_to_resource_recorded` function (and are discovered by that when sasview is run - that's how the docs are found at runtime)

src/sas/system/resources.py
```diff
@@ -326,37 +332,16 @@ def _extract_resource_tree_adjacent(self, src: str | Path, dest: Path) -> bool:
...
-            # specified path does not exist or is not a directory
+        resource = self._path_to_resource_adjacent(src, accept=_ResType.DIR)
+        if not resource:
             return False
 
-        return found
+        with importlib.resources.as_file(resource) as source_path:
+            destination_path = Path(dest)
+            destination_path.mkdir(exist_ok=True, parents=True)
+            shutil.copytree(source_path, destination_path, dirs_exist_ok=True)
```
llimeht, line 342: I remember spending too much time on dealing with the edge cases where the resource was inside a `whl`, `zip` or whatever `pyinstaller` might choose to do to a resource on any given platform. It might also have been written when we supported Python < 3.12 and there are useful improvements here that enable `as_file` to work with directories.

This looks right ... and using a new library function to delete a pile of code is a good outcome. I'm just nervous about how much time has been repeatedly sunk into unbreaking finding the docs...

(While doing the iteration manually is sucky, the code is also not slow enough to worry about - it is used to extract all the docs and example_data from sasdata and sasmodels and build time and is basically instantaneous.)

llimeht (changes requested):
I haven't looked at the errors that are coming from the test suite (recently) so not yet sure what failure is being addressed here; the changes to finding the docs in site-packages don't look right though. The other refactoring and improvements look good.

src/sas/system/resources.py
pkienzle: The problem is that `importlib.metadata.files("sasview")` is giving different results depending on whether the sasview root is on the python path.

When running python from the sasview directory the metadata files contains the code in `src/sas` but not the files in `docs/user`. Running outside the sasview directory it is the opposite.

Try it with
```sh
$python -c "import importlib.metadata; print('\n'.join(str(s) for s in importlib.metadata.files('sasview')[:20]))"
```

Even when running pytest from another directory pytest inserts the directory with pyproject.toml into the python path. With the wrong package data the resource tests always fail.

This PR also affects the docs. With it I'm able to use the help button when running from the sasview directory on an editable install. Without it I need to change to a different directory when starting sasview to see the help. This is of marginal use since the docs are pegged to when the install was run. If you are trying to view changes to the docs you will need to rerun the install each time.

Since "recorded" is checked first, the check for sas/docs in site-packages should do no harm.

I'm not going to spend anymore time trying to resolve the problems with the sasview directory on the path. I'm okay with deleting the site-packages hack and live with local pytest errors and broken help, or having someone else figure out how to remove the sasview root from the python path when using an editable install.

---

Without this PR the errors are:
```sh
$ pytest test/system/utest_resources.py  
...

self = <utest_resources.TestAdjacent object at 0x1073736f0>, input_type = <function _as_path at 0x107c3dc60>

    def test_path_to_resource_directory(self, input_type):
        """Test extracting single resource adjacent to code in module"""
        resources = sas.system.resources.ModuleResources("sas")
    
        source = "docs/user"
        src = input_type(source)
    
>       pth = resources.path_to_resource_directory(src)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

test/system/utest_resources.py:141: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <sas.system.resources.ModuleResources object at 0x10df4f5f0>, src = 'docs/user'

    def path_to_resource_directory(self, src: str | PurePath) -> Path:
        """Provide the filesystem path to a file resource directory
    
        If the resource is already available on the filesystem, then provide
        the path to it directly; if it is not available, then a NotADirectoryError
        is raised and the caller should extract into a filesystem location that
        they can clean up after use.
    
        Note that completely empty directories might not be found by this method
        as they are not recorded in the dist-info/RECORD.
        """
        src = _clean_path(src).rstrip("/")
        path = self._path_to_resource_recorded(src + "/**")
        if path:
            # this will be a entry within the directory or possibly even a subdirectory
            # need to walk up to find the entry we really want
            match = str(Path(f"/{self.module}/{src}"))  # pathlib.Path directories don't end in /
            for part in itertools.chain([path], path.parents):
                if str(Path(part)).endswith(match):
                    return part
            # if the upwards search falls through to here, then something went wrong
            # with the original filtering of the RECORD data, but allow it to try
            # importlib.resources before raising the exception
    
        path = self._path_to_resource_adjacent(src, accept=_ResType.DIR)
        if path:
            return path
    
>       raise NotADirectoryError(f"Resource directory {src} not found in module {self.module}")
E       NotADirectoryError: Resource directory docs/user not found in module sas

src/sas/system/resources.py:182: NotADirectoryError
======================================================================================================================== 12 failed, 16 passed in 0.88s ========================================================================================================================
```

llimeht: Thanks for the extra info - this is quite mysterious.

> The problem is that `importlib.metadata.files("sasview")` is giving different results depending on whether the sasview root is on the python path.

I can't reproduce this. The python one-liner gives the same files no matter where I run it from if I run it from a venv with an editable install.
 
> When running python from the sasview directory the metadata files contains the code in `src/sas` but not the files in `docs/user`. 

I don't have a `docs/user` directory at all; I do have a `docs/sphinx-docs/source/user/` though. Is that what you mean or is it a different doc tree that is in play here?

> This PR also affects the docs. With it I'm able to use the help button when running from the sasview directory on an editable install. Without it I need to change to a different directory when starting sasview to see the help. 

I definitely don't see that. Viewing the docs from an editable install should be working for a while now - I don't remember which PR that was, but it was definitely part of the requirements.

> Without this PR the errors are:
> 
> ```shell
> $ pytest test/system/utest_resources.py  

That passes with no issues here at all.

As a cross-check, I tried a fresh clone and a fresh venv and those tests pass, and sasview can find its help files without issues.

PR #4107, opened 2026-09-02
Be able to create trends on the data explorer.

jamescrake-merani:
## Description

This PR makes it possible to create trends on the data explorer.

## How Has This Been Tested?

You need to load some data with metadata. I have been using the Mumag data.

- Go to File -> Advanced Load
- Make sure you've set the 'Dataset Type' to '1 D I vs Q'
- Click on 'Select File', and load some Mumag data. If you've got sasdata cloned on the 'refactor_24_trend_new' branch, you can go to 'test/mumag', and load the data in one of the files.
- Click on 'Edit Metadata', and scroll down to the magnetic metadata. The Mumag data files in that directory follow the format [counting_index]\_[applied_magnetic_field]\_[saturation_magnetization]\_[demagnetizing_field]. Make sure to select the right part of the file for each metadata.
- Click 'Save', and then click 'Done' to load the files.
- Select all the files you just loaded, right click, and select 'Make Trend'. (This option should always appear when you've selected more than one data object.)
- Click on 'Add Axes'.
- Expand 'raw', 'root', 'magnetic', and select all of the metadata under 'magnetic'.
- Click 'Select', and then click 'Make Trend'
- You should now see the trend created on the data explorer.
- For each individual data item in the trend, you should be able to right click on it, click 'View Data', and it will show the values for the trend axes (only if its in a trend, which is a good way to test this has all worked).

You can't actually do anything with a trend yet, but my plan next is to work on integration of the Mumag tool as a perspective which will take advantage of the work here.

### DIFF
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -25,7 +25,7 @@ pyparsing
 PySide6
 pytools
 qtconsole
-sasdata @ git+https://github.com/SasView/sasdata.git@refactor_24
+sasdata @ git+https://github.com/SasView/sasdata.git@refactor_24_trend_new
 sasmodels
 scipy
 siphash24
--- a/src/sas/data_explorer_tree.py
+++ b/src/sas/data_explorer_tree.py
@@ -146,7 +146,10 @@ class DataExplorerTree(QTreeWidget):
             box.show()
 
     def showViewData(self):
-        viewer = DataViewer(self.currentTrackedDatum)
+        associated = self._data_manager.get_association(self.currentTrackedDatum)
+        if isinstance(associated, Trend):
+            trend = associated
+        viewer = DataViewer(self.currentTrackedDatum, trend)
         viewer.exec()
 
     @property
--- a/src/sas/data_explorer_tree.py
+++ b/src/sas/data_explorer_tree.py
@@ -24,6 +24,9 @@ from sas.refactored import Perspective
 def tracked_data_name(data: TrackedData) -> str:
     if isinstance(data, SasData):
         return data.name
+    if isinstance(data, Trend):
+        # TODO: This is temporary. We need a way for a Trend to represent itself.
+        return 'Trend'
     else:
         return data.formatName
 
--- a/src/sas/data_explorer_tree.py
+++ b/src/sas/data_explorer_tree.py
@@ -143,7 +143,7 @@ class DataExplorerTree(QTreeWidget):
                         errors.append(err)
             case "make_trend":
                 trend_data = cast(list[SasData], self.currentTrackedData)
-                trend_creation_dialog = TrendCreation(trend_data, self._data_manager.default_trend_name)
+                trend_creation_dialog = TrendCreation(trend_data, self._data_manager.default_new_trend_name)
                 creation_result = trend_creation_dialog.exec()
                 if creation_result == QDialog.DialogCode.Accepted:
                     self._data_manager.register_trend(trend_creation_dialog.proposed_trend)
--- a/src/sas/data_explorer_tree.py
+++ b/src/sas/data_explorer_tree.py
@@ -133,6 +133,9 @@ class DataExplorerTree(QTreeWidget):
                         self._data_manager.make_association(to_perspective, datum)
                     except ValueError as err:
                         errors.append(err)
+            case "make_trend":
+                trend_data = cast(list[SasData], self.currentTrackedData)
+                self._data_manager.make_trend(trend_data)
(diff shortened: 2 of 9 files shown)


src/sas/refactored.py
```diff
@@ -77,3 +78,12 @@ class Theory:
     # at the current SasView codebase, it seems they are all just Data1Ds with
     # nothing else special.
     pass
+
+class NamedTrend(Trend):
```
krzywon, line 82: Shouldn't this be in sasdata?

src/sas/data_explorer_tree.py
```diff
@@ -1,28 +1,34 @@
...
     if isinstance(data, SasData):
         return data.name
+    elif isinstance(data, NamedTrend):
+        return data.name
+    elif isinstance(data, Trend):
+        # Trends created in the data explorer should be named trends. If they're
+        # not, that indicates something has gone wrong. We won't fail here, but
+        # instead log a warning.
+        logging.warning("Trend doesn't have a name. This shouldn't be happening.")
+        return 'Unnamed Trend'
```
krzywon, line 31: Should all trends have a default naming scheme?

src/sas/qtgui/MainWindow/TrendCreation.py
```diff
@@ -0,0 +1,122 @@
...
+
+        self.table.show()
+
+    def handle_add_axes(self):
+        # TODO: This is assuming all the data in target data has the same
+        # metadata, but this of course is not guaranteed. I see two ways of
+        # resolving this:
+        #
+        # 1. Validate that the objects do share all the same metadata.
+        # 2. Only show metadata in the explorer which the data objects share.
```
krzywon, line 96: Option 2 is ideal. One simple implementation I see is to create a tree of the meta data namespaces available in the first data object and then compare that to every other, removing any namespace that isn't available in each subsequent data object. If the dictionary is ever empty, 

```
# Data class method
def get_meta_data_keys():
   ...

# TrendCreation class method
def find_common_meta_data():
    keys = self.target_data[0].get_meta_data_keys()
    for data in self.target_data[1:]:
        for key in keys:
            # Check each key exists in the data
            # Keep a tally of non-existent keys
        # Remove non-existent keys, reset the list of keys, and break out early if keys is empty
```

src/sas/refactored.py
jamescrake-merani: The reason I included it here is because the name is at the moment only relevant to the data explorer so it knows what to call the trend. But I don't feel too strongly about it being here so I can move it to SasData. I'll probably do that at the same time I make some other changes to trends.

src/sas/data_explorer_tree.py
jamescrake-merani: At the moment, when the trend creation dialog is launched, the trend will get the name "Trend {n}", where n is based on how many trends there already are in the data explorer. The dialog will error if you try to have an empty name, so the user should never see 'Unnamed Trend' but I left it in there so the whole thing doesn't fall apart if that does happen.

src/sas/qtgui/MainWindow/TrendCreation.py
jamescrake-merani: This sounds like a good way of doing it. I'll have a go at implementing it, potentially tomorrow if I have time.

jamescrake-merani:
In the absence of any specific changes requested in the technical meeting, and given that trends have now been merged to `refactor_24` in sasdata, I'm now setting this as ready to review. I didn't get time to look into Jeff's comment regarding the metadata tree. I'm quite busy this week with graduate induction but hopefully I'll have time to look at this on Friday.

In the meantime, another review would of course be welcome.

src/sas/qtgui/MainWindow/TrendCreation.py
jamescrake-merani: I've just pushed some changes to resolve it. I ended up implementing it in the MetadataExplorer file, as now it can accept a list of SasData objects, and check the metadata from them.

build_tools/requirements.txt
```diff
@@ -25,7 +25,7 @@ pyparsing
 PySide6
 pytools
 qtconsole
-sasdata @ git+https://github.com/SasView/sasdata.git@refactor_24
+sasdata @ git+https://github.com/SasView/sasdata.git@refactor_24_trend_improvements
```
krzywon, line 28: Merge the sasdata branch and change this back before merging.

krzywon (approved):

PR #4111, opened 2026-09-03
be very careful with app-wide event filters. #4109

### LINKED ISSUE #4109: Plugin Model Error
**Describe the bug**
A user is seeing the following error while trying to use a pure python F(Q) plugin model coupled with the hardsphere structure factor. The user set the volfrac to 0.001 to see the dilute solution and the error appeared, and continued to appear every calculation after until a new model was selected in the fit tab. Going back to the original model, and the errors were no longer thrown. It's unclear if this is specific to F(Q) models or if it applies to all plugin models that can accept a structure factor.

> ERROR: Traceback (most recent call last): File "sas\qtgui\Utilities\ModelEditors\TabbedEditor\PluginDefinition.py", line 431, in eventFilter File "sas\qtgui\Utilities\ModelEditors\TabbedEditor\PluginDefinition.py", line 360, in _findParentTable TypeError: 'GuiManager' object is not callable

**To Reproduce**
1. Create a pure python F(Q) that can be coupled with a structure factor.
2. Calculate the model using the default parameters.
3. Change the volume fraction to something small (say 0.0001).
4. See the error above.
5. Any change in the values will continue to throw this error until a new model is selected.

**Expected behavior**
No error should occur for values within the default limits.

 - Version: 6.2.0 (unclear if previous versions are affected)

rozyczko:
## Description

Walking up the parentage hierarchy while the app-wide event filter is active is not healthy.
This PR fixes this in PluginDefinitions

Fixes #4109 

## How Has This Been Tested?

Local Win11 test

## Review Checklist:

**Installers**

### DIFF
--- a/src/sas/qtgui/Utilities/ModelEditors/TabbedEditor/TabbedModelEditor.py
+++ b/src/sas/qtgui/Utilities/ModelEditors/TabbedEditor/TabbedModelEditor.py
@@ -142,6 +142,9 @@ class TabbedModelEditor(QtWidgets.QDialog, Ui_TabbedModelEditor):
         """
         if self.is_modified and self.saveClose():
             return
+        app = QtWidgets.QApplication.instance()
+        if app is not None:
+            app.removeEventFilter(self)
         event.accept()
 
     def eventFilter(self, obj: QtCore.QObject, event: QtCore.QEvent) -> bool:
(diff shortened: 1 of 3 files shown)


src/sas/qtgui/Utilities/ModelEditors/TabbedEditor/TabbedModelEditor.py
```diff
@@ -150,11 +153,13 @@ def eventFilter(self, obj: QtCore.QObject, event: QtCore.QEvent) -> bool:
         """
         if event.type() != QtCore.QEvent.KeyPress:
             return super(TabbedModelEditor, self).eventFilter(obj, event)
-        if event.key() == QtCore.Qt.Key_Escape:
-            if isinstance(obj, QtWidgets.QDialog) and (obj == self or self.isAncestorOf(obj)):
-                return True
-            else:
-                return False
+        if event.key() != QtCore.Qt.Key_Escape:
+            return super(TabbedModelEditor, self).eventFilter(obj, event)
```
DrPaulSharp: Can you rely on short circuiting to combine these lines as: 
```
if event.type() != QtCore.QEvent.KeyPress or event.key() != QtCore.Qt.Key_Escape:
    return super(TabbedModelEditor, self).eventFilter(obj, event)
```

DrPaulSharp (commented):
This code is a clear improvement, however I have not been able to reproduce the bug from #4109. If someone can verify the bug is resolved that would be very helpful.

src/sas/qtgui/Utilities/ModelEditors/TabbedEditor/TabbedModelEditor.py
DrPaulSharp: i.e., these two lines, with those immediately above.

Issue #3993, opened 2026-06-04
Leftover crumbs when renaming a model


butlerpd:
**Describe the bug**
I think this goes in the sasview repo but it might be a sasmodel issue.  In testing the alpha3's new polyhedron models I noticed that the polyhedron category contains **TWO** versions of the truncated octahedron model: `octahedron_truncated` and  `truncated_octathedron`.  Only the second one works. The firsrt looks like a relic of a first version before it was renamed to meet sasview naming convention? It seems somehow still in the list somewhere but the code no longer is. It returns a Key Error:

```phython
17:10:25 - ERROR: Traceback (most recent call last): File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1195, in onSelectModel File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1368, in respondToModelStructure File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 1985, in SASModelToQModel File "sas\qtgui\Perspectives\Fitting\FittingWidget.py", line 2032, in fromModelToQModel KeyError: 'octahedron_truncated'
```

**SasView version (please complete the following information):**
 - Version: 6.2.0a3

**Operating system (please complete the following information):**
 - OS: Windows 11

**Installation method (please complete the following information):**
 - Installed: from github release page artifact. Ran that installer locally to install on my machine

krzywon:
The bug and the title are two different things and I think we should track this, but your title isn't what is really wrong here. The issue is your `category.json` file is now polluted with a model you no longer have in your installation, but SasView does not have a mechanism to handle this. Currently, plugin models are suppressed if they don't exist or can't be loaded, but with an error. We should do something similar for built-in models as well.

krzywon:
Suggested title: `Check that all models exist on startup`

butlerpd:
So your suggesting this only happens on machines which ran/tested installer that once DID have models with those names packaged? So if a package has models that don't exist in the json file they get added, but if some exits in the json file that don't exist in the installer they do not get removed?  If so then your title makes some sense. I can try later on a different machine to make sure that is the case.

krzywon:
Correct. Alternately, delete your `categories.json` file and see what happens. Notes on how to do this - https://github.com/SasView/sasview/pull/3961#issuecomment-4595510688

DrPaulSharp:
[...]

PR #4118, opened 2026-09-09
3993: Remove non-existent models from the FittingWidget

krzywon:
## Description

This adds a check to the fitting widget to suppress models from the model selector if they are listed in `categories.json`, but do not exist in sasmodels.

Refs #3993 (maybe fixes, but there is a separate documentation issue we might need to address)

## How Has This Been Tested?

New unit test

## Review Checklist:

**Installers**


src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1930,8 +1930,13 @@ def regenerateModelDict(self) -> None:
...
         self.by_model_dict = defaultdict(list)
-        for category in self.master_category_dict:
-            for (model, enabled) in self.master_category_dict[category]:
+        master_category_list_iterator = self.master_category_dict.copy()
+        for category in master_category_list_iterator:
+            for (model, enabled) in master_category_list_iterator[category]:
+                if model not in MODELS:
+                    logging.warning(f"The model {model} is no longer available in sasmodels. It will be ignored.")
+                    self.master_category_dict[category].remove([model, enabled])
+                    continue
```
rozyczko: if you have two bad models in a row, the second one goes through. You should rebuild each list and never delete while looping

```python
{"Cylinder": [a, x, y, b]}   # x and y missing from sasmodels
→ {"Cylinder": [a, y, b]}     # y is still there
```

src/sas/qtgui/Perspectives/Fitting/FittingWidget.py
```diff
@@ -1930,8 +1930,13 @@ def regenerateModelDict(self) -> None:
         along with the enabled mapping
         """
         self.by_model_dict = defaultdict(list)
-        for category in self.master_category_dict:
-            for (model, enabled) in self.master_category_dict[category]:
+        master_category_list_iterator = self.master_category_dict.copy()
+        for category in master_category_list_iterator:
```
rozyczko: The removal below only happens in memory, so the user's categories.json still have the bad entries. The warning comes back every time readCategoryInfo runs, once per new fit tab and again on every category change.

src/sas/qtgui/Perspectives/Fitting/UnitTesting/FittingWidgetTest.py
```diff
@@ -162,6 +162,19 @@ def testSelectCategoryDefault(self, widget):
         #Test what is current text in the combobox
         assert fittingWindow.cbCategory.currentText() == FittingWidget.CATEGORY_DEFAULT
 
+    def testNonExistantModels(self, widget):
```
DrPaulSharp, line 165: Typo - should be "Existent"

DrPaulSharp (approved):
As discussed, this should help with part of the problem and hence be included in 6.2.1, but let's keep the issue open given the documentation problems.

PR #4124, opened 2026-09-17
Allow MDIArea windows be detached/reattached on demand #4094

### LINKED ISSUE #4094: Be able to pop out windows from the MDiArea
This is an idea I've mentioned in meetings in the past. Currently, subwindows in SasView (perspectives, plots etc) are contained within the MDiArea but in previous versions of SasView, perspectives were floating in their own separate windows. The aim here is to provide both options: the user can continue to have their subwindows in the MDiArea, but they can also pop them out if they want to. This is especially useful if user is working across two monitors, as they could move subwindows to other monitors.

After doing some looking, it appears QT's MDiArea does not support this feature itself, but it should be possible to implement this manually by reassigning the parent of the window.

rozyczko:
## Description

Introduced a `WorkspaceManager` abstraction to manage all workspace windows, both attached and detached, enabling windows (plots, perspectives, panels) to be detached into their own top-level windows and reattached to the workspace. This affects how windows are added, shown, closed, minimized, and activated. 
Windows (plots, perspectives, panels) now remember their size and position separately for attached and detached states, and restore to their previous state when reattached or detached again. Closing and minimizing behavior is updated for detached windows

Fixes #4094 

## How Has This Been Tested?

Local win11 checks

## Review Checklist:

- [X] Documentation changes are **in this PR**

**Installers**


docs/sphinx-docs/source/user/menu_bar.rst
```diff
@@ -43,12 +43,46 @@ Window
...
+*SasView* workspace. Any of them can be detached into its own window, which can then be moved
+anywhere on the desktop, for example onto a second monitor. A detached window keeps all of its
+content and settings, and can be attached back to the workspace at any time.
+
+To detach a window, do one of the following:
+
+- click the icon at the top-left corner of the window and select *Detach from Workspace*;
+- make the window active and select *Window > Detach Window from Workspace*, or press
+  [Ctrl]-[Shift]-[D];
+- for a plot, right-click on the plot and select *Detach from Workspace*.
```
krzywon, line 67: Is there not a drag-and-drop option? I would argue that is the most intuitive way to detach/attach.

src/sas/qtgui/MainWindow/UnitTesting/Quarantine/GuiManagerTest.py
```diff
@@ -36,6 +37,7 @@ def __init__(self, parent=None):
                 # define workspace for dialogs.
                 self.workspace = QMdiArea(self)
                 self.setCentralWidget(self.workspace)
+                self.workspace_manager = WorkspaceManager(self.workspace, self)
```
krzywon, line 40: This test file is currently ignored by `pytest`.

src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -1271,14 +1300,58 @@ def actionArrange_Icons(self):
...
+        workspace_manager.activate(target)
+
+    def actionDetachWindow(self):
+        """
+        Detach the active window from the workspace, or attach it back
+        """
+        workspace_manager = self.workspace_manager
+        widget = workspace_manager.active_widget()
+        if widget is None:
+            return
```
krzywon, line 1336: Walrus operators save lines!

`if widget := workspace_manager.active_widget() is None:`

docs/sphinx-docs/source/user/menu_bar.rst
```diff
@@ -43,12 +43,46 @@ Window
...
+  [Ctrl]-[Shift]-[D];
+- for a plot, right-click on the plot and select *Attach to Workspace*.
+
+A window remembers its size and position separately inside the workspace and as a detached window,
+and returns to the ones it last had in each. The first time a window is detached, it keeps its
+current size. A window that is maximised or minimised when you detach or attach it returns to its
+previous normal size when restored. Closing a detached analysis perspective minimises it, as in the
+workspace. When you switch to another perspective and
+back, the perspective reappears where you left it, attached or detached. Detached windows are closed
+when *SasView* exits; their placement is not saved in project files.
```
krzywon, line 82: This could use some clarification. Reading this, I assumed if one perspective was detached, when I switched perspectives, the new perspective would open in the same position as the one being hidden. If I read it this way, others may as well.

`When SasView opens, all perspectives are attached or detached, as selected in the general preferences, and switching from one perspective to another may result in some perspectives attach and others detached.` - I don't like the wording, but something to this effect.

src/sas/qtgui/Utilities/Preferences/GeneralPreferencesWidget.py
```diff
@@ -0,0 +1,32 @@
+from sas.system import config
+
+from .PreferencesWidget import PreferencesWidget
+
+
+class GeneralPreferencesWidget(PreferencesWidget):
```
krzywon, line 6: This same class was created in #4125, but that was merged into main. What kinds of conflicts will this result in?

src/sas/qtgui/MainWindow/WorkspaceManager.py
```diff
@@ -0,0 +1,798 @@
...
+    explicit hide and is ignored.
+    """
+
+    def __init__(self, container: _HostMixin):
+        super().__init__(container)
+        self._container = container
+        # True while the hosted widget's close handler is running
+        self.dispatching = False
+
+    def eventFilter(self, obj, event):
```
krzywon, line 93: I know there was a discussion on global event filters, but this is one case I think it makes sense.

krzywon (commented):
Overall, this seems to work pretty well and is a welcome feature. I think this might be something that could go into our next feature release, so a rebase against main would be helpful.

Suggestions I have and Issues I've noted while testing:
- Attach buttons are a welcome feature. Can we add Detach buttons as well?
- Icon on detachable windows is Qt, not the SasView icon. The icon on detached windows is correct.
- Box sum window is resized after reattaching.
- A detached slicer parameters window does not reopen detached after closing it.

Side note - this branch is not up-to-date with main, so some GUI tests are failing on Windows, that do not fail on Ubuntu (magnetic models and mpl defaults). That said, a rebase might help to resolve this.

src/sas/qtgui/MainWindow/WorkspaceManager.py
```diff
@@ -0,0 +1,798 @@
...
+        widget = self._active
+        if widget is None:
+            return None
+        container = self._hosts.get(widget)
+        if container is None or container.isHidden():
+            return None
+        return widget
+
+    def mdi(self) -> QMdiArea:
+        return self._mdi
```
jamescrake-merani, line 616: Should this not be a property?

src/sas/qtgui/Perspectives/Inversion/UnitTesting/InversionPerspectiveTest.py
```diff
@@ -201,14 +213,20 @@ def testRemoveData(self, widget):
     @pytest.mark.skip(reason="2026-02: Freezing on launch")
     def testClose(self, widget):
         """ Test methods related to closing the window """
+        window, manager = host_in_workspace(widget)
```
jamescrake-merani, line 216: This test was previously marked as skip. Perhaps its worth checking that the changes fix the issue that was spotted before?

jamescrake-merani (changes requested):
I don't have many comments on the code to add onto what Jeff has already said. However, I do have a concern about the 'WorkspaceManager' class. It probably works fine for what we have at the moment in the main SasView application, but when it comes to the refactored data explorer, we already have a new data manager which keeps track of perspectives, and plots, both of which the new Workspace Manager also tracks. I think in the refactoring project, I would like to see the Workspace Manager, and Data Manager merged (or perhaps the Workspace Manager becomes a component of the Data Manager) because there needs to be one source of truth for everything, otherwise we risk the state of both managers diverging, and bugs creeping in with this.

With this in mind, I propose that:
1. As Jeff has proposed, this PR is rebased onto main, and appears in the next SasView feature release as I think this PR is ready for that.
2. A separate issue is made to integrate this work into the refactoring project's Data Explorer, which I (or another volunteer) can take a look at separately.

On the second point, I hope to get round to writing some documentation for the refactoring project which should hopefully make it clearer as to how the new data manager is structured as I appreciate that at the moment its just something I've spoken about in meetings but we don't have anything written down to explain it other than the code itself.

src/sas/qtgui/MainWindow/WorkspaceManager.py
rozyczko: Indeed. For the workspace manager we need this global

src/sas/qtgui/MainWindow/UnitTesting/Quarantine/GuiManagerTest.py
rozyczko: oh... right. removed

docs/sphinx-docs/source/user/menu_bar.rst
rozyczko: Modified the text

src/sas/qtgui/Utilities/Preferences/GeneralPreferencesWidget.py
rozyczko: we will see when rebasing onto `main`...

docs/sphinx-docs/source/user/menu_bar.rst
rozyczko: Yes, addressed in a separate PR, since I want to make sure this is in the shape we want. Please have a look at https://github.com/SasView/sasview/pull/4143

src/sas/qtgui/MainWindow/WorkspaceManager.py
rozyczko: The rest of the WorkspaceManager API uses methods (active_widget(), container_of(), hosted_widgets()), as do Qt's own getters (QMdiSubWindow.mdiArea(), widget()).
I think that making just this one a property would be inconsistent...

rozyczko:
I addressed all the issues by @jamescrake-merani and @krzywon 
Now I am going to rebase the changes against current `main` and make a separate PR, linking this PR.

Issue #4112, opened 2026-09-04
Reimplementation of Mumag as a perspective


jamescrake-merani:
Mumag is currently implemented as a tool in SasView. It is to be reimplemented as a perspective, replacing its own logic, and data structures for loading in data with `SasData`, and `Trend` from the refactoring project.

Issue #4132, opened 2026-09-18
Plotting issues with mumag


jamescrake-merani:
**Describe the bug**

Some of the numbers on the axes are not visible on the plot, either because they have been cut off, or because they overlap over other numbers. This was spotting in work on the refactoring project but it is not a new bug. 

**To Reproduce**
1. Tools -> Mumag fitter
2. Load data
3. Fit
4. See plots.

**Screenshots**

<img width="887" height="783" alt="Image" src="https://github.com/user-attachments/assets/9927a7a3-607d-4db0-b2ae-8ec96903d1ba" />

<img width="905" height="804" alt="Image" src="https://github.com/user-attachments/assets/dbc00018-b228-4275-a37c-f7be56f2e8ab" />

**SasView version:**
 - Version: 6.2.0

**Operating system:**
 - OS: Tested on both Windows 10, and Ubuntu 24.04 LTS

**Installation method (please complete the following information):**
 - Installed: Pip for refactoring; Flatpak for 6.2.0

**Additional context**
There's a question about how high priority this is to fix given that the eventual goal is to integrate Mumag's plotting with the plotting refactor, but some of the Mumag plotting may need to be added to the plotting refactor so we might still need to consider this in that work.

Issue #4129, opened 2026-09-18
Mumag fitter fails to fit 'NdFeB_parallel_Bick_et_al'


jamescrake-merani:
**Describe the bug**
The Mumag tool fails to fit the data found in this directory:
'sasview/src/sas/qtgui/Utilities/MuMag/SANSData/NdFeB_parallel_Bick_et_al/'

With the error:

```
10:59:03 - ERROR: Fitting failed - are the parameters correct? FitFailure("A = -1.865191706480297e-11 (ValueError('array must not contain infs or NaNs'))")
```

This happens in the refactoring project as well but it actually appears to be an existing bug as I can reproduce it on SasView 6.2.0

**To Reproduce**
1. Go to Tools -> Mumag Fitter
2. Click on 'Import Data'
3. Load the data in the directory specified above

**SasView version:**
 - Version: 6.2.0

**Operating system:**
 - OS: Ubuntu 24.04 LTS

**Installation method (please complete the following information):**
 - Installed: Flatpak & Pip

**Additional Context**

I had a quick glance at the files themselves, and I cannot see anything obviously wrong with them. This was one of the datasets that was provided with Mumag, so I assume it was intended to work with it.

PR #4133, opened 2026-09-18
Refactor24 mumag perspective

jamescrake-merani:
## Description

This PR reimplemments Mumag as a perspective. In doing so:

- Mumag now uses quantities, and trends from sasdata.
- Perspectives can specify which type of data they accept. The dummy perspective accepts only `SasData`, whereas Mumag accepts only `Trend`

Known issues that will be fixed in separate branches:

- Help button doesn't work. The new DataManager needs to reimplement this.
- Saving data no longer works. I thought it would be better to wait for the work on exporting data.
- Pre-existing Mumag issues (#4129 , #4132)

Fixes #4112 

## How Has This Been Tested?

All data that was provided previously to test with Mumag works, except from 'NdFeB_parallel_Bick_et_al' but as noted above, this appears to be a pre-existing issue. The data now needs to be loaded as a trend. You can find instructions on loading in trends in the trend creation PR: #4107  

Since Mumag is a perspective in the new data explorer, you can now have multiple instances of Mumag running at the same time, and they appear to work fine.

## Review Checklist:

**Installers**


src/sas/qtgui/Utilities/MuMag/datastructures.py
```diff
@@ -116,7 +93,7 @@ class SweepOutput(Generic[T]):
 class FitResults:
     """ Output the MuMag fit """
     parameters: FitParameters
-    input_data: list[ExperimentalData]
```
krzywon, line 119: nitpick: The parameter name doesn't need to change here, but I understand why you did it.

src/sas/qtgui/Utilities/MuMag/MuMag.py
```diff
@@ -82,27 +86,31 @@ def __init__(self, parent=None):
         # Set visibility
         self.hide_everything()
 
-    def importData(self):
-        """ Callback for the import data button """
+    @property
+    def trend(self) -> Trend | None:
+        if len(self.associatedData) == 0:
+            return None
+        return cast(Trend, self.associatedData[0])
```
krzywon, line 93: Why are you only returning the first entry?

src/sas/qtgui/Utilities/MuMag/MuMag.py
```diff
@@ -122,22 +130,29 @@ def hide_everything(self):
...
+        colors = pl.cm.jet(np.linspace(0, 1, len(self.trend.data)))
 
-        for i, datum in enumerate(self.data):
+        for i, datum in enumerate(self.trend.data):
 
-            self.data_axes.loglog(datum.scattering_curve.x,
-                      datum.scattering_curve.y,
-                      linestyle='-', color=colors[i], linewidth=0.5,
-                      label=r'$B_0 = ' + str(datum.applied_field) + '$ T')
+            self.data_axes.loglog(datum.abscissae.axes[0].value,
```
krzywon, line 144: The majority of your changes were either typing related or ensuring you get the value of one or more Quantity parameters. The former is expected with the changes you've made, but the latter seems like an extra step that the Quantity class should be able to handle. Calling `datum.ordinate` should return `datum.ordinate.value` so developers don't have to remember to always explicitly ask for the value.

krzywon (commented):
Ideally, the MuMag package would move into the Perspectives package. Haven't had a chance to test functionality yet. Most code changes seem reasonable.

src/sas/qtgui/Utilities/MuMag/MuMag.py
jamescrake-merani: `associatedData` is a property which returns all the data associated with the perspective. However, Mumag (unlike Dummy) does not support having multiple data sent to it hence

```python
    @property
    @override
    def supports_multiple_data(self) -> bool:
        return False
```

So associatedData should either be of length 0 (i.e. it hasn't had a trend sent to it yet), or 1 (it has one trend sent to it). The data explorer will stop the user from sending another trend to Mumag thus we can assume the trend is the first associated data.

I do see where you are coming from though. One idea I have is to make a separate property for getting only one data object, and this property can also error when a developer tries to do it on a perspective which supports having multiple data sent to it. That would probably be for a separate branch though, perhaps combined with some other changes to the data manager to make it more intuitive for developers.

src/sas/qtgui/Utilities/MuMag/MuMag.py
jamescrake-merani: So in case where you are doing simple calculations, a Quantity object can be treated as a drop-in replacement for a float value, except it will handle all the unit conversions as we'd expect. The problem, however, is that libraries (like plotting libraries) will see a Quantity, and not know what to do with it. The `value` field just gets the raw float/numpy array value without any unit information on top of it, which other libraries can then handle.

This isn't ideal I know, and we may need to think about changes to the Quantity class. But just making `datum.ordinate` return `datum.ordinate.value` means we would lose information about the unit during calculations where that's important. I wonder if its possible to do some Python wizardry where we can make libraries treat Quantity objects as floats while still retaining all the additional information Quantity objects have.

jamescrake-merani:
> Ideally, the MuMag package would move into the Perspectives package

Yes, I didn't think about that but you're right. I think perhaps its best to do that in another branch as I'm not sure whether GitHub is smart enough to retain the diffs even when the file itself has been moved.

jamescrake-merani:
Rebased to refactor_24, and this is now ready for review.

jamescrake-merani:
Rebased to refactor_24, and this is now ready for review.

PR #4134, opened 2026-09-21
Merge data explorer into the main refactor 24 branch.

jamescrake-merani:
As discussed in the last meeting, to keep the branches more organised, I am requesting to merge data explorer into the main refactor branch so that the refactor branch can be the main point for merges.

This brings all of the work in data explorer (including ascii reader, data loader interaction with sasdata etc) into the main refactor branch.


src/sas/qtgui/MainWindow/GuiManager.py
```diff
@@ -154,7 +172,10 @@ def addWidgets(self):
         self.loadAllPerspectives()
 
         # Add FileDialog widget as docked - Perspectives must be loaded to ensure default perspective is shown
-        self.filesWidget = DataExplorerWindow(self._parent, self, manager=self._data_manager)
+        # self.filesWidget = DataExplorerWindow(self._parent, self, manager=self._data_manager)
+        # TODO: Is this a good opportunity to change this name? dataExplorer would be better I think.
+        self.filesWidget = NewDataExplorer(self._data_manager ,self._parent)
```
DrPaulSharp, line 177: Good question. Maybe consider this again now before merging this branch in.

src/sas/data_manager.py
```diff
@@ -0,0 +1,120 @@
...
+        return isinstance(obj, t)
+    else:
+        return False
+
+# Making this a QObject so it'll support events.
+class NewDataManager(QObject):
+    # Don't mutate these directly, or scary bad stuff will happen.
+    _all_data_entries: list[TrackedData]
+    associations: list[tuple[TrackedData, TrackedData]]
+    # These all take in 'object' because PySide6 can't handle handle having
```
DrPaulSharp: Is this comment right?

DrPaulSharp (approved):
All looks good. I've a couple of points to consider:

- It may be time to consider improving some temporary names in this branch. I'm thinking  of the files `refactored.py`, `refactored_data_explorer.py` and the classes `NewDataExplorer` and `NewDataManager`. The time to rename them might not yet be right, but worth thinking about at this point.
- It may be worth considering how this branch, and merging it, will impact on other branches submitted recently. The example that made me think of this is the use of `temp_xml_reader` in `MetadataExplorer`, which is renamed in SasView/sasdata#235.

jamescrake-merani:
> It may be time to consider improving some temporary names in this branch. I'm thinking of the files refactored.py, refactored_data_explorer.py and the classes NewDataExplorer and NewDataManager. The time to rename them might not yet be right, but worth thinking about at this point.

Yes that is true. My suggestion is that I make a new issue regarding this, and we move it when we feel it is appropriate. That work would also involve removing the previous data explorer. Tbh, I think we are coming closer to the point where this should be done so I might suggest we do this sooner rather than later.

> It may be worth considering how this branch, and merging it, will impact on other branches submitted recently. The example that made me think of this is the use of temp_xml_reader in MetadataExplorer, which is renamed in https://github.com/SasView/sasdata/pull/235.

My plan is to make the necessary changes once Jeff has completed all of his work on importers. There will be quite a few things that need to be changed, so I'm just waiting for it all to be ready so I can do it all in one go.

src/sas/data_manager.py
jamescrake-merani: Good spot. I must have gotten distracted in the middle of writing that comment lol.

src/sas/qtgui/MainWindow/GuiManager.py
jamescrake-merani: I'm a bit worried about potential merge conflicts if this renamed before other work is merged. I suggest that I create an issue after merging this branch, get other things merged into the base refactor branch, and then make a separate branch to rename this.

